PR #33, opened 2022-01-18
video: use a better vbv-bufsize & correct software bitrate calculation

cfajardo:
## Description

Change cherry-picked from @psyke83 repository. It has a real impact on stream quality on my various tests.

* Increase vbv-bufsize to 1/10 of requested bitrate. The previous value
  was too low, which was resulting in poor encoding quality and failure to
  stabilize at the requested bitrate. Setting to 1/10 of bitrate is a
  good compromise, as it avoids quality loss but also prevents bandwidth
  spikes far above the bitrate/vbv-maxrate.
* With vbv-bufsize set to a more appropriate value, testing shows that
  the average bitrate vs client-requested bandwidth overshoots by ~20%.
  Adjust for this scenario in the software encoding case only.

### Screenshot

N/A

### Issues Fixed or Closed

N/A

## Type of Change

## Checklist

- [X] I have added or updated the documentation blocks for new or existing components

### DIFF
--- a/sunshine/video.cpp
+++ b/sunshine/video.cpp
@@ -930,9 +930,9 @@ std::optional<session_t> make_session(const encoder_t &encoder, const config_t &
   }
 
   if(video_format[encoder_t::CBR]) {
-    auto bitrate        = config.bitrate * 1000;
+    auto bitrate        = config.bitrate * (hardware ? 1000 : 800); // software bitrate overshoots by ~20%
     ctx->rc_max_rate    = bitrate;
-    ctx->rc_buffer_size = bitrate / config.framerate;
+    ctx->rc_buffer_size = bitrate / 10;
     ctx->bit_rate       = bitrate;
     ctx->rc_min_rate    = bitrate;
   }


ReenigneArcher:
@cfajardo thanks for submitting this! I'd like to give the original author a change to submit it themselves if they feel that's necessary.

@psyke83 if you'd like to submit this yourself, please let me know or just open a separate PR.

Thanks!

psyke83:
@ReenigneArcher 

Thanks for the heads up. I've no issue with this PR as-is; just keep in mind that I already submitted this change via my own PR to upstream, just in case that causes you issues when merging/rebasing (if/when it gets accepted).

ReenigneArcher:
> Thanks for the heads up. I've no issue with this PR as-is; just keep in mind that I already submitted this change via my own PR to upstream, just in case that causes you issues when merging/rebasing (if/when it gets accepted).

Sounds good. Thanks again! Also this repo is an import of the original project, not a fork. So as far as I know any changes here would need to be copied into a fork of the main repo and then submitted. I decided to use an import instead of a fork because forks cannot be searched and it makes it more difficult to submit PRs. Not sure what's going on with Loki, but hopefully nothing serious. The goal here is to make things more manageable and somewhat automated.

cfajardo:
Tbh I felt uncomfortable with publishing this PR as it's not my own code. Thanks @psyke83 for your openness.

TheElixZammuto (approved):
Haven't tested this, but LGTM since the explication seems good enough and the modification is very simple

ReenigneArcher:
Just for reference: this is the same as https://github.com/loki-47-6F-64/sunshine/pull/288

PR #34, opened 2022-01-19
UI Web Fixes

TheElixZammuto:
## Description

Minor fixes to the Web UI, updated messages to make sure they are correct
## Type of Change

## Checklist

### DIFF
--- a/assets/web/pin.html
+++ b/assets/web/pin.html
@@ -28,14 +28,14 @@
     fetch("/api/pin", { method: "POST", body: b })
       .then((response) => response.json())
       .then((response) => {
-        if (response.status) {
+        if (response.status && response.status === "true") {
           document.querySelector(
             "#status"
           ).innerHTML = `<div class="alert alert-success" role="alert">Success! Please check Moonlight to continue</div>`;
         } else {
           document.querySelector(
             "#status"
-          ).innerHTML = `<div class="alert alert-danger" role="alert">PIN does not match, please check if it's typed correctly</div>`;
+          ).innerHTML = `<div class="alert alert-danger" role="alert">Pairing Failed: Check if the PIN is typed correctly</div>`;
         }
       });
   });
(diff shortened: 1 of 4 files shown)


TheElixZammuto:
> Pull requests must be made to the `nightly` branch. Thanks.

Good bot. Fixed by changing the base branch

assets/web/pin.html
```diff
@@ -28,14 +28,14 @@ <h1 class="my-4">PIN Pairing</h1>
     fetch("/api/pin", { method: "POST", body: b })
       .then((response) => response.json())
       .then((response) => {
-        if (response.status) {
+        if (response.status && response.status === "true") {
```
ReenigneArcher, line 31: Probably only need one or the other here.

I don't know exactly `response.status` is returning. If it's a string then the second one is probably what is needed over the first one as `if (response.status)` would be Truthy if any string is present.

https://stackoverflow.com/a/23061950/11214013

Also, if `response.status` is a string, it might be worth changing to `response.status.toLowerCase()`

Assuming a string, I'd suggest:
```javascript
if (response.status.toLowerCase() === "true") {
```

assets/web/welcome.html
```diff
@@ -2,12 +2,11 @@
...
     <h1 class="mb-0">Welcome to Sunshine!</h1>
     <p class="mb-0 align-self-start">
-      Before Getting Started, write down below these credentials
+      Before Getting Started, we need you to make a new username and password for accessing the Web UI.
     </p>
   </header>
   <div class="alert alert-warning">
-    These Credentials down below are needed to access the rest of the
-    application.<br />
+    These Credentials down below are needed to access Sunshine's Web UI.<br />
```
ReenigneArcher, line 9: I'd suggest changing this to:

`The credentials below are needed to access Sunshine's Web UI.`

ReenigneArcher (commented):
Reviewed

assets/web/pin.html
TheElixZammuto: I decided to make it more explicit by combining both toString and toLowerCase

ReenigneArcher (approved):

PR #38, opened 2022-01-23
Add version argument

ReenigneArcher:
## Description
- Add version argument

### Screenshot
image

### Issues Fixed or Closed
- Fixes Fixes loki-47-6F-64/sunshine#244

## Type of Change

## Checklist


ReenigneArcher:
This is just a first step.

We can merge or address this first. Would like to automate the version number during build so the CMakeLists.txt doesn't need to be manually updated before each release. Not sure the best way to accomplish that at the moment.

Some ideas I have:
- Use the changelog_parser and pass in the version as an argument. The issue being that the verify changelog action https://github.com/SunshineStream/Sunshine/blob/870b6e26753810f628608c0cf173f1f781da85db/.github/workflows/create_package.yml#L21 only runs on push into master branch or PR targeting master branch, so the variable wouldn't be available for the other builds. These are the known lines (at a minimum) that need to be updated:
  - https://github.com/SunshineStream/Sunshine/blob/ebc581e16519ae19fa8da146443a004cdb9c6b74/.github/workflows/create_package.yml#L71
  - https://github.com/SunshineStream/Sunshine/blob/ebc581e16519ae19fa8da146443a004cdb9c6b74/.github/workflows/create_package.yml#L203
  - https://github.com/SunshineStream/Sunshine/blob/ebc581e16519ae19fa8da146443a004cdb9c6b74/scripts/build-private.sh#L31
- Possibly when it's not a push or PR into master branch, use the commit tag. Not sure if it even accepts non semantic version numbers. If not, maybe it's possible to have a default value when it's not an official release, something like `0.0.0`
- Or, just take caution and ensure the `CMakeLists.txt` is updated whenever we update the `CHANGELOG.md`.
  - For PR or push into master... add a step in the build process to run `sunshine --version`, parse the output, and compare it to the changelog parser output.

TheElixZammuto:
Probably not possible to add the commit hash a version unfortunately, see here: https://gitlab.kitware.com/cmake/cmake/-/issues/16716

Maybe at the moment adding the check in the PR phase is a better option due to the limitations in CMake

ReenigneArcher:
I've added a version test step for the AppImage build job. It would be somewhat redundant to add to each build job, also a little bit more complicated to add to to the dockerfile builds (not quite sure how to set an environment variable on the host using the dockerfiles without doing more research).

TheElixZammuto (approved):

PR #53, opened 2022-02-13
WIP: WebSocket API

TheChoconut:
## Description

Remove Web UI in its current state, if wished it may be moved to another location.
I have coded another Web UI project that's currently not available but will be available in the future on GitHub.
Instead of having a REST API, I have decided to move to Web Sockets. The primary reason is the ability to send a pin_request from the server and show a PIN dialog directly. But this may include more useful functionality in the future, I don't think it's a bad idea.

The primary idea is to have a WebSocket talk to the sunshine backend and move the UI separate from the sunshine. WebSockets should be supported in virtually any UI possible, so while I have this website here talking to it, it could have also been a native application written in Qt (though I do not know how to write Qt apps). When it comes to certificates though, it is required to sign a certificate against a CA that has all the possible IP addresses and hostnames... I made some kind of a script by batching together some code online, but I need some help and advice on that. If I install that CA in the Windows Certificate Store, all browsers will accept the connection.

Talking to the API is now different, since every request must be a JSON string containing the type of request and data you want to provide with the request.

You can try building this and access the Web UI here: https://sleepy-hypatia-c15080.netlify.app/

You currently need to add an exception to the self-signed certificate (read [here](https://www.ibm.com/docs/en/zos/2.3.0?topic=interface-certificate-error-in-mozilla-firefox-browser)).
It is not responsive and not all features have been ported.

I would like some feedback and comments on this idea.

## TODO / Ideas

- [ ] Port all features to new Web UI
- [ ] Certificate scripts (creating and importing to Windows certificate store)
- [ ] Build into sunshinesvc and make sunshine reloadable via web UI?
- [ ] Proxy support docs?

### Screenshot
[https://streamable.com/otludf](https://streamable.com/otludf)

## Type of Change

## Checklist


sunshine/main.cpp
```diff
@@ -103,14 +102,11 @@ int entry(const char *name, int argc, char *argv[]) {
     return 0;
   }
 
-  http::save_user_creds(config::sunshine.credentials_file, argv[0], argv[1]);
-
   return 0;
 }
 } // namespace gen_creds
```
ReenigneArcher, line 107: The entire namespace `gen_creds` should probably be removed.

ReenigneArcher (changes requested):
I re-opened the PR and changed the branch over to `nightly`. Could you rebase your `master` branch? Otherwise it would be fine if you submit another PR, whichever you prefer.

The sample `sunshine.conf` file will also need to be updated to remove/add updated variables.

Would it be difficult to have an option for which web UI to use? This could allow a smooth transition.

```txt
interface = 0

where 0 = WebServer
          1 = WebSocket API
```

TheElixZammuto:
WebSockets are a much worse protocol for these types of applications. They are less mature, there are less tools/library for interacting (while with HTTP Api are pratically universal and already integrated in Sunshine since also GFE uses them - so less technicall debt).

If the reason is simply showing the PIN when asked, why not simply use HTTP Server Side events? https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

Also having a root CA is a bad idea since any other software with any hostname can appear trusted, which is a big no no for me.  I don't know if there alternative (we can start by having as a supported hostname only localhost - but is not enough)

Like I said many times before, removing the Web UI altogheter from the project instead of improving the already existing one is just a way to alienate the more casual user target which then  is forced to juggle between different UIs with different protocols supported, etc...

TheChoconut:
Closing this since WebSockets are not needed and maybe overkill for this purpose.

PR #55, opened 2022-02-16
Add clang linter

ReenigneArcher:
## Description
- Add clang linter
- Add dependabot (github action dependencies only)

## Type of Change

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "nightly"
+    open-pull-requests-limit: 20
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -1,0 +1,23 @@
+name: clang-format-lint
+
+on:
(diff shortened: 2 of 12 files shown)


ReenigneArcher:
The modified c++ files are the results of the linter. The action uploads them as an artifact.

TheElixZammuto:
What happens if the linter finds errors in the code style? Does it fail the pipeline or it just provides the changed files?

ReenigneArcher:
> What happens if the linter finds errors in the code style? Does it fail the pipeline or it just provides the changed files?

It's a matrix strategy. So when "inplace" is true the workflow succeeds and the artifacts are uploaded. Then it runs with "inplace" as false and will fail the workflow if there are errors.

The latest failure should show that as an example.

TheElixZammuto:
> > What happens if the linter finds errors in the code style? Does it fail the pipeline or it just provides the changed files?
> 
> It's a matrix strategy. So when "inplace" is true the workflow succeeds and the artifacts are uploaded. Then it runs with "inplace" as false and will fail the workflow if there are errors.
> 
> The latest failure should show that as an example.
Gotcha, so it's the best of both worlds! ^_^

TheElixZammuto (approved):

PR #62, opened 2022-02-22
Move to Web API

TheChoconut:
## Description
Remove Web UI in its current state, if wished it may be moved to another location.
I have coded another Web UI project, but the old one can also be updated to support this API.

I really wanted to clean up the current code. Having a function for handling every web UI file is really messy. Also, the current code requires a lot of copy&paste since every request needs to be printed for debugging, checked for authentication etc. Now, my code is also not that clean but if you have any suggestions on how to improve it, I'd be thankful.

Added new features:
- support for [SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) 
- password/token authentication

Password checking is implemented. The default password is created on run and it encrypts a new file, "sunshine_creds.data" using AES GCM. The password is printed in stdout. When authenticating using a password, the file is decrypted using a partial password hash (SHA-256) sent from the client. If the password is correct, it will be successfully decrypted and the full hash will be compared.

#### Password / token authentication
The API is going to be available to any application that's running on the host and without encryption, it's always going to be insecure. So this password / token authentication is implemented to secure that. It is expected that the client won't leak its access token.

#### Security
The added security is currently easily broken since everything is handled via HTTP. For that reason, it is hardcoded that requests can only be made from localhost or 127.0.0.1. Using the API outside of localhost should be handled via reverse proxy or a similar solution.

## Type of Change

## Checklist


ReenigneArcher:
@TheChoconut the linter is failing. The artifacts include the modified files and are available here: https://github.com/SunshineStream/Sunshine/actions/runs/1878913676

ReenigneArcher:
Additionally, I'm not able to load the UI on Windows (haven't tested other OSes). Nothing is printed to console to indicate what the problem may be.

image

TheChoconut:
The UI is removed, it cannot be accessed through the endpoint. The only available endpoint is the API. The idea is to host the UI somewhere else. Currently trying to work out using [Tauri](https://tauri.studio/) to create a simple native-like app that hosts native webview that would display the UI. Currently, it still doesn't allow making API requests with a self-signed certificate but maybe a rust binding could be done... Maybe someone else has more experience in this?

ReenigneArcher:
> The UI is removed, it cannot be accessed through the endpoint. The only available endpoint is the API. The idea is to host the UI somewhere else. Currently trying to work out using [Tauri](https://tauri.studio/) to create a simple native-like app that hosts native webview that would display the UI. Currently, it still doesn't allow making API requests with a self-signed certificate but maybe a rust binding for fetch API could be done... Maybe someone else has more experience in this?

Ah, okay. That makes sense. Looking forward to see more progress on this!

ReenigneArcher:
This probably won't be included in v0.14.0, but I would like to work on the new UI relatively soon after that. Can you fix up these issues? May want to wait until after v0.14.0 to do a rebase.

Some of the linux builds are failing due to missing boost-locale dependencies (this may be a non factor after v0.14.0). Also, some clang lint errors (you can download these artifacts to get the automatically corrected files).

ReenigneArcher:
If you have a change of opinion on the signing the CLA please let me know and I will re-open the PR.

PR #66, opened 2022-02-24
Initial support for MacOS

abusse:
## Description

This patch set introduces initial support for MacOS as third major host platform. It relies on the VideoToolbox framework for audio and video processing, which enables hardware accelerated processing of the stream on most platforms. Audio capturing requires third party tools as MacOS does not offer the recording of the audio output like the other platforms do. The commit enables most features offered by Sunshine for MacOS with the big exception of gamepad support. The patch sets was tested by a few volunteers, which allowed to remove some of the early bugs. However, several bugs especially regarding corner cases have probably not surfaced yet. 

Besides instructions how to build from source, the commit also adds a Portfile that allows a more easy installation. After available on the release branch, a pull request for the Portfile in the MacPorts project is planned.

## Type of Change

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,22 @@ if(WIN32)
 		PQOS_FLOWID=UINT32*
 		QOS_NON_ADAPTIVE_FLOW=2)
 endif()
+if(APPLE)
+	macro(ADD_FRAMEWORK fwname appname)
+    find_library(FRAMEWORK_${fwname}
+    NAMES ${fwname}
+    PATHS ${CMAKE_OSX_SYSROOT}/System/Library
+    PATH_SUFFIXES Frameworks
+    NO_DEFAULT_PATH)
+    if( ${FRAMEWORK_${fwname}} STREQUAL FRAMEWORK_${fwname}-NOTFOUND)
+        MESSAGE(ERROR ": Framework ${fwname} not found")
+    else()
+        TARGET_LINK_LIBRARIES(${appname} "${FRAMEWORK_${fwname}}/${fwname}")
+        MESSAGE(STATUS "Framework ${fwname} found at ${FRAMEWORK_${fwname}}")
+    endif()
+	endmacro(ADD_FRAMEWORK)
+endif()
+
 add_subdirectory(third-party/moonlight-common-c/enet)
 add_subdirectory(third-party/Simple-Web-Server)
 
@@ -23,7 +39,9 @@ include_directories(third-party/miniupnp)
 
 find_package(Threads REQUIRED)
 find_package(OpenSSL REQUIRED)
-set(Boost_USE_STATIC_LIBS ON)
+if(NOT APPLE)
+	set(Boost_USE_STATIC_LIBS ON)
+endif()
 find_package(Boost COMPONENTS log filesystem REQUIRED)
 
 list(APPEND SUNSHINE_COMPILE_OPTIONS -Wall -Wno-missing-braces -Wno-maybe-uninitialized -Wno-sign-compare)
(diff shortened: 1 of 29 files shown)


assets/apps_mac.json
```diff
@@ -0,0 +1,6 @@
+{
+	"env":{
+		"PATH":"$(PATH):$(HOME)/.local/bin"
+	},
+	"apps":[ ]
```
ReenigneArcher, line 5: - Is it possible to add a Steam BigPicture placeholder, like we do for Windows and Linux?
- Does the Desktop show up by default, like on Windows? Or do we need to add it like on Linux?

assets/info.plist
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleIdentifier</key>
```
ReenigneArcher, line 5: Reduce indent

assets/info.plist
```diff
@@ -0,0 +1,12 @@
...
+    <string>com.mycorp.sunshine</string>
```
ReenigneArcher, line 6: maybe change to `com.github.sunshinestream.sunshine` ?

sunshine/video.cpp
```diff
@@ -538,13 +538,49 @@ static encoder_t vaapi {
...
+  DEFAULT,
+  
+  nullptr
+};
+#endif
+
 static std::vector<encoder_t> encoders {
+#ifndef __APPLE__
   nvenc,
+#endif
```
ReenigneArcher, line 575: Is this removing nvenc from windows and linux?

abusse: No, it's the other way around, it enables it for all platforms except MacOS

assets/info.plist
abusse: ack

assets/info.plist
abusse: it's a tap, so I assume you mean two spaces instead?

assets/apps_mac.json
abusse: I have to research that. Currently the implementation only supports streaming the entire desktop of the primary screen. So that answers also the second question, the desktop will show up by default.

assets/info.plist
abusse: made it two spaces...

ReenigneArcher: > it's a tap, so I assume you mean two spaces instead?

Just thought the tags within the `dict` tag should be consistent. Probably better if they are all tabs (or 4 spaces).

abusse: looks like I was a little faster 😅 Since we have two spaces everywhere for indentation, I assumed that would be right here as well

ReenigneArcher: That works! Thanks!

CMakeLists.txt
```diff
@@ -352,7 +410,6 @@ list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
 		libminiupnpc-static
 		${CBS_EXTERNAL_LIBRARIES}
 		${CMAKE_THREAD_LIBS_INIT}
-		stdc++fs
```
abusse, line 355: hum...can someone check if this is the reason, why the Linux and Windows builds fail?

ReenigneArcher: What is interesting is the AppImage is building. Based on that I'm guessing it would still build on the Ubuntu/Debian Dockerfiles would still build. I have fast fail on for the Dockerfiles build job.

CMakeLists.txt
```diff
@@ -23,7 +39,9 @@ include_directories(third-party/miniupnp)
 
 find_package(Threads REQUIRED)
 find_package(OpenSSL REQUIRED)
-set(Boost_USE_STATIC_LIBS ON)
```
ReenigneArcher, line 26: I'm wondering if this is the reason for the build failures. There's a reference to `boost` directly before the error message.

https://github.com/SunshineStream/Sunshine/runs/5325277331?check_suite_focus=true#step:4:2515
https://github.com/SunshineStream/Sunshine/runs/5325278122?check_suite_focus=true#step:4:3114

CMakeLists.txt
abusse: The strange thing is that this is the only change that applies also to none Apple platforms. Everything else should only be active with the MacOS build. I think I have to setup a Linux VM on the weekend and debug that in more detail.

TheElixZammuto: My only guess is that `if(!APPLE)` is not working as it should and Boost is trying to linking itself as a dynamic library instead of a static one.  Isn't the correct definition `if(NOT APPLE)` ?

ReenigneArcher: > My only guess is that `if(!APPLE)` is not working as it should and Boost is trying to linking itself as a dynamic library instead of a static one. Isn't the correct definition `if(NOT APPLE)` ?
> 
> EDIT: At least according to https://cmake.org/cmake/help/v3.3/command/if.html

I agree although not an expert on CMake, I am basing my opinion on the fact the errors seem to be related to boost.

abusse: Since we also use `NOT` instead of `!` further down below in the file, I would say it is safe to assume that `NOT` is the right way to do it. I will push an update in a second.

abusse: Looks like that did it 😊

ReenigneArcher (approved):

ReenigneArcher:
@TheElixZammuto you good with this one? Okay to make a release after this? v0.13.0 or v0.12.1?

TheElixZammuto:
I don't have any macOS Machine, so I cannot test this. If someone has already tested this and can confirm it does work I'm more than happy to merge this. I guess 0.13.0 is the best option at the moment

ReenigneArcher:
> I don't have any macOS Machine, so I cannot test this. If someone has already tested this and can confirm it does work I'm more than happy to merge this. I guess 0.13.0 is the best option at the moment

Okay, sounds good. I'm more worried about anything breaking on Win/Linux. I'll give the windows build a test. Going to be a little bit before I can test things on linux side (been having trouble getting it to work in a vm).

I believe there's been quite a few testers of the original mac branch, here's the original discussion about it as well (https://github.com/loki-47-6F-64/sunshine/issues/127).

PR #77, opened 2022-02-27
Intel QuickSync support

caioavidal:
## Description

- Updated prebuilt ffmpeg library
- Add qsv encoder support

I have been trying to implement intel quicksync encoder to this project. The encoder itself seems to be working fine as I tested outputting the screen captured to a file. The problem is that moonlight seems to not recognize IDR frames. It waits forever for it and then disconnects. 
I'm attaching here the output from ffprobe.

Output from Moolight QT:
00:00:19 - SDL Info (0): FFmpeg-based video decoder chosen
00:00:19 - SDL Info (0): Dropping window event during flush: 6 (1920 1080)
00:00:20 - SDL Info (0): Received first video packet after 2100 ms
00:00:20 - SDL Info (0): Waiting for IDR frame
00:00:20 - SDL Info (0): Waiting for IDR frame
00:00:20 - SDL Info (0): Waiting for IDR frame

I have tested on moonlight android, ios and windows app and all of them once connected are giving black screen.

Please let me know what you guys think.

## Type of Change
[frame_qsv.txt](https://github.com/SunshineStream/Sunshine/files/8149518/frame_qsv.txt)


CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ if(WIN32)
 	enable_language(RC)
 	set(CMAKE_RC_COMPILER windres)
 	file(
-		DOWNLOAD "https://github.com/TheElixZammuto/sunshine-prebuilt/releases/download/1.0.0/pre-compiled.zip" "${CMAKE_CURRENT_BINARY_DIR}/pre-compiled.zip"
+		DOWNLOAD "https://github.com/caioavidal/sunshine-prebuilt/releases/download/1.0.0/pre-compiled.zip" "${CMAKE_CURRENT_BINARY_DIR}/pre-compiled.zip"
 		TIMEOUT 60
 		EXPECTED_HASH SHA256=5d59986bd7f619eaaf82b2dd56b5127b747c9cbe8db61e3b898ff6b485298ed6)
```
ReenigneArcher, line 55: Please adjust... See here:

https://github.com/SunshineStream/Sunshine/pull/50/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL35-R37

third-party/cbs/cbs.c
```diff
@@ -1,1050 +1,1050 @@
-/*
```
ReenigneArcher, line 1: If the file is not needed, delete it.

sunshine/video.cpp
```diff
@@ -575,6 +602,7 @@ static std::vector<encoder_t> encoders {
 #endif
 #ifdef _WIN32
   amdvce,
+  quicksync,
```
ReenigneArcher, line 605: Would this encoder work on Linux and/or Mac as well?

ReenigneArcher (changes requested):
In addition to the other comments... does anything need to change in the example conf file? Anything in the webui in order to enable this encoder when not using automatic mode?

third-party/cbs/cbs.c
caioavidal: Deleted

sunshine/video.cpp
caioavidal: Not sure, only tested on Windows. I can test it on linux and mac once I get this working well

CMakeLists.txt
caioavidal: I've added web ui tab for quicksync and also added to encoder's select. for now I need to keep this Download url because the new one from @PapyKahan doesn't have libmfx.

assets/sunshine.conf
```diff
@@ -1,318 +1,30 @@
-# If no external IP address is given, Sunshine will attempt to automatically detect external ip-address
```
ReenigneArcher, line 1: can't push your personal conf file to the project

assets/web/config.html
```diff
@@ -629,6 +625,35 @@ <h1 class="my-4">Configuration</h1>
...
+            <option value="7">Very Fast</option>
+            <option value="6">Faster</option>
+            <option value="5">Fast</option>
+            <option value="4">Medium</option>
+            <option value="3">Slow</option>
+            <option value="2">Slower</option>
+            <option value="1">Very Slow</option>
+          </select>
+        </div>
+      </div>
```
ReenigneArcher, line 644: @TheElixZammuto seems for every encoder we add a new setting for preset... can we have a single setting for preset that applies to whatever encoder is selected? I believe the possible values are the same for all of them.

CMakeLists.txt
ReenigneArcher: Okay, that's fine for now

assets/sunshine.conf
caioavidal: yes, noticed now. reverting

assets/web/config.html
TheElixZammuto: Unfortunately this is not the case since x264,amf, nvenc and qsv all have different preset options

third-party/cbs/cbs.c
TheElixZammuto: @ReenigneArcher why was this file deleted? As of https://github.com/SunshineStream/Sunshine/blob/master/third-party/cbs/cbs.c the file is present inside the repository and it will be removed by this PR

TheElixZammuto:
My tests unfortunately were unfruitful:

`[2022:02:28:21:02:29]: Info: Trying encoder [quicksync]
[2022:02:28:21:02:29]: Info:
Device Description : Intel(R) HD Graphics 5500
Device Vendor ID   : 0x00008086
Device Device ID   : 0x00001616
Device Video Mem   : 128 MiB
Device Sys Mem     : 0 MiB
Share Sys Mem      : 4036 MiB
Feature Level      : 0x0000B100
Capture size       : 1920x1080
Offset             : 0x0
Virtual Desktop    : 1920x1080
[2022:02:28:21:02:29]: Warning: Failed to set realtime GPU priority. Please run application as administrator for optimal performance.
[2022:02:28:21:02:29]: Info: Color coding [Rec. 601]
[2022:02:28:21:02:29]: Info: Color range: [JPEG]
[2022:02:28:21:02:30]: Error: First packet type is not an IDR frame
[2022:02:28:21:02:30]: Info:
Device Description : Intel(R) HD Graphics 5500
Device Vendor ID   : 0x00008086
Device Device ID   : 0x00001616
Device Video Mem   : 128 MiB
Device Sys Mem     : 0 MiB
Share Sys Mem      : 4036 MiB
Feature Level      : 0x0000B100
Capture size       : 1920x1080
Offset             : 0x0
Virtual Desktop    : 1920x1080
[2022:02:28:21:02:30]: Warning: Failed to set realtime GPU priority. Please run application as administrator for optimal performance.
[2022:02:28:21:02:30]: Info: Color coding [Rec. 601]
[2022:02:28:21:02:30]: Info: Color range: [JPEG]
[2022:02:28:21:02:30]: Error: First packet type is not an IDR frame
[2022:02:28:21:02:30]: Info:
Device Description : Intel(R) HD Graphics 5500
Device Vendor ID   : 0x00008086
Device Device ID   : 0x00001616
Device Video Mem   : 128 MiB
Device Sys Mem     : 0 MiB
Share Sys Mem      : 4036 MiB
Feature Level      : 0x0000B100
Capture size       : 1920x1080
Offset             : 0x0
Virtual Desktop    : 1920x1080`

Windows 11 build 22000, Intel Core i5 5005U, Driver Version 20.19.15.5063

sunshine/video.cpp
```diff
@@ -474,6 +475,34 @@ static encoder_t amdvce {
...
+
+static encoder_t quicksync {
+  "quicksync"sv,
+  { FF_PROFILE_H264_HIGH, FF_PROFILE_HEVC_MAIN },
+  AV_HWDEVICE_TYPE_QSV,
+  AV_PIX_FMT_QSV,
+  AV_PIX_FMT_NV12, AV_PIX_FMT_P010,
+  {
+    {
+      { "forced-idr"s, 1 },
```
TheElixZammuto, line 488: According to https://stackoverflow.com/questions/66921104/how-to-force-qsv-generate-idr-when-using-qsv-encoding-in-ffmpeg  and https://gist.github.com/nico-lab/597edc466ec54d0b6ce9dbf263da766a , the option is actually `forced_idr`

third-party/cbs/cbs.c
caioavidal: Because new ffmpeg library already has cbc.c implementation. I had to update ffmpeg and also added libmfx for qsv support

sunshine/video.cpp
```diff
@@ -1810,6 +1859,8 @@ platf::mem_type_e map_dev_type(AVHWDeviceType type) {
     return platf::mem_type_e::dxgi;
   case AV_HWDEVICE_TYPE_VAAPI:
     return platf::mem_type_e::vaapi;
+  case AV_HWDEVICE_TYPE_QSV:
```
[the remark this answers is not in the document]
caioavidal: Yes, I have tried that approach but with no success. The only way I could have this working was copying to system ram. I will investigate it later. I'm now focusing on making it work

sunshine/video.cpp
caioavidal: Yes, but have tried forced_idr as well. Actually I could manage it to send only idr frames but then getting:
00:10:10 - SDL Info (0): Unrecoverable frame 740637842: lost FEC blocks 1 to 3
00:10:10 - SDL Info (0): Unrecoverable frame -1801679895: lost FEC blocks 1 to 3
00:10:10 - SDL Info (0): Unrecoverable frame 1140981760: 0+2=2 received < 659 needed

I have tried many options to solve this "Waiting for IDR frame" error but none of them seemed to work

cgutman: If you capture the raw encoder output to a file, I can take a look at the bitstream and figure out what's going on.

Does it affect both H.264 and HEVC?

sunshine/video.cpp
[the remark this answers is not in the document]
cgutman: It looks like hwcontext_qsv.c can do it, but you might need to derive your QSV hwcontext from a D3D11VA hwcontext in order for it to work.

After that, should be as simple as taking the D3D11VA frame we fake up [here](https://github.com/SunshineStream/Sunshine/blob/4b658cd86ba0b7520f6f7cb5b5f07d6775845a81/sunshine/platform/windows/display_vram.cpp#L411-L421) and doing:
```
AVFrame* qsv_frame = av_frame_alloc();
qsv_frame->format = AV_PIX_FMT_QSV;
av_hwframe_map(qsv_frame, dxgi_frame, AV_HWFRAME_MAP_READ);
```

sunshine/video.cpp
caioavidal: @cgutman yep, both of them. I can send you the file

caioavidal: not sure it would help you but I was comparing software encode bitstream against qsv and it looks like nal_ref_idc is different. For instance SPS, PPS and IDR are 1 on qsv while it is 3 on software bitstream.

caioavidal: ok I finally got video and frames are running fine on moonlight
I basically needed to do this:

```
    payload_new = replace(payload, "\000\000\000\001\'"sv, "\000\000\000\001g"sv);
    payload     = { (char *)payload_new.data(), payload_new.size() };

    payload_new = replace(payload, "\000\000\000\001("sv, "\000\000\000\001h"sv);
    payload     = { (char *)payload_new.data(), payload_new.size() };

    payload_new = replace(payload, "\000\000\001%"sv, "\000\000\001e"sv);
    payload     = { (char *)payload_new.data(), payload_new.size() };
```

It replaces the nal_ref_idr of key frames to 3

cgutman: Very interesting. Old versions of Moonlight did care about `nal_ref_idc` (in violation of the H.264 and HEVC specification), but recent versions (within the last year or so) shouldn't care. The fix was https://github.com/moonlight-stream/moonlight-common-c/commit/b528867466fe780aa7b3f5309f4fd8dfa3b06765

I wonder if what's going on is that Sunshine itself is breaking the bitstream (or rather the fixups it's trying to perform aren't working because the original payload doesn't have the expected bytes).

cgutman: I looked at output.h264 and it looks like the issue is that the NALU for the IDR image has the 3-byte Annex B start code rather than the 4-byte Annex B start code that Moonlight expects for the start of a frame (and which is used for the SPS, PPS, and SEI NALUs).

image

I think using a 3-byte start code is legal, but just changing moonlight-common-c to accept would expose bugs in some clients (iOS at least) that assume the NALU start prefixes are always 4 bytes long. Handling both 3 and 4 byte start prefixes would complicate the logic we use on iOS to convert from Annex B to AVCC format. I think it makes sense to continue to fix this up on the Sunshine side, but we can make the logic more robust to avoid getting confused if the first NALU byte varies between decoders.

caioavidal: yeah, makes total sense to me. thanks for you support.

sunshine/video.cpp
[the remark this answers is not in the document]
caioavidal: Trying to implement that but getting segmentation fault on av_hwframe_map. I noticed qsv child device is using D3D9Ex. I'm trying to fix that.

[WARNING: defaulting child_device_type to AV_HWDEVICE_TYPE_DXVA2 for compatibility with old commandlines. This behaviour will be removed in the future. Please explicitly set device type via "-init_hw_device" option.
2022:03:01:18:33:06]: Info: Color coding [Rec. 601]
[2022:03:01:18:33:06]: Info: Color range: [JPEG]
[AVHWDeviceContext @ 0000024400399980] Using D3D9Ex device.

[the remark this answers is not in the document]
caioavidal: Tried to derive from D3D11VA but unfortunately getting
[AVHWDeviceContext @ 000001a860c0b3c0] Error setting child device handle: -16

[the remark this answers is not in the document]
caioavidal: @cgutman just posted a question on stackoverflow. If you have any clue on that pls let me know. thanks
https://stackoverflow.com/questions/71356395/derive-qsv-hwdevice-from-d3d11va-hwdevice-using-ffmpeg

caioavidal:
> My tests unfortunately were unfruitful: https://gist.github.com/TheElixZammuto/da1f117f1b01003356cffa5f2833bb79
> 
> Windows 11 build 22000, Intel Core i5 5005U, Driver Version 20.19.15.5063

Hey @TheElixZammuto. Could you try again? We have new commits now.

TheElixZammuto:
So, I did some tests on the same machine as above, connected on Ethernet to the client.
While now Intel Codec gets detected, performance is definetly not on par with the AMD encoder, and more similar to the software encoder (in terms of latency)

2022-03-20 18_23_21-Gestione attività
This is he resource consumption of Sunshine on the Intel machine
2022-03-20 18_24_35-Moonlight
And this is the consumption of Sunshine on the AMD machine.

My gut is telling me that some calculations are still done on the CPU rather than on GPU (due to the much higher memory usage), but I'm not sure of what it could be (but I would guess it's something related to [this](https://github.com/SunshineStream/Sunshine/pull/77/files#diff-b755390cb4f797659bdce2e4e7332af74197ed07d9f5ebe7b30939db1bd72ef9R1886)?

caioavidal:
> So, I did some tests on the same machine as above, connected on Ethernet to the client. While now Intel Codec gets detected, performance is definetly not on par with the AMD encoder, and more similar to the software encoder (in terms of latency)
> 
> 2022-03-20 18_23_21-Gestione attività This is he resource consumption of Sunshine on the Intel machine 2022-03-20 18_24_35-Moonlight And this is the consumption of Sunshine on the AMD machine.
> 
> My gut is telling me that some calculations are still done on the CPU rather than on GPU (due to the much higher memory usage), but I'm not sure of what it could be (but I would guess it's something related to [this](https://github.com/SunshineStream/Sunshine/pull/77/files#diff-b755390cb4f797659bdce2e4e7332af74197ed07d9f5ebe7b30939db1bd72ef9R1886)?

Yes. this is due to scale and hwframe_transfer_data. I'm trying to implement it using d3d11 device. Getting device failed from mfx lib while trying to encode frames.

ReenigneArcher:
> EDIT: Probably we need FFMPEG 5.0 for this

In that case, we probably need to get ffmpeg 5.0 done in a separate PR. I'm told that using cpack will include the dependencies for the user, so that will simplify build/versus run time dependencies.

Maybe we address ffmpeg 5.0 after this? #139

Saijin-Naib:
Is there still movement on this front (pending major FFMPEG upgrade)? If so, I'd like to help test as I can since I am using Intel/QuickSync on Linux, and plan on getting the Intel Arc dGPUs as soon as they launch.

ReenigneArcher:
> Is there still movement on this front (pending major FFMPEG upgrade)? If so, I'd like to help test as I can since I am using Intel/QuickSync on Linux, and plan on getting the Intel Arc dGPUs as soon as they launch.

ffmpeg still needs to be upgraded (on all platforms/builds)...

YTOAA:
Intel qsv encoding is not yet supported in Windows.

ReenigneArcher:
> Intel qsv encoding is not yet supported in Windows.

This was PR was never completed or merged.

YTOAA:
> > Intel qsv encoding is not yet supported in Windows.
> 
> This was PR was never completed or merged.

Do you think it's going to be hard to see this feature this year?

ReenigneArcher:
We're working to update to ffmpeg5. From my understanding quick sync won't be very useful for sunshine because it's slower.

YTOAA:
> We're working to update to ffmpeg5. From my understanding quick sync won't be very useful for sunshine because it's slower.

However, Intel graphics encoding using vaapi on Linux was quite satisfying to me.

caioavidal:
> We're working to update to ffmpeg5. From my understanding quick sync won't be very useful for sunshine because it's slower.

It might be faster than sw encoding at least

aandaluz:
Devs are now upgrading ffmpeg to 5.x series in https://github.com/LizardByte/Sunshine/pull/509

PR #108, opened 2022-03-20
WIP: Disconnect Commands 

8run0:
## Description
Additional functionality to have a list of commands that can be run when the termination of a stream is issued.

### Screenshot

disconnectcmds

## Type of Change

## Checklist


sunshine/process.cpp
```diff
@@ -182,7 +184,27 @@ void proc_t::terminate() {
...
+
+for(; _undo_it != std::end(proc.disconnect_cmds); ++_undo_it) {
+    auto &cmd = _undo_it->do_cmd;
+
+    BOOST_LOG(info) << "Executing: ["sv << cmd << ']';
+    auto ret = exe(cmd, _env, _pipe, ec);
+
+    if(ec) {
+      BOOST_LOG(error) << "Couldn't run ["sv << cmd << "]: System: "sv << ec.message();
+      return 0;
```
ReenigneArcher, line 200: `return false;`

sunshine/process.cpp
```diff
@@ -182,7 +184,27 @@ void proc_t::terminate() {
...
+    }
+
+    if(ret != 0) {
+      BOOST_LOG(error) << '[' << cmd << "] failed with code ["sv << ret << ']';
+      return 0;
```
ReenigneArcher, line 205: `return false;`

sunshine/process.cpp
```diff
@@ -147,6 +147,8 @@ int proc_t::running() {
 
 void proc_t::terminate() {
   std::error_code ec;
+  int app_id = _app_id;
+
```
ReenigneArcher, line 151: why not just use `_app_id` instead of creating a new variable?

sunshine/process.cpp
ReenigneArcher: also, should there be a `return` at all within a for loop?

sunshine/process.cpp
```diff
@@ -182,7 +184,27 @@ void proc_t::terminate() {
       std::abort();
     }
   }
+ auto &proc = _apps[app_id];
```
ReenigneArcher, line 188: `auto &proc = _apps[_app_id];`

ReenigneArcher (changes requested):

sunshine/process.cpp
8run0: Removed returns.

sunshine/process.cpp
8run0: as on line 156 _app_id is set to -1 this was to prevent any side effects from original behaviour

sunshine/process.cpp
8run0: _app_id is -1 at this point.

sunshine/process.cpp
8run0: removed

8run0:
Merge nightly.

assets/apps_linux.json
```diff
@@ -7,7 +7,10 @@
 		"name":"Low Res Desktop",
 		"prep-cmd":[
 		{ "do":"xrandr --output HDMI-1 --mode 1920x1080", "undo":"xrandr --output HDMI-1 --mode 1920x1200" }
```
ReenigneArcher, line 9: can you indent this line?

assets/apps_linux.json
```diff
@@ -7,7 +7,10 @@
...
-		]
+		],
+		"disconnect-cmd":[
+			{ "do":"xrandr --output HDMI-1 --mode 1920x1080", "undo":"xrandr --output HDMI-1 --mode 1920x1200" }
+			]
```
ReenigneArcher, line 13: un-indent this line?

ReenigneArcher (changes requested):
A few more small tweaks if you don't mind.

salty2011:
@8run0 just trying to get some better context to your feature. Is this proposed feature for linux specifically or across platforms?

Asking because this capability did exist at least from a windows perspective.

ReenigneArcher:
> @8run0 just trying to get some better context to your feature. Is this proposed feature for linux specifically or across platforms?
> 
> Asking because this capability did exist at least from a windows perspective.

By the way @8run0 , I have requested additional users to review PRs in the moonlight discord and @salty2011 volunteered.

README.md
```diff
@@ -240,6 +240,13 @@ All shortcuts start with CTRL + ALT + SHIFT, just like Moonlight
 		- undo <optional>: Run after the application has terminated
 			- This should not fail considering it is supposed to undo the 'do' commands.
 			- If it fails, Sunshine is terminated
+	- disconnect-cmd: A list of commands to be run once the stream has started disconnecting
```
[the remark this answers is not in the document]
ReenigneArcher: My previous comment no longer applies as readme has been refactored and documentation moved to docs; however documentation likely still needs to be updated.

KuleRucket:
As far as I can work out the 4 are:

- prep do - on connect (A)
- prep undo - to revert on failure to connect (B)
- disconnect do - on disconnect (C)
- disconnect undo - on failure to disconnect (D)

It seems to me that there are only 2 states you would ever want to be in which are when you are connected via a client and when you are not. When you are connected you might want to run xrandr to set the desktop display to match your client. When you are disconnected you might want the desktop resolution to match your local monitor.

If that's true then for (A) and (C) you want the disconnected state and for (B) and (D) you want the connected state so you would always set A==C and B==D, so you would not really need the 4 different states.

Why not just run the existing "prep - undo" commands when you disconnect as well as when you fail to connect? This would avoid the need to introduce new configuration.

ReenigneArcher:
@KuleRucket thanks for digging into this one. 

Perhaps there could some more advanced use cases besides setting resolutions. Maybe a user wants to copy a file to another location, send a webhook, etc.

The contributor hasn't been responding on here or on discord so not sure we're going to get this one merged either way.

PR #113, opened 2022-03-28
Update scripts readme to mention submodules

JeremyKennedy:
## Description

The build script requires the filesystem has the submodules pulled locally, but the default `git clone` command does not pull these. I've added a note in the scripts readme reminding users to do this.

## Type of Change

This is just a small documentation update.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -15,6 +15,9 @@ These scripts use docker under the hood, as such, they can only be used to compi
 sudo apt install docker
 ```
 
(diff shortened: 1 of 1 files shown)


HomerSp:
The Compilation steps in the readme do have --recurse-submodules in the git clone command, however, it probably can't hurt to have it separately as well. The bigger issue seems to be that the readme still uses loki's repository in some places, but those could/should probably be updated in a separate PR.

ReenigneArcher:
> The Compilation steps in the readme do have --recurse-submodules in the git clone command, however, it probably can't hurt to have it separately as well. The bigger issue seems to be that the readme still uses loki's repository in some places, but those could/should probably be updated in a separate PR.

Most likely I'll remove that stuff from the readme anyway and just have it in the wiki.

ReenigneArcher:
Superseded by #134

PR #125, opened 2022-04-10
AVPacket Fix

### LINKED ISSUE #124: Segfault in video::encode
### Describe the Bug

I am getting a segfault when trying to run the app.
Stacktrace:
```gdb
#0  0x00007ffff6675967 in av_buffer_unref () at /usr/lib/libavutil.so.57
#1  0x00007ffff69cdeb8 in av_packet_unref () at /usr/lib/libavcodec.so.59
#2  0x00007ffff6b0e29b in avcodec_receive_packet () at /usr/lib/libavcodec.so.59
#3  0x00005555557162a3 in video::encode(long, video::session_t&, AVFrame*, std::shared_ptr<safe::post_t<safe::queue_t<std::unique_ptr<video::packet_raw_t, std::default_delete<video::packet_raw_t> > > > >&, void*) ()
#4  0x0000555555718ec6 in video::validate_config(std::shared_ptr<platf::display_t>&, video::encoder_t const&, video::config_t const&) ()
#5  0x000055555571964e in video::validate_encoder(video::encoder_t&) ()
#6  0x000055555571abee in video::init() ()
#7  0x0000555555639dab in main ()
```

### Expected Behavior

No crash

### Additional Context

packet.get()->buf is == 0 at this point.

Manjaro Linux (Kernel 5.17)

### Architecture

x64

### Sunshine Version

Nightly

### GPU Type

AMD

### GPU Model

6900x

### GPU Driver/Mesa Version

mesa-git

(KVM?)

Logical-sh:
## Description
This PR changes the codebase to wrap AVPacket instead of extending it.

According to ffmpeg's documentation: 

>sizeof(AVPacket) being a part of the public ABI is deprecated. once av_init_packet() is removed, new packets will only be able to be allocated with [av_packet_alloc()](https://ffmpeg.org/doxygen/trunk/group__lavc__packet.html#gaaf85aa950695631e0217a16062289b66), and new fields may be added to the end of the struct with a minor bump.

I believe this to mean extending AVPacket is fragile and prone to break if things are changed in ffmpeg

In this case, the extended struct did not clear the memory for AVPacket.opaque_ref, so it was not guaranteed to be correctly set to nullptr.

That meant that when encode was called it would try to free a random pointer (since opaque_ref is a AVBuffer) and segfault.
While the issue in question could have been solved by setting that field to nullptr in the constructor, I though changing the codebase to use AVPacket the proper way was the correct move here instead to increase the resiliency of this code.

I have tested this with vaapi encoder with no issues in video (I am having a separate audio issue I plan on looking at) but do not have any other encoders setup to test. (Software encoding is giving me a separate problem, already documented as an issue.)

Note that my C++ is rough, but I believe this should still keep all the correct memory ref/ownership to prevent leaks (and I did not see an increase in memory usage after letting a video steam for some time.)

### Issues Fixed or Closed
- Fixes #124

## Type of Change

## Checklist

### DIFF
--- a/sunshine/stream.cpp
+++ b/sunshine/stream.cpp
@@ -895,7 +895,8 @@ void videoBroadcastThread(udp::socket &sock) {
     auto session = (session_t *)packet->channel_data;
     auto lowseq  = session->video.lowseq;
 
-    std::string_view payload { (char *)packet->data, (size_t)packet->size };
+    auto av_packet = packet->av_packet;
+    std::string_view payload { (char *)av_packet->data, (size_t)av_packet->size };
     std::vector<uint8_t> payload_new;
 
     auto nv_packet_header = "\0017charss"sv;
@@ -904,7 +905,7 @@ void videoBroadcastThread(udp::socket &sock) {
 
     payload = { (char *)payload_new.data(), payload_new.size() };
 
-    if(packet->flags & AV_PKT_FLAG_KEY) {
+    if(av_packet->flags & AV_PKT_FLAG_KEY) {
       for(auto &replacement : *packet->replacements) {
         auto frame_old = replacement.old;
         auto frame_new = replacement._new;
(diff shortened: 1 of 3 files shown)


comwizz2:
I redid the PR to fix an extraneous debugging include (iostream) and to properly apply clang-format!

TheElixZammuto:
Working as intended on libx264 and AMF on Windows, but I would wait for some NVEnc feedback before merging

TheElixZammuto (changes requested):
I would wait some feedback from NVEnc before merging this

Logical-sh:
Rebased to fix the merge conflict with the new try catch blocks.

KuleRucket:
I have tested this PR using nvenc (with nvfbc) and it is working fine for me. I don't see any difference in the logs and behaviour compared to the nightly branch.

PR #139, opened 2022-04-23
WIP: Moving to cpack in order to unify installers across all platforms

### LINKED ISSUE #61: Consider new packaging strategy
### Describe the Bug

Currently gen-deb.in is the only method of packaging, which seems to be a manually packaged deb file.

### Expected Behavior

Leverage cmake built in methods to create packages where possible:

- bundle (mac): https://cmake.org/cmake/help/latest/cpack_gen/bundle.html
- deb (debian/linux): https://cmake.org/cmake/help/latest/cpack_gen/deb.html
- dmg (mac): https://cmake.org/cmake/help/latest/cpack_gen/dmg.html
- rpm (redhat/linux): https://cmake.org/cmake/help/latest/cpack_gen/rpm.html

### Additional Context

Full list of cpack generators: https://cmake.org/cmake/help/latest/manual/cpack-generators.7.html

na

### Architecture

na

### Sunshine Version

0.12.0

### GPU Type

na

### GPU Model

na

### GPU Driver/Mesa Version

na

_No response_

ABeltramo:
## Description

I'm moving package distribution to `cpack` which will unify the current `.deb` and `.rpm` generation and it'll enable creation of Windows Installers and Mac OSX bundles.  
This should reduce considerably duplication of code.

Currently work in progress, needs testing on real deployments.

### Issues Fixed or Closed

- Fixes #61

## Type of Change

## Checklist


ABeltramo:
I'll modify the docs to reflect the change, if someone wants to try this out it's as simple as:
```bash
mkdir build # I like to put all the generated files inside here, so that we don't clutter the root
cd build
cmake ../
cpack -G DEB # you can also try RPM
```

CMakeLists.txt
```diff
@@ -448,3 +446,76 @@ foreach(flag IN LISTS SUNSHINE_COMPILE_OPTIONS)
...
+set(CPACK_PACKAGE_FILE_NAME "${CMAKE_PROJECT_NAME}_${VERSION}_${CPACK_DEBIAN_PACKAGE_ARCHITECTURE}")
+set(CPACK_STRIP_FILES YES)
+
+# Installation destination dir
+if(UNIX AND NOT APPLE)
+	set(CPACK_SET_DESTDIR true)
+	set(CMAKE_INSTALL_PREFIX "/etc/sunshine")
+endif()
+
+include(CPack)
```
ReenigneArcher, line 521: add new line at end of files

assets/linux-deb/conffiles
```diff
@@ -0,0 +1,2 @@
+/etc/sunshine/sunshine.conf
```
ReenigneArcher, line 1: `sunshine.conf` probably doesn't even need to be included (or even exist in the repo) anymore, as I cleared it out and moved all the documentation to the sphinx docs. What do you think?

CMakeLists.txt
```diff
@@ -448,3 +446,76 @@ foreach(flag IN LISTS SUNSHINE_COMPILE_OPTIONS)
...
+	install(FILES "${SUNSHINE_ASSETS_DIR}/85-sunshine-rules.rules" DESTINATION "/etc/udev/rules.d")
+	install(TARGETS sunshine RUNTIME DESTINATION "/usr/bin")
+	install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sunshine.service" DESTINATION "/usr/lib/systemd/user")
+	install(DIRECTORY "${SUNSHINE_ASSETS_DIR}/shaders/opengl" DESTINATION "shaders")
+
+	# Pre and post install
+	set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA
+		"${SUNSHINE_ASSETS_DIR}/linux-deb/preinst;${SUNSHINE_ASSETS_DIR}/linux-deb/postinst;${SUNSHINE_ASSETS_DIR}/linux-deb/conffiles")
+	set(CPACK_RPM_PRE_INSTALL_SCRIPT_FILE "${SUNSHINE_ASSETS_DIR}/linux-deb/preinst")
+	set(CPACK_RPM_POST_INSTALL_SCRIPT_FILE "${SUNSHINE_ASSETS_DIR}/linux-deb/postinst")
```
ReenigneArcher, line 494: What do you think of using `post_install` and `pre_install` for the file names? I think it might make it more clear what they are, especially for non native English speaking contributors.

Also, any issue with adding `.sh` as the file extensions?

ReenigneArcher (changes requested):
Thanks for working on this! Once you're happy with everything, CI will also need to be updated.

Linux builds: https://github.com/SunshineStream/Sunshine/blob/c4054c75a7fdd0c336db326af5a77171db072e49/.github/workflows/CI.yml#L146

Windows build: https://github.com/SunshineStream/Sunshine/blob/c4054c75a7fdd0c336db326af5a77171db072e49/.github/workflows/CI.yml#L197

The final action in each job expects the release asset to be in the `artifacts` directory.

assets/linux-deb/conffiles
ABeltramo: It's your call, even if it's safe to be removed it could be helpful for people to look at the defaults..

Let me know what you prefer!

CMakeLists.txt
ABeltramo: Those names are tied to the `.deb` format but I can probably rename then in the root and bundling them correctly during the build.

ReenigneArcher: Okay, no problem then. Leaving as is, is okay too.

assets/linux-deb/conffiles
ReenigneArcher: These are the contents of the `sunshine.conf` after I moved everything to the docs.

```txt
# See our documentation at https://sunshinestream.readthedocs.io/en/latest/about/advanced_usage.html
```

ReenigneArcher: Based on our discord conversation, let's leave the `sunshine.conf`.

.github/workflows/CI.yml
```diff
@@ -147,43 +147,29 @@ jobs:
...
-        include:  # package these differently
-          - distro: fedora_35
-            package: '-p'
-            extension: rpm
-
     steps:
       - name: Checkout
         uses: actions/checkout@v3
         with:
           submodules: recursive
```
ReenigneArcher, line 155: Need to install build time dependencies. Can likely combine the the appimage build here as well.

.github/workflows/CI.yml
```diff
@@ -226,23 +213,14 @@ jobs:
       - name: Build Windows
         shell: msys2 {0}
         run: |
+          mkdir artifacts
+
```
ReenigneArcher, line 217: won't need this here

.github/workflows/CI.yml
```diff
@@ -226,23 +213,14 @@ jobs:
...
-          7z a sunshine-windows.zip tools\audio-info.exe
-          7z a sunshine-windows.zip tools\sunshinesvc.exe
-          7z a sunshine-windows.zip ..\tools\install-service.bat
-          7z a sunshine-windows.zip ..\tools\uninstall-service.bat
-          cd ..
-          mkdir artifacts
-          move "sunshine-windows-build\sunshine-windows.zip" "artifacts"
+          cpack
+          
+          cp Sunshine__.exe ../artifacts
```
ReenigneArcher, line 223: Let's keep the package step as we discussed. Would be good to have a portable version as well.

Probably doesn't make much difference, but I think it would be better (slightly faster) to use `move`... also let's rename the file to `sunshine-windows-installer.exe`

Probably makes most sense to have both `cpack` and the `move` in the package step.

```yaml
      - name: Package Windows
        run: |
          cd sunshine-windows-build

          cpack

          del ..\assets\apps_linux.json
          7z a sunshine-windows.zip ..\assets
          7z a sunshine-windows.zip sunshine.exe
          7z a sunshine-windows.zip tools\dxgi-info.exe
          7z a sunshine-windows.zip tools\audio-info.exe
          7z a sunshine-windows.zip tools\sunshinesvc.exe
          7z a sunshine-windows.zip ..\tools\install-service.bat
          7z a sunshine-windows.zip ..\tools\uninstall-service.bat

          mkdir artifacts
          move "sunshine-windows.zip" "../artifacts"
          move "Sunshine__.exe" "../artifacts/sunshine-windows-installer.exe"
```

ReenigneArcher (changes requested):
Getting close!

CMakeLists.txt
```diff
@@ -448,3 +446,119 @@ foreach(flag IN LISTS SUNSHINE_COMPILE_OPTIONS)
...
+# Common options
+set(CPACK_PACKAGE_NAME "SunshineStream")
+set(CPACK_PACKAGE_VENDOR "CMake.org")
+set(CPACK_PACKAGE_CONTACT "https://github.com/SunshineStream/Sunshine")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "https://github.com/SunshineStream/Sunshine")
+set(CPACK_PACKAGE_DESCRIPTION "Gamestream host for Moonlight")
+set(CPACK_PACKAGE_HOMEPAGE_URL "https://github.com/SunshineStream/Sunshine")
+set(CPACK_RESOURCE_FILE_LICENSE ${PROJECT_SOURCE_DIR}/LICENSE)
+set(CPACK_PACKAGE_ICON ${PROJECT_SOURCE_DIR}/sunshine.png)
+set(CPACK_PACKAGE_FILE_NAME "${CMAKE_PROJECT_NAME}_${VERSION}_${CPACK_DEBIAN_PACKAGE_ARCHITECTURE}")
```
ReenigneArcher, line 463: I'm guessing this needs to change to prevent this (same with start menu entry).

image

CMakeLists.txt
```diff
@@ -448,3 +446,119 @@ foreach(flag IN LISTS SUNSHINE_COMPILE_OPTIONS)
...
+# CPACK
+####
+
+# Common options
+set(CPACK_PACKAGE_NAME "SunshineStream")
+set(CPACK_PACKAGE_VENDOR "CMake.org")
+set(CPACK_PACKAGE_CONTACT "https://github.com/SunshineStream/Sunshine")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "https://github.com/SunshineStream/Sunshine")
+set(CPACK_PACKAGE_DESCRIPTION "Gamestream host for Moonlight")
+set(CPACK_PACKAGE_HOMEPAGE_URL "https://github.com/SunshineStream/Sunshine")
```
ReenigneArcher, line 460: `https://sunshinestream.github.io`

.github/workflows/CI.yml
```diff
@@ -218,37 +237,51 @@ jobs:
...
+          # Installers
+          mv Sunshine__.exe artifacts/sunshine-windows-installer.exe
+          
+          # Standalone
+          mv ../assets/web artifacts/web
+          mv tools artifacts/tools
+          mv ../assets/shaders/directx artifacts/shaders/directx
+          mv ../assets/apps_windows.json artifacts/apps_windows.json
+          mv ../assets/sunshine.conf artifacts/sunshine.conf
+          mv sunshine.exe artifacts/sunshine.exe
```
ReenigneArcher, line 272: This won't work like this. We need only 2 files in the artifacts folders.

- installer exe
- zip (portable version as before)

This is because every file single in the assets directory will be uploaded as a release asset.

My code snip before should work (except the mkdir artifacts needed to be moved)

.github/workflows/CI.yml
```diff
@@ -218,37 +237,51 @@ jobs:
...
+      - name: Upload Sunshine Installer
+        if: ${{ github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: sunshine-windows-installer
+          path: artifacts/sunshine-windows-installer.exe
```
ReenigneArcher, line 278: This won't be necessary because just using `path: artifacts` will upload all the files that exist in that folder (this will be the same behavior for the create release action)

.github/workflows/CI.yml
```diff
@@ -218,37 +237,51 @@ jobs:
...
+      - name: Upload Sunshine executable
         if: ${{ github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
         uses: actions/upload-artifact@v3
         with:
-          name: sunshine-windows
-          path: artifacts/
+          name: sunshine-windows-standalone
+          path: artifacts/standalone
```
ReenigneArcher, line 284: Revert this

.github/workflows/CI.yml
```diff
@@ -32,7 +32,6 @@ jobs:
   check_versions:
     name: Check Versions
     runs-on: ubuntu-latest
-    needs: check_changelog
```
ReenigneArcher, line 35: Why was this changed? It's required here.

Check changelog runs first (doesn't run except on PRs or Pulls into master branch). Makes sure that the version in the changelog was updated and is not equal to the latest released version.

Then check versions runs... makes sure that the version number in the CMakeLists.txt was updated.

.github/workflows/CI.yml
```diff
@@ -51,49 +50,73 @@ jobs:
         echo Within 'CMakeLists.txt' change "project(Sunshine VERSION $cmakelists_version)" to "project(Sunshine VERSION ${{ needs.check_changelog.outputs.next_version_bare }})"
         exit 1
 
+  build_linux:
+    name: Build Sunshine on Linux
+    runs-on: ubuntu-20.04
+    needs: [check_versions, check_changelog]
```
ReenigneArcher, line 56: Just `check_changelog`

.github/workflows/CI.yml
```diff
@@ -51,49 +50,73 @@ jobs:
...
+          mkdir -p build
+          cd build
+  
+          cmake "-DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE" "-DSUNSHINE_EXECUTABLE_PATH=$SUNSHINE_EXECUTABLE_PATH" "-DSUNSHINE_ASSETS_DIR=$SUNSHINE_ASSETS_DIR" "-DSUNSHINE_ENABLE_WAYLAND=$SUNSHINE_ENABLE_WAYLAND" "-DSUNSHINE_ENABLE_X11=$SUNSHINE_ENABLE_X11" "-DSUNSHINE_ENABLE_DRM=$SUNSHINE_ENABLE_DRM" "-DSUNSHINE_ENABLE_CUDA=$SUNSHINE_ENABLE_CUDA" "../" -DCMAKE_INSTALL_PREFIX=/usr
+          make -j ${nproc} DESTDIR=AppDir
+
   build_appimage:
     name: AppImage
     runs-on: ubuntu-20.04
-    needs: check_changelog
```
ReenigneArcher, line 57: Needs `check_changelog` as well

ReenigneArcher (changes requested):
There might be more, but this is what I found so far.

.github/workflows/CI.yml
ReenigneArcher: This should be pretty close https://github.com/SunshineStream/Sunshine/pull/139#discussion_r860248861

.github/workflows/CI.yml
ABeltramo: The problem with this is that Github will make one `.zip` file that contains both the `.exe` installer and the `.zip` with the sunshine standalone package resulting in a 20MB zip to be donwloaded that basically contains everything twice just packaged in two different formats (see: https://github.com/ABeltramo/Sunshine/actions/runs/2249492914).

I think it's better to split them in two separate 10MB artifacts so that users can download only the one needed.

ABeltramo:
Current status: 
 - rebased on top of latest `nightly`
 - fixed Github Actions to generate all packages, see: [latest action](https://github.com/SunshineStream/Sunshine/actions/runs/2249589789)
   - Windows installer and standalone zip should be good to go
   - Linux `.deb` and `.rpm` needs to be tested, `appimage` should be left unchanged (unfortunately it's not supported by cpack)
   - OSX bundle `.dmg` can be generated but I can't run Sunshine on my Macbook M1 (segfault), also executing a `.app`, while possible, feels odd because it doesn't open up a terminal automatically like on Windows.
 - Generated packages should be tested on different platforms, volunteers?

This could be reviewed and merged if you are in a rush to release this, but I would like to cleanup `scripts/` to have one single docker container that can build  `.deb` and `.rpm` so that users can still compile without having to install extra packages on the host.  

There's also probably some documentation to be updated, I haven't looked into it yet.

.github/workflows/CI.yml
ReenigneArcher: It only does it for artifacts though, not releases. Users are rarely downloading artifacts.

Just trying to keep the CI more simple as it's already overly complex.

PR #144, opened 2022-04-30
Add cpack

### LINKED ISSUE #61: Consider new packaging strategy

### LINKED ISSUE #174: Add `OPTION+="static_node=uinput"` to linux udev rules
https://github.com/SunshineStream/Sunshine/blob/e7ec6050d9a820549d556bf80e6dcb17460ee996/src_assets/linux/misc/85-sunshine-rules.rules#L1

see here: https://github.com/ValveSoftware/steam-for-linux/issues/4794#issuecomment-327006061

ReenigneArcher:
## Description
- Replaces #139
- Use cpack to generate:
  - Linux
    - deb (tested and working on Ubuntu 20.04 VM with X11)
    - rpm (tested and working on Fedora 36 VM with Xorg)
  - MacOS
    - dmg (needs tested)
    - standalone zip (needs tested)
  - Windows
    - installer (tested and working on Windows 11)
    - standalone zip (tested and working on Windows 11)
- Adjust packaging of Linux AppImage (tested and working on Ubuntu 20.04 VM with X11)
- Reorganize assets

### Issues Fixed or Closed
- Fixes #61 
- Fixes #174 

## Type of Change

## Checklist


ReenigneArcher:
Todo:

- [ ] update documenation
- [ ] test deb on various distros
- [ ] test rpm on various distros
- [ ] test/review windows standalone archive

ReenigneArcher:
@abusse do you mind having a look at this MacOS build? Any idea what I missed? https://github.com/SunshineStream/Sunshine/runs/6243004490?check_suite_focus=true

abusse:
@ReenigneArcher It looks like `libopusenc` is not pulling in `opus`, which is a dependency on my machine. So first thing to try is to add `opus`. Looking further into it, it might even be enough to pull in only `opus`. As far as my quick check showed, we are not using `libopusenc` anywhere (anymore).

abusse:
One more remark: I haven't tested the build on MacOS11 for some time, since I'm only have a MacOS 12 install available rn. I'm not sure if it still builds there.

ReenigneArcher:
> @ReenigneArcher It looks like `libopusenc` is not pulling in `opus`, which is a dependency on my machine. So first thing to try is to add `opus`. Looking further into it, it might even be enough to pull in only `opus`. As far as my quick check showed, we are not using `libopusenc` anywhere (anymore).

Hmm, I think the same issue when using`opus` instead of `libopusenc`. https://github.com/SunshineStream/Sunshine/runs/6248084124?check_suite_focus=true#step:4:399

ReenigneArcher:
Kubuntu 22.04 (and likely every distro) looking for assets in the wrong directory. I probably broke something.

```txt
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot copy file: No such file or directory [/home/runner/work/Sunshine/Sunshine/assets/sunshine.conf] [./sunshine.conf]
Aborted (core dumped)
```

ABeltramo:
> Kubuntu 22.04 (and likely every distro) looking for assets in the wrong directory. I probably broke something.
> 
> ```
> terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
>   what():  filesystem error: cannot copy file: No such file or directory [/home/runner/work/Sunshine/Sunshine/assets/sunshine.conf] [./sunshine.conf]
> Aborted (core dumped)
> ```

This is happening [here](https://github.com/SunshineStream/Sunshine/blob/4b658cd86ba0b7520f6f7cb5b5f07d6775845a81/sunshine/config.cpp#L909) where `SUNSHINE_DEFAULT_DIR` hasn't been set so it'll default to `SUNSHINE_ASSETS_DIR` which by default it's set to `${CMAKE_CURRENT_SOURCE_DIR}/assets` which on the Github runner will resolve to `/home/runner/work/Sunshine/Sunshine/assets/`.

I'll have to check your latest changes but you should either set `SUNSHINE_DEFAULT_DIR` to the expected destination after install or fix `SUNSHINE_ASSETS_DIR` (this might be more challenging because it's used all over the cmake build file)

CMakeLists.txt
```diff
@@ -440,11 +437,144 @@ if(NOT DEFINED CMAKE_CUDA_STANDARD)
...
+
+	# Pre and post install
+	set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA
+		"${SUNSHINE_ASSETS_DIR}/linux-deb/preinst;${SUNSHINE_ASSETS_DIR}/linux-deb/postinst;${SUNSHINE_ASSETS_DIR}/linux-deb/conffiles")
+	set(CPACK_RPM_PRE_INSTALL_SCRIPT_FILE "${SUNSHINE_ASSETS_DIR}/linux-deb/preinst")
+	set(CPACK_RPM_POST_INSTALL_SCRIPT_FILE "${SUNSHINE_ASSETS_DIR}/linux-deb/postinst")
+
+	# Dependencies
+	set(CPACK_DEB_COMPONENT_INSTALL ON)
+	set(CPACK_DEBIAN_PACKAGE_DEPENDS "libssl1.1, libavdevice58, libboost-thread1.67.0 | libboost-thread1.71.0 | libboost-thread1.74.0, libboost-filesystem1.67.0 | libboost-filesystem1.71.0 | libboost-filesystem1.74.0, libboost-log1.67.0 | libboost-log1.71.0 | libboost-log1.74.0, libpulse0, libopus0, libxcb-shm0, libxcb-xfixes0, libxtst6, libevdev2, libdrm2, libcap2")
```
ABeltramo, line 568: A user from the Discord channel reported that libssl3 (the default on Ubuntu 22.04) should work anyway. We should probably change the dependency definition to be `libssl1.1 | libssl3.0`

abusse:
@ReenigneArcher I see that the MacOS build works now? What was the issue (just curious)?

ReenigneArcher:
> @ReenigneArcher I see that the MacOS build works now? What was the issue (just curious)?

The build works. We reverted how Cmake was finding homebrew packages.

However, I don't think the packages built are working. Would you be willing to test?

ReenigneArcher:
> 

I can set the default directory. What should it be on Linux? Guessing... `/usr/local/sunshine`?

abusse:
> However, I don't think the packages built are working. Would you be willing to test?

Yes, sure I am. However, I have to see when I find some time.

ReenigneArcher:
> > 
> 
> I can set the default directory. What should it be on Linux? Guessing... `/usr/local/sunshine`?

Set `DEFAULT_SUNSHINE_DIR` as `/etc/sunshine` for Linux and MacOS builds.

abusse:
> Set `DEFAULT_SUNSHINE_DIR` as `/etc/sunshine` for Linux and MacOS builds.

For MacOS, this is definitely a very bad idea as is is not supposed to contain third party files. `/usr/local` or `/opt/local` are more suited prefixes without having looked into it in more detail.

ReenigneArcher:
> > Set `DEFAULT_SUNSHINE_DIR` as `/etc/sunshine` for Linux and MacOS builds.
> 
> For MacOS, this is definitely a very bad idea as is is not supposed to contain third party files. `/usr/local` or `/opt/local` are more suited prefixes without having looked into it in more detail.

@abusse @ABeltramo okay, I selected the `/etc/sunshine` based off of two things. The assets directory in the Portfile (https://github.com/SunshineStream/Sunshine/blob/13c2dce3f11470e2d38c663918d748cc41cb1382/Portfile#L32) and the paths in this file (https://github.com/SunshineStream/Sunshine/blob/add-cpack/assets/linux-deb/preinst). Based on your comment and some light reading I think that I'll make some adjustments later. I think separating out the default config files form the assets will make things a bit easier.

I'm thinking assets should go in `/usr/local/sunshine/.assets`, and config in `/usr/local/sunshine/config`... what do you think? Note the assets folder is prefixed with a period.

ReenigneArcher:
@cazzoo This PR is attempting to move the AppImage to a single file instead of an archive. Would you mind testing the AppImage of this build? https://github.com/SunshineStream/Sunshine/actions/runs/2303786058

ReenigneArcher:
Tried to move Linux build to Ubuntu 18.04 in order to improve compatibility more older distros, but ran into a small problem. https://github.com/TheAssassin/appimagelint/issues/28

cazzoo:
@ReenigneArcher, when running from your appimage (with my folders from my appImage): 
```
~/Downloads> ./sunshine.AppImage
Setting $HOME to /home/caz/Downloads/sunshine.AppImage.home
Setting $XDG_CONFIG_HOME to /home/caz/Downloads/sunshine.AppImage.config
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot copy file: No such file or directory [/etc/sunshine/sunshine.conf] [./sunshine.conf]
zsh: IOT instruction (core dumped)  ./sunshine.AppImage
```
When starting appImage from a subfolder without home&config folders:
```
~/Downloads/test>  ./sunshine.AppImage
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot copy file: No such file or directory [/etc/sunshine/sunshine.conf] [./sunshine.conf]
zsh: IOT instruction (core dumped)  ./sunshine.AppImage
```

cazzoo:
> 

Why not using ~/.config/sunshine for linux builds?

ReenigneArcher:
> Why not using ~/.config/sunshine for linux builds?

We've been discussing how to define the config directory. I think the issue is defining a home folder on the github runner will become static to whatever it is on the running instead of the user's environment. Perhaps I am missing something.

CMakeLists.txt
```diff
@@ -440,11 +434,160 @@ if(NOT DEFINED CMAKE_CUDA_STANDARD)
...
+	# Pre and post install
+	set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA
+		"${SUNSHINE_SOURCE_ASSETS_DIR}/linux/deb/preinst;${SUNSHINE_SOURCE_ASSETS_DIR}/linux/deb/postinst;${SUNSHINE_SOURCE_ASSETS_DIR}/linux/deb/conffiles")
+	set(CPACK_RPM_PRE_INSTALL_SCRIPT_FILE "${SUNSHINE_SOURCE_ASSETS_DIR}/linux/deb/preinst")
+	set(CPACK_RPM_POST_INSTALL_SCRIPT_FILE "${SUNSHINE_SOURCE_ASSETS_DIR}/linux/deb/postinst")
+
+	# Dependencies
+	set(CPACK_DEB_COMPONENT_INSTALL ON)
+	set(CPACK_DEBIAN_PACKAGE_DEPENDS "libssl1.1 | libssl3.0, libavdevice58, libboost-thread1.67.0 | libboost-thread1.71.0 | libboost-thread1.74.0, libboost-filesystem1.67.0 | libboost-filesystem1.71.0 | libboost-filesystem1.74.0, libboost-log1.67.0 | libboost-log1.71.0 | libboost-log1.74.0, libpulse0, libopus0, libxcb-shm0, libxcb-xfixes0, libxtst6, libevdev2, libdrm2, libcap2")
+	set(CPACK_RPM_PACKAGE_REQUIRES "libssl==1.1, libavdevice>=58, libboost-thread>=1.67.0, libboost-filesystem>=1.67.0, libboost-log>=1.67.0, libpulse>=0, libopus>=0, libxcb-shm>=0, libxcb-xfixes>=0, libxtst>=0, libevdev>=2.0, libdrm>=2.0, libcap>=2.0")
```
ReenigneArcher, line 586: `libopus` should be `libopusenc` according to this: https://packages.fedoraproject.org/search?query=opus

Probably okay to set the version to `>=0.2` as it was released in 2018. `0.2.1` would be safer yet as it's the current version

abusse:
> @abusse @ABeltramo okay, I selected the `/etc/sunshine` based off of two things. The assets directory in the Portfile (
> 
> https://github.com/SunshineStream/Sunshine/blob/13c2dce3f11470e2d38c663918d748cc41cb1382/Portfile#L32
> 
> ) 

The Portfile has a prefix in front of it that is never `/` but commonly `/opt/local` for global installs or something in the users path, e.g., `$HOME/.macports/` for user installs. However, I saw that you opted for `/usr/local/sunshine` which is probably the choice that makes most sense.

ReenigneArcher:
> The Portfile has a prefix in front of it that is never `/` but commonly `/opt/local` for global installs or something in the users path, e.g., `$HOME/.macports/` for user installs. However, I saw that you opted for `/usr/local/sunshine` which is probably the choice that makes most sense.

@abusse Mind having a look at the changes I made to the Portfile? Based on your comment about the prefix, maybe I should make some additional adjustments.

https://github.com/SunshineStream/Sunshine/pull/144/files#diff-eeef86f2bbcec4ff31eb43b21c39524d77832db4033a4bd1f9b320d102c95daf

abusse:
Yes you probably should, since `$prefix` is already something like `/usr/local/` only for all ports install. So I would probably still put the config somewhere in `${prefix}/etc`, but I have no idea what would be a good folder for the assets. Do they only contain what's necessary for the web frontend? If so, maybe somewhere in  `${prefix}/www` could be an option?

ReenigneArcher:
> Yes you probably should, since `$prefix` is already something like `/usr/local/` only for all ports install. So I would probably still put the config somewhere in `${prefix}/etc`, but I have no idea what would be a good folder for the assets. Do they only contain what's necessary for the web frontend? If so, maybe somewhere in `${prefix}/www` could be an option?

For Mac, the assets is what's necessary for the web ui, default box art to display in moonlight, and the plist file (although that probably could be removed from the assets directory). I don't know if there will be more in the future. Windows and Linux both have shaders in the assets.

How about `${prefix}/etc/sunshine/config` and `${prefix}/etc/sunshine/.assets` ?

abusse:
Sounds reasonable, however, is there any particular reason why the `assets` folder has to be hidden?

ReenigneArcher:
> Sounds reasonable, however, is there any particular reason why the `assets` folder has to be hidden?

It does not need to be hidden, I just figured it would avoid users from doing anything to those files. I can go with unbidden though. I'm not really sure what the best practice is on Linux and Mac for this.

abusse:
So first of all, I think we should have the same structure for all Unix-like systems, minus a prefix. So what goes, e.g., under `/etc` in Linux should be under `/opt/local/etc` for Macports. Having thought about that for only 5min and read [this],(https://unix.stackexchange.com/questions/47436/why-is-the-root-directory-on-a-web-server-put-by-default-in-var-www) I would put the other stuff somewhere under `$PREFIX/var/`.

ReenigneArcher:
@cazzoo I'm still running into issues with the AppImage paths. I've tried a bunch of combinations, but nothing seems to be work properly. Not sure if you have any experience with this, but I'm at a loss and created an issue on linuxdeploy. https://github.com/linuxdeploy/linuxdeploy/issues/201

cazzoo:
Gotcha, I have no idea I never experienced it already but if I manage to find some time to throw an eye i'll let you know. Cmake is also new to me. 
I'll follow issue opened in linuxdeploy for personal sake.

cazzoo:
@ReenigneArcher did you check if you can use Permissions of the Install Directory directive (https://cmake.org/cmake/help/v3.0/command/install.html). I'm not sure but maybe it can help to overcome the issue you face during the appImage build

ReenigneArcher:
@cazzoo I did not try the permissions. According to this (https://docs.appimage.org/packaging-guide/from-source/native-binaries.html?highlight=relative%20path#cmake) the `CMAKE_INSTALL_PREFIX` must be set to `/usr`; however I had little luck when doing that.

I'm guessing this is actually the issue, but don't know how involved it would be to refactor. https://docs.appimage.org/packaging-guide/from-source/native-binaries.html?highlight=relative%20path#packaging-existing-binaries-or-manually-packaging-everything (see the Warning).

cazzoo:
Hi @ReenigneArcher ,
Did you have time to continue on the appImage?
Do you need some help for something in particular?

ReenigneArcher:
> Hi @ReenigneArcher , Did you have time to continue on the appImage? Do you need some help for something in particular?

@cazzoo I have time, but I don't know how to fix it. I'm thinking that the way sunshine actually looks for files needs to change as the AppImage documentation states the program should look for files in a relative manner. Are you on the discord? We could discuss there if you like.

ReenigneArcher:
Some notes for improving documentation as part of this PR.

- Add `sphinx-copybutton`
- `conf.py`

```python
extensions = [
    ...
    'sphinx_copybutton'
    ...
]
```

- remove github maintainer badge from third party packages, only keep AUR maintainer badge
- add urls to badges on third party packages page
- on usage page, rename `apps.json`
- adjust setup/linux section (these are instructions for building, not general usage)
- setup/application list, adjust quoting on second bullet

ReenigneArcher:
@cazzoo regarding the AppImage. I have it working again. It's still not a single file, but those improvements can be made in a separate PR. Additional context here: https://github.com/SunshineStream/Sunshine/issues/151

Further discussion on the AppImage should be moved there, unless it directly relates to changes made by this PR. Thanks!

ReenigneArcher:
@abusse hoping you can help on this...

The Portfile is configured with CMakeLists now. I was able to use it on a vm and successfully install sunshine. Within the vm I was able to use `sudo port pkg sunshine` and `sudo port dmg sunshine`. Thought I would try to accomplish the same in the github runner, but getting some pretty unhelpful errors.

I tried on macos-11 and macos12, same result.

https://github.com/SunshineStream/Sunshine/runs/6853691942?check_suite_focus=true#step:7:5166

CMakeLists.txt
```diff
@@ -440,11 +434,161 @@ if(NOT DEFINED CMAKE_CUDA_STANDARD)
...
+
+	# Pre and post install
+	set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA
+		"${SUNSHINE_SOURCE_ASSETS_DIR}/linux/misc/preinst;${SUNSHINE_SOURCE_ASSETS_DIR}/linux/misc/postinst;${SUNSHINE_SOURCE_ASSETS_DIR}/linux/misc/conffiles")
+	set(CPACK_RPM_PRE_INSTALL_SCRIPT_FILE "${SUNSHINE_SOURCE_ASSETS_DIR}/linux/misc/preinst")
+	set(CPACK_RPM_POST_INSTALL_SCRIPT_FILE "${SUNSHINE_SOURCE_ASSETS_DIR}/linux/misc/postinst")
+
+	# Dependencies
+	set(CPACK_DEB_COMPONENT_INSTALL ON)
+	set(CPACK_DEBIAN_PACKAGE_DEPENDS "libssl1.1 | libssl3, libavdevice58, libboost-thread1.67.0 | libboost-thread1.71.0 | libboost-thread1.74.0, libboost-filesystem1.67.0 | libboost-filesystem1.71.0 | libboost-filesystem1.74.0, libboost-log1.67.0 | libboost-log1.71.0 | libboost-log1.74.0, libpulse0, libopus0, libxcb-shm0, libxcb-xfixes0, libxtst6, libevdev2, libdrm2, libcap2")
```
ReenigneArcher, line 586: Use `openssl` instead of `libssl1.1` or `libssl3` ?

src_assets/linux/misc/85-sunshine-rules.rules
```diff
@@ -0,0 +1 @@
+KERNEL=="uinput", GROUP="input", MODE="0660", OPTION+="static_node=uinput"
```
ReenigneArcher, line 1: `KERNEL=="uinput", GROUP="input", MODE="0660", OPTIONS+="static_node=uinput"`

docs/source/about/usage.rst
```diff
@@ -55,39 +56,77 @@ Setup
...
 
       nano /etc/udev/rules.d/85-sunshine-input.rules
 
-   Input the following contents:
-   ``KERNEL=="uinput", GROUP="input", MODE="0660"``
+   Input the following contents.
+
+   .. code-block::
+
+      KERNEL=="uinput", GROUP="input", MODE="0660", OPTION+="static_node=uinput"
```
ReenigneArcher, line 85: `KERNEL=="uinput", GROUP="input", MODE="0660", OPTIONS+="static_node=uinput"`

docs/source/about/usage.rst
```diff
@@ -55,39 +56,77 @@ Setup
...
+
+      .. code-block::
+
+         [Unit]
+         Description=Sunshine Gamestream Server for Moonlight
 
-   #. Copy it to the users systemd: ``cp sunshine.service ~/.config/systemd/user/``
-   #. Starting
+         [Service]
+         ExecStart=~/sunshine.AppImage
```
ReenigneArcher, line 102: add table for ExecStart command

docs/source/about/installation.rst
```diff
@@ -21,53 +21,127 @@ Docker
...
 -----
 .. image:: https://img.shields.io/github/issues/sunshinestream/sunshine/os:macos?logo=github&style=for-the-badge
    :alt: GitHub issues by-label
 
-#. Install `MacPorts <https://www.macports.org>`_
-#. Download the `Portfile <https://github.com/SunshineStream/Sunshine/blob/master/Portfile>`_ from this repository to
-   ``/tmp``
-#. In a terminal run ``cd /tmp && sudo port install``
-#. The first time you start Sunshine, you will be asked to grant access to screen recording and your microphone.
+Disk Image File option:
```
ReenigneArcher, line 104: comment out all except portfile option

docs/source/about/installation.rst
```diff
@@ -21,53 +21,127 @@ Docker
...
+
+      .. code-block:: bash
+
+         systemctl --user start sunshine
+
+.. Note:: If screencasting fails with Wayland, you may need to run the following to force screencasting with X11.
+
+      .. code-block:: bash
+
+         sudo setcap -r $(readlink -f $(which sunshine))
```
ReenigneArcher, line 43: Remove this section as it's covered in usage.

PR #157, opened 2022-05-09
f35: add libcap-devel to enable KMS code

KarlVogel:
## Description

The Fedora 35 docker image doesn't have libcap development files, hence the CMake checks won't enable KMS support. (SUNSHINE_ENABLE_DRM depends on the libcap dev files)

### DIFF
--- a/scripts/Dockerfile-fedora_33
+++ b/scripts/Dockerfile-fedora_33
@@ -21,6 +21,8 @@ RUN dnf -y update && \
         openssl-devel \
         opus-devel \
         pulseaudio-libs-devel \
+        libcap-devel \
+        libdrm-devel \
         rpm-build \
     && dnf clean all \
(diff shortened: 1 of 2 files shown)


ReenigneArcher:
These dockerfiles will likely be going away, see here: https://github.com/SunshineStream/Sunshine/pull/145

But I'll leave this PR open for now just in case.

ReenigneArcher:
@KarlVogel is this library also necessary in the Fedora 33 dockerfile?

Also, please check https://github.com/SunshineStream/Sunshine/pull/150 and let me know if you approve.

KarlVogel:
@ReenigneArcher just checked, F33 also needs it and it also needs libdrm-devel. F35 had libdrm-devel but it was probably pulled in from another package as dependency. I'll update the PR.

Issue #168, opened 2022-05-27
sound distorted using Pipewire


adolfotregosa:
### Describe the Bug

Any audio comes distorted no matter what stream setting is used in moonlight. I have no idea here to start to fix this.
Any help is appreciated.

Both Host and guest are running manjaro with pipewire + wireplumber.

### Sunshine Host Operating System and Version

Manjaro

### Architecture

64

### Sunshine Version

0.13

### GPU Type

nvidia 

### GPU Model

3060

### GPU Driver/Mesa Version

515.43.04

### Capture Method (Linux Only)

X11

adolfotregosa:
First clue. The problem only exists while using pipewire. 
Pulseaudio works fine

KuleRucket:
This started happening to me and I remembered getting a pipewire update this week. So I rolled back to a snapshot from a week ago and the sound was fine again. I then held back the update of pipewire + pipewire-pulse, updated the rest and it's still working.

KuleRucket:
https://github.com/PipeWire/pipewire/blob/master/NEWS

Highlights:
  - Add 44.1KHz to allowed samplerates. The server can now switch by
    default between 48KHz and 44.1KHz.

So maybe now Sunshine is trying to use a different sampling rate than Moonlight and we need to do something to set it explicitly?

KuleRucket:
Thinking about it, if it works with pulseaudio then this a probably a problem with the pipeline-pulse translation. I guess the best solution would be to implement the option to use pipeline natively.

adolfotregosa:
> Thinking about it, if it works with pulseaudio then this a probably a problem with the pipeline-pulse translation. I guess the best solution would be to implement the option to use pipeline natively.

pipewire 0.3.52 same issue. If I can actually help on fixing this, please do ask

thegamingrestoration:
> This started happening to me and I remembered getting a pipewire update this week. So I rolled back to a snapshot from a week ago and the sound was fine again. I then held back the update of pipewire + pipewire-pulse, updated the rest and it's still working.

Do you still have the snapshot from a week ago? Could you attach it here?

KuleRucket:
It's a btrfs snapshot of my root filesystem and wouldn't be much use to you. You can just downgrade the packages though if you want to revert an update (https://wiki.archlinux.org/title/downgrading_packages).

schlegp:
[...]

PR #214, opened 2022-06-19
Integrate AUR package `sunshine-git`

### LINKED ISSUE #39: Publish to AUR using aurpublish
**Describe the bug**

There are two sunshine packages on AUR, manually maintained.
  - https://aur.archlinux.org/packages/sunshine
  - https://aur.archlinux.org/packages/sunshine-git

**Expected behavior**

- Build the AUR package on PRs into nightly and master.
- Automate the AUR package update when a push is made into master branch.
  - Hopefully `pkgver` can be automated easily (we already get version automatically from the CHANGELOG file)
    - https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=sunshine#n4
    - https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=sunshine-git#n5
  - url also should be updated to match this repo
- This file needs to be updated (https://github.com/SunshineStream/Sunshine/blob/nightly/.github/workflows/CI.yml)

ReenigneArcher:
## Description

- Add configurable `PKGBUILD` for AUR
- Add `optdepends` to `PKGBUILD`
- Build `PKGBUILD` file in docker container
- Create `.SRCINFO` file automatically
- Validate `PKGBUILD` file using `namcap`
- Use action to publish to AUR automatically [sunshine-git](https://aur.archlinux.org/packages/sunshine-git)

### Issues Fixed or Closed

- Fixes #39 

## Type of Change

## Checklist

- [ ] I have added or updated the docstring/documentation-blocks for new or existing methods/components


KuleRucket:
Few comments:
1) In keeping with the purpose of -git builds, I think it should point to the nightly branch. So:
source=("$pkgname::git+https://github.com/SunshineStream/sunshine.git#branch=nightly"
2) -D LIBSWSCALE_INCLUDE_DIR= and LIBAVUTIL_LIBRARIES= have duplicates
3) Adding optional dependencies optdepends=('cuda' 'libdrm' 'libcap') would help people wanting to use those without impacting people who don't
4) I don't think patching systemd-user-config.patch is necessary. It could just use our service spec without patches
5) udev.rules just seems to duplicate our own udev file so is also unnecessary

ReenigneArcher:
>     1. In keeping with the purpose of -git builds, I think it should point to the nightly branch. So:
>        source=("$pkgname::git+https://github.com/SunshineStream/sunshine.git#branch=nightly"

We discussed this on discord, and I agree. I don't know if it's possible to rename the repo on the AUR... maybe `sunshine-nightly` would be better?

Is there any benefit to adding the commit? As it will be tightly integrated with the github workflow, I can add the specific commit.

>     2. -D LIBSWSCALE_INCLUDE_DIR= and LIBAVUTIL_LIBRARIES= have duplicates

Which one should be removed?

>     3. Adding optional dependencies optdepends=('cuda' 'libdrm' 'libcap') would help people wanting to use those without impacting people who don't

I don't see any issue with this.

>     4. I don't think patching systemd-user-config.patch is necessary. It could just use our service spec without patches

Agree, just trying to build the existing version for now... will update it after I get it to at least build.

>     5. udev.rules just seems to duplicate our own udev file so is also unnecessary

Same comment as 4... also it's outdated since v0.14.0.

KuleRucket:
> Which one should be removed?`

They both appear twice in the sunshine-git PKGBUILD: 

-D LIBAVCODEC_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
-D LIBAVCODEC_LIBRARIES=/usr/lib/ffmpeg4.4/libavcodec.so \
-D LIBAVDEVICE_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
-D LIBAVDEVICE_LIBRARIES=/usr/lib/ffmpeg4.4/libavdevice.so \
-D LIBAVFORMAT_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
-D LIBAVFORMAT_LIBRARIES=/usr/lib/ffmpeg4.4/libavformat.so \
-D LIBAVUTIL_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
-D LIBAVUTIL_LIBRARIES=/usr/lib/ffmpeg4.4/libavutil.so \
-D LIBSWSCALE_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
**-D LIBAVUTIL_LIBRARIES=/usr/lib/ffmpeg4.4/libavutil.so** \
**-D LIBSWSCALE_INCLUDE_DIR=/usr/include/ffmpeg4.4** \
-D LIBSWSCALE_LIBRARIES=/usr/lib/ffmpeg4.4/libswscale.so

KuleRucket:
> Is there any benefit to adding the commit?

No I don't think so

ReenigneArcher:
> 1. In keeping with the purpose of -git builds, I think it should point to the nightly branch. So:
>        source=("$pkgname::git+https://github.com/SunshineStream/sunshine.git#branch=nightly"

Been studying this a bit more, and I changed my opinion. Based on most of my reading, `-git` is usually just indicating that the package is built from a git repo. It doesn't mean it should point to a development branch or anything like that.

Currently the way both `sunshine` and `sunshine-git` are setup, they actually both point to `master`, which is redundant as `master` never gets an unreleased version.

Anyway, therefore I believe that `sunshine-git` should become the default "stable" version. `sunshine` should probably be removed altogether... I'm pretty sure the "maintainer" doesn't even use sunshine.

Development `PKGBUILD` and `.SRCINFO` files will be available as artifacts from workflow runs if anyone needs to test a development version. Also, decided to add the commit, as it will allow testing the `PKGBUILD` for PRs. In the future I would like to setup `nightly` releases, not just for the AUR, but for general packages. The CI workflow is getting pretty complex though, so need to think about the best way to do this.

Anyway, I tested the last build `PKGBUILD` on a fresh manjaro vm. If I built with `makepkg -fi`, everything worked except being able to modify `sunshine.conf` and `apps.json`. I believe this is the an existing issue in the AUR anyway. Probably needs an adjustment to CMakeLists to solve, and might even be a problem on other distros/packages, so I don't think I'll worry about fixing it in this PR.

KuleRucket:
> 

In all of the other projects where I use the -git version the development is done on the master branch and so when developers change something, you get an update whether it works or not. For example lutris-git, pegasus-frontend-git or libretro-vice-git are all like this. The git version is not normally the stable version but is the dev version. Sunshine is a bit unusual in that only stable versions are pushed to the master branch.

If we keep -git pointing to master there is actually never a difference between sunshine and sunshine, so as you say one is superfluous. Normally, you expect to get a stable versioned release from the non-git and the dev version from sunshine-git. This is why the function for generating the package version in -git AURs is based on the latest tag + number of commits since the tag. If is so that you get a new package for every commit.

ReenigneArcher:
Okay... Still trying to get control of "sunshine"... It might help if more people put in a quest to have it orphaned so I can adopt it.

KuleRucket:
I was going to grab ownership when the previous guy quit but the current one got in there first. The current owner was interacting with this github but seems to have disappeared.

KuleRucket:
Actually there's a good demo of -git AURs in action in #168. The -git AURs on pipewire give you the opportunity to get the fix before the release has gone out without manually building yourself.

packaging/linux/aur/PKGBUILD
```diff
@@ -0,0 +1,55 @@
...
+pkgver=@PROJECT_VERSION@@SUNSHINE_SUB_VERSION@
+pkgrel=1
+pkgdesc="@PROJECT_DESCRIPTION@"
+arch=('x86_64' 'i686')
+url=@PROJECT_HOMEPAGE_URL@
+license=('GPL3')
+
+depends=('boost-libs' 'ffmpeg4.4' 'libpulse' 'libevdev' 'libx11' 'libxcb' 'libxfixes' 'libxrandr' 'libxtst' 'openssl' 'opus' 'udev')
+makedepends=('git' 'cmake' 'boost' 'make')
+optdepends=('cuda' 'libdrm' 'libcap')
```
ReenigneArcher, line 14: Should we add dependency version requirements? https://wiki.archlinux.org/title/PKGBUILD#depends

ReenigneArcher: Also, descriptions of when the optional dependencies are needed: https://wiki.archlinux.org/title/PKGBUILD#optdepends

ReenigneArcher: image

@sitiom care to elaborate on this? should it be required or optional?

packaging/linux/aur/PKGBUILD
```diff
@@ -0,0 +1,55 @@
...
+        -D LIBAVDEVICE_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
+        -D LIBAVDEVICE_LIBRARIES=/usr/lib/ffmpeg4.4/libavdevice.so \
+        -D LIBAVFORMAT_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
+        -D LIBAVFORMAT_LIBRARIES=/usr/lib/ffmpeg4.4/libavformat.so \
+        -D LIBAVUTIL_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
+        -D LIBAVUTIL_LIBRARIES=/usr/lib/ffmpeg4.4/libavutil.so \
+        -D LIBSWSCALE_INCLUDE_DIR=/usr/include/ffmpeg4.4 \
+        -D LIBSWSCALE_LIBRARIES=/usr/lib/ffmpeg4.4/libswscale.so
+
+    make -C build
```
ReenigneArcher, line 50: image

@KuleRucket I can add this, unless there's a specific reason not to

packaging/linux/aur/PKGBUILD
sitiom: Optional because it enables the discoverability of Sunshine on the local network? Or should that be required?

sitiom: It would make more sense as a required dependency.

ReenigneArcher: Added `avahi`

stephensrmmartin:
I just stumbled across this after working to get the sunshine-git PKGBUILD working for .14 (I'm hwkiller - in the comments who posted a revised version in pastebin).

What you are doing here is awesome! Thank you for supporting the AUR, and especially for making it part of the build system itself (I have not done this, and I probably should!).

Just one question:

```
-D SUNSHINE_ASSETS_DIR="local/sunshine/assets" \
-D SUNSHINE_CONFIG_DIR="local/sunshine/config" \
```

I think these should be put into `/usr/share/sunshine`; or *preferably* assets should be in `/usr/share/sunshine`, and config should be in either `/usr/lib/sunshine` or `/etc/sunshine/` (depending on whether the configs are *used* by default, or simply there as a *prototype* at runtime to be copied; i.e., if you are loading the config, then /etc; if you are copying it to XDG config, then /usr/lib/ would be a better bet for configs).

... rather than `/usr/local`. `/usr/local` shouldn't really be touched by packages (even AUR ones).

Likewise, udev rules from packages should, from what I understand, go into `/usr/lib/udev/rules.d/`.

`/etc/udev/rules.d` is meant for overrides or custom configs by the user.

packaging/linux/aur/PKGBUILD
KuleRucket: > Should we add dependency version requirements? https://wiki.archlinux.org/title/PKGBUILD#depends
Possibly, however on Arch everyone is supposed to be up to date with latest packages anyway so it probably doesn't add much.

> Also, descriptions of when the optional dependencies are needed
Yes I agreed that would be useful.

ReenigneArcher:
> Note that for the change:
> make -C build -j ${nproc}
> 
> This should normally be set globally in /etc/makepkg.conf:
> /etc/makepkg.conf:MAKEFLAGS="-j $(nproc)"
> 
> So should not go into the individual PKGBUILD files
> 
> 

Okay, I'll revert this.

ReenigneArcher:
@stephensrmmartin thank you, that explanation makes sense.

Please let me know if you agree or disagree with this by placing a 👍 or 👎 

| os         | configs                                | assets                                | udev rules                  |
|--------- | ----------------------------- | ---------------------------- | ----------------------- |
| Linux     | /etc/sunshine                     | /usr/share/sunshine           | /usr/lib/udev/rules.d |
| MacOS  | /usr/local/sunshine/config | /usr/local/sunshine/assets | n/a ?                          |

At the moment I believe the configs are used in the same location you that they are installed to... unless the user specifies a different config and apps.json path... then the default ones are copied over.

In this case, I am hoping the definitions on Linux will be acceptable for any distro (Debian, Red Hat, Arch, etc.), and MacOS according to your comment would likely be the same for BSD if we ever create a package for that.

stephensrmmartin:
The linux and macos paths look good to me. My only remaining question is whether the /etc/sunshine config directory is there to have configs that are actually loaded, or if it has configs that are example configs which are copied into ~/.config/sunshine/ (or, generally, XDG_CONFIG_HOME/sunshine). If it is a 'prototypical' config that is copied when the user runs the sunshine user service, then it should belong in either /usr/lib/sunshine or /usr/share/sunshine (the choice is unclear, I've seen it both ways). But this is such minutiae.

Regardless though, the choice of assets, udev rules, binary, and configs are acceptable regardless of the distro at this point I believe. I don't even foresee anyone quabbling about a prototypical config being put into /etc/ (I'm not sure there's a particular 'rule', per se, about prototype configs belonging in /etc or elsewhere, it just seems like they tend to wind up in /usr/share or /usr/lib).

The macos paths should be similar to freeBSD I believe. Netbsd has different paths; openbsd may have different paths too (I'm not as familiar with these two). I would not personally worry about these latter cases. The number of people who want to run sunshine on netbsd or openbsd would surely be so vanishingly small, that if they truly wanted to port it, they can feel free to submit a PR.

ReenigneArcher:
> My only remaining question is whether the /etc/sunshine config directory is there to have configs that are actually loaded, or if it has configs that are example configs which are copied into ~/.config/sunshine/ (or, generally, XDG_CONFIG_HOME/sunshine)

This is how I believe it works.

`./sunshine` -> uses sunshine.conf and apps.json as they exist
`./sunshine ~/sunshine.conf` will copy the existing config to the destination (or maybe it just writes a brand new config)

I've had a lot of issues with this, and ideally these files don't even need to exist, and should be baked into the source code and created when needed.

The default config is blank anyway, and always has been (previously everything was commented out and just served as documentation)... I have since moved the documentation to read the docs.

I just don't have the time to dig into the source code to change this behavior.

stephensrmmartin:
> > My only remaining question is whether the /etc/sunshine config directory is there to have configs that are actually loaded, or if it has configs that are example configs which are copied into ~/.config/sunshine/ (or, generally, XDG_CONFIG_HOME/sunshine)
> 
> This is how I believe it works.
> 
> `./sunshine` -> uses sunshine.conf and apps.json as they exist `./sunshine ~/sunshine.conf` will copy the existing config to the destination (or maybe it just writes a brand new config)
> 
> I've had a lot of issues with this, and ideally these files don't even need to exist, and should be baked into the source code and created when needed.
> 
> The default config is blank anyway, and always has been (previously everything was commented out and just served as documentation)... I have since moved the documentation to read the docs.
> 
> I just don't have the time to dig into the source code to change this behavior.
> 
> The one reason I wouldn't write the files to the home directory is that sunshine can really only be run from one user simultaneously, and probably should only have one config.

Alright; well if /etc has a functional config, then keep the config in there.

As for per-user config, I believe it's fairly standard practice to have a 'default' config copied over to $XDG_CONFIG_HOME/sunshine/sunshine.conf if it doesn't exist, and if it's run as a user rather than a system account.

Personally, I think it makes sense to have per-user configs, especially if we are typically running it as a systemd user service (e.g., systemctl --user start sunshine). That behavior doesn't strike me as odd at all.

Having a 'default' config consist of nothing but commented-out, default options is also pretty common as far as I know.

istori1:
I have been a bit confused on this given Sunshine already saves some configuration into the users home.

image

Since it is running as a user service, it will be standard practice to place the config into ~/.config/sunshine

I believe @ReenigneArcher wants a system configuration, in which case I would suggest to use /var/lib/sunshine for that, and making the sunshine sub-folder only accessible by the input group.

Since there is already data stored on ~/.config/sunshine we should be either saving all the configuration there or changing the Sunshine code to do all config data in /var/lib/sunshine for a system wide writable configuration.

Ideally there should be a read only default config somewhere in /etc that gets copied to ~/.config/sunshine and then edit that user configuration.

ReenigneArcher:
@istori1 I don't want a system configuration necessarily, although that does make sense to me since running more than one instance of Sunshine will be problematic... likely the ports will already be in use, most moonlight clients do not allow a custom port range, and of course hardware limitations. With that said, I'm okay with it using the home directory.

It's possible I may have broke something when refactoring for the packaging updates previously... this was a massive PR. Do you know if the config or apps files were saved there in previous versions?

stephensrmmartin:
> @istori1 I don't want a system configuration necessarily, although that does make sense to me since running more than one instance of Sunshine will be problematic... likely the ports will already be in use, most moonlight clients do not allow a custom port range, and of course hardware limitations. With that said, I'm okay with it using the home directory.
> 
> It's possible I may have broke something when refactoring for the packaging updates previously... this was a massive PR. Do you know if the config or apps files were saved there in previous versions?

I have those files too, but the aur package currently has a service file that literally does that copy if the configs aren't already in the config directory. As in, the user service file checks for XDG_CONFIG_HOME/sunshine and if it doesn't exist, copies files from the packaged config directory.

I suspect that a system (or system user) running sunshine wouldn't even work, no? I haven't tried, but I imagine there'd be inherent limitations to what can be streamed depending on which X display is open, or which Wayland session is active. It's not like a vnc server is running. It's ffmpeg recording an active session, no?

stephensrmmartin:
> > I have those files too, but the aur package currently has a service file that literally does that copy if the configs aren't already in the config directory. As in, the user service file checks for XDG_CONFIG_HOME/sunshine and if it doesn't exist, copies files from the packaged config directory.
> 
> I see some issues with this approach: https://aur.archlinux.org/cgit/aur.git/tree/systemd-user-config.patch?h=sunshine#n10
> 
>     * It's copying the entire assets directory (shaders, web, configs, etc.)... that's a complete waste
> 
>     * I think the copied files have the original permissions, so they are still read only
> 
>     * It's only using the copied `sunshine.conf`, not the `apps.json`, and still points to the default `apps.json` (read only)
> 
>     * It's unnecessary as sunshine will already copy the default `sunshine.conf` if you start sunshine like this `sunshine <preferered conf file>`

I totally agree; it's definitely a hack (I'm not sure who originated  that hack; if it was the AUR submitter/maintainer, or upstream at some point).
The service file shouldn't do any copying. That should be up to sunshine itself, based on whether you have a config already.

I imagine the best solution would be to 1) Use /etc/sunshine/* if $XDG_CONFIG_HOME/sunshine/* is not populated and 2) When a user changes a setting, to save the config to $XDG_CONFIG_HOME/sunshine/; 3) When $XDG_CONFIG_HOME/sunshine/sunshine.conf exists, use it *by default* and *unless* they explicitly specify another config via command line.

ReenigneArcher:
On a side note... anyone know how long it takes arch to respond to requests? I know they have a lot, but I would expected this to get a response by now?

https://aur.archlinux.org/packages/sunshine

https://lists.archlinux.org/pipermail/aur-requests/2022-June/073138.html
https://lists.archlinux.org/pipermail/aur-requests/2022-June/073186.html

ReenigneArcher:
Not sure how to make the build have different prefixes... It didn't like this...

```txt
        -D CMAKE_INSTALL_PREFIX="/" \
        -D SUNSHINE_ASSETS_DIR="usr/share/sunshine" \
        -D SUNSHINE_CONFIG_DIR="etc/sunshine" \
```

https://github.com/SunshineStream/Sunshine/runs/7035419321?check_suite_focus=true#step:7:990

So going to use this until the paths can be addressed better in a separate PR.

```txt
        -D CMAKE_INSTALL_PREFIX="/usr" \
        -D SUNSHINE_ASSETS_DIR="share/sunshine/assets" \
        -D SUNSHINE_CONFIG_DIR="share/sunshine/config" \
```

.github/workflows/CI.yml
```diff
@@ -52,6 +52,100 @@ jobs:
...
+      - name: Upload Artifacts
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: sunshine-linux-aur
+          path: artifacts/
+
+      - name: Publish AUR package
+        if: ${{ env.aur_publish == 'true' }}
+        uses: KSXGitHub/github-actions-deploy-aur@master  # assets arg not in latest release
```
ReenigneArcher, line 137: v2.3.0 has been released now

ReenigneArcher:
Perhaps I'm too impatient, but I'm kind of losing faith that the aur people will orphan the repo for us.

What option do you guys want.

1. Use `sunshine-git` (this will be indicated in our docs that this is the official one, and the third party one will remain how it is in the docs)
2. Create a new repo `sunshinestream-sunshine`

Let me know.

stephensrmmartin:
Does this automated version push to aur on every master commit? Or is it only after a git release?

ReenigneArcher:
> Does this automated version push to aur on every master commit? Or is it only after a git release?

Every push to master, which will be when we decide to make a release as we push everything into `nightly` first.

stephensrmmartin:
> > Does this automated version push to aur on every master commit? Or is it only after a git release?
> 
> Every push to master, which will be when we decide to make a release as we push everything into `nightly` first.

Then sunshine-git is fine to use.

I think other aur packages do this too. There's pkg, pkg-git, pkg-nightly-git, etc. Even if release coincides with master commits, it's appropriate to use -git. I don't think -git necessarily requires that it's tracking every commit ever pushed into a project at any given time. It tends to mean the master branch of the upstream repo, no matter what that project's master policy is. 
If people want a truly bleeding edge version, they're free to switch the branch in the pkgbuild themselves.

ReenigneArcher:
> I think other aur packages do this too. There's pkg, pkg-git, pkg-nightly-git, etc. Even if release coincides with master commits, it's appropriate to use -git. I don't think -git necessarily requires that it's tracking every commit ever pushed into a project at any given time. It tends to mean the master branch of the upstream repo, no matter what that project's master policy is.
> If people want a truly bleeding edge version, they're free to switch the branch in the pkgbuild themselves.

I would agree with that based on my research so far, although some didn't agree in the discord, haha.

My understanding is that `-git` strictly indicates that it's building from a git repo. We also will have the `PKGBUILD` as an artifact on CI workflow runs if users want to download from there.

ReenigneArcher:
Unless there are any objections, this will be merged soon... Following the next push to `master` `sunshine-git` will be updated automatically.

csolisr:
Pinging @ReenigneArcher to inform that the main AUR package has been orphaned and ready for takeover. https://aur.archlinux.org/packages/sunshine

ReenigneArcher:
Thanks. I've adopted it.

packaging/linux/aur/PKGBUILD
NyaomiDEV: There's a specific reason not to:

https://github.com/archlinux/svntogit-packages/blob/e89aff010015264dd9992ea0ad220c8839280a34/trunk/makepkg.conf#L49

The user chooses whether to export the `-j` flag through the `MAKEFLAGS` variable in the `makepkg` configuration file.

ReenigneArcher: @NyaomiDEV why you are commenting on a merged PR, let alone a resolved conversation?

PR #319, opened 2022-08-10
ci: build all before releasing

sitiom:
## Description
Upon creating a new release, all binaries should be available. This should fix https://github.com/LizardByte/Sunshine/discussions/318#discussioncomment-3362357

The https://github.com/softprops/action-gh-release action is used for releasing and the action from this org is no longer used as the action is only run once and not reused elsewhere.

### Issues Fixed or Closed

- Fixes https://github.com/LizardByte/Sunshine/discussions/318#discussioncomment-3362357

## Type of Change

## Checklist

- [ ] I have added or updated the docstring/documentation-blocks for new or existing methods/components

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -235,15 +235,6 @@ jobs:
           name: sunshine-linux-flatpak
           path: artifacts/
 
-      - name: Create Release
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
-        uses: LizardByte/.github/actions/create_release@master
-        with:
-          token: ${{ secrets.GH_BOT_TOKEN }}
-          next_version: ${{ needs.check_changelog.outputs.next_version }}
-          last_version: ${{ needs.check_changelog.outputs.last_version }}
-          release_body: ${{ needs.check_changelog.outputs.release_body }}
-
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
This does not resolve the issue at hand. The winget workflow is failing even with the `.exe` present.

sitiom:
> This does not resolve the issue at hand. The winget workflow is failing even with the `.exe` present.

Which shouldn't, and should be a bug in the action if it does so. cc @vedantmgoyal2009

But anyways, this fixes the case where there `.exe` is not present (which happened on the first run).

vedantmgoyal2009:
On re-runs, the release object in the GitHub context payload does not get updated, and hence, the action can't find `.exe` URL.

ReenigneArcher:
> > On re-runs, the release object in the GitHub context payload does not get updated, and hence, the action can't find `.exe` URL.
> 
> Oh alright, that explains it. @ReenigneArcher Should I make a manual update to winget-pkgs for `0.14.1` for a while?

Yes, for now. I don't plan to create another release just to update winget-pkgs.

sitiom:
Ok. https://github.com/microsoft/winget-pkgs/pull/69256

PR #471, opened 2022-11-02
Add install scripts (Flatpak)

istori1:
## Description
Add install script for mouse permission.

## Type of Change

## Checklist


packaging/linux/flatpak/install-mouse
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+cp /app/share/sunshine/udev/rules.d/85-sunshine.rules $HOME/.config/sunshine
+flatpak-spawn --host pkexec sh -c "sudo usermod -a -G input $USER && cp $HOME/.config/sunshine/85-sunshine.rules /etc/udev/rules.d"
```
ReenigneArcher, line 4: I think we don't need `cp` and `rm` commands. Not tested though.

Would this work?
```bash
flatpak-spawn --host pkexec sh -c "sudo usermod -a -G input $USER && cat "/app/share/sunshine/udev/rules.d/85-sunshine.rules" | sudo tee /etc/udev/rules.d/85-sunshine.rules
```

packaging/linux/flatpak/dev.lizardbyte.sunshine.yml
```diff
@@ -208,3 +208,6 @@ modules:
       - sed -r -z -i -e
         's/("((do)|(undo)|(cmd)|(detached))"\s*:\s*\[?\n*\s*")(.*")/\1flatpak-spawn --host \7/gm'
         /app/share/sunshine/apps.json
+      - install -D packaging/linux/flatpak/install-mouse /app/bin/install-mouse
+      - install -D packaging/linux/flatpak/remove-mouse /app/bin/remove-mouse
```
ReenigneArcher, line 212: Can you add the `.sh` extension to these? I know it's not necessary to function but will help anyone using an IDE to develop.

If you add a directory like `packaging/linux/flatpak/scripts/`, could you just install the entire directory and not have a new line for each script? Or alternatively could it be a single script with arguments you can pass to it, like the AppRun for AppImage?

packaging/linux/flatpak/scripts/install-user-service.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+mkdir -p ~/.config/systemd
+mkdir -p ~/.config/systemd/user
+cp /app/share/sunshine/systemd/user/sunshine.service $HOME/.config/systemd/user/sunshine.service
+sed -i 's/\/app\/bin\/sunshine/flatpak run dev.lizardbyte.sunshine\nExecStop=flatpak kill dev.lizardbyte.sunshine/g' $HOME/.config/systemd/user/sunshine.service
```
ReenigneArcher, line 6: You can use another symbol to separate the `sed` command, so you don't need to escape so much.  You can use any symbol you like.
Example:

```bash
sed -i 's%/app/bin/sunshine%flatpak run dev.lizardbyte.sunshine\nExecStop=flatpak kill dev.lizardbyte.sunshine%g' $HOME/.config/systemd/user/sunshine.service
```

I think we should move the `sed` command to the manifest. If a user runs the `install-user-service.sh` script more than once it could have unintended results.

Also, has the `sed` command been tested?

packaging/linux/flatpak/scripts/install-mouse.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+UDEV=$(cat /app/share/sunshine/udev/rules.d/85-sunshine.rules)
+flatpak-spawn --host pkexec sh -c "usermod -a -G input $USER && echo '$UDEV' > /etc/udev/rules.d/85-sunshine.rules"
```
ReenigneArcher, line 4: Was this tested? Does `$USER` evaluate properly?

packaging/linux/flatpak/scripts/remove-mouse.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+flatpak-spawn --host pkexec sh -c "gpasswd -d $USER input && rm /etc/udev/rules.d/85-sunshine.rules"
```
ReenigneArcher, line 3: Was this tested? Does `$USER` evaluate properly?

istori1: Yes

packaging/linux/flatpak/scripts/install-mouse.sh
istori1: Yes

packaging/linux/flatpak/scripts/additional-install.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# User Service
+mkdir -p ~/.config/systemd
```
ReenigneArcher, line 4: This line is not needed since `-p` will make the entire tree if needed. So Next line will create this.

ReenigneArcher (changes requested):
One small change request to simplify it a bit.

Can you also update the docs on how to use these?

Thanks!

docs/source/about/installation.rst
```diff
@@ -116,12 +116,25 @@ Flatpak Package
...
       .. code-block:: bash
 
          flatpak run dev.lizardbyte.sunshine
 
+      Sunshine Service
+
+      .. code-block:: bash
+
+         systemctl --user start sunshine
+
```
ReenigneArcher, line 137: This is already covered in the `usage` page. I probably wouldn't add it here since we don't have it for the other linux versions.

docs/source/about/installation.rst
```diff
@@ -104,17 +104,17 @@ Flatpak Package
    :alt: GitHub issues by-label
 
 #. Install `Flatpak <https://flatpak.org/setup/>`_ as required.
-#. Download ``sunshine.flatpak`` and run the following code.
+#. Download ``sunshine_x86_64.flatpak`` and run the following code.
```
ReenigneArcher, line 107:
```rst
Download ``sunshine_{arch}.flatpak`` and run the following code.

.. Note:: Be sure to replace ``{arch}`` with the architecture for your operating system.
```

docs/source/about/installation.rst
```diff
@@ -104,17 +104,17 @@ Flatpak Package
...
 
    System level (recommended)
       .. code-block:: bash
 
-         flatpak install --system sunshine.flatpak
+         flatpak install --system ./sunshine_x86_64.flatpak
```
ReenigneArcher, line 112: `sunshine_{arch}.flatpak`

docs/source/about/installation.rst
```diff
@@ -104,17 +104,17 @@ Flatpak Package
...
 
    User level
       .. code-block:: bash
 
-         flatpak install --user sunshine.flatpak
+         flatpak install --user ./sunshine_x86_64.flatpak
```
ReenigneArcher, line 117: `sunshine_{arch}.flatpak`

ReenigneArcher (changes requested):

docs/source/about/installation.rst
```diff
@@ -104,17 +104,24 @@ Flatpak Package
...
    User level
       .. code-block:: bash
 
-         flatpak install --user sunshine.flatpak
+         flatpak install --user ./sunshine_{arch}.flatpak
+
+   Additional installation (required)
+      .. code-block:: bash
+
+         flatpak run --command=additional-install.sh dev.lizardbyte.sunshine
```
ReenigneArcher, line 124: Add this code to each sub section (system and user level) instead of making a new sub section.

And if you agree to rename the scripts, change the name here as well.

packaging/linux/flatpak/scripts/additional-install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
ReenigneArcher, line 1: can we just rename to `install.sh` ?

packaging/linux/flatpak/scripts/remove-additional-install.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
ReenigneArcher, line 1: can we just rename to `remove.sh`?

docs/source/about/installation.rst
```diff
@@ -130,7 +137,8 @@ Start:
 Uninstall:
    .. code-block:: bash
 
-      flatpak uninstall --delete-data sunshine.flatpak
+      flatpak run --command=remove-additional-install.sh dev.lizardbyte.sunshine
```
ReenigneArcher, line 140: If you agree to rename the scripts, change the name here as well.

ReenigneArcher (changes requested):
Minor changes requested.

Also, is it possible to have a single script with arguments, like the AppRun for AppImage? https://github.com/LizardByte/Sunshine/blob/master/packaging/linux/AppImage/AppRun

If not, no issue with having two scripts.

ReenigneArcher (approved):

Issue #412, opened 2022-10-17
Capturing images by callback, rather than pulling Causes Desktop issues.


Theo-Technologies:
### Describe the Bug

The Commit Capturing images by callback, rather than pulling seen here: https://github.com/LizardByte/Sunshine/commit/7ddf8bbe94ceb31dcc411dc28206399d81396a05

Causes issues when using the desktop. If the mouse is not moving the fps will drop dramatically and extreme lag will occur. If the mouse is constantly moving then there is no lag. This only occurs on the desktop or using non GPU intensive applications like the web browser or Microsoft Word. Previous to this commit, the desktop would function fine.

### Expected Behavior

We need the desktop to be stable at all times. Whether this be a different callback method or a small fix.

### Host Operating System

Windows

### Operating System Version

Windows 10 Pro

### Architecture

64 bit.

### Sunshine Version

0.14.0

### GPU Type

Nvidia

### GPU Model

A4000

### GPU Driver/Mesa Version

n/a

OliverJFedorko:
> This seems relevant. [loki-47-6F-64/sunshine#253](https://github.com/loki-47-6F-64/sunshine/issues/253)

I have seen this before. I have noticed that the callback method is far superior on slower internet connections with better stability when gaming though.

ReenigneArcher:
@Theo-Technologies while I asked you to open this, it appears to be a duplicate. Can you add your details to #122 ?

Issue #387, opened 2022-10-04
Low FPS/High delay on desktop


JFMugen:
### Describe the Bug

While writing something or just browsing images for example, delay increases as much as 1-2 seconds.

### Expected Behavior

Constant 60 fps without any delay. 

### Additional Context

I am using Cat6 wire to connect client and host so I have no network limit. 

### Host Operating System

Windows

### Operating System Version

Win10 Home 21H2  -  19044.2006

### Architecture

64

### Sunshine Version

0.14.1

### GPU Type

Nvidia

### GPU Model

3060

### GPU Driver/Mesa Version

516.94

Jake-Jensen:
Just commenting here, but this: "I am using Cat6 wire to connect client and host so I have no network limit." means functionally nothing. If your NIC is only 10/10, it doesn't matter what wire you're using, it's 10/10.

ReenigneArcher:
While correct, I don't believe this is a network issue. This is a known issue on Windows desktops when few pixels are changing.

ReenigneArcher:
This is a duplicate of #412 and #122 so closing.

PR #485, opened 2022-11-08
video: sync capture: resend frame if needed to avoid capture delay

### LINKED ISSUE #122: AMD AMF: Last frames not showing up when screen updates are ending  [Windows]
### Describe the Bug

When using the *H.264* or the *H.265* AMD codec the last frames of a screen refresh aren't transmitted and get "lost". The screen is only refreshed when the mouse is moved or something else is displayed on the screen.

That means for example:

* when typing you have a delay of one letter / until the text cursor is blinking again
* you can't imemdiately see when marking some text with a double click
* when scrolling a website the content makes a "jump" when moving the mouse / the screen is updated
* the mouse feels a litte bit laggy because of the delay
* in Games every *~45* seconds the screen hangs for a second. But i can't really explain why this does happen or if its connected to this issue

For demonstrational purporses I have created a [simple demo video](https://drive.google.com/file/d/12aKzgClu7n1pm-moCPYhwoZ6-jlhCaqQ/view?usp=sharing) showing the correct and excpected result on the left side and the output of moonlight on the right side. Mabe its helpful to understand the problem.

### Expected Behavior

All frames are transmitted and showing in real time.

### Additional Context

When using software encoding or VAAPI in Linux this issue ist not present. Also on my old RX 570 this issue did not occure, but there were other problems ... 😅

You can improve the problem drasticly with activating the tracking overlay with a transparency of 99% and a sampling rate of 0,25 seconds. Because then all 0,25 seconds the screen ist updated, and you can't actually feel the listed problems earlier.

Windows 11

### Architecture

64 Bit

### Sunshine Version

0.12

### GPU Type

AMD

### GPU Model

RX 6600 XT

### GPU Driver/Mesa Version

22.3.1 (Newest)

_No response_

psyke83:
## Description
* platform: replace generic 1000ms capture delay with real frame delay
* for sync encoding, ensure that a frame is sent within frame to frame interval
regardless of timeout status.

This resolves the issue in which the last captured frame is delayed which
is especially noticeable in low framerate content (such as desktop streaming)
on certain encoders.

Please note that I have only verified capture delay as fixed on Windows via AMF, so nvenc and Linux capture needs to be validated before merging.

### Issues Fixed or Closed
Fixes #122, #387, #412

## Type of Change

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -45,6 +45,7 @@ bl::sources::severity_logger<int> error(4);   // Recoverable errors
 bl::sources::severity_logger<int> fatal(5);   // Unrecoverable errors
 
 bool display_cursor = true;
+bool force_callback = false;
 
 using text_sink = bl::sinks::asynchronous_sink<bl::sinks::text_ostream_backend>;
 boost::shared_ptr<text_sink> sink;
--- a/src/main.h
+++ b/src/main.h
@@ -12,6 +12,7 @@
 
 extern util::ThreadPool task_pool;
 extern bool display_cursor;
+extern bool force_callback;
 
 extern boost::log::sources::severity_logger<int> verbose;
 extern boost::log::sources::severity_logger<int> debug;
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -573,14 +573,16 @@ capture_e display_vram_t::capture(snapshot_cb_t &&snapshot_cb, std::shared_ptr<:
     }
     next_frame = now + delay;
 
-    auto status = snapshot(img.get(), 1000ms, *cursor);
+    auto status = snapshot(img.get(), force_callback ? std::chrono::duration_cast<std::chrono::milliseconds>(delay * 2) : 1000ms, *cursor);
     switch(status) {
     case platf::capture_e::reinit:
     case platf::capture_e::error:
       return status;
     case platf::capture_e::timeout:
-      std::this_thread::sleep_for(1ms);
-      continue;
+      if(!force_callback) {
+        std::this_thread::sleep_for(1ms);
+        continue;
+      }
     case platf::capture_e::ok:
       img = snapshot_cb(img);
       break;
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -240,6 +240,7 @@ enum flag_e {
   H264_ONLY         = 0x02, // When HEVC is too heavy
   LIMITED_GOP_SIZE  = 0x04, // Some encoders don't like it when you have an infinite GOP_SIZE. *cough* VAAPI *cough*
   SINGLE_SLICE_ONLY = 0x08, // Never use multiple slices <-- Older intel iGPU's ruin it for everyone else :P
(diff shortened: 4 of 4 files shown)


ReenigneArcher:
Thank you for this PR (and all of your contributions)! Are you on our discord server https://app.lizardbyte.dev/discord? We can ask for testers there. On Linux I can test X11 with software encoding only at this point but to be honest I never noticed these delays in that scenario before.

psyke83:
I joined the discord (same username as here on github).

From looking at the encoders listed in `video.cpp`, the only encoder that need to be tested to see if the frame delay issue is fixed is nvenc on Windows; I have no NVIDIA hardware available to test. The other platform drivers should not be impacted by the change in capture delay, but I was also able to verify that Linux VAAPI capture still works OK in addition to AMF/software on Windows.

istori1:
I tested https://www.testufo.com/framerates#count=1&background=stars&pps=960 on a build with these changes. It looks like the motion is affected. It jumped more than before.

OS: Linux (Flatpak on Fedora 36)
Capture Method: KMS (X11)
GPU: Intel
Encoder: h264_vaapi

psyke83:
@istori1 

Thanks, I'll look into it. I'll need to reinstall Linux on my desktop first, so it will take some time. If some encoders are too sensitive to the low capture interval, we can use a conditional check for setting the delay such as `resend_frame ? std::chrono::duration_cast<std::chrono::milliseconds>(delay) : 1000ms`. This would preserve the same behaviour as before the patch on async encoders.

If you use moonlight-qt, it might help to enable the statistics (Ctrl + Alt + S) and see if the incoming framerate is reduced from normal.

ReenigneArcher:
@psyke83 I had an idea, maybe not a good one, but... Would it be possible to have a setting for each application that could be named "desktop mode" (or "game mode")? Then depending on the application and the setting it would stream in the "mode" more desirable for either desktop or gaming. I don't know how difficult it would be to implement on a per app basis.

psyke83:
I actually found a proper fix to this. Will update the PR soon after I research if the issue can also fix nvenc encoding.

psyke83:
I've refreshed the patches. I may have jumped the gun in saying that I found the proper fix, unfortunately. 

In the case of AMF, it seems that the encoder buffers some frames even though "ultralowlatency" is supposed to encode a frame immediately after a frame is submitted. Reducing the buffer from 16 to 1 eliminates the delay, but the encoder also can no longer run at 60fps. Setting buffers to 2 allows the encoder to run at full speed and reduces latency, but doesn't eliminate the delayed frame issue completely, so by forcing a callback when a timeout the span of 3 frames passes, it greatly improves the situation to the point that I can barely notice any delay on desktop content, and I would be absolutely comforable to use Sunshine as a VNC replacement.

The refreshed patch will only change behaviour for `amdvce` and `nvenc` on Windows, so the latter is what needs to be tested.

psyke83:
This PR (or any other kind of hack) will be unnecessary if https://github.com/LizardByte/ffmpeg-prebuilt/pull/19 is merged and Sunshine is linked against newly patched prebuilts. Note that the patch needs to be present on the master branch for it to take effect.

psyke83:
Closing this test PR; correct proper fix is here: https://github.com/LizardByte/build-deps/pull/18

PR #507, opened 2022-11-19
video: reduce nvenc delay/async_depth to 0

psyke83:
## Description

If `delay` is not set manually, it defaults to INT_MAX, which is then reduced to number of surfaces - 1, which can potentially cause encode delay.

See: 
https://github.com/FFmpeg/FFmpeg/blob/eb1e359a14e04367239a784506e1ae414512a104/libavcodec/nvenc_hevc.c#L110-L111
https://github.com/FFmpeg/FFmpeg/blob/eb1e359a14e04367239a784506e1ae414512a104/libavcodec/nvenc.c#L961
https://forums.developer.nvidia.com/t/nvenc-hevc-ultra-low-latency-with-ffmpeg-libraries-what-should-be-my-expectations/143954/5

Note that I don't own NVIDIA hardware, so this change needs testing by someone else.

## Type of Change

## Checklist

### DIFF
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -418,6 +418,7 @@ static encoder_t nvenc {
   AV_PIX_FMT_NV12, AV_PIX_FMT_P010,
   {
     {
+      { "delay"s, 0 },
       { "forced-idr"s, 1 },
       { "zerolatency"s, 1 },
       { "preset"s, &config::video.nv.preset },
@@ -428,6 +429,7 @@ static encoder_t nvenc {
   },
   {
     {
+      { "delay"s, 0 },
       { "forced-idr"s, 1 },
       { "zerolatency"s, 1 },
       { "preset"s, &config::video.nv.preset },


Isalight:
Hi!
Quickly tested this build.
Windows 11, 3060 Ti
Checked with 90 fps and 40 Mbps.
Works fine.
Does this build include changes from here https://github.com/LizardByte/Sunshine/pull/485? I noticed that, as in that PR (https://github.com/LizardByte/Sunshine/pull/485), the desktop works fine, without frame delays. But unlike from resend-frame PR, according to the statistics overlay, FPS drops to 1-2 if there is no movement on the screen. Whereas with frame-resend PR fps drops to ~30 if nothing happens.

psyke83:
@Isalight 

Thanks for testing. #485 is a hack to workaround the issue of ffmpeg buffering multiple frames seen in nvenc and amfenc, but this PR appears to be the legitimate fix (albeit for nvenc only). If this PR allows the encoder throttles down to 1fps at idle but doesn't have delayed frames or excessive lag, I would consider the issue fixed.

As for #485, I am going to try some experiments with local builds of ffmpeg to properly remove the encoder delay with the amfenc codec... or, worst case scenario, I now know that I can whatever ugly workaround may be needed to only apply to amfenc in future. Thanks.

Isalight:
Ok! If there are new changes I can test AMD too.

pinealservo:
I just tried out this build too, and it resolved the problems I was having with screen updates when streaming my Win10 desktop. Am using a 3080 w/nvenc.

Docmine17:
### Left: PC    /    Right: Raspberry pi 3

sunshine git + this pull

https://user-images.githubusercontent.com/67164552/207441824-42925692-8f08-4214-93db-fd61d35831a3.MOV

0.15 .rpm

https://user-images.githubusercontent.com/67164552/207441833-49407d20-d43a-45f6-adba-9927f1160d0f.MOV

The difference is noticeable, in version 0.15 the latency is higher, making gameplay difficult.


Spec: 
Fedora 37 Workstation, GTX 1050 Ti

Chronometer:
https://cronometronline.com.br/

Docmine17:
With this pull request the latency was around ~16-29ms

psyke83:
Thanks for doing the latency testing - very informative!

I believe that 0.15 was operating with `async_depth` set to 3 frames (assuming `number_of_surfaces` was 4, as reported by the user on the nvidia forums linked at the beginning of the PR), which for 60fps streaming could add up to ~50.1ms (~16.7ms x 3) additional latency.  It now appears to be 1-2 frames latency vs the host.

PR #515, opened 2022-11-27
Avoid using functions deprecated in OpenSSL 3.0

cgutman:
## Description
This PR fixes deprecation warnings in crypto.cpp due to use of legacy low-level SHA256/RSA APIs. I have replaced these with the high-level EVP equivalents as suggested by the OpenSSL 3.0 migration guide.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/crypto.cpp
+++ b/src/crypto.cpp
@@ -4,8 +4,6 @@
 #include <openssl/pem.h>
 
 namespace crypto {
-using big_num_t = util::safe_ptr<BIGNUM, BN_free>;
-// using rsa_t = util::safe_ptr<RSA, RSA_free>;
 using asn1_string_t = util::safe_ptr<ASN1_STRING, ASN1_STRING_free>;
 
 cert_chain_t::cert_chain_t() : _certs {}, _cert_ctx { X509_STORE_CTX_new() } {}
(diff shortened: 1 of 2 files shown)


cgutman:
There are still deprecated functions in use in the `Simple-Web-Server` submodule. If you fork that into the `LizardByte` org, I can submit a PR to fix those too.

ReenigneArcher:
Should we try to use the original Simple Web Server instead of trying to maintain a fork? It seems to still be actively developed. https://gitlab.com/eidheim/Simple-Web-Server

I've also forked Loki's fork, here: https://github.com/LizardByte/Sunshine-Simple-Web-Server

cgutman:
It looks like some of the things we needed to patch in (like local endpoint address info) are now available upstream.

We may still need some out-of-tree patches, but we should at least rebase our fork on upstream. I'll play with the upstream repo and see what it takes to make it usable for us.

ReenigneArcher:
Perhaps change the submodule to point to upstream, and then we can apply any required patches similar to this (https://github.com/LizardByte/build-deps/pull/18/files) where needed?

cgutman:
It turns out that we can actually migrate to the upstream version without needing any patches. See #517.

PR #571, opened 2022-12-17
Add IPv6 streaming support

### LINKED ISSUE #234: IPv6 not supported
### Describe the Bug

Currently, IPv6 is not used by any of the network sockets.

If you run `netstat -lnp | grep sunshine`, you can see all of the sockets bound to `0.0.0.0`, not `[::]`. Also, if you look at https://github.com/SunshineStream/Sunshine/blob/master/sunshine/network.cpp, you can see it explicitly only binds with AF_INET.

On the server side, my ISP at puts me behind an IPv4 CGNAT, but has full end-to-end IPv6 connectivity.

And on the client side, both locations where I'd use it (either at work, or via my mobile) are also IPv6–only. (VPNs are not always feasible due to corp policy reasons.)

### Expected Behavior

The Moonlight client supports IPv6. I expected to be able to connect to Sunshine using IPv6, but wasn't able to due to its lack of support.

### Additional Context

_No response_

Fedora 36

### Architecture

x86_64

### Sunshine Version

0.14.0

### GPU Type

AMD

### GPU Model

RX 6900XT

### GPU Driver/Mesa Version

22.0.1

Wayland

cgutman:
## Description
This PR introduces a new "Address Family" option in the "Advanced" settings page which can be used to select between IPv4-only and IPv4+IPv6 modes for streaming. IPv4-only is the default to match the current behavior.

I've also implemented IPv6 Firewall Control via UPnP (based on the GSv6Fwd code) which is supported by some routers. This will automatically open the IPv6 firewall to allow communication when UPnP is enabled, similar to how we create port mappings via UPnP for IPv4. Some routers that don't support UPnP IPv6 firewall control support PCP for pinholes instead, but I will implement that in a future PR. For now, users can open the IPv6 firewall manually in those cases.

The dual address family mode depends on OS support for the `IPV6_V6ONLY` socket option which allows us to have a single socket that can communicate with IPv4 and IPv6 endpoints. There is currently no IPv6-only option due to the need to plumb support for that `IPV6_V6ONLY` option into `Simple-HTTP-Server`. On OSes like OpenBSD that lack support for `IPV6_V6ONLY` , IPv6+IPv4 mode behaves like IPv6-only.

Note: The web UI is dual-stack, however it doesn't accept on-link interface addresses as LAN. Any GUA will be treated as a WAN address even if it's on-link.

### Screenshot
Screenshot from 2022-12-16 18-27-26

### Issues Fixed or Closed

- Fixes #234 
- Fixes #390 

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -7,55 +7,28 @@
 #include <algorithm>
 
 using namespace std::literals;
+
+namespace ip = boost::asio::ip;
+
 namespace net {
-  // In the format "xxx.xxx.xxx.xxx/x"
-  std::pair<std::uint32_t, std::uint32_t>
-  ip_block(const std::string_view &ip);
-
-  std::vector<std::pair<std::uint32_t, std::uint32_t>> pc_ips {
-    ip_block("127.0.0.1/32"sv)
+  std::vector<ip::network_v4> pc_ips_v4 {
+    ip::make_network_v4("127.0.0.0/8"sv),
   };
-  std::vector<std::tuple<std::uint32_t, std::uint32_t>> lan_ips {
-    ip_block("192.168.0.0/16"sv),
-    ip_block("172.16.0.0/12"sv),
-    ip_block("10.0.0.0/8"sv),
-    ip_block("100.64.0.0/10"sv),
-    ip_block("169.254.0.0/16"sv)
+  std::vector<ip::network_v4> lan_ips_v4 {
+    ip::make_network_v4("192.168.0.0/16"sv),
+    ip::make_network_v4("172.16.0.0/12"sv),
+    ip::make_network_v4("10.0.0.0/8"sv),
+    ip::make_network_v4("100.64.0.0/10"sv),
+    ip::make_network_v4("169.254.0.0/16"sv),
   };
 
-  std::uint32_t
-  ip(const std::string_view &ip_str) {
-    auto begin = std::begin(ip_str);
-    auto end = std::end(ip_str);
-    auto temp_end = std::find(begin, end, '.');
-
-    std::uint32_t ip = 0;
-    auto shift = 24;
-    while (temp_end != end) {
-      ip += (util::from_chars(begin, temp_end) << shift);
-      shift -= 8;
-
-      begin = temp_end + 1;
-      temp_end = std::find(begin, end, '.');
-    }
-
-    ip += util::from_chars(begin, end);
-
-    return ip;
-  }
-
-  // In the format "xxx.xxx.xxx.xxx/x"
-  std::pair<std::uint32_t, std::uint32_t>
-  ip_block(const std::string_view &ip_str) {
-    auto begin = std::begin(ip_str);
-    auto end = std::find(begin, std::end(ip_str), '/');
-
-    auto addr = ip({ begin, (std::size_t)(end - begin) });
-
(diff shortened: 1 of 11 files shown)


src/network.cpp
```diff
@@ -10,7 +10,7 @@ namespace net {
 std::pair<std::uint32_t, std::uint32_t> ip_block(const std::string_view &ip);
 
 std::vector<std::pair<std::uint32_t, std::uint32_t>> pc_ips {
-  ip_block("127.0.0.1/32"sv)
+  ip_block("127.0.0.0/8"sv)
```
cgutman, line 13: This was a little drive-by fix. IPv4 loopback is the whole `127.0.0.0/8` subnet, not just `127.0.0.1`.

psyke83:
Tested on my IPv6 router:

* On server: enabled IPv4+IPv6
* On client: log of `flatpak run com.moonlight_stream.Moonlight` shows both IPv4 and IPv6 addresses; connects fine.
* On client with IPv4 temporarily disabled (`ip address remove ...`): log shows IPv6 address only; connects fine.

psyke83:
Update: there seems to be a problem with establishing a connection after disconnecting and reconnecting multiple times (testing with mixed IPv4+IPv6 on client).

The Moonlight log doesn't mention any errors related to connecting, but the GUI pops up this message:

`Starting control stream establishment failed: Error 110
Check your firewall and port forwarding rules for port(s) UDP 47999`

It seems that I can't connect the server again until I manually restart the service.

Server is Windows 10 22H2, client is Ubuntu 22.10.

i8372370:
好耶

BigBrainAFK:
The WebUI can be made accessible via IPv6 if we refer to ULA specified in RFC 4913 which gives a range of non globally routable IPv6 addresses and how they are structured.

dongzheng9:
Web UI: [127.0.0.1] -- not authorized        TAT

euroclydon37:
I know this is a PR and not an issue or discussion, but I just wanted to say thank you for working on this and I wish I had the adequate CPP knowledge to contribute.

src/network.cpp
euroclydon37: I love that we have ~16 million loopback addresses. lol

euroclydon37:
Also, I built a beefy server with 2 gpus and I'm passing them each through to VMs and letting a couple friends use them remotely Stadia-style. Without IPv6 support, I'll have to do some much-more-complicated networking configuration.

All that to say that I'm totally willing to do some testing to help this along when you're ready.

Again, thank you a ton for working on this and I'm sure you're prioritizing a lot of different work streams. So no pressure whatsoever. Just expressing interest in the feature and trying to be of some assistance. lol

cgutman:
Now that the major dependency of IPv6 is done in https://github.com/LizardByte/Sunshine/pull/1569 and new versions of the [GameStream IPv6 Forwarder](https://github.com/moonlight-stream/GS-IPv6-Forwarder/releases) and [Moonlight Internet Hosting Tool](https://github.com/moonlight-stream/Internet-Hosting-Tool/releases) are shipped that no longer interfere with IPv6 on the default GameStream ports, IPv6 support should be ready now.

euroclydon37:
You're the best!

PR #584, opened 2022-12-19
api: switch to http api, add config schema and clean up

TheChoconut:
## Description
Move Web UI out of support on main executable and switched to an API solution. This should allow different UI clients to interact with Sunshine instead of just one web site.

I have coded another Web UI project, but the old one can also be updated to support this API.

I really wanted to clean up the current code. Having a function for handling every web UI file is really messy. Also, the current code requires a lot of copy&paste since every request needs to be printed for debugging, checked for authentication etc. Now, my code is also not that clean but if you have any suggestions on how to improve it, I'd be thankful.

Added new features:
- HTTP-only API
- Only allow access to IPs from local address space (localhost, 127.0.0.1, no LAN, no WAN)
- added support for [SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) 
- added a config schema

#### Security
No added security. Now everything is handled via HTTP. For that reason, it is hardcoded that requests can only be made from localhost or 127.0.0.1. Using the API outside of localhost should be handled via reverse proxy or a similar solution.

## Type of Change

## Checklist


TheChoconut:
This PR should do fewer things at once. It will not support i18n, I would consider adding i18n support later. The target for this PR is only to introduce API as a solution. The old website's logic needs to be rewritten for the new API.

I would also consider removing the website from this repo and moving it to a new one... Eventually, the main goal would be to make a UI for Sunshine that is newbie friendly and presents basic settings in a user-friendly way.

ReenigneArcher:
> This PR should do fewer things at once. It will not support i18n, I would consider adding i18n support later. The target for this PR is only to introduce API as a solution. The old website's logic needs to be rewritten for the new API.
> 
> I would also consider removing the website from this repo and moving it to a new one... Eventually, the main goal would be to make a UI for Sunshine that is newbie friendly and presents basic settings in a user-friendly way.

Sounds good! There is a bunch of stuff already in this repo to help with localization. You just need to set your strings so they are identified as strings to translate. I think this is all you need to do... https://docs.lizardbyte.dev/projects/sunshine/en/latest/contributing/localization.html#extraction

Then the localization can happen on CrowdIn, by the community. I've tested this logic in a python project after you suggested it previously, and it's pretty cool to be honest. Our website is also localized on CrowdIn.

TheChoconut:
Okay, looks like you guys updated submodules and I haven't kept up. Somehow I was working on a wrong branch so it was compiling just fine on master but nightly has new submodules so build failed. Working on fixing this.

TheElixZammuto:
>Password checking is implemented. The default password is created on run and it encrypts a new file, "sunshine_creds.data" >using AES GCM. The password is printed in stdout. When authenticating using a password, the file is decrypted using a partial >password hash (SHA-256) sent from the client. If the password is correct, it will be successfully decrypted and the full hash will > be compared.

I don't understand the rationale behind this implementation and what advantages brings to the table instead of the current method of authentication

Also using stdout breaks Windows Services since it would be inaccessible from the user when the default installation also creates the service

TheElixZammuto:
> The current solution cannot be used because it's not standard in APIs to have to read a header in order to authenticate. Mostly you would have a token that is used to identify a user and that token should be secret so no one else can access the API but the user. Now currently you have a password because remembering tokens would be more difficult.

Personally I have used many APIs that uses the Authentication Header to authenticate. While more complex apps uses OpenID/OAuth2 or similar solutions, they always use the Authentication Header of a random (or JWT) token in practice. Basic Authentication is fairly easy to consume and implement.

> This solution is not secure because HTTP data is not encrypted. So if data was intercepted, you can read the token and get full access to the API. In a local environment, HTTP should be safe enough. But if HTTP is moving through a router on LAN or WAN then it's a security risk. So in these cases, reverse proxy or a similar solution is recommended.

I also read that by default the API is exposed to localhost only, is that right? If so I don't see any security problem since it stays on the machine. I also reccomend the reverse proxy approach. Hopefully If the user if smart enough to setup a reverse proxy, I think is smart enough to install certbot and enable https.

> I decided to use this method because it allows Sunshine to never know or store the user password. If you get access to the file, you need to decrypt the file in order to get the hash required to authenticate. Does that improve security, I am not sure.

The current approach does already hash and salt the user's password so it doesn't get stored on the computer storage. Also keep in mind that when you send the password hash over the server instead of the plain text, the hash becomes the actual password.  

> If anyone has a better idea on authentication, let me know. I think that the current solution has the same flaws but it seems pointless to invent a secure authentication method that is basically HTTPS (encrypting and decrypting all data based on public and private key encrption).

I would add an IP allowlist system to the already existing system. By default only localhost can access the API/UI. Users can add their reverse proxy IP (with all the protections in place) or their trusted devices IP if needed.

ReenigneArcher:
> I also read that by default the API is exposed to localhost only, is that right?

The original documentation from the old config example stated that the `/pin` endpoint was only available on the localhost by default, while the Web UI was available on LAN by default. If the documentation is to be believe, I would assume the rest of the API probably also only works on localhost by default as well.

TheElixZammuto:
btw I'm more than happy to improve the frontend side of things with the changes provided here

TheChoconut:
> The current approach does already hash and salt the user's password so it doesn't get stored on the computer storage. Also keep in mind that when you send the password hash over the server instead of the plain text, the hash becomes the actual password.

Yeah, hash as a password was the goal as you never interact with the password of the user, as some users use one password on multiple services (a very bad practice). But this updated method brings no additional security to the authentication process. I think the right way to improve security would be to implement a challenge-response authentication protocol. But that is a bit too much for me, I would leave that for another PR.

I still like that we are using tokens for most of the requests as they cannot be reused after they expire. But since they are also not secure, I can consider removing them.

I can implement the Basic protocol. If you think that's the right way, I would remove the token system and leave only Basic authentication using the Authorization header.

TheElixZammuto:
Keep in mind, my rationale for the comments I made is to make sure that we are not adding so much effort and complexity to the authentication system that it actually goes against us in security and complexity (it's very easy to screw up sadly).
I don't know if others agree with me, but I think what we should do is:

1) Transforming or opening a PR for the API improvements (so Config schema and SSE)
2) Opening an RFC (over here or Discord) to think with the community what should be the best approach to authentication (hopefully both users and developers would give their insight here)
3) Implementing point 2 :^)

What do you guys think?

TheChoconut:
That's exactly what I just did. I reverted back to Basic authentication and only changed the following:
- move to HTTP-only
- only allow IP from local address space (no LAN, no WAN)
- add config schema
- add support for SSE

I also made the API a little bit cleaner but I would like to add some sort of self-commenting code so that an API doc can be generated, if something like that exists?

kayakyakr:
@TheChoconut No one responded to your last message, but if it's self-documenting cpp code you're looking for, I think Doxygen is considered the universal standard: https://www.doxygen.nl/

ReenigneArcher:
> @TheChoconut No one responded to your last message, but if it's self-documenting cpp code you're looking for, I think Doxygen is considered the universal standard: https://www.doxygen.nl/

I've read Doxygen can be integrated with Sphinx, so that would be a great option in my opinion.

PR #600, opened 2022-12-21
Launch games under the correct user account on Windows

cgutman:
## Description
This implements a new `platf::run_unprivileged()` function that can be used (on Windows, currently) to launch a process as the active console user (including their environment variables).

In order to make this work, I had to fix issues with the way we derive the working directory if one is not provided. That required adding a dependency on `libboost-program-options` for `boost::program_options::split_unix()` to perform proper command-line parsing to retrieve the actual binary name.

This could definitely use some testing since there are a _ton_ of use cases to cover.

### Issues Fixed or Closed
Resolves #521 

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/process.cpp
+++ b/src/process.cpp
@@ -10,10 +10,12 @@
 
 #include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/program_options/parsers.hpp>
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
 
 #include "main.h"
+#include "platform/common.h"
 #include "utility.h"
 
 namespace proc {
@@ -35,7 +37,7 @@ void process_end(bp::child &proc, bp::group &proc_handle) {
   proc.wait();
 }
 
-int exe(const std::string &cmd, bp::environment &env, file_t &file, std::error_code &ec) {
+int exe_with_full_privs(const std::string &cmd, bp::environment &env, file_t &file, std::error_code &ec) {
   if(!file) {
     return bp::system(cmd, env, bp::std_out > bp::null, bp::std_err > bp::null, ec);
   }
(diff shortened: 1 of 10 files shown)


psyke83:
I believe that something may be wrong with path handling (but it applies to this PR and not #599). I tested with and without the complementary PR in case it helps to troubleshoot.

Test case (3 created test applications)
* Add Notepad1, with command as `notepad`
* Add Notepad2, with command as `C:\Windows\System32\notepad.exe`
* Add Notepad3, with command as `C:\Windows\System32\notepad.exe` and working directory as `C:\Windows\System32\`

Results:
* Nightly: Notepad1 fails (`Warning: Couldn't run [notepad.exe]: System: The filename, directory name, or volume label syntax is incorrect`). Notepad2 and Notepad3 work. I presume this is expected.
* #599: identical to nightly.
* #600: Notepad1 and Notepad3 works. Notepad2 fails with an error in Moonlight-QT: `GeForce Experience returned error: Malformed GFE XML (missing root element) (Error -1)`, but nothing logged in Sunshine.
* #599 + #600: identical to #599 alone.

In summary, this PR makes `notepad` available (as it is in `$Path`), but `C:\Windows\System32\notepad.exe` fails to launch unless you manually add the working directory (which should not be necessary according to the Web UI's description).

src/process.cpp
```diff
@@ -35,31 +37,42 @@ void process_end(bp::child &proc, bp::group &proc_handle) {
...
   return bp::system(cmd, env, bp::std_out > file.get(), bp::std_err > file.get(), ec);
 }
 
-int proc_t::execute(int app_id) {
-  if(!running() && _app_id != -1) {
-    // previous process exited on its own, reset _process_handle
-    _process_handle = bp::group();
+boost::filesystem::path find_working_directory(const std::string &cmd, bp::environment &env) {
+  // Parse the raw command string into parts to get the actual command portion
+  std::vector<std::string> parts = boost::program_options::split_unix(cmd);
```
psyke83, line 50: The failure testcase (Notepad2) seems not to proceed beyond this function.

psyke83: I think that the `working_path` is not being detected correctly in other cases too.

e.g.: Command `D:\Software\CinebenchR20\Cinebench.exe` with no working path defined.

Unlike Notepad, this app will launch correctly (because it's *not* in the system path?), but if I add some debugging to the code, it seems that the working directory is being set to `c:/` instead of the correct path `D:\Software\CinebenchR20\`.

cgutman: Yep, the reason is dumb. I am using `split_unix()` which treats `\` as an escape character. Using `split_winmain()` instead should take care of it.

ReenigneArcher: Is it difficult to correct the path separator automatically if the user enters the wrong one?

cgutman: It's non-trivial to try to sort out what the user intended, especially given the variety of ways you can enclose paths with spaces. I don't think that's likely to be a big issue but we can revisit if it turns out to be.

cgutman:
OK, I think I've taken care of all the issues.

- I've changed the behavior for detached commands to use the given working directory as discussed with @ReenigneArcher on Discord (previously working directory was unset for those commands)
- Fixed working directory lookup issues caused by incorrect command parsing (Notepad1, 2, and 3 from @psyke83 's test should be working now)
- Switched from ANSI to Unicode functions for process and environment creation. I verified Unicode support is working with non-ASCII working directories, commands, and environment variables (both in the user environment block and from apps.json in `env`).
- Fixed Flatpak build by adding missing `program_options` library)
- Fixed `clang-format` issue (seems to disagree with my local `clang-format` but I'll defer to the bot)

ReenigneArcher:
> Fixed clang-format issue (seems to disagree with my local clang-format but I'll defer to the bot)

If you have any suggestions for changes to the `.clang-format` file, I'm happy to discuss. That file was provided to me early after I took over sunshine. One thing I personally don't like is the alignment of variable values, but I guess that's normal in c++.

ReenigneArcher:
> Several users on the discord had tested this build and I don't recall any issues. We would likely get more testing if merged into nightly.

This PR is not ready to merge.

cgutman:
`platf::run_unprivileged()` changes:
- Added the `ImpersonateLoggedOnUser()` call to ensure that access checks are done against the user token rather than our SYSTEM token. This also allows launching apps on a network share or network drive.
- Addressed handle inheritance issues in a cleaner way (using `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` rather than manually disabling inheritance on Sunshine's own std handles)
- Added `CREATE_BREAKAWAY_FROM_JOB` to handle changes in #602 that would otherwise terminate the child process if SunshineSvc died
- Added `CREATE_NEW_CONSOLE` to handle the case where log output is not redirected and incorrectly ends up in Sunshine's own log file
- Removed `bp::child()` fallback if `CreateProcessAsUser()` fails. I left that fallback in place just in case `CreateProcessAsUser()` failed when running as non-admin, but that case seems to work just fine (`CreateProcessAsUser()` and `ImpersonateLoggedOnUser()` using your own user token requires no additional permissions).

Log file fixes:
- Fixed Unicode log file path handling on Windows (previously, attempts to use non-ANSI log file paths would silently fail)
- Added `_SH_DENYNO` to ensure reopening log files for applications that are still running will not fail with a sharing violation.

These changes also fix the crash that we saw with UWP notepad.exe with the previous version of this PR.

Michoko92:
Awesome job! 👍

exalented:
With the 17 release I can no longer launch an app on linux:
`Platform:` Arch / Sway / Nvidia, running AUR package.
```
Warning: run_unprivileged() is not yet implemented for this platform. The new process will run with Sunshine's permissions.
terminate called after throwing an instance of 'boost::process::process_error'
  what():  killpg(2) failed in terminate: Invalid argument
Aborted (core dumped)
```

PR #634, opened 2022-12-27
Logs inside the WebUI

TheElixZammuto:
## Description

Description
Logs are now stored inside the filesystem other than the stdout. This allows the WebUI to fetch and display the logs inside the "troubleshooting" section. This is especially useful on Windows when Sunshine runs as a Service (which is the default option now)

This needs some more testing on Linux platforms

### Screenshot

web-ui-logs

### Issues Fixed or Closed

N/A

## Type of Change

## Checklist

## Branch Updates


src_assets/common/assets/web/troubleshooting.html
```diff
@@ -90,3 +123,14 @@ <h2>Unpair All Clients</h2>
...
+<style>
+  .troubleshooting-logs {
+    white-space: pre;
+    font-family: monospace;
+    overflow: auto;
+    max-height: 500px;
+    min-height: 500px;
+    font-size: 16px;
+  }
+</style>
```
ReenigneArcher, line 136: Can you add a blank line to the end of the file?

src/main.cpp
```diff
@@ -171,6 +178,9 @@ int main(int argc, char *argv[]) {
 
   boost::shared_ptr<std::ostream> stream { &std::cout, NoDelete {} };
   sink->locked_backend()->add_stream(stream);
+  auto path         = std::filesystem::temp_directory_path();
```
ReenigneArcher, line 181: What do you think of saving the log next to the config/app files?

ReenigneArcher (commented):
This probably needs to be changed as well: https://github.com/LizardByte/Sunshine/blob/70ae7a2fa9b33173eb949c86ec42a96e7b9e8c25/tools/sunshinesvc.cpp#L65

What do you think of also adding a config option to customize the log file directory?

Would it be difficult to make a rolling log file? i.e. rename the file when it reaches a certain size? Could have 5 files before removing the oldest.

src/main.cpp
KuleRucket: For Linux temp_directory_path() goes to /tmp I think. This would be better than to ~/.config/sunshine/. Better still would be to make it a config setting so it could be directed to /var/log or whatever the user prefers.

TheElixZammuto: Okay, I'm adding the config path for logs, with the sunshine directory being the default one

src_assets/common/assets/web/troubleshooting.html
```diff
@@ -39,16 +35,26 @@ <h2>Unpair All Clients</h2>
...
     </div>
   </div>
+  <!--Logs-->
+  <div class="card p-2 my-4">
+    <div class="card-body">
+      <h2>Logs</h2>
+      <br />
+      <div class="d-flex justify-content-between align-items-baseline py-2">
+        <p>See the logs uploaded by Sunshine</p>
+        <input type="text" class="form-control" v-model="logFilter" placeholder="Find..." style="width: 300px">
```
ReenigneArcher, line 51: I like the idea of this feature! Is it difficult to add a copy button, like this?
image

Wouldn't necessarily need to be inside the log text container.

ReenigneArcher:
Additionally, can you update the docs?

At the bottom of `docs/source/about/advanced_usage.rst` can you add?

```rst
log_path
^^^^^^^^^^^^^^^^

**Description**
   The path where the sunshine log is stored.

**Default**
   ``sunshine.log``

**Example**
   .. code-block:: text

      log_path = sunshine.log
```

src_assets/common/assets/web/troubleshooting.html
TheElixZammuto: Done

PR #640, opened 2022-12-27
windows: Fix audio when host is not using 48 KHz sample rate

### LINKED ISSUE #233: No Audio on host when audio output bitrate is anything other than 48000hz
### Describe the Bug

No audio on host when audio device is set to output any bitrate other than 48khz.

### Expected Behavior

Audio works across different bitrates. (maybe detect and downsample bitrate?)

### Additional Context

only tested with Play audio on PC option enabled in moonlight client.

Windows 10

### Architecture

64bit x86

### Sunshine Version

0.14.0

### GPU Type

AMD

### GPU Model

RX 470

### GPU Driver/Mesa Version

22.5.1

_No response_

cgutman:
## Description
Sunshine requires 48 KHz sample rate audio for compatibility with the Opus encoder. If the primary audio device on the host is not set to 48 KHz, audio will be glitchy or unavailable entirely due to a lack of a compatible 48 KHz loopback format.

Fortunately, Windows includes a built-in resampler that can transparently resample any loopback format to 48 KHz for us. All we have to do is ask for the resampler when creating our audio client and it will do the magic for us.

Relevant docs: https://learn.microsoft.com/en-us/windows/win32/coreaudio/audclnt-streamflags-xxx-constants

### Issues Fixed or Closed
Fixes #233

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/audio.cpp
+++ b/src/platform/windows/audio.cpp
@@ -81,262 +81,10 @@ public:
   PROPVARIANT prop;
 };
 
-class audio_pipe_t {
-public:
-  static constexpr auto stereo     = 2;
-  static constexpr auto channels51 = 6;
-  static constexpr auto channels71 = 8;
-
-  using samples_t = std::vector<std::int16_t>;
-  using buf_t     = util::buffer_t<std::int16_t>;
-
-  virtual void to_stereo(samples_t &out, const buf_t &in) = 0;
-  virtual void to_51(samples_t &out, const buf_t &in)     = 0;
-  virtual void to_71(samples_t &out, const buf_t &in)     = 0;
-};
-
-class mono_t : public audio_pipe_t {
-public:
-  void to_stereo(samples_t &out, const buf_t &in) override {
-    auto sample_in_pos = std::begin(in);
-    auto sample_end    = std::begin(out) + out.size();
-
-    for(auto sample_out_p = std::begin(out); sample_out_p != sample_end;) {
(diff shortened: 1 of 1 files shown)


psyke83:
I need to check in more detail (currently testing in a loud room with TV playing in background), but it seems that this change has forced my stereo headphones (@48KHz, so shouldn't be resampled) into mono on my system.

psyke83:
When the flags you added are used, all formats specified in this list are accepted:

https://github.com/LizardByte/Sunshine/blob/347cb965238af0a12b9203c1997c94e8523c8a19/src/platform/windows/audio.cpp#L348-L385

Since the first item is mono, that's why my (16 bit 48000Hz 2ch) sound was being forced into mono. Reordering the list so that mono is not the first test is also not ideal, as `make_audio_client()` will now accept any format specification. So, depending on how you arrange the list order, my stereo config could be tricked into resampling fake 7:1 audio, or a surround setup could be tricked into being downsampled to stereo or mono.

cgutman:
Yep, I found the same thing. I ripped out all the audio channel conversion logic and just passed the stream config we want straight into `IAudioClient::Initialize()`. That's _way_ simpler than all the stuff we were doing before and fixes the format selection logic too.

cgutman:
Pushed a fix for a bug when Sunshine tries to configure Steam Virtual Speakers for surround output where it could wedge the device into an incompatible output format.

I think this is ready to go now 🤞

PR #647, opened 2022-12-28
Graceful termination on shutdown, logoff, and service stop

cgutman:
## Description
This PR addresses several issues handling service stop, shutdown, and logoff on Windows:
- Moves Sunshine further back in the priority of processes to be terminated to keep connectivity as long as possible during logoff. This allows interaction with the screen that appears when apps are blocking shutdown.
- Adds handling for `WM_ENDSESSION` to allow Sunshine to gracefully close itself when the user session is ending. Unfortunately, we have to use the hidden window solution rather than the simpler `SetConsoleCtrlHandler()` API because the display capture APIs we use cause `gdi32.dll` to be loaded.
- Sends a Ctrl-C to gracefully terminate Sunshine when SunshineSvc is stopped or the system is shutting down. This requires a helper process because we can't attach to Sunshine's console from the service since it's in a different session.
- Fixes a breakage introduced by #602 that prevented Sunshine from spawning again after a logout+login cycle

### Issues Fixed or Closed
Fixes Sunshine not reverting back to the default audio output after logoff, shutdown, or service stop while streaming

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -209,6 +209,13 @@ int main(int argc, char *argv[]) {
       return;
     }
 
+#ifdef _WIN32
+    // If this is running from a service with no console window, don't wait for user input to exit
+    if(GetConsoleWindow() == NULL) {
+      return;
+    }
+#endif
+
     task_pool.cancel(force_shutdown);
 
     std::cout << "Sunshine exited: Press enter to continue"sv << std::endl;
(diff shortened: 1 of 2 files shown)


cgutman:
I ended up finding some additional bugs after doing some testing in another branch that is fixing/removing `std::abort()` calls.

KuleRucket:
It would be great if the "Press enter to continue" was disabled for Linux all together. You would not normally need to hold a terminal window open in Linux so this has always been mildly irritating.

ReenigneArcher:
> It would be great if the "Press enter to continue" was disabled for Linux all together. You would not normally need to hold a terminal window open in Linux so this has always been mildly irritating.

Same for Windows to be honest.

PR #660, opened 2023-01-01
Enable parallel encoding for NVENC and AMF

cgutman:
## Description
This PR introduces the required use of the `ID3D10Multithread` interface and enables `PARALLEL_ENCODING` for NVENC and AMF encoders to hopefully improve performance for capturing Vulkan applications.

This works fine for me (no warnings with the debug runtime using NVENC), but I'd like a test from @psyke83 on AMD to verify that it works fine there and see if it improves performance for Gravity Mark in Vulkan mode.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I want maintainers to keep my branch updated

PR #667, opened 2023-01-01
Fix unexpected format switching in display_vram

cgutman:
## Description
I think I finally figured out what's going on with the unexpected capture format changes. It looks like DWM will cheat by using a surface that is in the desktop format until a frame is returned that that has `LastPresentTime != 0` (which requires the DWM to convert the surface into the format we asked for).

Due to this behavior, #654 introduced a bug that would cause the surface format to constantly flip-flop between the requested capture format (when the desktop is updated) and the desktop format (when returning a cursor-only update). This format-switching happened constantly when requesting `DXGI_FORMAT_R10G10B10A2_UNORM` which led to very poor performance due to constantly reinitializing the capture session. This issue wasn't visible when we tested #654 because `display_vram` reported support for both R8G8B8A8 and B8G8R8A8 desktop formats, so the desktop format always matched the capture format.

To address this, I've added an intermediate texture (similar to the staging texture `display_ram` uses) to store a copy of the last captured desktop frame. If `LastPresentTime == 0`, we copy from that frame instead of `src`.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display.h
+++ b/src/platform/windows/display.h
@@ -185,6 +185,8 @@ public:
   vs_t scene_vs;
 
   gpu_cursor_t cursor;
+
+  texture2d_t last_frame_copy;
 };
 } // namespace platf::dxgi
 
--- a/src/platform/windows/display_base.cpp
+++ b/src/platform/windows/display_base.cpp
@@ -292,22 +292,10 @@ int display_base_t::init(int framerate, const std::string &display_name) {
 
   //FIXME: Duplicate output on RX580 in combination with DOOM (2016) --> BSOD
   {
-    dxgi::output1_t output1 {};
-    dxgi::output5_t output5 {};
-
     // IDXGIOutput5 is optional, but can provide improved performance and wide color support
+    dxgi::output5_t output5 {};
     status = output->QueryInterface(IID_IDXGIOutput5, (void **)&output5);
-    if(FAILED(status)) {
-      BOOST_LOG(warning) << "Failed to query IDXGIOutput5 from the output"sv;
-    }
-
-    status = output->QueryInterface(IID_IDXGIOutput1, (void **)&output1);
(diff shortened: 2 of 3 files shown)


psyke83:
This commit as-is doesn't seem to cause any noticeable problems with my AMD card & SDR only setup, but it has a bad interaction when tested together with #660 (so feel free to disregard in case you are aware):

When connecting to the host with no significant movement on-screen, the first frame received will look OK, but moving the mouse will trigger a full black frame to display and then the desktop image again, which repeats with each successive mouse movement for perhaps 3-5 seconds, and then the video will appear to stabilize with no further black frames being inserted. There is no log activity when this happens.

cgutman:
Ok, thanks for testing. I think I'm going to take a different approach with #660 and try using 2 separate ID3D11Device objects (one for capture and one for encode) then pass textures between them using `D3D11_RESOURCE_MISC_SHARED_NTHANDLE` and `ID3D11Device1::OpenSharedResource1()`.

cgutman:
While I was redoing #660, I figured out what the problem was and it was indeed a bug in this PR, though one that's only likely to show up with `PARALLEL_ENCODING` because `DEFAULT` only uses a single image. I forgot to call `complete_img()` in the newly added codepath, so we'd pass back an image with texture inside.

cgutman:
I pushed an additional fix for an issue in #654 that I discovered when testing my parallel encoding work. We don't actually want to fall back to `DuplicateOutput()` if `DuplicateOutput1()` is available.

There are temporary reasons that desktop duplication fails (such as being in the middle of a modeset operation or the secure desktop being displayed when not running as service). If we happen to fail our 2 tries of `DuplicateOutput1()` and succeed in the call to `DuplicateOutput()`, then we'll be stuck with suboptimal capture performance until something else triggers the capture to reinitialize.

PR #668, opened 2023-01-01
Use separate encoding and capture devices to enable parallel encoding and capture

cgutman:
## Description
This PR modifies `hwdevice_t` to use a separate `ID3D11Device` and `ID3D11DeviceContext` to allow encoding to happen in parallel with capture without violating DXGI synchronization requirements. Each `img_t` has a texture created on the capture-side by `display_vram_t::complete_img()` which is then opened via handle on the encoding side by `hwdevice_t::share_img()`. 

The only concurrency bottleneck now is `display_vram_t::snapshot()` and `hwdevice_t::convert()` where they use a `IDXGIKeyedMutex` to synchronize access to the same `img` texture. Since `PARALLEL_ENCODING` creates a pool of images, this is unlikely to a problem in practice.

This PR temporarily includes the changes in #667 since they are necessary for this to work. Once #667 is merged, I'll rebase this to remove that commit.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display.h
+++ b/src/platform/windows/display.h
@@ -18,6 +18,10 @@
 namespace platf::dxgi {
 extern const char *format_str[];
 
+// Add D3D11_CREATE_DEVICE_DEBUG here to enable the D3D11 debug runtime.
+// You should have a debugger like WinDbg attached to receive debug messages.
+auto constexpr D3D11_CREATE_DEVICE_FLAGS = D3D11_CREATE_DEVICE_VIDEO_SUPPORT;
+
 template<class T>
 void Release(T *dxgi) {
   dxgi->Release();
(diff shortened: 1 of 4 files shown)


psyke83:
Before commenting on performance, the stuttering that was present in #660 and black frame issue in the the older version of #667 are no longer present.

Testing with 1440p host resolution, capture 1440p/60 @ 90MBps. GravityMark's real performance stays in the range of 80-130fps, average ~110fps (never any major dips).
* Vulkan: average capture is ~55fps with two very noticeable dips down to 35fps (asteroid field close-ups).
* D3D12: average capture is 45-50fps.
* D3D11: average capture is solid 60fps (very small dips on scene change)

Further testing with AMD's built-in Chill framerate limiter:
* Vulkan via Chill @ 60fps min/max: capture rate runs at a solid 60fps with no noticeable dips.
* Vulkan via Chill @ 80fps min/max: capture rate runs at a solid 60fps, perhaps or or two minor dips to 55fps.
* D3D11 via Chill @ 60fps min/max: capture rate runs at ~50fps with frequent dips [N.B. this also happens without this  on latest nightly, so it's not an indication of something wrong with this PR]
* D3D11 via Chill @ 80fps min/max: capture rate runs at solid 60fps.
* D3D12: n/a (Chill cannot limit framerate in GravityMark with this API)

In summary, it seems 1) uncapped rates may be causing GPU contention that results in the encoder throttling in Vulkan and D3D12 (but not D3D11), and 2) the D3D11 capped @60fps tests result result seems quite counter-intuitive (but the poor performance in that case is unrelated to this PR).

I need to test some more with other non-Vulkan applications, but so far it looks very promising and is definitely an improvement compared to nightly.

psyke83:
Updated PR (with call to SetGPUThreadPriority() for second device) seems to perform more or less the same as before.

I also tested some 4K/60 host+client capture. Current nightly will slightly miss the target, hitting about 57fps in Cyberpunk 2077 or even when browsing https://www.vsynctester.com/, but the PR improves the capture rate to reach the 60fps target. Will continue testing and let you know if I encounter any issues.

cgutman:
This PR is ready for merge.

PR #715, opened 2023-01-06
Generated app id with hashed input data

brad-richardson:
## Description
Generate stable app id based on name and image data hash. This allows us to stop persisting ids altogether and prevents us filling up moonlight's cache if ids are changing often.

To test:
- [x] Changing image data without changing path
- [x] Removing image path
- [x] Duplicate apps with same name + image path
- [x] No apps

### Issues Fixed or Closed
- Should fix bug reported on Moonlight discord

## Type of Change

## Checklist

## Branch Updates


src/process.cpp
```diff
@@ -233,47 +238,16 @@ std::vector<ctx_t> &proc_t::get_apps() {
...
-  }
-
-  // check if specified image exists
-  std::error_code code;
-  if(!std::filesystem::exists(app_image_path, code)) {
-    // return default box image if image does not exist
-    return default_image;
+  auto image_path = validate_app_image_path(app_image_path);
+  if(image_path == DEFAULT_APP_IMAGE_PATH) {
+    BOOST_LOG(info) << "Couldn't find non-default app image for ID ["sv << app_id << ']';
```
cgutman, line 248: We probably don't want to print a log message like this unless the user specified an image and we couldn't load it.

src/confighttp.cpp
```diff
@@ -316,13 +315,16 @@ void saveApp(resp_https_t response, req_https_t request) {
...
-      id = util::generate_int32(1, std::numeric_limits<std::int32_t>::max());
+    // Moonlight checks the id of an item to determine if an item was changed, recompute
+    auto possible_ids = proc::calculate_app_id(inputTree.get<std::string>("name"), inputTree.get<std::string>("image-path"), inputTree.get<int>("index"));
+    if(ids.count(std::get<0>(possible_ids)) == 0) {
+      // Avoid using index to generate id if possible
+      inputTree.put("id", std::get<0>(possible_ids));
+    }
+    else {
+      // Fallback to include index on collision
+      inputTree.put("id", std::get<1>(possible_ids));
```
cgutman, line 326: This ID written into apps.json isn't used anymore, is it? The code in `parse()` ignores it and uses `calculate_app_id()` instead (as it should). We should remove this and just not write an ID field into apps.json at all.

brad-richardson: Yeah, I forgot we just follow the same `parse()` logic after this, I'll remove it

src/process.cpp
```diff
@@ -355,6 +326,114 @@ std::string parse_env_val(bp::native_environment &env, const std::string_view &v
...
+    return DEFAULT_APP_IMAGE_PATH;
+  }
+
+  // image is a png, and not in assets directory
+  // return only "content-type" http header compatible image type
+  return app_image_path;
+}
+
+std::optional<std::string> calculate_sha256(const std::string &filename) {
+  EVP_MD_CTX *mdctx = EVP_MD_CTX_new();
```
cgutman, line 367: This needs to be freed. You can use the `crypto:: md_ctx_t` defined in crypto.h to handle this:
```
crypto::md_ctx_t ctx { EVP_MD_CTX_create() };
```

cgutman (commented):
Other than the leak, LGTM

PR #716, opened 2023-01-06
Migrate Windows config files into specific config directory

cgutman:
## Description
This moves config and other data files out of the root of our Program Files directory into a specific `config` directory. Only that new config directory has RW access for the Users group now.

I've tested this locally, but it would be nice for someone else to grab the installer from the CI job to make sure it all works seamlessly for them too.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -532,6 +532,9 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
     install(DIRECTORY "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/misc/service/"
             DESTINATION "scripts"
             COMPONENT service)
+    install(DIRECTORY "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/misc/migration/"
+            DESTINATION "scripts"
+            COMPONENT assets)
 
     # Sunshine assets
     install(DIRECTORY "${SUNSHINE_SOURCE_ASSETS_DIR}/common/assets/"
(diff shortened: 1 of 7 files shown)


src_assets/windows/misc/migration/migrate-config.bat
```diff
@@ -0,0 +1,43 @@
...
+if exist "%OLD_DIR%\credentials\" (
+    if not exist "%NEW_DIR%\credentials\" (
+        move "%OLD_DIR%\credentials" "%NEW_DIR%\"
+    )
+)
+
+rem Migrate the covers directory
+if exist "%OLD_DIR%\covers\" (
+    if not exist "%NEW_DIR%\covers\" (
+        move "%OLD_DIR%\covers" "%NEW_DIR%\"
```
ReenigneArcher, line 37: I haven't tested, but I think this would be a breaking change.

This is an example app where the cover finder feature is used.
```json
        {
            "name": "Halo",
            "output": "",
            "cmd": "",
            "image-path": ".\/covers\/igdb_740.png"
        }
```

Anything like the `sed` command we can use in batch to patch the `apps.json`? We obviously don't want to include another software just for this, but maybe we can invoke a simple powershell command from the bat script? https://www.computerhope.com/forum/index.php?topic=142440.0

I'd definitely prefer to keep the majority in batch, so if it can't be a one liner powershell command, maybe add a `patch-apps.ps` script. Wouldn't be a bad idea to also patch the program files in env/path that you previously fixed as well.

Or instead of patching apps.json we could handle it in `process.cpp`.

Or instead of over complicating this, we can just document that user's need to fix their image paths.

src_assets/windows/misc/migration/migrate-config.bat
```diff
@@ -0,0 +1,43 @@
...
+    )
+)
+
+rem Remove log files
+del "%OLD_DIR%\*.txt"
+del "%OLD_DIR%\*.log"
```
ReenigneArcher, line 43: new line at end of file, please

src_assets/windows/misc/migration/migrate-config.bat
cgutman: The Powershell solution worked

CMakeLists.txt
```diff
@@ -549,12 +552,16 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
     set(CPACK_PACKAGE_INSTALL_DIRECTORY "${CPACK_PACKAGE_NAME}")
 
     # Extra install commands
-    # Sets permissions on the installed folder so that we can write in it
+    # Restores permissions on the install directory
+    # Migrates config files from the root into the new config folder
+    # Sets permissions on the config folder so that we can write in it
```
brad-richardson, line 557: I noticed the default steam config writes its output to `Program Files/Sunshine/steam.txt` as well, do we need to migrate that?

brad-richardson (commented):
Worked for me upgrading from v0.16, just one question

CMakeLists.txt
cgutman: Nah, I think we can just nuke it. Honestly, we probably don't need to store that log output anywhere.

ReenigneArcher: To clarify, it will write them to the `config` directory moving forward, correct?

cgutman: Our template `apps.json` doesn't specify a path (on any platform), so it just goes into Sunshine's current directory (if it's writeable). On Windows, that's the directory with Sunshine.exe. I assume it's `/usr/bin` or something on Linux. Neither are particularly great locations for log files.

If the current directory is not writable, the app still launches but the log file is not written.

cgutman: Removed `steam.txt` from the default apps.json per our discussion on Discord.

PR #721, opened 2023-01-07
Enable Mouse Keys while streaming for systems with no mouse plugged in

cgutman:
## Description
This PR introduces a workaround for the issue where Windows systems with no physical mouse connected would not display the mouse cursor (OS behavior since Windows 8). As a workaround, we can enable Mouse Keys for the duration of the streaming session which will force the mouse cursor to be displayed. Fortunately for us, Mouse Keys doesn't see input from `SendInput()` so we don't have to do anything special to make the number pad work either.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/misc.cpp
+++ b/src/platform/windows/misc.cpp
@@ -28,6 +28,9 @@ using namespace std::literals;
 namespace platf {
 using adapteraddrs_t = util::c_ptr<IP_ADAPTER_ADDRESSES>;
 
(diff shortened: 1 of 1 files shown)


psyke83:
Tested with mouse driver disabled on Windows 10 - seems to be working perfectly.

Mouse Keys is enabled (if not already so) when stream start, making the cursor visible in this situation. Mouse Keys setting is also restored to the original setting before stream began.

Issue #167, opened 2022-05-26
Mouse movement is not even close to 1:1 with the host and wonky/console port-y, or just not good.


thiccelf83:
### Describe the Bug

On the host system (windows) mouse movement is within the parameters of the settings for it. When streaming a computer, windows or linux, mouse movement is far from the same. even if the settings and os on the guest and hoist are matching. I have experienced this with geforce experience too and it seems to be an inherit issue of the software, or maybe just moonlight.

It's unusable in some shooters, those with inherit input lag issues especially such as the early battlefield series. but just in general, I can never perform when streaming (locally always) as I would on my pc, because the mouse movement is off. Its accelerated and weird, its like how console port aiming is at times.

This is probably a issue with the client and not the streaming software, but I am not a developer so i wouldnt know.

One more thing is, it seems that this is somewhat different every session.

### Expected Behavior

mouse movement should be more or less 1:1 with the host system within reasonableamounts and not strangely accelerated in some weird emulated hard to predict way. 

### Additional Context

I've been told before this is unfixable and an inherit problem with gamestream.

 but this is sunshine. 

### Sunshine Host Operating System and Version

Windows 10

### Architecture

64 bit

### Sunshine Version

0.13.0

### GPU Type

nvidia

### GPU Model

gtx 980 ti

### GPU Driver/Mesa Version

27.21.14.5671

cgutman:
With GFE, the problem is that it turns on mouse acceleration and adjusts your mouse pointer speed when streaming starts. That’s the “unfixable” issue. 

Sunshine doesn’t do that, but there have historically been other mouse issues on systems such multiple monitors. I’m not sure if any of those still exist these days.

Are you using multiple monitors on your host?

What client OS are you running Moonlight on? I assume you have “remote desktop optimized mouse mouse” off in Moonlight, correct?

thiccelf83:
Hi, thanks for replying. I am not using GFE anymore, just sunshine. 

Yes, that is off. Its less wonky, more that it is accelerated and weird and not 1:1, if I matc hsettings, its a little better. but its not like native, and the difference is too great to be chalked up to latency on a local network like mine where performance is fine.

It feels emulated

thiccelf83:
Is this related to my cursor not showing without trails on any version of sunshine

wezel:
Hi, I've encountered a similar issue on a Linux (pop-os 22.04) host.
[...]

PR #727, opened 2023-01-08
Windows: disable enhanced pointer precision

psyke83:
## Description

Disabling enhanced pointer precision (mouse acceleration) helps alleviate potential double acceleration caused by the host.

As with Mouse Keys, the previous setting will be restored on stream exit.

### Issues Fixed or Closed
Potentially resolves: #167 (depends on whether host still has "Enhance pointer precision" left to the Windows default of enabled).

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/misc.cpp
+++ b/src/platform/windows/misc.cpp
@@ -28,7 +28,9 @@ using namespace std::literals;
 namespace platf {
 using adapteraddrs_t = util::c_ptr<IP_ADAPTER_ADDRESSES>;
 
-bool enabled_mouse_keys = false;
+int previous_mouse_state[3];
+bool disabled_mouse_accel = false;
+bool enabled_mouse_keys   = false;
 MOUSEKEYS previous_mouse_keys_state;
 
 std::filesystem::path appdata() {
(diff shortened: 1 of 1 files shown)


cgutman:
I'm a bit uneasy about making a change like this. The way this is supposed to work is that Moonlight should send unaccelerated mouse deltas, so they should be accelerated once at most (by the host PC). This allows the app running on the host to opt in or out of acceleration as desired.

Some clients do send pre-accelerated deltas like Android and Qt on macOS, but I think we should handle those via some special control message that informs the host that the deltas are pre-accelerated (assuming we can't just un-accelerate them by  applying the inverse of the acceleration curve or something) then we can turn off acceleration on the host just for those apps.

If we do disable mouse acceleration ourselves, we have to watch out for games and other apps that manipulate it too. It's possible that we save the current state of mouse acceleration when we start, then a game also disables it (saving the previous state as disabled), then the Moonlight session ends and we apply _our_ saved value (which re-enables it), and now the game is incorrectly running with mouse acceleration enabled.

psyke83:
OK, what you're saying makes sense. Even when I completely disable acceleration on my client system (GNOME's "flat" acceleration profile), mouse movement via Moonlight still has an unacceptable level of acceleration when enhanced pointer precision is enabled, perhaps due to a large difference in DPI settings between the host and client mouse. I will just keep the setting disabled for my own sanity and deal with a sluggish mouse on the host.

PR #753, opened 2023-01-11
Switch to faster nanors Reed-Solomon implementation

cgutman:
## Description
This PR switches from the Reed-Solomon implementation in moonlight-common-c to a faster one in https://github.com/sleepybishop/nanors

Right now I'm just building with the default `autoshim.h` flags which means SSSE3 for x86_64 and NEON for aarch64 (since nanors doesn't support runtime CPU feature detection), but that's still good enough for huge performance improvement over the old code.

### Screenshot
Profile with old RS implementation:
old_prof

Profile with new RS implementation (sample counts will vary, so just look at the relative values vs WSASendTo):
image

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,3 +42,7 @@
 	path = third-party/ffmpeg-macos-aarch64
 	url = https://github.com/LizardByte/build-deps
 	branch = ffmpeg-macos-aarch64
+[submodule "third-party/nanors"]
+	path = third-party/nanors
+	url = https://github.com/sleepybishop/nanors.git
+	branch = master
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -337,8 +337,8 @@ configure_file(version.h.in version.h @ONLY)
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
 set(SUNSHINE_TARGET_FILES
-        third-party/moonlight-common-c/reedsolomon/rs.c
-        third-party/moonlight-common-c/reedsolomon/rs.h
+        third-party/nanors/rs.c
+        third-party/nanors/rs.h
         third-party/moonlight-common-c/src/Input.h
         third-party/moonlight-common-c/src/Rtsp.h
         third-party/moonlight-common-c/src/RtspParser.c
@@ -385,6 +385,9 @@ set(SUNSHINE_TARGET_FILES
 
 set_source_files_properties(src/upnp.cpp PROPERTIES COMPILE_FLAGS -Wno-pedantic)
 
+set_source_files_properties(third-party/nanors/rs.c
+        PROPERTIES COMPILE_FLAGS "-include deps/obl/autoshim.h -ftree-vectorize")
+
 # Pre-compiled binaries
 if(WIN32)
     set(FFMPEG_PREPARED_BINARIES "${CMAKE_CURRENT_SOURCE_DIR}/third-party/ffmpeg-windows-x86_64")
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -42,3 +42,6 @@
 	path = third-party/ffmpeg-macos-aarch64
 	url = https://github.com/LizardByte/build-deps
 	branch = ffmpeg-macos-aarch64
+[submodule "third-party/nanors"]
+	path = third-party/nanors
+	url = https://github.com/sleepybishop/nanors.git
```
ReenigneArcher, line 47: I know this repo only has one branch, but would you mind adding specifying the branch anyway?

cgutman:
I've submitted the change (with tests) upstream as https://github.com/sleepybishop/nanors/pull/4.

cgutman:
The PR was accepted, so I've changed to point back to the upstream repo. This should be ready to merge now.

CMakeLists.txt
```diff
@@ -385,6 +385,8 @@ set(SUNSHINE_TARGET_FILES
 
 set_source_files_properties(src/upnp.cpp PROPERTIES COMPILE_FLAGS -Wno-pedantic)
 
+set_source_files_properties(third-party/nanors/rs.c PROPERTIES COMPILE_FLAGS "-include deps/obl/autoshim.h")
```
sleepybishop, line 388: not sure how to turn on auto vectorization in MSVC but with `gcc`/`clang`
adding `-ftree-vectorize` gives you simd accelerated xor for [this loop](https://github.com/sleepybishop/nanors/blob/master/deps/obl/oblas_lite.c#L232). I didn't use intrinsics explicitly for xor since the I let the compiler handle it.

cgutman: Are you suggesting `-ftree-vectorize` instead of including `autoshim.h` or in addition to?

sleepybishop: in addition to, the autoshim will just choose the proper intrinsic code for the platform, auto vectorization needs to be enabled for faster xor's.

after a bit of reading I think msvc will auto vectorize with `/02` which is the default for release builds, so maybe no changes required at least for windows builds

cgutman: Our Windows builds are using MinGW, so `-ftree-vectorize` should be correct across the board.

sleepybishop: hmm, didn't seem to work, 

```
objdump -S ./CMakeFiles/sunshine.dir/third-party/nanors/rs.c.o|less
```

```
 18d:   4c 8b 65 d8             mov    -0x28(%rbp),%r12
        for (; ap < ae; ap++, bp++)
 191:   eb 14                   jmp    1a7 <obl_axpy+0x4c>
            *ap ^= *bp;
 193:   0f b6 13                movzbl (%rbx),%edx
 196:   41 0f b6 04 24          movzbl (%r12),%eax
 19b:   31 d0                   xor    %edx,%eax
 19d:   88 03                   mov    %al,(%rbx)
        for (; ap < ae; ap++, bp++)
```

ideally you want at least `pxor` being used, with `-include deps/obl/autoshim.h -ftree-vectorize -msse4 -O2` you should get something like
```
12d:   00 00 00 
           *ap ^= *bp;
130:   f3 0f 6f 04 07          movdqu (%rdi,%rax,1),%xmm0
135:   f3 0f 6f 14 06          movdqu (%rsi,%rax,1),%xmm2
13a:   66 0f ef c2             pxor   %xmm2,%xmm0
13e:   0f 11 04 07             movups %xmm0,(%rdi,%rax,1)
       for (; ap < ae; ap++, bp++)

```

cgutman: Yep, the issue is the optimization flags were broken for C code unless you manually specified an optimized `CMAKE_BUILD_TYPE`. I have a fix in #777.

Issue #373, opened 2022-09-18
Error: Couldn't import RGB Image: 00003002


nougatbyte:
### Describe the Bug

Trying to connect to a Kubuntu 20.04 host results in instant „Connection terminated“.

### Expected Behavior

Working stream.

### Additional Context

Hi,

I tried running Sunshine on Kubuntu 20.04 on the host and ipadOS/Moonlight on the client.
My setup is a little weird but as this works on GFE on the same setup I guess it‘s a bug.

I run a Proxmox VM with a passed through 770 GTX. As soon as I disconnect the virtual monitor and connect via Moonlight I get „Connection Terminated“ immedeatly. The GPU is connected to a monitor.

Since there is an open pull request which seems to fix a similar error I built that version but it doesnt seem to fix my problem. (https://github.com/LizardByte/Sunshine/pull/345)

Host OS: Kubuntu 20.04
Client: Current Moonlight on ipadOS
GPU: 770 GTX
Drivers: Most recent proprietary Nvidia drivers

Thank you for your engangement!

Edit: Also tried a 980 which sadly leads to the same symptoms.

### Host Operating System

Linux

### Operating System Version

Kubuntu 20.04

### Architecture

64bit

### Sunshine Version

v0.14.1

### GPU Type

Nvidia

### GPU Model

Geforce 770 GTX

### GPU Driver/Mesa Version

474.141.03

### Capture Method (Linux Only)

X11

### Relevant log output

```Shell
[2022:09:18:23:02:01]: Info: /CN=NVIDIA GameStream Client -- verfied
[2022:09:18:23:02:02]: Info: /CN=NVIDIA GameStream Client -- verfied
[2022:09:18:23:02:02]: Info: /CN=NVIDIA GameStream Client -- verfied
[2022:09:18:23:02:02]: Info: CLIENT CONNECTED
[2022:09:18:23:02:02]: Warning: Couldn't expose some properties for card: /dev/dri/card0
[2022:09:18:23:02:02]: Info: Screencasting with KMS
[2022:09:18:23:02:02]: Warning: Couldn't expose some properties for card: /dev/dri/card0
[2022:09:18:23:02:02]: Info: Found monitor for DRM screencasting
[2022:09:18:23:02:02]: Error: Couldn't import RGB Image: 00003002
[2022:09:18:23:02:02]: Info: Color coding [Rec. 601][2022:09:18:23:02:02]: Info: Color range: [MPEG]
```

tproulund:
[...]

PR #649, opened 2022-12-29
Fix wayland capture on nvidia

jhgarner:
## Description
When using Sway on a Nvidia device, the following error occurs: https://discord.com/channels/804382334370578482/1053022079039058001

To summarize, moonlight displays a black screen after connecting and an opengl error is printed to the sunshine logs. The logs are here: https://gist.github.com/jhgarner/5627b7e48aae7f8b1d9744e3f9011f11

I can consistently reproduce and fix the issue by compiling with and without the code in this branch, but I can't explain why this code fixes it. As far as I can tell, this is just debug logging. My guess is that it's forcing the driver to actually flush parts of the command buffer? Swapping the code out with a glFinish call doesn't fix it though. Maybe Nvidia is just broken? 

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I want maintainers to keep my branch updated

Issue #404, opened 2022-10-14
Green Screen


istori1:
### Describe the Bug

Sometimes when I resume session I get a green screen and this message in Sunshine `Error: GL: /run/build/sunshine/src/platform/linux/graphics.cpp:528: [00000502]`

### Expected Behavior

See the screen instead of a green screen.

### Additional Context

Closing and resuming session again will make it go away.

### Host Operating System

Linux

### Operating System Version

Fedora 36

### Architecture

x86_64

### Sunshine Version

nightly

### GPU Type

Intel

### GPU Model

Intel Geminilake (Gen 9)

### GPU Driver/Mesa Version

Flatpak

### Capture Method (Linux Only)

KMS

Issue #210, opened 2022-06-17
Error Installing RPM 0.14.0 on Fedora Silverblue 36 - '/usr/local' unsupported


dannymate:
### Describe the Bug

When trying to install the rpm for sunshine 0.14.0 on Fedora Silverblue 36 (rpm-ostree install <sunshine.rpm>) I get the error:
```
error: Importing package 'sunshinestream': Analyzing /usr/local: Unsupported path; see https://github.com/projectatomic/rpm-ostree/issues/233
```

### Expected Behavior

RPM should be layered onto my install.

### Additional Context

As Silverblue is an immutable OS it currently doesn't support installing into the /usr/local. You can see the issue here: https://github.com/projectatomic/rpm-ostree/issues/233.

### Sunshine Host Operating System and Version

Fedora Silverblue 36

### Architecture

64 bit

### Sunshine Version

0.14.0

### GPU Type

AMD iGPU + nVIDIA dGPU

### GPU Model

Nvidia RTX 3050

### GPU Driver/Mesa Version

510.68.02

### Capture Method (Linux Only)

Pipewire

ReenigneArcher:
You can set the destination paths yourself if you build yourself.

jeremyvisser:
In fairness, it is pretty unusual for an RPM to install to `/usr/local`, so I think the implied proposal to install to /usr instead is reasonable. Would you accept a pull request that changes the prefix to `/usr`?

Another strange RPM behaviour is the postinstall logic in [`src_assets/linux/misc/postinst`](https://github.com/SunshineStream/Sunshine/blob/master/src_assets/linux/misc/postinst) which backs up and restores the config file. RPM has a built-in flag for this: `%config(noreplace)` ([doc](https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html)). But admittedly CPack makes doing this a bit harder than RPM itself ([`CPACK_RPM_USER_FILELIST`](https://cmake.org/cmake/help/latest/cpack_gen/rpm.html#variable:CPACK_RPM_USER_FILELIST) shenanigans).

ReenigneArcher:
I will be changing the paths to `/usr/share/sunshine` based on discussion in another PR.

I don't know if cpack has an option for that, but we're sticking with cpack for the packages it supports as it greatly simplifies our packaging process.

ReenigneArcher:
Working on updates to the paths in #330 . Will `/usr/share/sunshine/` be acceptable for RPM?

PR #761, opened 2023-01-12
Fix Error: Couldn't import RGB Image: 00003002 (#373)

manteuffel723:
## Description
surface_descriptor_t::fds (graphics.h) did not get properly initialized in platf::kms::display_t::refresh.
That resulted in egs::import_source reading values from uninitialized variable/array elements from surface_descriptor_t and passing them to eglCreateImage.
eglCreateImage then fails with error code 3002 (EGL_BAD_ACCESS) (see https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import.txt)

This fix ensures that surface_descriptor_t::fds does not contain uninitialized array elements after a successfull call to platf::kms::dislplay_t::refresh. Elements corresponding to invalid handles get set to -1.
That way the check that was already in place in egs::import_source `fd < 1` (line 490 graphics.h)  properly prevents passing invalid/unsupported/unwanted attributes to eglCreateImage. This check was basically dead code before this fix.

### Issues Fixed or Closed

Fixes Error: Couldn't import RGB Image: 00003002 #373`

## Type of Change

## Checklist

## Branch Updates


KuleRucket:
Is it possible this fixes #453 as well?

src/platform/linux/kmsgrab.cpp
```diff
@@ -583,6 +583,7 @@ class display_t : public platf::display_t {
 
     for(int y = 0; y < 4; ++y) {
       if(!fb->handles[y]) {
+        sd->fds[y] = -1;
```
ReenigneArcher, line 586: Could you please add a comment above your change, describing what this is doing?

src/platform/linux/kmsgrab.cpp
```diff
@@ -583,6 +583,7 @@ class display_t : public platf::display_t {
...
         // It's not clear wheter there could still be valid handles left.
```
ReenigneArcher, line 587: I know this wasn't your change, but since you're changing this anyway, can you fix `whether`?

Deytron:
I just tested the fix, and I can confirm that this fixes both #373 and #453 (00003002 and 00003004 respectively)

## Test environment

OS: EndeavourOS Linux x86_64 
Kernel: 6.1.4-zen2-1-zen 
Resolution: 1920x1080 
DE: Plasma 5.26.5 
WM: kwin-wayland
CPU: AMD Ryzen 5 3600X (12) @ 4.000GHz 
GPU: NVIDIA GeForce RTX 2060 Rev. A 
Memory: 32GB

### What works

- NVENC HEVC encoder seems to work
- Mouse and keyboard both work
- Latency is close to zero in LAN
- Sound works

### What does not work

- May be me, but the web UI has a blank page and gives me (localhost : not authorized)
- I get a black screen on the client
- This errors pops up in terminal :

`[2023:01:13:11:39:24]: Error: GL: /home/romain/Sunshine/src/platform/linux/graphics.cpp:529: [00000501]`

Line 529 only contains `gl_drain_errors;` , so I guess the error lies somewhere else

Interestingly, Sunshine tries to capture an xrandr window even on Wayland

KuleRucket:
This fixes #453 for me too. For the first time ever I can screencast with KMS under wayland on Nvidia

KuleRucket:
The graphics.cpp Error: GL: problem is reported in #404

manteuffel723:
> The graphics.cpp Error: GL: problem is reported in #404

What error do you get? 00000501 (GL_INVALID_VALUE) like Deytron or 00000502 (GL_INVALID_OPERATION) as in #404?

Looks like calls to:
- BindTexture
- GenTextures
   
can cause these errors. Conditions for these Errors are documented in https://registry.khronos.org/OpenGL-Refpages/gl4/.
Maybe you can do some printf debugging to see, where exactly the error comes from.  There are 2 calls to BindTexture in import_source. I didn´t find documentation for EGLImageTargetTexture2DOES, so maybe even after that call.
Additionally there are calls to GenTextures and BindTexture in tex_t::make (line 26 graphics.cpp) that is called from import_source.

I don´t have a nvidia gpu, so i cannot reprduce it. Screencasting works just fine on my amd card now.

KuleRucket:
> What error do you get? 00000501 (GL_INVALID_VALUE) like Deytron or 00000502 (GL_INVALID_OPERATION) as in #404?

You're right I didn't notice the difference. I get 00000501 like Deytron above.

Deytron:
I will be able to investigate in the following week, I'll keep you updated

zastrixarundell:
Compiling this on fedora silverblue 37 and running under an Ubuntu 22.04 container I managed to not get #453 but the screen on the client was black with my cursor. I did get `Screencasting with X11` rather than KMS.

This issue might be just on my side because of the container mess as I can't install on my system because of #210.

Deytron:
I don't know if this will help much, but here's some debugging I did :  

```
[2023:01:22:22:18:31]: Error: GL: /home/romain/Sunshine/src/platform/linux/graphics.cpp:534: [00000501]

eglCreateImage returned 0x7f1378379c91

egl_display: 0x7f137800b4f0
attribs: 0x7f139d7f9910
rgb->tex[0]: 1
rgb->xrgb8: 0x7f1378379c91
rgb->tex[0]: 1
rgb->xrgb8: 0x7f1378379c91
```

Deytron:
For reference, #649 should have fixed the GL errors, but it didn't fix it on my end

zastrixarundell:
Another thing I noticed with a multi-monitor setup (3 monitors horizontally aligned) is that when I move the cursor on the monitor at `+0+0` position I get a weird mouse-following artifact. Keep in mind that after I connected on moonlight I did NOT move the mouse of the client but rather I moved it on the host. 

So the video feed manages to capture the cursor but not anything below it.

I had a similar issue when I installed openbox without a compositor on a clean fedora installation.

https://user-images.githubusercontent.com/30195425/214153681-a2ce9f8e-9730-42c3-a9ab-2eafde38629c.mp4

ReenigneArcher:
Am I correct in that this solves 00003002, and only removes the error for 00003004?

zastrixarundell:
I can't say for 00003002 but on a wayland + nvidia setup it does remove the error 00003004 and for the first time I can initialize the stream. The video stream itself doesn't work how it should but everything else does (audio and input capture).

Deytron:
> Am I correct in that this solves 00003002, and only removes the error for 00003004?

Yes, this is correct, both 0003002 and 00003004 are solved. I guess we should open a new issue concerning the new GL error.

Deytron:
Created issue #838 for further investigation and better tracking

KuleRucket:
I think this fix should be merged. At least the behaviour is better than it was.

PR #766, opened 2023-01-13
Add timers for snapshot/encode in verbose mode

KuleRucket:
## Description
This PR measures the time to snapshot, convert and encode frames and log the information as verbose logging. Although it adds a very small amount of processing when running normally, I find the changes extremely useful for objectively comparing performance between different settings. This can also be useful to diagnose problems when people are not getting the performance we expect.

### Screenshot
n/a

### Issues Fixed or Closed
n/a

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/cuda.cpp
+++ b/src/platform/linux/cuda.cpp
@@ -591,6 +591,7 @@ public:
   }
 
   platf::capture_e snapshot(platf::img_t *img, std::chrono::milliseconds timeout, bool cursor) {
+    auto start = std::chrono::steady_clock::now();
     if(cursor != cursor_visible) {
       auto status = reinit(cursor);
       if(status != platf::capture_e::ok) {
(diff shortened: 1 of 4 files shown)


KuleRucket:
Alternatively I could replace the /1000.0 with a cast to milli:
```
auto diff = stop-start;
BOOST_LOG(verbose) << "nvfbc snapshot in "sv << std::chrono::duration <double, std::milli> (diff).count() << "ms"sv;
```

KuleRucket:
What also could be cool is to report the encode time within a sunshine protocol extension so that is can be seen in the client statistics

cgutman:
> What also could be cool is to report the encode time within a sunshine protocol extension so that is can be seen in the client statistics

I think that's probably the right direction to go for this. GFE can send [both small and large frame headers](https://github.com/moonlight-stream/moonlight-common-c/blob/ef9ad529a493d699724d84a189cc1899afcc2d72/src/VideoDepacketizer.c#L845-L856) and right now Moonlight just completely ignores the extra data in the large frame header. We could use some of that space to store a couple of `uint16_t` values in there for capture and encode time.

That would give the client per-frame encode and capture statistics that it can display in the overlay. The problem with logging for performance metrics is that the logging itself can affect performance (and in interesting non-linear ways, like if one particular log message happens to trigger a flush to disk or something).

KuleRucket:
> The problem with logging for performance metrics is that the logging itself can affect performance (and in interesting non-linear ways, like if one particular log message happens to trigger a flush to disk or something).

This is true for end to end performance but what is happening year is that distinct portions are being measured with no logging being performed in between. Measuring the start/end time of an encode then logging the result doesn't affect the time it took to encode.

PR #768, opened 2023-01-13
Versioning improvements

TheElixZammuto:
## Description
**This PR adds an improved versioning system.**
1. CMake has been modified to add support to https://github.com/nocnokneo/cmake-git-versioning-example, so it uses `git describe`  to get the git tag, the current commit and if the commit tree is dirty.

2. The Sunshine version has been prepended to the stdout for troubleshooting.

3. In the index.html file of the Web UI, communication with the GitHub API has been added, for fetching the latest version and letting the user know if a new version is available. release users only get notices of releases, while nightly users get both release and the last successful CI in the Ci.yaml

For making the tag sync with the current master (0.17), I've merged master into nightly (there are only some merge commits)

### Screenshot

sun1
sun2

## Type of Change

## Checklist

## Branch Updates


src/main.cpp
```diff
@@ -96,7 +96,7 @@ int entry(const char *name, int argc, char *argv[]) {
 */
 namespace version {
 int entry(const char *name, int argc, char *argv[]) {
-  std::cout << PROJECT_NAME << " version: v" << PROJECT_VER << std::endl;
+  std::cout << PROJECT_NAME << " version: " << PROJECT_VER << std::endl;
```
ReenigneArcher, line 99: Unless you're strongly against this, could you leave the "v" prefix as it matches the tags on releases?

src_assets/common/assets/web/index.html
```diff
@@ -35,3 +73,43 @@ <h2>Legal</h2>
...
+      },
+      hasNewNightly() {
+        if (!this.version || !this.nightlyData || this.version.split("-g").length === 1) return false;
+        if (this.version.indexOf("dirty") !== -1) return false;
+        let commit = this.version.split("-g")[1].split("-")[0];
+        return this.nightlyData.head_sha.indexOf(commit) !== 0;
+      }
+    }
+  });
+</script>
```
ReenigneArcher, line 115: please add empty blank line at end of file

ReenigneArcher (commented):
Could you also move `version.h.in` to `./src`, and rename to `version.h`.

I might be wrong, but I think if we put it in `./src` we won't even need to configure this file in CMakeLists.txt like I set it up originally.

Also, there's a small typo in line 8 of `version.h.in`, `PROJECT_VER_PATCH`.

src/main.cpp
TheElixZammuto: Done

src_assets/common/assets/web/index.html
TheElixZammuto: Done

TheElixZammuto:
> Could you also move version.h.in to ./src, and rename to version.h.
I've moved the file but kept the suffix, since the file is an intermediate source file. version.h is generated by Cmake

> Also, there's a small typo in line 8 of version.h.in, PROJECT_VER_PATCH.
Fixed

TheElixZammuto:
It seems I've broke the build, fixing this

TheElixZammuto:
Build should be fixed now!

ReenigneArcher:
Great! Just a few indentations to fix now in `CMakeLists.txt`

ReenigneArcher:
> For making the tag sync with the current master (0.17), I've merged master into nightly (there are only some merge commits)

When I updated the branch, I did so with a re-base and it removed the merge commits. I didn't know that would happen, but now I have a question.

Is this going to work reliably when building nightly builds (or PR builds), as we don't typically merge `master` back into `nightly`?

TheElixZammuto:
The PR uses git for tag handling (it uses the `git describe --tags --dirty` command), so I think that if you tag a commit that is not present on nightly, the tag is not applied. e.g. before my merges, the version of the nightly was a much older then 0.17.0. (iirc was 0.11.0 ?).  For nightly this works correctly but of course the UI does show a much older version

TheElixZammuto:
The PR has now been updated to use the version number inside the CMakeLists.txt. 
Nightly and dirty versions of sunshine are still handled based on the latest commit and if the commit tree is dirty or not.

CMakeLists.txt
```diff
@@ -10,6 +10,50 @@ and Nvidia GPUs for hardware encoding. Software encoding is also available. You
...
+    execute_process(
+        COMMAND ${GIT_EXECUTABLE} diff --quiet --exit-code
+        #WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+        RESULT_VARIABLE GIT_IS_DIRTY
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    if(NOT GIT_DESCRIBE_ERROR_CODE)
+        MESSAGE(Sunshine Branch: ${GIT_DESCRIBE_BRANCH})
+        if(NOT GIT_DESCRIBE_BRANCH STREQUAL "master")
+            set(PROJECT_VERSION ${PROJECT_VERSION}-g${GIT_DESCRIBE_VERSION})
```
ReenigneArcher, line 45: I think this (and L49) is causing the error with the AUR - https://github.com/LizardByte/Sunshine/actions/runs/4091756865/jobs/7056202851#step:7:88

What if we replace the `-` with a `.`?

Then probably don't set the subversion for the AUR as we currently do? Would also require a change here in these places:

-https://github.com/LizardByte/Sunshine/blob/f8819d32e324a02421fe838baadd000210b61d4a/packaging/linux/aur/PKGBUILD#L5
- https://github.com/LizardByte/Sunshine/blob/f8819d32e324a02421fe838baadd000210b61d4a/.github/workflows/CI.yml#L171
- https://github.com/LizardByte/Sunshine/blob/f8819d32e324a02421fe838baadd000210b61d4a/.github/workflows/CI.yml#L191
- https://github.com/LizardByte/Sunshine/blob/f8819d32e324a02421fe838baadd000210b61d4a/.github/workflows/CI.yml#L198
- https://github.com/LizardByte/Sunshine/blob/f8819d32e324a02421fe838baadd000210b61d4a/.github/workflows/CI.yml#L212

Subversion on AUR was more to handle nightly releases, which we're not doing currently anyway.

ReenigneArcher:
Todo:

- [ ] fix cmake lint errors
- [ ] ensure .git directory is not excluded in dockerignore file

ReenigneArcher:
@TheElixZammuto I fixed a couple of things in the builds, but still getting a dirty git tree in the docker builds. Any idea? I suppose it's not copying the `.git` directory correctly, although I'm not sure why.

PR #798, opened 2023-01-18
Fix streaming to multiple clients from hardware encoder on Windows

### LINKED ISSUE #795: Regression: Streaming to multiple simultaneous clients seems to be broken
### Describe the Bug

With `channels > 1` in `sunshine.conf`, Sunshine is supposed to allow multiple devices to connect to the same stream. However, this appears to be currently broken on Windows. If one client is already in the stream, and then a second client joins, _both_ clients show nothing but black after that. The Sunshine log is flooded with this message when this occurs:

```
[2023:01:17:03:52:37]: Error: Failed to acquire encoder mutex [0x887A0001]
```

This is a recent regression. I was able to fix the problem by reverting the following three commits in order: e2fce257b53626b4f21cfc7132a1e94f781a24e1, 00aa23b3424701bbae1fa73315a22ea3d10231ed, and 0439d7a83a0deac7578043cfdf8fddadd4b46799 The reverts all apply cleanly to b2fe0423d2894aed889dd51a83dd361a72bb4231, when done in the right order. The oldest of these commits is only about two weeks old, so this regression is very recent.

### Operating System Version

Windows 11 21H2 (Build 22000.1455)

### Architecture

64 bit

### GPU Type

Nvidia

### GPU Model

GeForce RTX 3090

### GPU Driver/Mesa Version

GeForce Game Ready Driver 528.02

cgutman:
## Description
#668 broke the multi-client streaming scenario because it assumes a single `img_t` is shared between a single capture device and a single encoding device. When using multiple clients, there is a new `hwdevice_t` for each independent stream, so there can be an arbitrary number of encoding devices sharing a single capture texture. This PR fixes this issue by keeping a map of encoder image context structures in the `hwdevice_t`, so each separate encoding device can open the same capture texture and mutex properly.

### Issues Fixed or Closed
Fixes #795

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display.h
+++ b/src/platform/windows/display.h
@@ -195,6 +195,8 @@ public:
   gpu_cursor_t cursor_xor;
 
   texture2d_t last_frame_copy;
+
+  std::atomic<uint32_t> next_image_id;
 };
 } // namespace platf::dxgi
 
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -101,15 +101,15 @@ struct img_d3d_t : public platf::img_t {
   render_target_t capture_rt;
   keyed_mutex_t capture_mutex;
 
-  // These objects are owned by the hwdevice_t's ID3D11Device
-  texture2d_t encoder_texture;
-  shader_res_t encoder_input_res;
-  keyed_mutex_t encoder_mutex;
-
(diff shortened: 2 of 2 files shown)


cathyjf:
I've tested this pull request and it solves the problem for me. Thanks for fixing this bug!

PR #825, opened 2023-01-24
Implement HDR support for Windows

cgutman:
## Description
The title pretty much says it all. This adds support for streaming HDR content from a Windows host.

I've tested this successfully with NVIDIA (RTX 2080), AMD (RX 6700 XT), and Intel (12600K IGP) GPUs. AMD and NVIDIA  encoders had no trouble encoding HEVC Main 10 in HDR at 4K 60 FPS, but Intel QuickSync seemed to struggle with HEVC Main 10 with notably degraded image quality compared to HEVC Main profile (this impacted both SDR and HDR modes).

### Screenshot
I'd attach one but I can't find a screenshot utility that handles HDR content properly

### Issues Fixed or Closed
Resolves #376

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/usage.rst
+++ b/docs/source/about/usage.rst
@@ -257,6 +257,18 @@ Considerations
   instead it simply starts a stream.
 - For the Linux flatpak you must prepend commands with ``flatpak-spawn --host``.
 
+HDR Support
+-----------
+Streaming HDR content is supported for Windows hosts with NVIDIA, AMD, or Intel GPUs that support encoding HEVC Main 10.
+You must have an HDR-capable display or EDID emulator dongle connected to your host PC to activate HDR in Windows.
+
+- Ensure you enable the HDR option in your Moonlight client settings, otherwise the stream will be SDR.
+- A good HDR experience relies on proper HDR display calibration both in Windows and in game. HDR calibration can differ significantly between client and host displays.
+- We recommend calibrating the display by streaming the Windows HDR Calibration app to your client device and saving an HDR calibration profile to use while streaming.
+- You may also need to tune the brightness slider or HDR calibration options in game to the different HDR brightness capabilities of your client's display.
+- Older games that use NVIDIA-specific NVAPI HDR rather than native Windows 10 OS HDR support may not display in HDR.
+- Some GPUs can produce lower image quality or encoding performance when streaming in HDR compared to SDR.
+
 Tutorials
 ---------
 Tutorial videos are available `here <https://www.youtube.com/playlist?list=PLMYr5_xSeuXAbhxYHz86hA1eCDugoxXY0>`_.
--- a/src/platform/windows/display.h
+++ b/src/platform/windows/display.h
@@ -151,6 +151,7 @@ protected:
   }
 
   const char *dxgi_format_to_string(DXGI_FORMAT format);
+  const char *colorspace_to_string(DXGI_COLOR_SPACE_TYPE type);
 
   virtual capture_e snapshot(img_t *img, std::chrono::milliseconds timeout, bool cursor_visible) = 0;
   virtual int complete_img(img_t *img, bool dummy)                                               = 0;
--- a/src/platform/windows/display_base.cpp
+++ b/src/platform/windows/display_base.cpp
@@ -598,6 +598,23 @@ bool display_base_t::get_hdr_metadata(SS_HDR_METADATA &metadata) {
   DXGI_OUTPUT_DESC1 desc1;
   output6->GetDesc1(&desc1);
 
+  // The primaries reported here seem to correspond to scRGB (Rec. 709)
+  // which we then convert to Rec 2020 in our scRGB FP16 -> PQ shader
+  // prior to encoding. It's not clear to me if we're supposed to report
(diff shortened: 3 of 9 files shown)


louis-bompart:
Hey @cgutman, should this be revised in the doc too?
https://github.com/LizardByte/Sunshine/blob/master/docs/source/gamestream/gamestream.rst#limitations

Also, a random question, does Sunlight implementation requires the host to be plugged in a HDR display, [like GameStream](https://www.nvidia.com/en-us/support/gamestream/hdr/) ?

cgutman:
> Hey @cgutman, should this be revised in the doc too? https://github.com/LizardByte/Sunshine/blob/master/docs/source/gamestream/gamestream.rst#limitations

Yep, I missed removing that.

I added an HDR section in the Usage page of the docs: https://docs.lizardbyte.dev/projects/sunshine/en/nightly/about/usage.html#hdr-support

> Also, a random question, does this implementation of HDR for Windows requires the host to be plugged in an HDR display, [like GameStream](https://www.nvidia.com/en-us/support/gamestream/hdr/) ?

Yes it does. It's an OS limitation on a few fronts. Windows will not allow HDR to be turned on at the OS-level without an HDR-capable display or EDID emulator dongle connected. Without HDR enabled, DWM will not composite the desktop in HDR, so any HDR content rendered in games or applications would be clamped to SDR range before we could capture it using the Desktop Duplication API.

I have an idea of how to work around these OS limitations, but it requires injecting a DLL into games to hook some DirectX APIs. It's doable, but there's a ton of complexity in that approach (not to mention potential anti-cheat and game stability issues). I will probably come back to this approach because capturing frames straight from the game has a number of other benefits, including being able to encode on the render GPU rather than the display GPU on multi-GPU systems, but that's why it's not supported for v1 (and AFAIK no other Windows software can do HDR capture on SDR display today).

LionsAd:
I wish the indirect display driver would support HDR. You can load a HDR EDID - similar to the NVIDIA data center driver, but nothing happens.

Some have suspected it is, because the wrong color space R8B8G8A8 is registered, but it’s unclear if Microsoft ever fixed that limitation.

It might be worth checking if the IDD driver can be hot-patched, before going down the route of intercepting the D3D calls - though as you said it might have other advantages.

FWIW, GameStream works fine with the IDD in SDR - did not have a chance to try SunShine, yet.

cgutman:
> Great work! On the HDR support.
> 
> I wish the indirect display driver would support HDR. You can load a HDR EDID - similar to the NVIDIA data center driver, but nothing happens.
> 
> Some have suspected it is, because the wrong color space R8B8G8A8 is registered, but it’s unclear if Microsoft ever fixed that limitation.

My understanding (mostly from [here](https://learn.microsoft.com/en-us/windows/win32/direct3darticles/high-dynamic-range)) is that the DWM basically has 2 different composition pipelines that it can choose to use based on the connected displays and HDR setting.

One is the standard SDR/8-bit composition pipeline where DWM converts/clamps the presented frames into `DXGI_FORMAT_B8G8R8A8_UNORM` sRGB colorspace at the time they are received from the application. In this mode, the DWM composition pipeline is B8G8R8A8 all the way to the final presentation that gets picked up by the display driver. Any HDR or wide color gamut content (like `DXGI_FORMAT_R10G10B10A2_UNORM` or `DXGI_FORMAT_R16G16B16A16_FLOAT`) would be clipped by DWM prior to compositing the final image and unavailable for capture via the Desktop Duplication API.

The second mode is the "Advanced Color" mode which is used when HDR is enabled or an SDR display is connected that has Advanced Color capabilities. In this mode, DWM will take the application's frames in whatever format they are provided and feed that into the composition process. When DWM composites in this mode, it will use scRGB FP16 as the format of its composition buffer which allows it to keep the extra color information available in WCG/HDR content as it passes the buffer to the driver for presentation.

The standard composition pipeline makes sense for the vast majority of systems, since FP16 is twice the size of B8G8R8A8. Compositing in FP16 all the time would waste power and GPU bandwidth on systems where the output is going to be clamped to B8G8R8A8 anyway. For this reason, DWM only enables Advanced Color pipeline in scenarios where the extra color data will actually make it to the display.

There would need to be a way to convince DWM to use the Advanced Color composition pipeline to even have a shot at capturing HDR/WCG content. Emulating an SDR WCG display _may_ work, but I suspect DWM is still going to clamp the output to [0.0f-1.0f] because the luminance behavior for SDR and HDR displays differs (1.0f is 80 nits on HDR displays and max display brightness on SDR displays). Unless the IDD also reports itself as an HDR10 display, games will probably not show the HDR option even if DWM is compositing to scRGB FP16.

> 
> It might be worth checking if the IDD driver can be hot-patched, before going down the route of intercepting the D3D calls - though as you said it might have other advantages.

Unless it can be done purely with registry changes, patching the driver is probably more trouble than doing the D3D/DXGI hooks because Windows kernel-mode code integrity rules are very strict. Drivers cannot (without an exploit) be patched in memory by a user-mode process.

> 
> FWIW, GameStream works fine with the IDD in SDR - did not have a chance to try SunShine, yet.

I think some people have reported that it works fine.

PR #829, opened 2023-01-24
Fix SimpleWeb arbitrary file read vulnurability

safijari:
## Description
Fix for [this](https://discord.com/channels/804382334370578482/1067020615472652308/1067020615472652308) using std::filesystem and computing relative paths.

## Type of Change

## Checklist

## Branch Updates


src/confighttp.cpp
```diff
@@ -247,23 +247,32 @@ void getSunshineLogoImage(resp_https_t response, req_https_t request) {
...
-  }
-  else {
-    std::string content = read_file((WEB_DIR + request->path).c_str());
-    response->write(content);
+  fs::path webDirPath(WEB_DIR);
+  fs::path nodeModulesPath(webDirPath / "node_modules");
+  auto filePath = fs::weakly_canonical(webDirPath / request->path);
+
+  // Don't do anything if file does not exist or is outside the node_modules directory
+  if (!isChildPath(filePath, nodeModulesPath) || !fs::exists(filePath)) {
```
ReenigneArcher, line 262: For this condition `if (!isChildPath(filePath, nodeModulesPath)` can we do

`response->write(SimpleWeb::StatusCode::client_error_bad_request, "Not Authorized");`

src/confighttp.cpp
```diff
@@ -246,23 +246,40 @@ void getSunshineLogoImage(resp_https_t response, req_https_t request) {
...
+  if(!fs::exists(filePath)) {
+    response->write(SimpleWeb::StatusCode::client_error_not_found);
   }
-  else if(boost::algorithm::iends_with(request->path, ".woff2") == 1) {
-    std::ifstream in((WEB_DIR + request->path).c_str(), std::ios::binary);
-    headers.emplace("Content-Type", "font/woff2");
-    response->write(SimpleWeb::StatusCode::success_ok, in, headers);
+  else if(!isChildPath(filePath, nodeModulesPath)) {
+    BOOST_LOG(warning) << "Someone requested a path " << filePath << " that is outside the node_modules folder";
+    response->write(SimpleWeb::StatusCode::client_error_bad_request, "Not Authorized");
```
TheElixZammuto, line 268: I would still log the attempt, but give 404 nevertheless. The attacker shouldn't know if the file exists or not

safijari: How about a straight up 400. We can do the 400 in the case that the path is outside the node modules even if it doesn't exist. That way there would be no way to map out the user's computer.

safijari: Ok I've made that change in my latest commit and it seems to work as expected. Lemme know your thoughts.

PR #830, opened 2023-01-24
Restart the systemd service in case of a crash

Max-Might:
Make the systemd service restart itself in case of a crash. A restart is attempted 5 seconds after a crash is detected. Additionally, if the service is restarted more than 5 times in a 500 second interval it is considered as failed and no more restart attempts are going to be made.

## Description
Sunshine crashed when I attempted to connect to a desktop session.
Not sure what the actual crash was. Nothing suspicious in the output logs.
Maybe systemd keeps the core dumps somewhere. I will open a new ticket if I find out more about the issue.

## Type of Change

## Branch Updates

### DIFF
--- a/docs/source/about/usage.rst
+++ b/docs/source/about/usage.rst
@@ -87,9 +87,13 @@ Sunshine needs access to `uinput` to create mouse and gamepad events.
 
             [Unit]
             Description=Sunshine self-hosted game stream host for Moonlight.
+            StartLimitIntervalSec=500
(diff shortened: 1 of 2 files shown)


ReenigneArcher:
Could you also update the docs. https://github.com/Max-Might/Sunshine/blob/patch-1/docs/source/about/usage.rst#linux

Max-Might:
@ReenigneArcher I updated the documentation.

PR #850, opened 2023-01-28
Display initialization and frame ownership fixes

cgutman:
## Description
This fixes some issues with capture initialization/reinitialization that could cause us to discard good frames. When we reinitialize encoding, the encoder thread will free any pending frames and encode an initial dummy frame. This could result in the first captured frame from the newly reinitialized display being dropped or immediately superseded by a dummy frame.

@psyke83 Please give this a try to ensure it doesn't regress #823. It should also fix the black screen issue when starting a stream or triggering a UAC prompt when no new frames are being generated.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -1273,6 +1273,13 @@ void encode_run(
   auto packets        = mail::man->queue<packet_t>(mail::video_packets);
   auto idr_events     = mail->event<bool>(mail::idr);
 
+  // Load a dummy image into the AVFrame to ensure we have something to encode
+  // even if we time out waiting on the first frame.
+  auto dummy_img = disp->alloc_img();
+  if(!dummy_img || disp->dummy_img(dummy_img.get()) || session->device->convert(*dummy_img)) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


psyke83 (approved):
I tested over 100 fullscreen transitions using the same testcase in #823 with this PR and couldn't reproduce the problem. Additionally, I couldn't trigger the `Capture size changed` message, but I'm guessing that the changes here might impact that. Seem fine to me.

psyke83:
I also tested the issue related to the dummy frame on first connect:

1. Open an image in Windows 10 Photos app & enter fullscreen.
2. Repeatedly disconnect and reconnect without moving the mouse or keyboard.

Prior to this PR, sometimes the stream would stay black indefinitely until some input triggers a screen refresh. With the PR, the new behaviour is that a black screen may occur on connect that lasts for 1 second followed by the proper desktop image. I assume this delay is likely due to `AcquireNextFrame` being set with a 1 second timeout and thus may not be an indication of anything wrong with this PR.

PR #883, opened 2023-02-03
Try to fix kms output index randomly switching order when using kms on linux

HomerSp:
This is an attempt to fix the monitor order randomly being wrong when using kms on Linux - for me the output_name would randomly be wrong, and I had to change it in the config file.
I've done this by sorting the crcts by id (so output_name = 0 in the config would choose the crtc with the lowest id etc), instead of relying on the order of the planes.
I don't know if this is the best way to do it, but it seems to work for me.
My initial plan was to sort them by the actual order the monitors were in, but it seems like the x and y values of both the crtcs and planes are 0, so there's no actual way of doing that it seems.
Maybe someone more knowledgeable with drm and kms has some input? 

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/kmsgrab.cpp
+++ b/src/platform/linux/kmsgrab.cpp
@@ -457,11 +457,33 @@ class display_t : public platf::display_t {
 public:
   display_t(mem_type_e mem_type) : platf::display_t(), mem_type { mem_type } {}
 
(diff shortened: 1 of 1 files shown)


HomerSp:
Seems this doesn't work consistently either, back to the drawing board...

PR #941, opened 2023-02-16
Add setting for suppressing input from mouse, keyboard, or gamepads

briankendall:
## Description
This PR adds three new settings that allow suppressing client input from the mouse, keyboard, and gamepads, one for each type of input. They can be toggled individually, allowing any combination of allowable input the user desires. This also allows for disabling client input entirely, putting Sunshine in a "view only" mode.

All settings are disabled by default, meaning there's no change in default behavior.

This feature was suggested here: https://github.com/LizardByte/Sunshine/discussions/822

### Screenshot
Screen Shot 2023-02-15 at 8 36 11 PM

### Issues Fixed or Closed
N/A

## Type of Change

## Checklist

Notes: code additions are very simple and I consider them sufficiently self explanatory to not require any comments.

I have not added any new methods or components.

## Branch Updates

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -951,9 +951,9 @@ void apply_config(std::unordered_map<std::string, std::string> &&vars) {
 
   string_restricted_f(vars, "gamepad"s, input.gamepad, platf::supported_gamepads());
 
-  bool_f(vars, "suppress_mouse", input.suppress_mouse);
-  bool_f(vars, "suppress_keyboard", input.suppress_keyboard);
-  bool_f(vars, "suppress_gamepad", input.suppress_gamepad);
+  bool_f(vars, "mouse", input.mouse);
+  bool_f(vars, "keyboard", input.keyboard);
+  bool_f(vars, "controller", input.controller);
 
   int port = sunshine.port;
   int_f(vars, "port"s, port);
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -408,6 +408,10 @@ input_t input {
     platf::supported_gamepads().front().data(),
     platf::supported_gamepads().front().size(),
   }, // Default gamepad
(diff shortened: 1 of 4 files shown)


src/config.cpp
```diff
@@ -947,6 +951,10 @@ void apply_config(std::unordered_map<std::string, std::string> &&vars) {
 
   string_restricted_f(vars, "gamepad"s, input.gamepad, platf::supported_gamepads());
 
+  bool_f(vars, "suppress_mouse", input.suppress_mouse);
+  bool_f(vars, "suppress_keyboard", input.suppress_keyboard);
+  bool_f(vars, "suppress_gamepad", input.suppress_gamepad);
```
ReenigneArcher, line 956: I think these should be named as follows:

- input.mouse
- input.keyboard
- input.controller (since input.gamepad is already used)

With the default value being enabled.

ReenigneArcher (changes requested):
Thank you for this PR! This will be a nice feature to add.

I would like to ask for a small adjustment though.

src/config.cpp
briankendall: That seems reasonable, and I can make that change. Would you prefer I add another commit, or amend the current one to keep it at just one commit?

ReenigneArcher: > Would you prefer I add another commit, or amend the current one to keep it at just one commit?

I would normally do an amend, but it's up to you. We will do a squash merge when we merge the PR so it will appear as a single commit anyway.

briankendall: @ReenigneArcher I've made the change, and kept it as a separate commit so it's easier to see what's changed from my last commit.

ReenigneArcher (approved):

PR #960, opened 2023-02-20
Support compiling for earlier releases of macOS

briankendall:
## Description
These changes broaden the macOS versions that Sunshine can be compiled for. There are only a few places in the codebase where a macOS API function that requires macOS 10.15 or later is used, and this PR makes it so that these functions are weakly linked and avoids using them when running in macOS 10.14 and earlier.

The other issue with compiling on macOS 10.14 is that its version of libc++ doesn't support some of the later C++ features used throughout the codebase. However, this can be overcome by compiling with a newer installation of LLVM / clang and statically linking against its copy of libc++.

With the changes in this PR, I was able to get Sunshine to compile for macOS 10.12-10.14 by installing clang 15 using MacPorts (along with Sunshine's other dependencies) and then configuring it with the following `cmake` command:

    export CC=/opt/local/libexec/llvm-15/bin/clang
    export CXX=/opt/local/libexec/llvm-15/bin/clang++
    cmake \
        -DCMAKE_CXX_FLAGS="-nostdinc++ -nostdlib++ -isystem /opt/local/libexec/llvm-15/include/c++/v1 -I /opt/local/include -I /opt/local/libexec/openssl11/include" \
        -DCMAKE_EXE_LINKER_FLAGS="-nostdlib /opt/local/libexec/llvm-15/lib/libc++.a /opt/local/libexec/llvm-15/lib/libc++abi.a /usr/lib/libSystem.dylib" \
        -DBOOST_ROOT=/opt/local/libexec/boost/1.80 \
        -DCMAKE_OSX_DEPLOYMENT_TARGET=10.12 \
        /path/to/sunshine

(Note that I had to manually supply to paths for both boost and OpenSSL.)

Even if Sunshine will perform worse in them than the latest macOS releases, I think it's important to support older releases of macOS at least in some capacity (even if it requires the user to compile it themselves) since it's very useful for them to have a fully open source streaming solution, as well as there being a lot of reasons to avoid the newer releases of macOS, including lack of support for older macs, poorer options for customization, Apple's continuing agenda of removing user control, and them just generally being frustratingly buggy.

Personally I'm still running macOS 10.14 and have no intention of updating any time soon, if I can manage it! And I'd still like to use Sunshine if I can.

### Screenshot
N/A

### Issues Fixed or Closed
N/A

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,7 +233,7 @@ elseif(APPLE)
             src/platform/macos/display.mm
             src/platform/macos/input.cpp
             src/platform/macos/microphone.mm
-            src/platform/macos/misc.cpp
+            src/platform/macos/misc.mm
             src/platform/macos/misc.h
             src/platform/macos/nv12_zero_device.cpp
             src/platform/macos/nv12_zero_device.h
@@ -620,6 +620,8 @@ endif()
 
 if(APPLE)
     target_link_options(sunshine PRIVATE LINKER:-sectcreate,__TEXT,__info_plist,${APPLE_PLIST_FILE})
+    # Tell linker to dynamically load these symbols at runtime, in case they're unavailable:
+    target_link_options(sunshine PRIVATE -Wl,-U,_CGPreflightScreenCaptureAccess -Wl,-U,_CGRequestScreenCaptureAccess)
 endif()
 
 foreach(flag IN LISTS SUNSHINE_COMPILE_OPTIONS)
--- a/src/platform/macos/av_audio.m
+++ b/src/platform/macos/av_audio.m
@@ -29,7 +29,6 @@
     return [AVCaptureDevice devicesWithMediaType:AVMediaTypeAudio];
 #pragma clang diagnostic pop
   }
-
 }
 
 + (NSArray<NSString *> *)microphoneNames {
--- a/src/platform/macos/av_audio.m
+++ b/src/platform/macos/av_audio.m
@@ -3,11 +3,33 @@
 @implementation AVAudio
 
 + (NSArray<AVCaptureDevice *> *)microphones {
-  AVCaptureDeviceDiscoverySession *discoverySession = [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:@[AVCaptureDeviceTypeBuiltInMicrophone,
-    AVCaptureDeviceTypeExternalUnknown]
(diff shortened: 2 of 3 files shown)


ReenigneArcher:
Thank you for this PR! While we cannot build older version in our CI (https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/) I do think it's good to have support for older hardware in some capacity.

P.S. Do you have any thoughts about statically linking deps during the macOS build? I think it would solve the issues we have with the pkg and dmg packages.

briankendall:
> P.S. Do you have any thoughts about statically linking deps during the macOS build? I think it would solve the issues we have with the pkg and dmg packages.

There's nothing wrong with using static linking to smooth out installation. The norm in macOS is to include all of the necessary binaries for an app or package in one way or another, rather than relying on something like a package manager. Using static linking certainly seems like the shortest route to getting there.

I'll mention that the other more "mac-like" way to do it would be to package Sunshine into an application bundle and then include all of its needed libraries inside of there, generally in the `Contents/Frameworks` subdirectory. With dependencies coming from a package manager, though, that gets really tricky since Sunshine's dependencies have their own tree of dependencies, and on top of having to gather every single dependency recursively, they'd all have to be altered using a tool like `install_name_tool` to change the paths referencing every one of those dynamic libraries from absolute paths to bundle-relative ones (so rather than linking to libraries somewhere inside of `/opt/local` it'd be something like `@executable_path/../Frameworks` or `@rpath`). I wonder if there's a tool out there that can do this automatically for binaries compiled using MacPorts or Homebrew?

That all said, my recommendation would still be to use static linking.

With static linking in place, the other thing that would be nice is to have Sunshine be installed into a self-contained location like `/Library/Application Support/Sunshine` rather than mixing it into `/opt/local`. I think that could be as simple as changing the default macOS install prefix to `/Library/Application Support/Sunshine` and then updating its launchd plist to execute it from there.

ReenigneArcher:
Could you fix the lint errors please?

ReenigneArcher (approved):
I haven't tested this, but the code is well documented and builds fine, so I approve and will merge this.

If you'd be interested in making a PR to statically link the macOS builds it would be really appreciated.

PR #977, opened 2023-02-25
Add support for global prep commands

pgrunzjr:
## Description
Add the ability to add a global set of Prep Commands that will run for _all_ applications. Global prep configuration is accessible in the UI through new tab on Applications page. Users can also enabled/disabled execution of global commands on a per applications basis. Default is Enabled.

### Screenshot
Global Command Preparations on Configuration -> General page
image

Per-App Toggle of global prep command execution
image

### Issues Fixed or Closed
NA

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -426,8 +426,8 @@ sunshine_t sunshine {
   platf::appdata().string() + "/sunshine.conf", // config file
   {},                                           // cmd args
   47989,
-  platf::appdata().string() + "/sunshine.log", // log file,
-  {}
+  platf::appdata().string() + "/sunshine.log",  // log file
+  {},                                           // prep commands
 };
 
 bool endline(char ch) {
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -8,6 +8,8 @@
 #include <boost/asio.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/json.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
 
 #include "config.h"
 #include "main.h"
@@ -765,24 +767,22 @@ void list_prep_cmd_f(std::unordered_map<std::string, std::string> &vars, const s
   std::string string;
   string_f(vars, name, string);
 
-  auto json = boost::json::parse(string);
+  std::stringstream jsonStream;
   
-  if (json.if_array()) {
-    auto list = json.as_array();
+  // We need to add a wrapping object to make it valid JSON, otherwise ptree cannot parse it.
+  jsonStream << "{\"prep_cmd\":" << string << "}";
 
-    for(auto &el : list) {
-      if (el.is_object()) {
-        auto cmd = el.as_object();
+  boost::property_tree::ptree jsonTree;
+  boost::property_tree::read_json(jsonStream, jsonTree);
 
-        input.emplace_back(
-          cmd.if_contains("do") ? cmd.at("do").as_string().c_str() : "",
-          cmd.if_contains("undo") ? cmd.at("undo").as_string().c_str() : ""
-        );
-      }
-      else {
-        continue;
-      }
-    }
+  for(auto &[_, prep_cmd] : jsonTree.get_child("prep_cmd"s)) {
+    auto do_cmd = prep_cmd.get<std::string>("do"s);
+    auto undo_cmd = prep_cmd.get<std::string>("undo"s);
+    
+    input.emplace_back(
+      std::move(do_cmd),
+      std::move(undo_cmd)
+    );
   }
 }
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -7,6 +7,7 @@
 
 #include <boost/asio.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/json.hpp>
 
 #include "config.h"
 #include "main.h"
@@ -425,7 +426,8 @@ sunshine_t sunshine {
   platf::appdata().string() + "/sunshine.conf", // config file
   {},                                           // cmd args
   47989,
-  platf::appdata().string() + "/sunshine.log", // log file
+  platf::appdata().string() + "/sunshine.log", // log file,
+  {}
 };
 
 bool endline(char ch) {
@@ -759,6 +761,30 @@ void list_string_f(std::unordered_map<std::string, std::string> &vars, const std
     input.emplace_back(begin, pos);
   }
 }
+void list_prep_cmd_f(std::unordered_map<std::string, std::string> &vars, const std::string &name, std::vector<prep_cmd_t> &input) {
+  std::string string;
+  string_f(vars, name, string);
+
+  auto json = boost::json::parse(string);
+  
+  if (json.if_array()) {
+    auto list = json.as_array();
+
+    for(auto &el : list) {
+      if (el.is_object()) {
+        auto cmd = el.as_object();
+
+        input.emplace_back(
+          cmd.if_contains("do") ? cmd.at("do").as_string().c_str() : "",
+          cmd.if_contains("undo") ? cmd.at("undo").as_string().c_str() : ""
+        );
+      }
+      else {
+        continue;
+      }
+    }
+  }
+}
 
 void list_int_f(std::unordered_map<std::string, std::string> &vars, const std::string &name, std::vector<int> &input) {
   std::vector<std::string> list;
@@ -902,6 +928,7 @@ void apply_config(std::unordered_map<std::string, std::string> &&vars) {
   string_f(vars, "external_ip", nvhttp.external_ip);
   list_string_f(vars, "resolutions"s, nvhttp.resolutions);
   list_int_f(vars, "fps"s, nvhttp.fps);
+  list_prep_cmd_f(vars, "global_prep_cmd", config::sunshine.prep_cmds);
 
   string_f(vars, "audio_sink", audio.sink);
   string_f(vars, "virtual_sink", audio.virtual_sink);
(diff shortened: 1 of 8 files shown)


src/process.cpp
```diff
@@ -128,27 +123,61 @@ int proc_t::execute(int app_id) {
     terminate();
   });
 
-  for(; _undo_it != std::end(proc.prep_cmds); ++_undo_it) {
-    auto &cmd = _undo_it->do_cmd;
+  // Execute global prep commands if enabled
+  if (_app.include_global_prep)
```
Nonary, line 127: What was the reasoning for an additional for loop in this case? Why not just create a new iterator by merging them together? You can decide to just not add it to the new iterator if the checkmark is false.

src/process.cpp
```diff
@@ -49,14 +49,6 @@ void process_end(bp::child &proc, bp::group &proc_handle) {
   proc.wait();
 }
 
-int exe_with_full_privs(const std::string &cmd, bp::environment &env, file_t &file, std::error_code &ec) {
```
Nonary, line 52: This is a great idea, but honestly this should be an entirely different pull request. Since it is more of a security concern in general, it's not ideal to mix multiple features or changes in one pull request generally.

src/process.cpp
```diff
@@ -203,25 +232,59 @@ void proc_t::terminate() {
...
+    
+    child.wait();    
+    auto ret = child.exit_code();
 
     if(ret != 0) {
       BOOST_LOG(warning) << "Return code ["sv << ret << ']';
     }
   }
+  
+  // Execute global prep commands if enabled
```
Nonary, line 260: Much like the previous comment, we probably don't need an extra loop here.

src/process.cpp
```diff
@@ -439,28 +502,49 @@ std::optional<proc::proc_t> parse(const std::string &file_name) {
...
       proc::ctx_t ctx;
 
-      auto prep_nodes_opt     = app_node.get_child_optional("prep-cmd"s);
-      auto detached_nodes_opt = app_node.get_child_optional("detached"s);
-      auto output             = app_node.get_optional<std::string>("output"s);
-      auto name               = parse_env_val(this_env, app_node.get<std::string>("name"s));
-      auto cmd                = app_node.get_optional<std::string>("cmd"s);
-      auto image_path         = app_node.get_optional<std::string>("image-path"s);
-      auto working_dir        = app_node.get_optional<std::string>("working-dir"s);
+      auto prep_nodes_opt      = app_node.get_child_optional("prep-cmd"s);
```
Nonary, line 540: You know now looking over this, we can probably just refactor this section entirely instead of spamming null checks for every variable. Nothing wrong with what you did here though, just brainstorming some ideas.

src/process.h
```diff
@@ -39,6 +39,8 @@ struct cmd_t {
  *    filename -- The output of the commands are appended to filename
  */
 struct ctx_t {
+  ctx_t() : include_global_prep(true) {}
```
Nonary, line 42: Preferably booleans should initialize at false, it's not a hard requirement but that is general convention. I would invert the boolean here to state disable_global_prep instead. That way you don't have to add an initializer and it matches with most of the booleans in this code base.

Nonary (commented):
I have some questions about these changes, I am not an expert at C++ but just curious on the changes as opposed to other common programing structures.

src/process.h
ReenigneArcher: I think they just didn't get around to updating it yet... We had the same discussion here: https://discord.com/channels/804382334370578482/1074377411069747250/1077688007659896832

src/process.cpp
pgrunzjr: I actually had them combined previously, before adding the ability to exclude them per app, but split them to make things easier to debug when adding that feature in. There's really no reason they couldn't be combined though, so I may change it back now that you've pointed it out.

src/process.cpp
pgrunzjr: Yeah, in hindsight, it probably should have been a separate PR. I'll let @ReenigneArcher make the call on it. I'm happy to split the execution changes into a separate PR if you want.

src/process.cpp
pgrunzjr: Agreed. It does look like #584 does some significant refactoring here.

src/process.h
pgrunzjr: Fixed in latest commit.

src/process.cpp
Nonary: I was just thinking when someone adds the capability to reference moonlight client variables (like client resolution, framerate, hdr setting, etc) it would be easier to implement if it's using an iterator. Multiple loops could work as well if you refactor it to use a function to do most of the work too

src/process.cpp
Nonary: Well, as long as he agrees with it, #998 is the issue you can link for that PR

src/process.cpp
pgrunzjr: I've updated it to load everything into a single vector for each app, so only a single loop is needed now.

Nonary:
The latest changes look really good to me, although granted I am not fluent in C++. Just from an object oriented perspective, it looks good to me. I think all we really need now is documentation updates and we can merge it into nightly from there.

src/process.cpp
ReenigneArcher: I agree that this change should be a separate PR.

src/process.cpp
ReenigneArcher: That PR still needs a lot of work, so I wouldn't base any changes on what's in that to be honest.

For one it completely removes the UI from Sunshine which is a no go.

src/confighttp.cpp
```diff
@@ -488,6 +488,45 @@ void uploadCover(resp_https_t response, req_https_t request) {
   outputTree.put("path", path);
 }
 
+void savePrepCmd(resp_https_t response, req_https_t request) {
```
ReenigneArcher, line 491: Unfortunately this is probably not an easy change, but... instead of adding a new endpoint and function... can we just use the saveConfig endpoint and attach them to the config object?

ReenigneArcher (changes requested):

Nonary:
@pgrunzjr The global prep commands tab isn't working with the latest commits. I click on it and nothing happens.

src/process.cpp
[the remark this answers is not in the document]
pgrunzjr: Created a separate PR and will remove these change form this one.

#1022

src/confighttp.cpp
pgrunzjr: Might be a bit of a challenge since sunshine.conf isn't JSON, but I should be able to make it work. Would we also want to move the global prep tab onto the config page then? I feel like that makes more sense anyway.

ReenigneArcher: > Would we also want to move the global prep tab onto the config page then?

Yea, I think it makes the most sense.

There's a couple of options that are like a json list (resolution and fps I believe), just going from memory, so that might not be 100% accurate.

pgrunzjr: Now stored in config and setup moved to the config page.
image

pgrunzjr:
> @pgrunzjr The global prep commands tab isn't working with the latest commits. I click on it and nothing happens.

Newer changes have moved global prep commands to the Configuration section and the Apps page tabs removed.

src/config.cpp
```diff
@@ -424,7 +425,8 @@ sunshine_t sunshine {
   platf::appdata().string() + "/sunshine.conf", // config file
   {},                                           // cmd args
   47989,
-  platf::appdata().string() + "/sunshine.log", // log file
+  platf::appdata().string() + "/sunshine.log", // log file,
```
ReenigneArcher, line 428: comma not necessary after comment

src/config.cpp
```diff
@@ -424,7 +425,8 @@ sunshine_t sunshine {
...
+  {}
```
ReenigneArcher, line 429: please add comment

ReenigneArcher (changes requested):
Requesting a few small changes. We're going to need to add boost/json to our build process as well. I will take care of that in another PR so you don't have to dig through all the CI and packaging files.

I like the idea of using that though, can probably move the `resolutions` and `fps` options to use it as well, instead of that manual serialize function. I have some tweaks to how the config is handled in the UI in #1023, but I'll merge yours first so you don't have to fix any conflicts.

Edit: Can you also update the documentation, specifically "advanced_usage.rst". This is where we document each setting.

Edit 2: Apparently there's no `libboost-json-dev` library for debian distros. Do we need to install something additional?

pgrunzjr:
> Edit 3: boost-json was introduced in Boost 1.75. This is going to be a problem since Ubuntu 22.04 is on 1.74, and we're building Ubuntu 20.04 with Boost 1.71.

For the purpose of this PR, I think I should be able to update it to use Boost PropertyTree for the JSON stuff for now. In the future, we could look into using another library for JSON, maybe https://github.com/nlohmann/json.

pgrunzjr:
PR updated with documentation updates and to use Boost PropertyTree instead of Boost JSON.

src_assets/common/assets/web/config.html
ReenigneArcher, line 1: I think we need one minor change due a change in #1023. Around L966, just add the default value for the global prep commands. That might make L1000 irrelevant... Anyway the idea with #1023 was if the setting in the UI matches the default then we won't write it to the config file.

pgrunzjr: Defaults have been added. L1000 still needs to exist, otherwise Vue doesn't have anything to bind too and the page doesn't show the new rows when you click the + button unless you manually refresh.

src/config.cpp
```diff
@@ -7,6 +7,9 @@
 
 #include <boost/asio.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/json.hpp>
```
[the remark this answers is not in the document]
pgrunzjr: Ah, missed removing that. Will fix momentarily.

ReenigneArcher (approved):

PR #1054, opened 2023-03-22
Add CREATE_WAITABLE_TIMER_HIGH_RESOLUTION define to avoid compile errors for low system versions

ggqshr:
## Description

Sunshine use CreateWaitableTimerEx api and CREATE_WAITABLE_TIMER_HIGH_RESOLUTION flag to replace busy loop frame capture
https://github.com/LizardByte/Sunshine/blob/0dfbcfdbc4a996fc6431387fd227525e7ec57684/src/platform/windows/display_base.cpp#L94-L103
As commented in the above code, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION is only available on higher versions of systems. The above code is able to handle version-compatible behavior correctly at runtime, but when I compile on a lower version of the system that will report the following error:
> error: 'CREATE_WAITABLE_TIMER_HIGH_RESOLUTION' was not declared in this scope; did you mean 'CREATE_WAITABLE_TIMER_MANUAL_RESET'

To avoid compile errors, we can add the following code just like https://github.com/python/cpython/issues/89592#issuecomment-1093932643
```c++
#ifndef CREATE_WAITABLE_TIMER_HIGH_RESOLUTION
#define CREATE_WAITABLE_TIMER_HIGH_RESOLUTION 0x00000002
#endif
```

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display_base.cpp
+++ b/src/platform/windows/display_base.cpp
@@ -24,6 +24,10 @@ using namespace std::literals;
 }
 namespace platf::dxgi {
 namespace bp = boost::process;
+// Avoid compilation errors for low system versions
+#ifndef CREATE_WAITABLE_TIMER_HIGH_RESOLUTION
+#define CREATE_WAITABLE_TIMER_HIGH_RESOLUTION 0x00000002
+#endif
 
 capture_e duplication_t::next_frame(DXGI_OUTDUPL_FRAME_INFO &frame_info, std::chrono::milliseconds timeout, resource_t::pointer *res_p) {
   auto capture_status = release_frame();


psyke83:
What version of Windows are you attempting to compile on? The definition is provided by mingw's `synchapi.h` header for  Vista or later. Given that Microsoft's Desktop Duplication API is only supported on Windows 8 or later, my guess is that you just need to update your [MinGW-w64](https://www.mingw-w64.org/) installation to a newer version that includes the definition in the aforementioned header.

ggqshr:
> What version of Windows are you attempting to compile on?

win10 21H2

You're right, This problem can be solved by update MYSY2, i tried it on another new computer(also win10 21H2),  but the computer where the compilation error occurred is not very convenient to upgrade the MYSY2 version.

psyke83:
The latest MSYS2 release still supports Windows 8.1, so I think it's reasonable that a more recent version is installed for the purpose of compiling Sunshine rather than adding duplicate definitions to the Sunshine code. It's probably a good idea to keep your toolchain as updated as possible, especially when you consider that you could be also compiling against insecure security libraries such as OpenSSL, etc.

I think this can be closed in lieu of updating MSYS2. Thanks for your understanding.

PR #1063, opened 2023-03-24
Provide ability to force a capture method via configuration.

KuleRucket:
## Description
In Linux sunshine will test nvfbc, kms, wayland and x11 capture methods and will choose the first one that works in that order. This change allows you to force a specific one as long as it is available. If no capture method is found or if you select one that cannot be initialized, then an error message will be raised. 

Documentation update to follow.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -752,6 +752,36 @@ hevc_mode
 
       hevc_mode = 2
 
+capture
+^^^^^^^
+
+**Description**
+   Force specific screen capture method. If set to automatic then Sunshine will attempt to find a capture method in the order of the table below, otherwise it will only attempt to capture via the method specified.
+
+   .. Caution:: Applies to Linux only.
+
+**Choices**
+
+.. table::
+   :widths: auto
+
+   =========  ===========
+   Value      Description
+   =========  ===========
+   nvfbc      Use NVIDIA Frame Buffer Capture to capture direct to GPU memory. This is usually the fastest method for NVIDIA cards. For GeForce cards it will only work with drivers patched with `nvidia-patch <https://github.com/keylase/nvidia-patch/>`_ or `nvlax <https://github.com/keylase/nvidia-patch/>`_.
+   wlr        Capture for wlroots based Wayland compositors via DMA-BUF.
+   kms        DRM/KMS screen capture from the kernel. This requires that sunshine has cap_sys_admin capability. See :ref:`Linux Setup <about/usage:setup`.
+   x11        Uses XCB. This is the slowest and most CPU intensive so should be avoided if possible.
+   =========  ===========
+   
+**Default**
+   Automatic
+
+**Example**
+   .. code-block:: text
+
+      capture = kms
+      
 encoder
 ^^^^^^^
 
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -756,7 +756,7 @@ capture
 ^^^^^^^
 
 **Description**
-   Force specific screen capture method. If set to automatic then Sunshine will attempt to find a capture method in the order of the table below, otherwise it will only attempt to capture via the method specified.
+   Force specific screen capture method.
 
    .. Caution:: Applies to Linux only.
 
@@ -775,7 +775,7 @@ capture
    =========  ===========
    
 **Default**
-   Automatic
+   Automatic. Sunshine will use the first capture method available in the order of the table above.
 
 **Example**
    .. code-block:: text
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -770,7 +770,7 @@ capture
    =========  ===========
    nvfbc      Use NVIDIA Frame Buffer Capture to capture direct to GPU memory. This is usually the fastest method for NVIDIA cards. For GeForce cards it will only work with drivers patched with `nvidia-patch <https://github.com/keylase/nvidia-patch/>`_ or `nvlax <https://github.com/keylase/nvidia-patch/>`_.
    wlr        Capture for wlroots based Wayland compositors via DMA-BUF.
-   kms        DRM/KMS screen capture from the kernel. This requires that sunshine has cap_sys_admin capability. See :ref:`Linux Setup <about/usage:setup`.
+   kms        DRM/KMS screen capture from the kernel. This requires that sunshine has cap_sys_admin capability. See :ref:`Linux Setup <about/usage:setup>`.
    x11        Uses XCB. This is the slowest and most CPU intensive so should be avoided if possible.
    =========  ===========
    
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -768,9 +768,13 @@ capture
    =========  ===========
    Value      Description
    =========  ===========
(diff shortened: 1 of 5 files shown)


KuleRucket:
While making this change I came across this:
```
        // On Wayland, using KMS, the cursor is unreliable.
        // Hide it by default
        display_cursor = false;
```
I'm not sure if this is still valid.

ReenigneArcher:
This is a nice idea! I'll review in detail tomorrow.

KuleRucket:
Documentation added. I'm not sure how to test if my reference to Linux Setup is working.

docs/source/about/advanced_usage.rst
```diff
@@ -735,6 +735,36 @@ hevc_mode
...
+   :widths: auto
+
+   =========  ===========
+   Value      Description
+   =========  ===========
+   nvfbc      Use NVIDIA Frame Buffer Capture to capture direct to GPU memory. This is usually the fastest method for NVIDIA cards. For GeForce cards it will only work with drivers patched with `nvidia-patch <https://github.com/keylase/nvidia-patch/>`_ or `nvlax <https://github.com/keylase/nvidia-patch/>`_.
+   wlr        Capture for wlroots based Wayland compositors via DMA-BUF.
+   kms        DRM/KMS screen capture from the kernel. This requires that sunshine has cap_sys_admin capability. See :ref:`Linux Setup <about/usage:setup>`.
+   x11        Uses XCB. This is the slowest and most CPU intensive so should be avoided if possible.
+   =========  ===========
```
ReenigneArcher, line 758: I don't know if this will work, but you can try this to improve readability?

```rst
.. table::
   :widths: auto

   =========  ===========
   Value      Description
   =========  ===========
   nvfbc      Use NVIDIA Frame Buffer Capture to capture direct to GPU memory. This is usually the fastest method for
              NVIDIA cards. For GeForce cards it will only work with drivers patched with
              `nvidia-patch <https://github.com/keylase/nvidia-patch/>`_ or
              `nvlax <https://github.com/keylase/nvidia-patch/>`_.
   wlr        Capture for wlroots based Wayland compositors via DMA-BUF.
   kms        DRM/KMS screen capture from the kernel. This requires that sunshine has cap_sys_admin capability.
              See :ref:`Linux Setup <about/usage:setup>`.
   x11        Uses XCB. This is the slowest and most CPU intensive so should be avoided if possible.
   =========  ===========
```

src/config.cpp
```diff
@@ -351,6 +351,7 @@ video_t video {
     -1,
   }, // vt
 
+  {},  // capture method
```
ReenigneArcher, line 354: Since this is just "capture" in the header, it probably makes sense to be the same here.

src/platform/linux/misc.cpp
```diff
@@ -8,13 +8,15 @@
 
 #include <fstream>
 
+// local includes
```
ReenigneArcher, line 11: Not sure if you picked up on this from some of my other code, but thanks for adding these. `// standard includes` at L1 too?

Normally I put "lib includes" before "local" unless there's some conflict with it.

src/platform/linux/misc.cpp
```diff
@@ -518,34 +520,44 @@ std::unique_ptr<deinit_t> init() {
...
   }
 #endif
+
 #ifdef SUNSHINE_BUILD_CUDA
-  if(verify_nvfbc()) {
-    sources[source::NVFBC] = true;
+  if(config::video.capture.empty() || config::video.capture == "nvfbc") {
+    if(verify_nvfbc()) {
+      sources[source::NVFBC] = true;
+    }
```
ReenigneArcher, line 528: How about this?

```cpp
if(verify_nvfbc() && (config::video.capture.empty() || config::video.capture == "nvfbc")) {
  sources[source::NVFBC] = true;
}
```

ℹ️ Applies to all these if statements.

src/platform/linux/misc.cpp
```diff
@@ -518,34 +520,44 @@ std::unique_ptr<deinit_t> init() {
...
-    sources[source::X11] = true;
+  if(config::video.capture.empty() || config::video.capture == "x11") {
+    if(verify_x11()) {
+      sources[source::X11] = true;
+    }
   }
 #endif
 
   if(sources.none()) {
+    BOOST_LOG(error) << "Unable to initialize capture method";
```
ReenigneArcher, line 560: Should this be `fatal` instead of `error`?

src_assets/common/assets/web/config.html
```diff
@@ -591,7 +591,21 @@ <h1 class="my-4">Configuration</h1>
           HEVC is more CPU-intensive to encode, so enabling this may reduce performance when using software encoding.
         </div>
       </div>
-      <!--Encoder -->
+      <!--Capture-->
+      <div class="mb-3" v-if="platform === 'linux'">
+        <label for="capture" class="form-label">Force a Specific Capture Method</label>
+        <select id="capture" class="form-select" v-model="config.capture">
+          <option value>Autodetect</option>
```
ReenigneArcher, line 598: I'm not an expert in HTML, but should this be `value=""`? They're probably equivalent, so just checking.

ReenigneArcher (changes requested):
Overall the changes looks fine to me. I just have one suggestion to reduce the nesting of if statements, and a few other very minor requested changes.

src/config.cpp
KuleRucket: ok I'll change it

src/platform/linux/misc.cpp
KuleRucket: I'll arrange them to match the others. I was just nudging it towards the standard not to mix another change in the PR.

src/platform/linux/misc.cpp
KuleRucket: I prefer the first way as it's more readable, but fine I'll change it.

src/platform/linux/misc.cpp
KuleRucket: I have something in mind for a follow up PR. Currently if you configure a capture method that does not work, Sunshine will shut down and you will lose the web UI. I would like it to avoid exiting so that you are still able to edit configuration even if the capture is not working.

src_assets/common/assets/web/config.html
KuleRucket: I copied it from the encoder setting below so I guess it's equivalent. I'll change both though to make it clearer.

src/platform/linux/misc.cpp
ReenigneArcher: It's okay if you want to leave it how you have it.

KuleRucket:
I think it's all cleaned up now

ReenigneArcher (approved):

docs/source/about/advanced_usage.rst
```diff
@@ -752,6 +752,40 @@ hevc_mode
...
+.. table::
+   :widths: auto
+
+   =========  ===========
+   Value      Description
+   =========  ===========
+   nvfbc      Use NVIDIA Frame Buffer Capture to capture direct to GPU memory. This is usually the fastest method for
+              NVIDIA cards. For GeForce cards it will only work with drivers patched with
+              `nvidia-patch <https://github.com/keylase/nvidia-patch/>`_
+              or `nvlax <https://github.com/keylase/nvidia-patch/>`_.
```
ReenigneArcher, line 774: Was this intentional to point to the keylase patch? Someone brought this up on discord.

KuleRucket: No it was supposed to be https://github.com/illnyang/nvlax

ReenigneArcher: Thanks, submitted #1205 to correct it.

PR #1073, opened 2023-03-25
Nightly Notification Bug Fixes

### LINKED ISSUE #1072: Release Notifications do not work for Nightly branch

- [X] This issue is present in the nightly release

### Describe the Bug

The version notification in the web UI is currently bugged and does not work for nightly releases of Sunshine. It offers to upgrade to stable and is unable to recognize when a new nightly release is available for download.

### Expected Behavior

If a new nightly release is available, it should suggest it.
If a stable release exists, don't suggest it unless it's higher version than the nightly release.

### Additional Context

_No response_

### Host Operating System

other, n/a

### Operating System Version

N/A

### Architecture

32 bit

0.18.4

### Package

other (fork of this repo)

### GPU Type

none (software encoding)

### GPU Model

N/A

### GPU Driver/Mesa Version

N/A

N/A

### Config

```Shell
N/A
```

### Apps

```json
N/A
```

### Relevant log output

```shell
N/A
```

Nonary:
## Description
This pull request fixes a bug in the code that prevents the version update from working properly for nightly builds. The changes made in this pull request ensure that the expected behavior is met, which is to suggest a new nightly release when available, and only suggest a stable release if it has a higher version number than the nightly release.

Changes Made:

Updated the hasNewStable method to check nightly builds by comparing the latest nightly version tag with the stable version, and only returning true if the latest stable version is higher than their current nightly version. If the user does not have a nightly version of Sunshine, the existing functionality of the method remains the same.

Fixed a bug in the hasNewNightly method by using pop() instead of indexing by -1. JavaScript does not use negative indexes, so that would have returned back undefined instead of their original intended design decision, which was to retrieve the last index of the version split.

### Issues Fixed or Closed
- Fixes #1072 

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src_assets/common/assets/web/index.html
+++ b/src_assets/common/assets/web/index.html
@@ -89,7 +89,7 @@
       try {
         this.version = (await fetch("/api/config").then((r) => r.json())).version;
         this.githubVersion = (await fetch("https://api.github.com/repos/LizardByte/Sunshine/releases/latest").then((r) => r.json()));
-        if (this.version.split(".").length > 1) {
+        if (!this.buildVersionIsStable()) {
           this.nightlyData = (await fetch("https://api.github.com/repos/LizardByte/Sunshine/actions/workflows/CI.yml/runs?branch=nightly&status=success&per_page=1").then((r) => r.json())).workflow_runs[0];
         }
       } catch(e){
@@ -128,7 +128,7 @@
     },
     methods: {
       buildVersionIsStable() {
-        return !this.version || !this.nightlyData || this.version.split(".").length === 3;
+        return this.version.split(".").length === 3 || !this.version || !this.nightlyData;
       }
     }
   });
--- a/src_assets/common/assets/web/index.html
+++ b/src_assets/common/assets/web/index.html
@@ -117,7 +117,7 @@
         return v !== this.version.split(".")[0];
       },
       nightlyBuildAvailable() {
-        if (this.buildVersionIsStable()) { return false };
+        if (this.buildVersionIsStable()) return false;
         // If built with dirty git tree, return false
         if (this.version.indexOf("dirty") !== -1) return false;
         // Get the commit hash
--- a/src_assets/common/assets/web/index.html
+++ b/src_assets/common/assets/web/index.html
@@ -12,12 +12,12 @@
       <div v-if="loading">
         Loading Latest Release...
       </div>
-      <div v-else-if="!hasNewNightly && !hasNewStable">
+      <div v-else-if="!nightlyBuildAvailable && !stableBuildAvailable">
         <div class="alert alert-success">
           You're running the latest version of Sunshine
         </div>
       </div>
-      <div v-if="hasNewNightly">
+      <div v-if="nightlyBuildAvailable">
         <div class="alert alert-warning">
           <div class="d-flex justify-content-between">
             <div class="my-2">A new <b>Nightly</b> Version is Available!</div>
@@ -27,7 +27,7 @@
           <pre>{{nightlyData.display_title}}</pre>
         </div>
       </div>
-      <div v-if="hasNewStable">
+      <div v-if="stableBuildAvailable">
         <div class="alert alert-warning">
           <div class="d-flex justify-content-between">
             <div class="my-2">A new <b>Stable</b> Version is Available!</div>
@@ -97,8 +97,7 @@
       this.loading = false;
     },
     computed: {
-      // Check for new stable version
-      hasNewStable() {
+      stableBuildAvailable() {
         // If we can't get versions, return false
         if (!this.githubVersion || !this.version) return false;
         // If built with dirty git tree, return false
@@ -109,7 +108,7 @@
         if (v.indexOf("v") === 0) v = v.substring(1);
 
         // if nightly, we do an additional check to make sure its an actual upgrade.
-        if (this.version.split('.').length > 3) {
+        if (!this.buildVersionIsStable()) {
           const stableVersion = this.version.split('.').slice(0, 3).join('.');
           return this.githubVersion.tag_name.substring(1) > stableVersion;
         }
@@ -117,16 +116,19 @@
         // return true if the version is different, otherwise false
         return v !== this.version.split(".")[0];
       },
-      hasNewNightly() {
-        // If we're not on a nightly build, just return false
-        // If length of version split is 3, we're on a stable build
-        if (!this.version || !this.nightlyData || this.version.split(".").length === 3) return false;
+      nightlyBuildAvailable() {
+        if (this.buildVersionIsStable()) { return false };
         // If built with dirty git tree, return false
         if (this.version.indexOf("dirty") !== -1) return false;
         // Get the commit hash
         let commit = this.version.split(".").pop();
         // return true if the commit hash is different, otherwise false
         return this.nightlyData.head_sha.indexOf(commit) !== 0;
+      }
+    },
+    methods: {
+      buildVersionIsStable() {
+        return !this.version || !this.nightlyData || this.version.split(".").length === 3;
       }
     }
   });
--- a/src_assets/common/assets/web/index.html
+++ b/src_assets/common/assets/web/index.html
@@ -107,10 +107,16 @@
         let v = this.githubVersion.name;
         // If the version starts with a v, remove it
         if (v.indexOf("v") === 0) v = v.substring(1);
+
+        // if nightly, we do an additional check to make sure its an actual upgrade.
+        if (this.version.split('.').length > 3) {
+          const stableVersion = this.version.split('.').slice(0, 3).join('.');
+          return this.githubVersion.tag_name.substring(1) > stableVersion;
+        }
+
         // return true if the version is different, otherwise false
         return v !== this.version.split(".")[0];
       },
-      // Check for new nightly version
       hasNewNightly() {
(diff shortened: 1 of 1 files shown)


src_assets/common/assets/web/index.html
```diff
@@ -107,21 +107,27 @@ <h2>Legal</h2>
...
-      // Check for new nightly version
       hasNewNightly() {
         // If we're not on a nightly build, just return false
         // If length of version split is 3, we're on a stable build
         if (!this.version || !this.nightlyData || this.version.split(".").length === 3) return false;
         // If built with dirty git tree, return false
         if (this.version.indexOf("dirty") !== -1) return false;
         // Get the commit hash
-        let commit = this.version.split(".")[-1];
+        let commit = this.version.split(".").pop();
```
ReenigneArcher, line 127: Using pop changes the original array and ends up going crazy.

image

This part of the code was never hit before though, due to a mistake when I finalized the PR that added the versioning notifications.

L92 needs to change to this.

`if (this.version.split(".").length === 4) {`

Elix was splitting the version with `-g`, but I had to change it to `.` to satisfy some build systems like Archlinux/AUR... and I must have missed changing that.

src_assets/common/assets/web/index.html
```diff
@@ -107,21 +107,27 @@ <h2>Legal</h2>
...
+        // if nightly, we do an additional check to make sure its an actual upgrade.
+        if (this.version.split('.').length > 3) {
+          const stableVersion = this.version.split('.').slice(0, 3).join('.');
+          return this.githubVersion.tag_name.substring(1) > stableVersion;
+        }
+
         // return true if the version is different, otherwise false
         return v !== this.version.split(".")[0];
       },
-      // Check for new nightly version
```
ReenigneArcher, line 113: Besides the fact the comment is obvious, is there another reason it was removed?

src_assets/common/assets/web/index.html
```diff
@@ -107,21 +107,27 @@ <h2>Legal</h2>
...
         // Get the commit hash
-        let commit = this.version.split(".")[-1];
+        let commit = this.version.split(".").pop();
         // return true if the commit hash is different, otherwise false
         return this.nightlyData.head_sha.indexOf(commit) !== 0;
       }
     }
   });
-</script>
+</script>
```
ReenigneArcher, line 133: Please undo this (re-add blank line at end of file).

src_assets/common/assets/web/index.html
```diff
@@ -107,21 +107,27 @@ <h2>Legal</h2>
         let v = this.githubVersion.name;
         // If the version starts with a v, remove it
         if (v.indexOf("v") === 0) v = v.substring(1);
+
+        // if nightly, we do an additional check to make sure its an actual upgrade.
+        if (this.version.split('.').length > 3) {
```
ReenigneArcher, line 112: I'd prefer this to be just set to `=== 4`. If it's more than 4 it's for sure not a nightly.

Or set all the `=== 4` to `> 3`.

ReenigneArcher (changes requested):
Thanks for the PR! A few changes requested. Once the requested changes are made, I'll get this merged right away.

src_assets/common/assets/web/index.html
Nonary: That was the main reason, it was a bit too obvious. It’s often suggested in clean code to remove obvious comments at times.

src_assets/common/assets/web/index.html
Nonary: Not sure what you mean, I had to add code statements and that shifted the script tag down a few lines?

src_assets/common/assets/web/index.html
Nonary: You’re right, this is more of an error on my part for not double checking that. It is best practices to make computed properties immutable, meaning if you mutated the value it doesn’t change the original.

That’s normally how it’s done on my projects anyway and due to my on conventions I made an incorrect assumption.

The fix is pretty simple either way.

Nonary: Actually looking at that, pop shouldn’t be changing original version here. I thought it was a computed property, oops.

Can you explain your setup so I can see if I can replicate this bug?

src_assets/common/assets/web/index.html
ReenigneArcher: The line number changing isn't the issue. Your IDE might be screwing with it. There should be a blank line at the end of each file though. You can see this on the files changed tab. https://github.com/LizardByte/Sunshine/pull/1073/files

image

src_assets/common/assets/web/index.html
ReenigneArcher: My screenshot shows what happens when L92 is fixed AND using the `pop`. If using the `[-1]` and fixing L92, the code works correctly.

This screenshot, below, is strictly changing L92 (doesn't include your fix for stable).

image

And without fixing L92, it just shows the "new stable version" message.

Nonary: I know, but I’m trying to figure out how you’re seeing that as it doesn’t happen on my side. 

The -1 does not work in JavaScript. That code statement is busted, it will return back undefined in all cases. 

Basically changing it to pop() doesn’t cause it to break, but rather some other lines of code in the existing code has a bug I can’t replicate on my side.

ReenigneArcher: Not sure what happened to cause the versioning to get all messed up in my original screenshot. `pop` seems to be okay now.

We still need to address L92 though.

Nonary: I thought I had included that fix already in my commits, not sure why it reverted back. I've pushed those requested changes as well as I refactored the names of the computed properties to reflect their intention as well, which would eliminate the purpose of the comments of the function itself.

src_assets/common/assets/web/index.html
```diff
@@ -89,16 +89,15 @@ <h2>Legal</h2>
       try {
         this.version = (await fetch("/api/config").then((r) => r.json())).version;
         this.githubVersion = (await fetch("https://api.github.com/repos/LizardByte/Sunshine/releases/latest").then((r) => r.json()));
-        if (this.version.split("-g").length > 1) {
+        if (this.version.split(".").length > 1) {
```
ReenigneArcher, line 92: Needs to be `> 3` or `=== 4`. Previously when splitting with `-g` there would be only one occurrence, so the length was 2... but if checking against `1` with a `.` it will always think it's a nightly version.

ReenigneArcher (changes requested):

src_assets/common/assets/web/index.html
ReenigneArcher: or `if (!this.buildVersionIsStable()) {`

ReenigneArcher: Alternatively, could have a similar method to `buildVersionIsStable` added to check if it's a nightly version.

Nonary: @ReenigneArcher That's basically what I did, I put !this.buildVersionIsStable() and changed the boolean logic to do the version check first, so that it doesn't short circuit and return a false positive.

Basically if you have a boolean check like this

a || b || c

If a is true, then it doesn't test b or c.

So that is why it was important to move the version check as the first condition in the logic basically.

src_assets/common/assets/web/index.html
```diff
@@ -89,16 +89,15 @@ <h2>Legal</h2>
       try {
         this.version = (await fetch("/api/config").then((r) => r.json())).version;
         this.githubVersion = (await fetch("https://api.github.com/repos/LizardByte/Sunshine/releases/latest").then((r) => r.json()));
-        if (this.version.split("-g").length > 1) {
+        if (!this.buildVersionIsStable()) {
```
ReenigneArcher, line 92: This is not enough to determine if it's a nightly version... and in fact will always return false since checking for stable version checks if `this.nightlyData` is populated.

image


But this will work.

Change L92 to `if (this.buildVersionIsNightly()) {`

And add a new method

```js
methods: {
      buildVersionIsStable() {
        return this.version.split(".").length === 3 || !this.version || !this.nightlyData;
      },
      buildVersionIsNightly() {
        return this.version.split("."). length === 4 || !this.version
      }
    }
```

These two changes result in the code working correctly.

image

Nonary: in JavaScript, booleans will short circuit. So it shouldn’t even test !nightlyData because the first check is checking the version split.

I’m not sure why you’re running into errors again though

Nonary: Never mind it’s not short circuiting in this context, I’ll fix it later in a few hours.

Nonary: My first thought here is there is no need to even check nightly data to determine if it’s nightly/stable

ReenigneArcher: Maybe not.

If it helps this is the logic that determines what version is used during compilation.

https://github.com/LizardByte/Sunshine/blob/a487fb31eaaa47c98bfab45eddaf34e33f2932ab/CMakeLists.txt#L14-L71

Nonary: They're still needed, sort of. The main reason is that UI might try to load before the data is fully loaded. It's mostly because of how the fetch requests are being done. I will refactor those at a later time but I fixed the issues now.

ReenigneArcher (approved):

PR #1080, opened 2023-03-26
Configurable Max Frame Count for Image Buffer (for VRAM optimization)

Nonary:
## Description
Provides an option for users to change the maximum size of the image buffer to save VRAM.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -414,6 +414,29 @@ dwmflush
 
       dwmflush = enabled
 
+
+max_frame_buffer_size
+-----------------
+
+**Description**
+   This option specifies the number of images to allocate in the image buffer. 
+   Increase the count improves stream stability at the expense of increased VRAM usage. 
+   If configured with fewer frames, VRAM usage will significantly decrease, but this may cause performance issues.
+   
+   .. Note:: If you're streaming at 4k with HDR enabled, the default setting will reserve roughly 1.8GB of VRAM.
+      It is recommended to reduce the preloaded frames, especially if streaming with those settings and playing modern games.
+      However, if you are using multiple streams at once it is recommended to leave this at default, or multiply it such as 3 * (Channel Count)
+
+**Default**
+   ``12``
+
+**Values**
+   An integer value between ``2`` and ``12``, inclusive.
+
+**Example**
+      max_frame_buffer_size = 12
+
+
 Audio
 -----
 
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -299,13 +299,18 @@ output_name
    .. Tip:: To find the name of the appropriate values follow these instructions.
 
       **Linux**
-         .. code-block:: bash
+         During Sunshine startup, you should see the list of detected monitors:
 
-            xrandr --listmonitors
+         .. code-block:: text
 
-         Example output: ``0: +HDMI-1 1920/518x1200/324+0+0  HDMI-1``
+            Info: Detecting connected monitors
+            Info: Detected monitor 0: DVI-D-0, connected: false
+            Info: Detected monitor 1: HDMI-0, connected: true
+            Info: Detected monitor 2: DP-0, connected: true
+            Info: Detected monitor 3: DP-1, connected: false
+            Info: Detected monitor 4: DVI-D-1, connected: false
 
-         You need to use the value before the colon in the output, e.g. ``0``.
+         You need to use the value before the colon in the output, e.g. ``1``.
 
       .. Todo:: macOS
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -331,6 +331,7 @@ namespace config {
     0,  // hevc_mode
 
     1,  // min_threads
+    12, // max_frame_buffer_size
     {
       "superfast"s,  // preset
       "zerolatency"s,  // tune
@@ -911,6 +912,7 @@ namespace config {
 
     int_f(vars, "qp", video.qp);
     int_f(vars, "min_threads", video.min_threads);
+    int_between_f(vars, "max_frame_buffer_size", video.max_frame_buffer_size, { 2, 12 });
     int_between_f(vars, "hevc_mode", video.hevc_mode, { 0, 3 });
     string_f(vars, "sw_preset", video.sw.sw_preset);
     string_f(vars, "sw_tune", video.sw.sw_tune);
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -331,7 +331,7 @@ namespace config {
     0,  // hevc_mode
 
     1,  // min_threads
-    12, // max_frame_buffer_size
+    12,  // max_frame_buffer_size
     {
       "superfast"s,  // preset
       "zerolatency"s,  // tune
--- a/src_assets/common/assets/web/config.html
+++ b/src_assets/common/assets/web/config.html
@@ -1016,7 +1016,7 @@
     "nv_tune": "ull",
     "origin_pin_allowed": "pc",
     "origin_web_ui_allowed": "lan",
-    "preloaded_frames": 12,
+    "max_frame_buffer_size": 12,
     "qsv_coder": "auto",
     "qsv_preset": "medium",
     "resolutions": "[352x240,480x360,858x480,1280x720,1920x1080,2560x1080,3440x1440,1920x1200,3860x2160,3840x1600]",
--- a/src_assets/common/assets/web/config.html
+++ b/src_assets/common/assets/web/config.html
@@ -573,6 +573,24 @@
           You need to use the value before the colon in the output, e.g. <b>1</b>.
         </div>
       </div>
+      <div class="mb-3">
+          <label for="qp" class="form-label">Maximum Frame Buffer Size</label>
+          <input
+            type="number"
+            class="form-control"
+            id="max_frame_buffer_size"
+            placeholder="12"
+            min="2"
+            max="12"
+            v-model="config.max_frame_buffer_size"
+          />
+          <div class="form-text">
+            At the expense of using more VRAM, you can adjust the maximum amount of images to cache in Sunshine. <br/>
+            If configured with fewer frames, VRAM usage will significantly decrease, but this may cause performance issues, such as lower framerates or instability. <br/>
+            Most systems can handle 3 frames without issues, and users are encouraged to experiment to find the perfect balance for their system based on their hardware capabilities. <br/>
+            If you are using the multiple session feature, it is recommended to keep this at 12 as those streams are more performance intensive.
+          </div>
+      </div>
     </div>
     <div v-if="currentTab === 'advanced'" class="config-page">
       <!--Port family-->
@@ -998,6 +1016,7 @@
     "nv_tune": "ull",
     "origin_pin_allowed": "pc",
     "origin_web_ui_allowed": "lan",
+    "preloaded_frames": 12,
     "qsv_coder": "auto",
     "qsv_preset": "medium",
     "resolutions": "[352x240,480x360,858x480,1280x720,1920x1080,2560x1080,3440x1440,1920x1200,3860x2160,3840x1600]",
--- a/src_assets/common/assets/web/config.html
+++ b/src_assets/common/assets/web/config.html
@@ -560,9 +560,17 @@
           v-model="config.output_name"
         />
         <div class="form-text">
-          xrandr --listmonitors<br />
-          Example output:
-          <pre>   0: +HDMI-1 1920/518x1200/324+0+0 HDMI-1</pre>
+          During Sunshine startup, you should see the list of detected monitors, e.g.:<br />
+          <br />
+          <pre style="white-space: pre-line;">
+            Info: Detecting connected monitors
+            Info: Detected monitor 0: DVI-D-0, connected: false
+            Info: Detected monitor 1: HDMI-0, connected: true
+            Info: Detected monitor 2: DP-0, connected: true
+            Info: Detected monitor 3: DP-1, connected: false
+            Info: Detected monitor 4: DVI-D-1, connected: false
+          </pre>
+          You need to use the value before the colon in the output, e.g. <b>1</b>.
         </div>
       </div>
     </div>
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -49,7 +49,7 @@ jobs:
 
     steps:
       - name: Automerging
-        uses: pascalgn/automerge-action@v0.15.5
+        uses: pascalgn/automerge-action@v0.15.6
         env:
           BASE_BRANCHES: nightly
           GITHUB_TOKEN: ${{ secrets.GH_BOT_TOKEN }}
--- a/.github/workflows/ci-qodana.yml
+++ b/.github/workflows/ci-qodana.yml
@@ -20,6 +20,8 @@ concurrency:
 jobs:
   qodana_initial_check:
     name: Qodana Initial Check
+    permissions:
+      actions: write  # required to use workflow dispatch on fork PRs
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -235,6 +237,8 @@ jobs:
     needs: [qodana_initial_check, qodana]
     runs-on: ubuntu-latest
     name: Dispatch Qodana
+    permissions:
+      actions: write  # required to use workflow dispatch on fork PRs
     if: ${{ needs.qodana_initial_check.outputs.files != '' }}
     steps:
       - name: Setup qodana publish inputs
--- a/.github/workflows/issues-stale.yml
+++ b/.github/workflows/issues-stale.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Stale
-        uses: actions/stale@v7
+        uses: actions/stale@v8
         with:
           close-issue-message: >
             This issue was closed because it has been stalled for 10 days with no activity.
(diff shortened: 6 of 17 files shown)


Nonary:
Please note that this is a draft and needs a few more fixes like validation, and removing the fact it was copy and pasted from another pull request. I am not sure how to flag PR as draft.

src/config.cpp
```diff
@@ -842,6 +843,7 @@ void apply_config(std::unordered_map<std::string, std::string> &&vars) {
 
   int_f(vars, "qp", video.qp);
   int_f(vars, "min_threads", video.min_threads);
+  int_f(vars, "preloaded_frames", video.preloaded_frames);
```
psyke83, line 846: You need to validate a minimum value of 2, otherwise Sunshine will crash. I `int_between_f` (see the line directly below).

src_assets/common/assets/web/config.html
```diff
@@ -517,6 +517,21 @@ <h1 class="my-4">Configuration</h1>
           <pre>   0: +HDMI-1 1920/518x1200/324+0+0 HDMI-1</pre>
         </div>
       </div>
+      <div class="mb-3">
+          <label for="qp" class="form-label">Preloaded Frame Count</label>
+          <input
+            type="number"
+            class="form-control"
```
psyke83, line 524: You should probably enforce a min (2) / max (12) here as well.

Nonary: That's the plan, main reason why I kept this as draft for the moment.

src_assets/common/assets/web/config.html
```diff
@@ -517,6 +517,21 @@ <h1 class="my-4">Configuration</h1>
...
+            id="min_fps_target"
```
ReenigneArcher, line 525:
```suggestion
            id="preloaded_frames"
```

ReenigneArcher (changes requested):
Will also need docs updated in `advanced_usage.rst` for the new config option.

src_assets/common/assets/web/config.html
ReenigneArcher: Example in the port option.

```html
        <input
          type="number"
          min="0"
          max="65529"
          class="form-control"
          id="port"
          placeholder="47989"
          v-model="config.port"
        />
```

src_assets/common/assets/web/config.html
```diff
@@ -517,6 +517,21 @@ <h1 class="my-4">Configuration</h1>
...
+            placeholder="3"
```
ReenigneArcher, line 526: Placeholder value should be the default.

```suggestion
            placeholder="12"
```

ReenigneArcher (changes requested):

Nonary:
You guys are giving me reviews before I flagged it for review, oof, brutal

ReenigneArcher:
> You guys are giving me reviews before I flagged it for review, oof, brutal
Sorry, ha... just trying to keep up with all the notifications.

docs/source/about/advanced_usage.rst
```diff
@@ -395,6 +395,32 @@ dwmflush
...
+   An integer value between ``2`` and ``12``, inclusive.
+
+**Example**
+   To set the number of preloaded frames to 6, add the following line to the configuration file:
+
+   .. code-block:: text
+
+      preloaded_frames = 6
+
+   This will preload 6 frames in the image buffer, reducing VRAM usage but potentially decreasing stream stability.
```
ReenigneArcher, line 422: To remain consistent with the rest of this documentation.

```suggestion
**Range**
   ``2-12``

**Example**
   .. code-block:: text

      preloaded_frames = 12
```

Nonary: I'll be honest, the docs need a complete overhaul because they're not very descriptive at all. But if we want it to be that way, sure I guess.

KuleRucket:
I think the documentation could be better on this one to prevent a lot of questions. For example you could explain what preloaded frames actually achieves (or provide a link that explains it). 

Also what behaviour manifests when it is set too low? "may cause performance issues" is a bit vague.

Nonary:
> I think the documentation could be better on this one to prevent a lot of questions. For example you could explain what preloaded frames actually achieves (or provide a link that explains it).
> 
> Also what behaviour manifests when it is set too low? "may cause performance issues" is a bit vague.

If people can't understand what stream stability and lowering the value may cause performance issues, they're really not going to understand anything we document anyway.

Nonary:
> > If people can't understand what the term stability means and that lowering the value may cause performance issues, they're really not going to understand anything we document anyway.
> 
> What sort of stability? Image quality? frame rate? connection? What sort of performance issues; encoding? capture? decoding? network?
> 
> I don't understand what it does from the description even though I have a vague understanding of what the code is doing.
> 
> EDIT: The main point I would be looking for is how do I know when I have set it too low? What behaviour should I look for?

We don't know, it's different for everyone. Many can run at 2 frames and experience 0 differences, while others can report odd behavior. Even if we did know the specific performance issue, it doesn't actually matter. What matters is the end result to the user, they would notice their streams performance has been impacted, such as more stuttering, less smooth... its not as good as it used to be basically.

Explaining it as network, encoding, capture, decoding, is irrelevant as the users would not understand how to debug that situation anyway.

KuleRucket:
It just sound to me that you are making a setting configurable but can't really explain what the purpose of preloading frames is. If that's the case then fine, but if not I would appreciate a better explanation.

Nonary:
> It just sounds to me that you are making a setting configurable but can't really explain what the purpose of preloading frames is. If that's the case then fine, but if not I would appreciate a better explanation.

No, we have a general idea of what the buffer is used for. What we don't know is if its even necessary to be set to 12 frames anymore. Maybe modern hardware is fast enough to where it doesn't matter, maybe it doesn't impact Windows users, but would for Linux or Mac. 

I would suggest that if you are confused with any setting in Sunshine you are welcome to ask in our discord, or you can experiment yourself and draw your own conclusions. Otherwise, I wouldn't change the setting at all if it doesn't benefit you in any way. 

For example, the documentation clearly states that this is more for optimizing the VRAM usage. So if you don't have any VRAM issues or have plenty of it, then I wouldn't bother changing this setting.

psyke83:
@Nonary

Not sure if you were kidding before, but no offense was intended by adding feedback to your draft. This looks like a potentially useful change.

Regarding the ongoing discussion over the purpose of the buffer size: there are two types of capture methods in Sunshine (sync and async, of which only the latter uses the 12 buffers relevant to this PR via the `captureThread` function). 

Until relatively recently, async capture was only used for `vaapi`, `nvenc` (only on Linux) and `software` encoding, but recent changes have enabled async capture for every encoder except for `videotoolbox` on Mac. It's possible that loki originally set an arbitrary value for the buffers without noticing the VRAM overhead it introduces for the hardware encoders. This overhead also didn't affect Windows hardware encoding at all until Cameron's recent performance optimizations that switched the Windows encoders to async capture.

For my GPU (RX 6600), 2 buffers seems to work fine at 4K60 without introducing any noticeable dropped frames in Moonlight's statistics, and reduces VRAM usage (from ~300MB to ~100MB at 1440p), but my card (& client) is not powerful enough to test the impact at even higher framerates or when the added overhead of HDR is introduced.

 We might also want to think about setting a lower default than 12, but feedback from users would be necessary; we have to keep in mind that almost all encoders (including software) are affected by this.

Nonary:
@psyke83 I was only jesting, no offense taken. Great minds often think a like, so it's nice to see that we're all thinking about the same thing. Makes me feel like I have a great mind, at least that's what I tell myself anyway HAH.

I do agree we could probably get away with dropping down the frames from 12, but without feedback it would be hard to do. Part of me wants to set it to 3 and then just see if crap hits the fan so to speak, but I also think we can realistically drop it to 6 and it wouldn't impact anyone except for Mac I guess?

cgutman:
I don't think we should expose this option users at all. The user is not going to be able to make a reasonable choice other than trying various settings. The answer might be different between resolution/FPS/bitrate settings or based on GPU/CPU load, so they can't even fully express the level of configurability needed in this single option. 

The capture thread itself already knows the answer of how many frames we need, because we can look at their `use_count()`. If we go to move to the next frame and it has `use_count() > 1`, then we know we need to allocate a new frame (up to some limit of maybe 6 or so). Since whatever led to the allocation of extra frames could be transient, we can periodically look through the frame list and reap some frames if it looks like we don't need all of them anymore.

Nonary:
> I don't think we should expose this option users at all. The user is not going to be able to make a reasonable choice other than trying various settings. The answer might be different between resolution/FPS/bitrate settings or based on GPU/CPU load, so they can't even fully express the level of configurability needed in this single option. 
> 
> 
> 
> The capture thread itself already knows the answer of how many frames we need, because we can look at their `use_count()`. If we go to move to the next frame and it has `use_count() > 1`, then we know we need to allocate a new frame (up to some limit of maybe 6 or so). Since whatever led to the allocation of extra frames could be transient, we can periodically look through the frame list and reap some frames if it looks like we don't need all of them anymore.

The ultimate goal was to make it dynamically adjust on its own. But I didn't have the knowledge for that and the current configuration is consuming a lot of VRAM and causing performance problems when streaming... so it's more of a temporary option until it's automated.

cgutman:
> The ultimate goal was to make it dynamically adjust on its own. But I didn't have the knowledge for that and the current configuration is consuming a lot of VRAM and causing performance problems when streaming... so it's more of a temporary option until it's automated.

Got it. In that case, I think we ought to do something even more simple than this and just lower the fixed count from 12 to 6. 6 frames of latency is already 100 ms at 60 FPS, so that's well beyond what most users are willing to tolerate.

I strongly suspect most systems don't need more than 2 or 3 frames in the pool. The big perf win of this async encode path is being able to parallelize capture and encoding operations, and you only need 1 or 2 frames to do that. If an encoder isn't able to keep up with the capture rate, throwing more buffers at the problem isn't going to improve throughtput anyway.

ns6089:
First, maybe call it "Capture buffer size (in frames)", or something along these lines. "Preloaded frames count" may make people assume it adds latency, which is not true for the most part.

And second, not sure if it's possible to fully automate. Because automatically increasing queue size may contribute to encoder throttling in the first place due to vram pressure.

Nonary:
@cgutman I am going to tackle it, you've provided an explanation of how it can be done. So I'll work on adjusting the round robin to be that way... I am totally fine with writing out the code and dedicating some time to it, I just really needed a high level explanation of how to do it so I appreciate that.

cgutman:
> And second, not sure if it's possible to fully automate. Because automatically increasing queue size may contribute to encoder throttling in the first place due to vram pressure.

The current round-robin system of picking the `img` to use for capture is actually a worst case scenario for this too. We're always using the least recently used `img` object, which is most likely to have been evicted to due to VRAM pressure.

I think we can put enough guardrails on the dynamic allocation code to prevent it from reaching a size where the VRAM usage starts to significantly impact performance. That along with picking the first unused frame will probably ensure we don't encounter this problem in practice.

Nonary:
And of course as soon as I say this, I realize a glaring bug, with the imgs not actually increasing in size... woops... will fix that in a moment

Nonary:
@cgutman okay, that should be fixed now. From my small scale of testing, it appears to work. I will try it for like an hour or two and see the results...

Nonary:
> First, maybe call it "Capture buffer size (in frames)", or something along these lines. "Preloaded frames count" may make people assume it adds latency, which is not true for the most part.
> 
> And second, not sure if it's possible to fully automate. Because automatically increasing queue size may contribute to encoder throttling in the first place due to vram pressure.

It does add latency, but rather very small amounts. I choose the word preloaded frames, because it is referring to a buffer and most people would be familiar with what a preloaded frame is as opposed to just calling what it really is. Although it seems like we might not even need the configuration anyway so.

I have an idea to workaround the second issue, if the concern is that it ends up going in a runaway state where it just keeps increasing the buffer size because the act of increasing the size is causing it to go into use_count() > 1 again... if I am understanding this correctly....

ReenigneArcher:
Updating w/ rebase to make this easier to fix conflicts after adjusting clang-format rules

Nonary:
Okay the rebase and clang changes broke everything, I am trying to figure out how to fix at the moment...

Nonary:
Yeah honestly it is so screwed up I will just recreate it again

ReenigneArcher:
> Okay the rebase and clang changes broke everything, I am trying to figure out how to fix at the moment...

https://discord.com/channels/804382334370578482/1074377411069747250/1083218068148330526

https://discord.com/channels/804382334370578482/1074377411069747250/1090088303031746761

Luckily, this is a small PR.

Nonary:
Okay this proof of concept is going to need a lot of work, basically the image buffer cannot be changed, at least the size.

First thing off the top of my head is to use a fixed sized, but null out the images we don't need anymore. But this will require refactoring quite a bit of the capture code and would have to add some extra security checks but it seems doable. I am not that experience with C++ but I will try this small scale before trying to get it to work like I originally intended it to be.

ns6089:
My take on how it can be approached.

Ditch the round-robin balancing altogether, I don't think it applies well to memory where you always want the most "hot" free region to operate on, not the least, due to how hardware caches are designed.

Decide the "normal" pool size, I think it's 2 for current code - 1 is sent to capture, 1 to decoder. Don't delete these two textures after use.

Now for the "boost" pool size. Honestly, I'd rather throttle the capture if encoder can't keep up and be done with that. [Present code](https://github.com/LizardByte/Sunshine/blob/master/src/video.cpp#L817) already does that.

But if do it anyway, set it to something like 4 or 6. And delete/recreate these "boost" textures after each use. But fighting positive feedback loops in effective blackboxes is not fun, if at all possible. And you might end up with fluctuating queues and choppy playback anyway.

ReenigneArcher:
Since this PR was opened in order to address high VRAM usage, perhaps, as a short term solution we just set the buffer to 6 to cut VRAM usage in half. We can ask for testers on the discord to ensure there's no regressions. There's 4 `@tester-<platform>` roles.

Nonary:
> Since this PR was opened in order to address high VRAM usage, perhaps, as a short term solution we just set the buffer to 6 to cut VRAM usage in half. We can ask for testers on the discord to ensure there's no regressions. There's 4 `@tester-<platform>` roles.

I am pretty sure with some slight modifications to this proof of concept, if we just have it scale up and never back down (to avoid the crash I am talking about) it would still work for most of our users. It's unlikely it would ever reach to the same amount of images. Perhaps maybe more polish in having it scale up only if it thread waits more than a few times.

There is relatively low risk of crash in the proof of concept as long as we do not scale down. I just need to double check security wise if what I am doing is safe to do basically.

Nonary:
Okay, I have tried all sorts of things and basically, the nullptr buffer worked the best. I tried thread safe round robins, rotating buffers, you name it... they all had issues with resizing the buffer even with proper synchronization, it would just break randomly. I am sure it's probably because I am not an expert with multithreading. 

But anyways, the current commit is a working proof of concept of a dynamically adjusting frame buffer that will increase up to 12 if it detects it needs to do so.

If there is anyone else here advanced in multithreading or optimizing it better, I am open to suggestions. End result is that it will most likely work fine for most people with just 2 in the buffer, but if not, it will go to 3,4,etc.

Nonary:
*sigh* nevermind, it is filling up the image buffer with all frames again.

At this point yeah I am just going to open a PR that will change the default down to like 4, until I figure this out or someone else deos.

ns6089:
Removed round-robin, added on-demand allocation (but no on-demand delete), decreased buffer size to 4 frames, added debug print to log every 10s to test every encoder.

If anyone's interested https://github.com/ns6089/Sunshine/compare/nightly...remove_round_robin

ReenigneArcher:
@ns6089 do you mind creating a draft pull request?

ns6089:
@ReenigneArcher Done https://github.com/LizardByte/Sunshine/pull/1098

Nonary:
Okay, I think my proof of concept is ready for testing.

I fixed the revolving buffer and in first testing, looks like I can get 4kHDR working with just 480MB of VRAM when previously it was taking 1.8GB

ns6089:
It appears we can forgo all those fancy large growing queues and simply limit the size to 3. At least until underlying threading code is changed.

No more than 3 is currently utilized even with throttled encoder, because only single conditional variable is used to transport captured images between capture and encoding threads.

``using img_event_t = std::shared_ptr<safe::event_t<std::shared_ptr<platf::img_t>>>;``
https://github.com/LizardByte/Sunshine/blob/master/src/video.cpp#L1302

ns6089:
As to where number 3 comes from. 1 image currently is used by encoder, 1 image is sitting in conditional variable waiting to be grabbed by encoder, 1 image is sent to next capture pass.

Nonary:
> It appears we can forgo all those fancy large growing queues and simply limit the size to 3. At least until underlying threading code is changed.
> 
> No more than 3 is currently utilized even with throttled encoder, because only single conditional variable is used to transport captured images between capture and encoding threads.
> 
> `using img_event_t = std::shared_ptr<safe::event_t<std::shared_ptr<platf::img_t>>>;` https://github.com/LizardByte/Sunshine/blob/master/src/video.cpp#L1302

So, basically have it set to 2 and then scale down dynamically. My latest proof of concept actually is pretty simple and was working even when scaling down as well

Nonary:
I've reverted my proof of concept and restored the old options, which was to offer an option in Sunshine to manually adjust it. The main goal is to make this dynamic and an option that doesn't require configuration from users. But doing so will require a lot of changes to Sunshine, and require testing. It also may not even work out in the end, you can follow the progress on #1098

I say we merge this in the meantime to give users an option to lower the frame count to save their VRAM as opposed to keeping it unmerged.

cgutman:
Superseded by #1098

PR #1095, opened 2023-03-28
Fix system tray Open Sunshine

KuleRucket:
## Description
In Linux, the action to Open Sunshine in the system tray did not work. The action calls "open URL", but really it should be "xdg-open URL".

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/system_tray.cpp
+++ b/src/system_tray.cpp
@@ -52,7 +52,7 @@ namespace system_tray {
   #else  // if unix
     // set working dir to user home directory
     working_dir = boost::filesystem::path(std::getenv("HOME"));
-    std::string cmd = R"(open ")" + url + R"(")";
+    std::string cmd = R"(xdg-open ")" + url + R"(")";
   #endif
 
     boost::process::environment _env = boost::this_process::environment();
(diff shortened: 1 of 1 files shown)


src/system_tray.cpp
```diff
@@ -52,7 +52,7 @@ namespace system_tray {
   #else  // if unix
     // set working dir to user home directory
     working_dir = boost::filesystem::path(std::getenv("HOME"));
```
ReenigneArcher, line 54:
```suggestion
  #elif defined(__linux__) || defined(linux) || defined(__linux)
    // set working dir to user home directory
    working_dir = boost::filesystem::path(std::getenv("HOME"));
```

src/system_tray.cpp
```diff
@@ -52,7 +52,7 @@ namespace system_tray {
...
-    std::string cmd = R"(open ")" + url + R"(")";
+    std::string cmd = R"(xdg-open ")" + url + R"(")";
   #endif
```
ReenigneArcher, line 56:
```suggestion
  #elif defined(__APPLE__) || defined(__MACH__)
    std::string cmd = R"(open ")" + url + R"(")";
  #endif
```

ReenigneArcher (changes requested):
For macOS, it should still be "open"?

I know the tray is disable in macOS right now due to a threading issue, but still want to make it as easy as possible to implement if we find a solution to that.

KuleRucket:
> For macOS, it should still be "open"?
> 
> I know the tray is disable in macOS right now due to a threading issue, but still want to make it as easy as possible to implement if we find a solution to that.

I'll take your word for it as I have no experience with macOS.

ReenigneArcher (approved):

PR #1098, opened 2023-03-29
Minimize vram usage

ns6089:
An attempt to minimize vram usage without introducing possible computational bottlenecks.
Still needs a lot of work, but so far the results look promising.

## Progress
- [x] Implement Windows platform
- [x] Test and document Windows platform
- [x] Implement Linux platform
- [x] Test and document Linux platform
- [x] Implement MacOS platform
- [ ] Test and document MacOS platform

Before (4K60 SDR)
2023-03-31 16_55_11-sunshine exe (15700) Properties

After (4K60 SDR)
2023-03-31 16_50_21-sunshine exe (12084) Properties

### DIFF
--- a/src/platform/macos/display.mm
+++ b/src/platform/macos/display.mm
@@ -37,38 +37,46 @@ namespace platf {
     }
 
     capture_e
-    capture(snapshot_cb_t &&snapshot_cb, std::shared_ptr<img_t> img, bool *cursor) override {
-      __block auto img_next = std::move(img);
-
+    capture(const push_captured_image_cb_t &push_captured_image_cb, const pull_free_image_cb_t &pull_free_image_cb, bool *cursor) override {
       auto signal = [av_capture capture:^(CMSampleBufferRef sampleBuffer) {
-        auto av_img_next = std::static_pointer_cast<av_img_t>(img_next);
-
         CFRetain(sampleBuffer);
 
         CVPixelBufferRef pixelBuffer = CMSampleBufferGetImageBuffer(sampleBuffer);
         CVPixelBufferLockBaseAddress(pixelBuffer, kCVPixelBufferLock_ReadOnly);
 
-        if (av_img_next->pixel_buffer != nullptr)
-          CVPixelBufferUnlockBaseAddress(av_img_next->pixel_buffer, 0);
+        std::shared_ptr<img_t> img_out;
+        if (!pull_free_image_cb(img_out)) {
+          // got interrupt signal
+          // returning false here stops capture backend
+          return false;
+        }
+        auto av_img = std::static_pointer_cast<av_img_t>(img_out);
 
-        if (av_img_next->sample_buffer != nullptr)
-          CFRelease(av_img_next->sample_buffer);
+        if (av_img->pixel_buffer != nullptr)
+          CVPixelBufferUnlockBaseAddress(av_img->pixel_buffer, 0);
 
-        av_img_next->sample_buffer = sampleBuffer;
-        av_img_next->pixel_buffer = pixelBuffer;
-        img_next->data = (uint8_t *) CVPixelBufferGetBaseAddress(pixelBuffer);
+        if (av_img->sample_buffer != nullptr)
+          CFRelease(av_img->sample_buffer);
+
+        av_img->sample_buffer = sampleBuffer;
+        av_img->pixel_buffer = pixelBuffer;
+        img_out->data = (uint8_t *) CVPixelBufferGetBaseAddress(pixelBuffer);
 
         size_t extraPixels[4];
         CVPixelBufferGetExtendedPixels(pixelBuffer, &extraPixels[0], &extraPixels[1], &extraPixels[2], &extraPixels[3]);
 
-        img_next->width = CVPixelBufferGetWidth(pixelBuffer) + extraPixels[0] + extraPixels[1];
-        img_next->height = CVPixelBufferGetHeight(pixelBuffer) + extraPixels[2] + extraPixels[3];
-        img_next->row_pitch = CVPixelBufferGetBytesPerRow(pixelBuffer);
-        img_next->pixel_pitch = img_next->row_pitch / img_next->width;
+        img_out->width = CVPixelBufferGetWidth(pixelBuffer) + extraPixels[0] + extraPixels[1];
+        img_out->height = CVPixelBufferGetHeight(pixelBuffer) + extraPixels[2] + extraPixels[3];
+        img_out->row_pitch = CVPixelBufferGetBytesPerRow(pixelBuffer);
+        img_out->pixel_pitch = img_out->row_pitch / img_out->width;
 
-        img_next = snapshot_cb(img_next, true);
+        if (!push_captured_image_cb(std::move(img_out), false)) {
+          // got interrupt signal
+          // returning false here stops capture backend
+          return false;
+        }
 
-        return img_next != nullptr;
+        return true;
       }];
 
       // FIXME: We should time out if an image isn't returned for a while
@@ -132,6 +140,7 @@ namespace platf {
         img->row_pitch = CVPixelBufferGetBytesPerRow(pixelBuffer);
         img->pixel_pitch = img->row_pitch / img->width;
 
+        // returning false here stops capture backend
         return false;
       }];
 
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -311,6 +311,16 @@ namespace platf::dxgi {
   public:
     int
     convert(platf::img_t &img_base) override {
+      // Garbage collect mapped capture images whose weak references have expired
+      for (auto it = img_ctx_map.begin(); it != img_ctx_map.end();) {
+        if (it->second.img_weak.expired()) {
+          it = img_ctx_map.erase(it);
+        }
+        else {
+          it++;
+        }
+      }
+
       auto &img = (img_d3d_t &) img_base;
       auto &img_ctx = img_ctx_map[img.id];
 
@@ -341,6 +351,9 @@ namespace platf::dxgi {
 
       // Release encoder mutex to allow capture code to reuse this image
       img_ctx.encoder_mutex->ReleaseSync(0);
+
+      ID3D11ShaderResourceView *emptyShaderResourceView = nullptr;
+      device_ctx->PSSetShaderResources(0, 1, &emptyShaderResourceView);
 
       return 0;
     }
@@ -650,12 +663,15 @@ namespace platf::dxgi {
       shader_res_t encoder_input_res;
       keyed_mutex_t encoder_mutex;
 
+      std::weak_ptr<const platf::img_t> img_weak;
+
       void
       reset() {
         capture_texture_p = nullptr;
         encoder_texture.reset();
         encoder_input_res.reset();
         encoder_mutex.reset();
+        img_weak.reset();
       }
     };
 
@@ -699,6 +715,9 @@ namespace platf::dxgi {
       }
 
       img_ctx.capture_texture_p = img.capture_texture.get();
+
+      img_ctx.img_weak = img.weak_from_this();
+
       return 0;
     }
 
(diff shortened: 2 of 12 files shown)


KuleRucket:
round_robin.h is still used by kmsgrab.cpp and wayland.cpp

ns6089:
Brought back `round_robin.h` file and rebased against latest `nightly`.

ns6089:
Now prioritizes last image for reuse and tries to un-allocate unused images beyond normal target.
Should be more or less done.

But I can't throttle encoder enough for capture to use more than 2-3 images, Either I made wrong assumptions, or threading code needs some corrections.

Will investigate further.

src/video.cpp
```diff
@@ -1343,6 +1352,12 @@ namespace video {
             BOOST_LOG(error) << "Could not convert image"sv;
             return;
           }
+          else {
+            if (dummy_img) {
+              // free dummy image as we don't need it anymore
+              dummy_img.reset();
```
cgutman, line 1358: We actually don't even need to wait until here to free it. We can free it right after `convert()`.

I did that in #1136

ns6089: Makes sense. One thing though, I still haven't verified that image destruction during capture is implemented correctly across all platforms. And on Linux three capture backends use `snapshot()` in `dummy_img()`. So I can cherry-pick your commit into this pull request and test it together, or you merge and I rebase. Your call, either is fine with me.

src/platform/windows/display_vram.cpp
```diff
@@ -346,6 +356,12 @@ namespace platf::dxgi {
       // Release encoder mutex to allow capture code to reuse this image
       img_ctx.encoder_mutex->ReleaseSync(0);
 
+      ID3D11RenderTargetView *emptyRenderTarget = nullptr;
+      device_ctx->OMSetRenderTargets(1, &emptyRenderTarget, nullptr);
+      device_ctx->RSSetViewports(0, nullptr);
```
cgutman, line 361: I don't think we need to reset the render targets or the viewport here. Those are associated with this `hwdevice_t` rather than the provided image.

src/video.cpp
```diff
@@ -791,24 +791,46 @@ namespace video {
...
+          }
+          break;
+        }
       }
-    }
+      if (img_out) {
+        // unallocate unused above normal buffer size
+        size_t index = 0;
+        for (auto &img : imgs) {
+          if (index >= capture_buffer_normal_size && img && img.use_count() == 1) img.reset();
```
cgutman, line 820: I think we should have a grace period before freeing the excess buffers to avoid thrashing alloc/free. I wrote logic like that in #1128 that you could use as a reference.

src/video.cpp
```diff
@@ -791,24 +791,46 @@ namespace video {
...
+          if (it != imgs.begin()) {
+            // move freshly allocated or unused img to the front of the list to prioritize its reusal
+            imgs.erase(it);
+            imgs.push_front(img_out);
+          }
+          break;
+        }
       }
-    }
+      if (img_out) {
```
cgutman, line 816: If we don't manage to find a free image, we should sleep and try again until we do get one. There can be transient conditions that might hang the encoder temporarily, but we should just wait for one to free up (like the current code does) rather than returning an error.

Making this function always succeed will also allow you to eliminate the error handling around this new callback which will simplify the code.

src/video.cpp
cgutman: I took a look through them and it looks like it should be safe. Capture backends that support `PARALLEL_ENCODING` need to have images that are multi-threading safe (as long as they aren't actually being captured and encoded at the same time, which the parallel encoding code prevents with the `use_count()` checks).

Capture might be happening concurrently, but we know that the particular image we're about to free is not going to be involved in capture or encoding.

src/video.cpp
ns6089: Yes, it would make sense. But I think it's better not to shove too much logic into single big pull request that already contains potentially breaking changes. Multiple thresholds with different timeouts, simple to implement because the code is localized in one space. Just not everything at the same time, I'm struggling with macOS code portion at the moment, for example. And that is blocking the rest of no matter how perfect logic we write.

src/platform/windows/display_vram.cpp
ns6089: I can remove it, if you think it can have negative effects on performance. Just having contexts clean of dangling references can save us debugging headache down the line, when in a year or so we decide to refactor or rewrite something.

src/video.cpp
ns6089: I can do that. With interrupt signal or variable captured by the callback. Giving control over sleeps to callers has its own pros, but going with as little changes unrelated to vram usage is the right call.

ns6089: If I find the signal :(
Variable won't cut it because everything is likely to happen in single capture thread.

ns6089: I guess I can check for `capture_ctx_queue->running()` https://github.com/ns6089/Sunshine/blob/remove_round_robin/src/video.cpp#L830 and hope it's enough to prevent potential deadlocks.

ns6089: Like this https://github.com/LizardByte/Sunshine/pull/1098/commits/4e5dfc3de217087891f80f8aef4f339c44244864

src/video.cpp
ns6089: Another example, we can save another frame of memory and do intelligent copy evasion for situations when there's no mouse cursor to blend. But that logic would be rather complex, so I decided to exclude it from this pull request.

src/video.cpp
ns6089: Cherry-picked into this pull request.

ns6089:
Rebased against latest nightly. With info prints into log every 10 seconds about buffer status (that commit should be removed before merge). I think it should be ready for testing, once I resolve all open github conversations with @cgutman.

src/platform/windows/display_vram.cpp
ns6089: Double checked, and yes it does make sense, Pushed a fix.

src/video.cpp
ns6089: Adding grace period should be as simply as this https://github.com/ns6089/Sunshine/commit/0c95cc93e1b34146275892c12f294c84c1e64042 (untested, but logic is simple enough)
But I'd still rather not include it with this particular pull request, unless you insist.

cgutman: That's fine I guess. In your testing are you seeing it stick pretty consistently at 1 image or do you see it go up and down?

ns6089: Goes up to 2 at the very beginning (may indicate error in sleep scheduling there by the way), then stays solidly at 1. But that's nvidia, I'm half expecting amd having problems there as well. That's partially why I've added log messages with extensive statistics, need to gather baseline for tuning auto-resize. But I don't think it should hurt having very slight most likely invisible overhead for a couple of days in nightly branch.

ns6089:
Improved info statistics to not only include min and max values, but also how often each individual number was hit.

ns6089:
Rebased against latest nightly.

src/video.cpp
ns6089: Either way, I've raised normal buffer size to 2 frames https://github.com/LizardByte/Sunshine/pull/1098/commits/7fa2c4f385152fd1daae03766a291bc568c5bc2d. Won't get downsized below this target. Nothing should block the merge now other than testing on Linux and macOS platforms. Temporary solution anyway, dynamic resize will be done here https://github.com/LizardByte/Sunshine/pull/1149.

ns6089:
Is it possible to trigger CI in such a way that it will have public downloadable artifacts? So we can distribute them to platform owners for testing.

ReenigneArcher:
They're public, but user needs to have a GitHub account and be logged in.

ns6089:
> They're public, but user needs to have a GitHub account and be logged in.

I guess most CI tasks need manual approval from maintainer? I'd like to start bothering people in respective discord channels with the builds.

src/video.cpp
Nonary: A safe choice, but in reality if it can't maintain one image and has to use 2, then we'd be safer going with the other pull requests that do the same thing with less changes. I am really hoping it can be brought down to 1, but if 2 is the minimum, it's still better than now I guess.

Nonary:
> > They're public, but user needs to have a GitHub account and be logged in.
> 
> I guess most CI tasks need manual approval from maintainer? I'd like to start bothering people in respective discord channels with the builds.

In the actions, there is a CI workflow. If all of the builds pass, it will generate artifacts you can use to share out to users.

Nonary:
I am not sure why they are requiring approval to build it, it has not done this before for my other pull requests

src/video.cpp
cgutman: I had thought the same thing initially, but I think this one has one very nice advantage over my simpler approach.

Because the new image is only retreived at the moment we're ready to capture into it, it means we can achieve an image pool of a single entry. Because the current code (and my #1128), latch the next image at the moment the previous one is submitted for encoding, it is limited to a 2 image minimum. There's basically no way we could manage to have just one image buffer without blocking the capture thread, since we need to grab the next one immediately after submitting the previous one.

cgutman:
> I am not sure why they are requiring approval to build it, it has not done this before for my other pull requests

Probably because I rebased the previous ones to kick the build off (and did so for this one too to drop your cherry-pick of my commit, which I just merged).

ns6089:
And Linux part of the pull request is broken. X11->nvenc path can't detect nvenc (nightly works fine). On the plus side I managed to reproduce it.

ns6089:
> And Linux part of the pull request is broken. X11->nvenc path can't detect nvenc (nightly works fine). On the plus side I managed to reproduce it.

Inheriting base `img_t` from `std::enable_shared_from_this<img_t>` breaks some low level memory serialization magic, because of course it is. I'll see what can be done. It wasn't necessary in the first place, just made `dummy_img()` call `snapshot()` on Linux/macOS less ugly.

ns6089:
And `NvFBC` refuses to work on principle, either stores raw pointers or implicitly expects subsequent `img_t` to be different, or both. At least wayland works in all its slow glory. Fun stuff, gonna be here for a while.

ns6089:
Figured out what was wrong with `NvFBC`. If you try to initialize capture on a new thread without calling `alloc_img()` from it at least once, capture initialization fails. I think texture creation there implicitly initializes cuda thread context. Will try to initialize the context explicitly, or just call `alloc_img()`.

ns6089:
I believe I went through all relevant Linux paths except for `wlroots` capture (my hardware is too proprietary to run `sway`). If someone confirms it and macOS capture, this should be safe to merge.

ReenigneArcher:
> I am not sure why they are requiring approval to build it, it has not done this before for my other pull requests

You're a member of the org, so do not need approval approval.

src/platform/macos/display.mm
```diff
@@ -37,38 +37,46 @@
...
-        auto av_img_next = std::static_pointer_cast<av_img_t>(img_next);
-
         CFRetain(sampleBuffer);
 
         CVPixelBufferRef pixelBuffer = CMSampleBufferGetImageBuffer(sampleBuffer);
         CVPixelBufferLockBaseAddress(pixelBuffer, kCVPixelBufferLock_ReadOnly);
 
-        if (av_img_next->pixel_buffer != nullptr)
-          CVPixelBufferUnlockBaseAddress(av_img_next->pixel_buffer, 0);
+        __block std::shared_ptr<img_t> img_out;
```
cgutman, line 47: `__block` isn't needed here. It's used if you're declaring something outside a block that you want to be able to modify from within the block (otherwise it would normally just be copied and changes wouldn't be reflected in the original).

ns6089: Fixed.

ns6089:
> Hello ns6089, what software do you use to see the gpu usage like on your screen? Thanks!

Process Hacker. Got rebranded into System Informer recently, but didn't try the new version. https://github.com/winsiderss/systeminformer

Fewnity:
> 

Nice thank you!

Issue #1117, opened 2023-03-31
Elevated Do/UNDO Broken in 0.19 (and nightly)


Nonary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug
A security fix was applied to prevent elevation of privileges when running Sunshine as a service and the elevation tool does not appear to work properly when using those commands. 

### Expected Behavior
Elevated commands should prompt for UAC, if the user does not have UAC enabled, they will not function.

### Host Operating System

Windows

### Operating System Version

11

### Architecture

64 bit

### Sunshine commit or version

0.19

### Package

Linux - AppImage

### GPU Type

Nvidia

### GPU Model

N/A

### GPU Driver/Mesa Version

N/A

### Capture Method (Linux Only)

N/A

### Config

```Shell
N/A
```

### Apps

```json
N/A
```

### Relevant log output

```shell
N/A
```

Issue #978, opened 2023-02-26
Remote Access Exploit - Credential Resets


Nonary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

You are able to reset the Sunshine credentials without elevated rights, which could allow systems that have been already compromised with user level privileges to getting full remote access and capability to pair new clients with the default settings of Sunshine.

This is mostly caused by the fact that the config directory is giving users modify and read access, which prevents the need for having administrator privileges.

### Expected Behavior

When resetting credentials, if the current user is not running Sunshine in an elevated context, it should reject the request telling them to run it again with administrator level privileges.

### Additional Context

Workaround is to go in advanced settings and configure a different location for the credentials, which you can then add deny permissions to users. 

It is considered a low security risk as it already requires the users machine to be compromised in some way. However, if users are doing something stupid, like sharing the sunshine directory via SMB to everyone on the local network you could exploit resetting the credentials by editing the app.json file to execute the password reset commands when user starts a stream.

### Host Operating System

Docker

### Operating System Version

Windows 11

### Architecture

32 bit

### Sunshine commit or version

0.18.0

### Package

Windows - installer

### GPU Type

Nvidia

### GPU Model

N/A

### GPU Driver/Mesa Version

N/A

### Config

```Shell
N/A
```

### Relevant log output

```shell
N/A
```

ReenigneArcher:
While this isn't completely invalid, I doubt any change to Sunshine is going to come about due to this.

If the system is already compromised there is nothing stopping them from just running the program as admin (or other hacks that I won't go into detail about here).

Regarding sharing the Sunshine config/installation directory on a network share... just don't do that.

Nonary:
Re-opening this as it is a valid exploit, being able to remotely access someone's computer and run commands elevated without their knowledge (assuming they are not at home at the time) is still a risk. I'll do some investigation on my own and submit a PR.

I'm basically going to go through the code base and fix as many security exploits I can find....

PR #1123, opened 2023-04-02
Elevated Commands Redesign

Nonary:
### Fixes
- #978 
- #1117 

## Description
This pull request aims to rework the way commands are executed, specifically on the Windows platform. The main goal is to provide a safe method for users to elevate permissions without being prompted for UAC permissions. By doing this, we also prevent potential privilege exploits that are currently possible in Sunshine due to its existing design.

The core issue with Sunshine is that it allows adding commands without administrator rights, enabling anyone with user-level permissions to reset credentials and run elevated commands without additional safety checks.

Although there is a proof of concept demonstrating that secure administrative rights can be achieved without refactoring a significant portion of the code base, it does not address the fundamental problem. Additionally, it does not provide an option for users to safely disable elevation prompts.

To achieve this goal, we will need to implement several high-level changes, which I will outline below.

## Sunshine as a Service is Mandatory
If using the installer, installing Sunshine as a service will no longer be optional. The primary reason for this change is to allow us to securely configure the apps.json file and require elevated permissions to modify it. Please note that this change will **not** affect portable installations of Sunshine.

## All Commands are Un-elevated
As Sunshine runs under the SYSTEM context, it would be unwise to let commands run under this account, as it can lead to issues with certain scripts that need access to the UI on your PC. For instance, if you're trying to create a pre-command and need to swap monitors before the stream starts, this would not work until we execute those commands under the user's profile.

## Elevation as an Option
After changing the permissions of the app.json file, we will need administrative access to make modifications to it. This means we can "safely" allow users to mark a command as "elevated." Once that is done, Sunshine will run any command marked with this flag under admin, launching as the current user's profile **without** any UAC prompts. If the current user's profile does not have admin privileges, then it would execute under the System context instead.

## Type of Change

## Checklist

## Branch Updates


cgutman:
> If the current user's profile does not have admin privileges, then it would execute under the System context instead.

In the "elevation requested" case, I think we should always use the token given by `WTSQueryUserToken()`. That will be the unrestricted token providing the full privileges of the current user. If that user is not an admin, I don't think we should promote it to admin.

> We are considering changing the way WebAPI operates, specifically to prevent it from running everything under the System account, which is a vector for exploits.
>
> However, I would like to gather feedback on how we should handle editing the apps.json file in this scenario. One option is to add a UAC prompt, but this would not be suitable if users are not physically present at their machines.

Any changes we do to potentially move the web service into a separate lower-privileged process should be invisible to users. That process would still communicate with the Sunshine service via some IPC interface that would allow changes to be made to settings without needing elevation.

ReenigneArcher (commented):
> We are considering changing the way WebAPI operates, specifically to prevent it from running everything under the System account, which is a vector for exploits.

What exploits? Or this is just a theoretical statement?

src/platform/windows/misc.cpp
```diff
@@ -900,4 +900,9 @@ namespace platf {
     return std::make_unique<qos_t>(flow_id);
   }
 
+  bp::child
+  run_priveleged(const std::string &cmd, boost::filesystem::path &working_dir, boost::process::environment &env, FILE *file, std::error_code &ec, boost::process::group *group) {
+    BOOST_LOG(warning) << "You are entering dangerous territory here";
```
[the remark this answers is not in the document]
Nonary: This is all placeholder code which was required in order to submit a pull request in the first place. Anything presented here is it final and not necessary to review.

This is a draft pull request and more changes will be coming in the coming days unless the feedback that comes in suggests we shouldn't do this

Nonary:
> > We are considering changing the way WebAPI operates, specifically to prevent it from running everything under the System account, which is a vector for exploits.
> 
> 
> 
> What exploits? Or this is just a theoretical statement?

Not theory, any input received by the web server is vulnerable to exploits if not validated. One vulnerably was fixed in previous pull requests and I'm sure I'll find others once I do a detailed security review over it when I get to that.

Ideally we wouldn't want a web service running as system but I don't have any other solutions that are effective enough to not be any difference to the user experience basically.

Nonary:
Okay, it's ready for feedback/review

src/platform/windows/misc.cpp
```diff
@@ -200,6 +202,71 @@ namespace platf {
...
+  HANDLE
+  duplicate_users_token_elevated() {
+    DWORD consoleSessionId;
+    HANDLE userToken, duplicateToken;
+    TOKEN_ELEVATION_TYPE elevationType;
+    DWORD dwSize;
+
+    // Close the userToken handle when it goes out of scope
+    auto token_close = util::fail_guard([&userToken]() {
+      CloseHandle(userToken);
```
cgutman, line 219: This won't be initialized prior to a successful call to `WTSQueryUserToken()`, so you'll need to move this below that call.

src/platform/windows/misc.cpp
```diff
@@ -454,26 +615,74 @@ namespace platf {
...
+ * @param working_dir The working directory for the new process
+ * @param env The environment variables to use for the new process
+ * @param file A file object to redirect the child process's output to (may be nullptr)
+ * @param ec An error code, set to indicate any errors that occur during the launch process
+ * @param group A pointer to a `bp::group` object to which the new process should belong (may be nullptr)
+ *
+ * @return A `bp::child` object representing the new process, or an empty `bp::child` object if the launch fails
+ */
+  bp::child
+  run_priviliged(const std::string &cmd, boost::filesystem::path &working_dir, boost::process::environment &env, FILE *file, std::error_code &ec, boost::process::group *group) {
```
cgutman, line 638: privileged

CMakeLists.txt
```diff
@@ -822,14 +822,19 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
...
+    set(CPACK_COMPONENT_SUNSHINESVC_REQUIRED true)
 
     # service scripts
     set(CPACK_COMPONENT_SERVICE_DISPLAY_NAME "service-scripts")
-    set(CPACK_COMPONENT_SERVICE_DESCRIPTION "Scripts to enable/disable the service.")
+    set(CPACK_COMPONENT_SERVICE_DESCRIPTION "Scripts for installing and enabling the service.")
     set(CPACK_COMPONENT_SERVICE_GROUP "scripts")
     set(CPACK_COMPONENT_SERVICE_DEPENDS sunshinesvc)
+    # This is required because we've changed the ACL for config to require admin to protect against EoP exploits.
+    set(CPACK_COMPONENT_SERVICE_REQUIRED true)
```
cgutman, line 837: We need to also get rid of the shortcut to Sunshine.exe, since that will launch without elevation.

I think we discussed on Discord turning that shortcut into a PS script that reads the config and launches the browser pointing to the correct admin port.

src/platform/windows/misc.cpp
```diff
@@ -334,6 +401,33 @@ namespace platf {
     return true;
   }
 
+  /**
+
+@brief Check if the current process is running with system-level privileges.
+@return true if the current process has system-level privileges, false otherwise.
+*/
+  bool
+  is_running_as_system() {
```
cgutman, line 410: Missing error checks in this function

src/platform/windows/misc.cpp
```diff
@@ -334,6 +401,33 @@ namespace platf {
...
+    // Retrieve the process token handle
+    HANDLE hToken;
+    OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &hToken);
+
+    PRIVILEGE_SET ps;
+    ps.PrivilegeCount = 1;
+    ps.Control = 0;  // check for any privilege
+    LookupPrivilegeValueW(NULL, L"SeDelegateSessionUserImpersonatePrivilege", &ps.Privilege[0].Luid);
```
cgutman, line 418: I'd prefer if we used an approach like `CreateWellKnownSid(WinLocalSystemSid)` and `CheckTokenMembership()`.

src/platform/windows/misc.cpp
```diff
@@ -200,6 +202,71 @@ namespace platf {
...
+
+    // Get the user token for the active console session
+    if (!WTSQueryUserToken(consoleSessionId, &userToken)) {
+      BOOST_LOG(debug) << "QueryUserToken failed, this would prevent commands from launching under the users profile.";
+      return nullptr;
+    }
+
+    // We need to know if this is an elevated token or not.
+    // Get the elevation type of the user token
+    GetTokenInformation(userToken, TokenElevationType, &elevationType, sizeof(TOKEN_ELEVATION_TYPE), &dwSize);
```
cgutman, line 238: Missing error checking

src/platform/windows/misc.cpp
```diff
@@ -200,6 +202,71 @@ namespace platf {
...
+      // Retrieve the administrator token that is linked to the limited token
+      if (!GetTokenInformation(userToken, TokenLinkedToken, reinterpret_cast<void *>(&linked_token), sizeof(TOKEN_LINKED_TOKEN), &dwSize)) {
+        DWORD lasterror = GetLastError();
+        // If the retrieval failed, log an error message and return null
+        BOOST_LOG(error) << "Request to elevate the users token had failed. Error: " << lasterror;
+        return nullptr;
+      }
+
+      // Since we need the elevated token, we'll replace it with their administrative token.
+      userToken = linked_token.LinkedToken;
```
cgutman, line 252: We need to close the original `userToken` handle before overwriting it

src/platform/windows/misc.cpp
```diff
@@ -200,6 +202,71 @@ namespace platf {
     return std::string(buffer, bytes);
   }
 
+  /**
+ * @brief A function to duplicate the current sessions user's token with elevated privileges
+ *
+ * @return A handle to the duplicated users token, or null if the duplication failed
+ */
+  HANDLE
+  duplicate_users_token_elevated() {
```
cgutman, line 211: You can change this to accept a parameter for whether it wants the elevated token or not, then you can get rid of `duplicate_shell_token()` and use this function for both cases.

src/platform/windows/misc.cpp
```diff
@@ -484,44 +693,84 @@ namespace platf {
...
+ * @param working_dir The working directory for the new process
+ * @param env The environment variables to use for the new process
+ * @param file A file object to redirect the child process's output to (may be nullptr)
+ * @param ec An error code, set to indicate any errors that occur during the launch process
+ * @param group A pointer to a `bp::group` object to which the new process should belong (may be nullptr)
+ *
+ * @return A `bp::child` object representing the new process, or an empty `bp::child` object if the launch fails
+ */
+  bp::child
+  run_unprivileged(const std::string &cmd, boost::filesystem::path &working_dir, bp::environment &env, FILE *file, std::error_code &ec, bp::group *group) {
```
cgutman, line 717: It looks like `run_unprivileged()` and `run_privileged()` are pretty similar. Rather than separating out the parts like `create_startup_info()` and `create_boost_child_from_results()`, I think you can consolidate them into a single run function that just takes a parameter of whether to elevate or not like you did with `run_command()`.

cgutman (changes requested):

CMakeLists.txt
```diff
@@ -729,15 +731,13 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
...
     # Migrates config files from the root into the new config folder
-    # Sets permissions on the config folder so that we can write in it
     # Install service
     SET(CPACK_NSIS_EXTRA_INSTALL_COMMANDS
             "${CPACK_NSIS_EXTRA_INSTALL_COMMANDS}
         IfSilent +2 0
         ExecShell 'open' 'https://sunshinestream.readthedocs.io/'
         nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\" /reset'
         nsExec::ExecToLog '\\\"$INSTDIR\\\\scripts\\\\migrate-config.bat\\\"'
-        nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\\config\\\" /grant:r Users:\\\(OI\\\)\\\(CI\\\)\\\(F\\\)'
```
cgutman, line 740: We should also deny reads to `config\credentials` for non-admin users

CMakeLists.txt
```diff
@@ -729,15 +731,14 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
...
     # Restores permissions on the install directory
     # Migrates config files from the root into the new config folder
-    # Sets permissions on the config folder so that we can write in it
     # Install service
     SET(CPACK_NSIS_EXTRA_INSTALL_COMMANDS
             "${CPACK_NSIS_EXTRA_INSTALL_COMMANDS}
         IfSilent +2 0
         ExecShell 'open' 'https://sunshinestream.readthedocs.io/'
         nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\" /reset'
+        nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\\config\\\\credentials\\\" /deny Users:\\\(R\\\)'
```
cgutman, line 740: It doesn't look like this does what we want because of inherited ACLs.

In my testing, this did the trick:
```
nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\\config\\\\credentials\\\" /inheritance:r'
nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\\config\\\\credentials\\\" /grant:r Administrators:\\\(OI\\\)\\\(CI\\\)\\\(F\\\)'
```

The first `icacls` invocation will disable inheritance, which strips all ACLs off and makes it accessible by nobody. The second one adds back access for the Administrators group.

src/platform/windows/misc.cpp
```diff
@@ -200,43 +202,78 @@ namespace platf {
...
-    DWORD shell_pid;
-    GetWindowThreadProcessId(shell_window, &shell_pid);
-    HANDLE shell_process = OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, shell_pid);
-    if (!shell_process) {
-      BOOST_LOG(error) << "Failed to open shell process: "sv << GetLastError();
-      return NULL;
+    // Get the user token for the active console session
+    if (!WTSQueryUserToken(consoleSessionId, &userToken)) {
+      BOOST_LOG(debug) << "QueryUserToken failed, this would prevent commands from launching under the users profile.";
+      CloseHandle(userToken);
```
cgutman, line 228: `userToken` won't be initialized on failure, so no need to close it.

CMakeLists.txt
Nonary: That's odd, not really sure why it worked on my machine then. When I tested it, it correctly denied permissions

src/platform/windows/misc.cpp
```diff
@@ -200,43 +202,78 @@ namespace platf {
...
+    // User has a limited token, this likely means they have UAC enabled.
+    if (elevated && elevationType == TokenElevationTypeLimited) {
+      TOKEN_LINKED_TOKEN linkedToken;
+      // Retrieve the administrator token that is linked to the limited token
+      if (!GetTokenInformation(userToken, TokenLinkedToken, reinterpret_cast<void *>(&linkedToken), sizeof(TOKEN_LINKED_TOKEN), &dwSize)) {
+        // If the retrieval failed, log an error message and return null
+        BOOST_LOG(debug) << "Retrieving linked token information failed: " << GetLastError();
+
+        //Maybe the user is not an admin here, so let's just return a normal token.
+        CloseHandle(linkedToken.LinkedToken);
```
cgutman, line 254: Same here, we shouldn't close on failure.

The general contract of most Windows APIs is that any output parameters are in an undefined state on failure, so we shouldn't be touching them. Internally, if the token was somehow received prior to another error, it will take care of closing the handle for you.

src/platform/windows/misc.cpp
```diff
@@ -200,43 +202,78 @@ namespace platf {
...
+        return userToken;
+      }
+
+      // Since we need the elevated token, we'll replace it with their administrative token.
+      CloseHandle(userToken);
+      userToken = linkedToken.LinkedToken;
+      CloseHandle(linkedToken.LinkedToken);
```
cgutman, line 261: We still need that `linkedToken.LinkedToken` handle, so we can't close it yet. (`HANDLE` is not a special RAII class that transfers ownership or anything. It's basically just like a raw pointer.)

The `token_close` object above will handle closing the linked token when we return from this function.

src/platform/windows/misc.cpp
```diff
@@ -454,74 +594,102 @@ namespace platf {
...
+      if (!merge_user_environment_block(env, user_token)) {
+        ec = std::make_error_code(std::errc::not_enough_memory);
+        return bp::child();
+      }
+
+      // Create the STARTUPINFOEXW object for the new process
+
+      if (elevated && is_running_as_system()) {
+        // Impersonate the current user and launch the new process
+        ec = impersonate_current_user(user_token, [&]() {
```
cgutman, line 657: We only need this single codepath, not the `!is_token_same_user_as_process(user_token)` one below. You won't need to check `elevated` because `retrieve_user_token()` already grabbed the correct token for you, and you don't have to check `is_running_as_system()` since you're already inside an `if (is_running_as_system())` codepath.

src/platform/windows/misc.cpp
```diff
@@ -395,46 +468,113 @@ namespace platf {
...
-    std::wstring start_dir = utf8_to_wide_string(working_dir.string());
+  /**
+ * @brief Impersonate the current user, invoke the callback function, then returns back to system context.
+ *
+ * @param user_token A handle to the user's token that was obtained from the shell
+ * @param callback A function that will be executed while impersonating the user
+ * @return An std::error_code object that will store any error that occurred during the impersonation
+ */
+  std::error_code
+  impersonate_current_user(HANDLE user_token, std::function<void()> callback) {
```
cgutman, line 526: Since there's only one codepath where you need to impersonate now, you can put this back in `run_command()`.

src_assets/common/assets/web/apps.html
```diff
@@ -290,6 +375,9 @@ <h4 class="modal-title">Covers Found</h4>
           this.$set(this.editForm, "detached", []);
         if (this.editForm["exclude-global-prep-cmd"] === undefined)
           this.$set(this.editForm, "exclude-global-prep-cmd", false);
+        if(this.editForm["elevated"] === undefined && this.platform == 'windows'){
```
cgutman, line 378: I think you want `===` for all these type coercion warnings

src/platform/windows/misc.cpp
```diff
@@ -200,43 +202,78 @@ namespace platf {
...
+    // Close the userToken handle when it goes out of scope
+    auto token_close = util::fail_guard([&userToken]() {
+      CloseHandle(userToken);
+    });
+
+    // We need to know if this is an elevated token or not.
+    // Get the elevation type of the user token
+    if (!GetTokenInformation(userToken, TokenElevationType, &elevationType, sizeof(TOKEN_ELEVATION_TYPE), &dwSize)) {
+      BOOST_LOG(debug) << "Retrieving token information failed: " << GetLastError();
+      CloseHandle(userToken);
```
cgutman, line 241: The fail guard is already going to close this for us

CMakeLists.txt
cgutman: Yeah, it just leaves a bunch of stuff around that makes me uneasy like `APPLICATION PACKAGE AUTHORITY\ALL APPLICATION PACKAGES` and `APPLICATION PACKAGE AUTHORITY\ALL RESTRICTED APPLICATION PACKAGES`, so I think UWP apps may still have access to it.

src/platform/windows/misc.cpp
```diff
@@ -395,46 +416,113 @@ namespace platf {
...
+    std::error_code ec;
+    // Impersonate the user when launching the process. This will ensure that appropriate access
+    // checks are done against the user token, not our SYSTEM token. It will also allow network
+    // shares and mapped network drives to be used as launch targets, since those credentials
+    // are stored per-user.
+    if (!ImpersonateLoggedOnUser(user_token)) {
+      auto winerror = GetLastError();
+      // Log the failure of impersonating the user and its error code
+      BOOST_LOG(error) << "Failed to impersonate user: "sv << winerror;
+      ec = std::make_error_code(std::errc::permission_denied);
```
cgutman, line 484: We should fail in this case rather than invoking the callback without impersonation.

src/platform/windows/misc.cpp
```diff
@@ -454,74 +542,83 @@ namespace platf {
...
-      if (!ImpersonateLoggedOnUser(shell_token)) {
-        auto winerror = GetLastError();
-        BOOST_LOG(error) << "Failed to impersonate user: "sv << winerror;
-        ec = std::make_error_code(std::errc::permission_denied);
+    // Convert cmd, env, and working_dir to the appropriate character sets for Win32 APIs
+    std::wstring wcmd = utf8_to_wide_string(cmd);
+    std::wstring env_block = create_environment_block(env);
+    std::wstring start_dir = utf8_to_wide_string(working_dir.string());
+
+    STARTUPINFOEXW startup_info = create_startup_info(file, ec);
```
cgutman, line 573: Need to check `ec` for an error

src/platform/windows/misc.cpp
```diff
@@ -200,101 +202,84 @@ namespace platf {
...
 
-    auto user = (PTOKEN_USER) HeapAlloc(GetProcessHeap(), 0, return_length);
-    if (!user) {
-      return nullptr;
-    }
+    if (elevated && elevationType == TokenElevationTypeDefault) {
+      // The token does not have a linked token, and user requested it to be elevated.
+      // This indicates that the user is not a member of the Administrators group.
+      BOOST_LOG(warning) << "This command requires elevation and the current user account logged in does not have administrator rights. "
+                         << "For security reasons Sunshine will retain the same access level as the current user and will not elevate it.";
```
cgutman, line 247: I'm curious if you hit this codepath on an admin account when UAC is disabled.

src/platform/windows/misc.cpp
```diff
@@ -200,101 +202,84 @@ namespace platf {
...
 
-    if (!GetTokenInformation(token, TokenUser, user, return_length, &return_length)) {
-      auto winerr = GetLastError();
-      BOOST_LOG(error) << "Failed to get token information: "sv << winerr;
-      HeapFree(GetProcessHeap(), 0, user);
-      return nullptr;
+      return userToken;
```
cgutman, line 249: We can't return `userToken` here (or below in the failure case for GetTokenInformation) since our fail_guard action is going to close it. Can you refactor these 2 checks to fall-through on failure to call `DuplicateTokenEx()` will the current `userToken`?

src/platform/windows/misc.cpp
Nonary: That code path is only hit if user has UAC disabled and is not an admin. 

Admins with UAC disabled would have TokenElevationType of Full
Users with UAC enabled would have a LimitedToken,
Users with UAC disabled would have Default

Also tested on my machine and confirmed code path is not hit.

src/platform/windows/misc.cpp
cgutman: Actually, `WTSQueryUserToken()` returns a primary token, so I think we can drop the `DuplicateTokenEx()` call entirely and just directly return the handle we got from `WTSQueryUserToken()` or `GetTokenInformation()`. I would probably ditch the fail guard and just add `CloseHandle()` to the failure paths where it's required.

Nonary:
@cgutman Any other changes needed? Been about two weeks so was just curious

cgutman:
Almost everything looks good in my testing. I'm seeing a bug though where trying to quit a running app from Moonlight never completes. It works in nightly, so it looks like a regression specific to this PR.

Nonary:
> Almost everything looks good in my testing. I'm seeing a bug though where trying to quit a running app from Moonlight never completes. It works in nightly, so it looks like a regression specific to this PR.

I am not observing this behavior, but I did pull some changes from nightly so maybe it was fixed in nightly? IDK. I am able to quit apps without issues.

cgutman (approved):

PR #1175, opened 2023-04-12
Don't copy to intermediate surface when there is no mouse cursor

ns6089:
## Description
Optimize fullscreen game capture path with copy evasion and save one frame of vram. Currently in draft, it works well, but lacks documentation and instant `std::variant` should be replaced with delayed surface destruction.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -101,8 +101,6 @@ namespace platf::dxgi {
   blob_t scene_NW_ps_hlsl;
 
   struct img_d3d_t: public platf::img_t {
-    std::shared_ptr<platf::display_t> display;
-
     // These objects are owned by the display_t's ID3D11Device
     texture2d_t capture_texture;
     render_target_t capture_rt;
@@ -1321,7 +1319,6 @@ namespace platf::dxgi {
     // Initialize format-independent fields
     img->width = width;
     img->height = height;
-    img->display = shared_from_this();
     img->id = next_image_id++;
 
     return img;
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -128,6 +128,52 @@ namespace platf::dxgi {
     };
   };
 
+  struct texture_lock_helper {
+    keyed_mutex_t _mutex;
+    bool _locked = false;
+
+    texture_lock_helper(const texture_lock_helper &) = delete;
+    texture_lock_helper &
+    operator=(const texture_lock_helper &) = delete;
+
+    texture_lock_helper(texture_lock_helper &&other) {
+      _mutex.reset(other._mutex.release());
+      _locked = other._locked;
+      other._locked = false;
+    }
+
+    texture_lock_helper &
+    operator=(texture_lock_helper &&other) {
+      if (_locked) _mutex->ReleaseSync(0);
+      _mutex.reset(other._mutex.release());
+      _locked = other._locked;
+      other._locked = false;
+      return *this;
+    }
+
+    texture_lock_helper(IDXGIKeyedMutex *mutex):
+        _mutex(mutex) {
+      if (_mutex) _mutex->AddRef();
+    }
+
+    ~texture_lock_helper() {
+      if (_locked) _mutex->ReleaseSync(0);
+    }
+
+    bool
+    lock() {
+      if (_locked) return true;
+      HRESULT status = _mutex->AcquireSync(0, INFINITE);
+      if (status == S_OK) {
+        _locked = true;
+      }
+      else {
+        BOOST_LOG(error) << "Failed to acquire texture mutex [0x"sv << util::hex(status).to_string_view() << ']';
+      }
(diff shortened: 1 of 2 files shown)


src/platform/windows/display_vram.cpp
```diff
@@ -910,97 +920,228 @@ namespace platf::dxgi {
...
+          ID3D11ShaderResourceView *emptyShaderResourceView = nullptr;
+          device_ctx->PSSetShaderResources(0, 1, &emptyShaderResourceView);
+        }
+        break;
+      }
+
+      case ofa::dummy_fallback: {
+        if (!pull_free_image_cb(img_out)) return capture_e::interrupted;
+        auto img = (img_d3d_t *) img_out.get();
+        if (complete_img(img, true)) return capture_e::error;
```
cgutman, line 1141: If this image was used before, it may have a cursor rendered into it. That's why the code was using `UpdateSubresource()` to clear it each time.

src/platform/windows/display_vram.cpp
```diff
@@ -910,97 +920,228 @@ namespace platf::dxgi {
...
+        }
+        else {
+          if (std::holds_alternative<texture2d_t>(last_frame_variant)) {
+            last_frame_action = lfa::replace_surface_with_img;
+          }
+          out_frame_action = ofa::forward_last_img;
+        }
+      }
+      else {
+        last_frame_action = lfa::nothing;
```
cgutman, line 972: Is this path actually reachable? It seems like we should only hit the dummy fallback for `capture_format == DXGI_FORMAT_UNKNOWN`. For other cases, we should always have something to give the caller.

src/platform/windows/display_vram.cpp
```diff
@@ -910,97 +920,228 @@ namespace platf::dxgi {
...
-      if (status != S_OK) {
-        BOOST_LOG(error) << "Failed to acquire capture mutex [0x"sv << util::hex(status).to_string_view() << ']';
-        return capture_e::error;
+      case lfa::replace_img_with_surface: {
+        auto p_img = std::get_if<std::shared_ptr<platf::img_t>>(&last_frame_variant);
+        if (!p_img) {
+          BOOST_LOG(error) << "Logical error at " << __FILE__ << ":" << __LINE__;
+          return capture_e::error;
+        }
+        auto [d3d_img, lock] = get_locked_platform_img(*p_img);
```
cgutman, line 1037: Do we need to lock the image here? It doesn't look like we use `d3d_img` in this codepath.

As an aside, this case will always be followed by `copy_last_surface_and_blend_cursor` which calls `pull_free_image_cb()`. Is it worth including a code path that allows us to just use the `img_t` we have here instead of having to call `pull_free_image_cb()` again?

cgutman:
Cool stuff, this is a nice clever use of the new `pull_free_image_cb()` stuff.

src/platform/windows/display_vram.cpp
ns6089: > I'll see if I can do it without making the code significantly less readable.

One more path on `output_frame_action` switch and one temporary variable, close to no duplication. Should be alright  https://github.com/ns6089/Sunshine/commit/8ecd15794fa6181e47a33fec4598fc134f0734bd

ns6089: And now thinking about it, we can't draw to this image exactly because it may be still used by encoder(s), locks don't help because they can race. Original code was correct.

src/platform/windows/display_vram.cpp
cgutman: Hmm, I think I would prefer to render the cursor. @psyke83 might have more comments about this, but I believe we do commonly hit this case where a game is loading and hasn't yet rendered its first frame. It might be confusing to the user to just see a black screen with nothing on it, especially because the mouse will be visible on the host display.

src/platform/windows/display_vram.cpp
cgutman: > We don't write to it, but we read it (copy to newly allocated surface). And the lock may be held by encoder thread at this point. I don't think it's allowed by d3d, there's evidence they wanted to include read-only locks, but didn't do it in the end for some reason.

Ah yes, I didn't see that for some reason.

> And now thinking about it, we can't draw to this image exactly because it may be still used by encoder(s), locks don't help because they can race. Original code was correct. Theoretically we can check for refcount, but I'd rather not introduce not obvious strong coupling here, it screams potential problems.

Indeed, since we were handing out references to this frame, the encoder might still have one. I agree, let's just keep that logic the way you had it in the original code.

src/platform/windows/display_vram.cpp
ns6089: > especially because the mouse will be visible on the host display

Makes sense, will do it.

ns6089: Done.

ns6089:
Delayed destruction of the intermediate surface for 10 seconds, and added comments.
I believe it's ready.

PR #1192, opened 2023-04-17
Extend packet header with frame processing latency

ns6089:
## Description
Utilize 2 unused bytes of packet short header for transferring frame latency to moonlight.

Can't really granulate this latency (into processing and encoding for example) because of asynchronous nature of gpu commands, at least to my knowledge and while keeping the code clean. And I don't expect anyone particularly caring for it.

moonlight-common-c pull request: https://github.com/moonlight-stream/moonlight-common-c/pull/77
moonlight-qt pull request: https://github.com/moonlight-stream/moonlight-qt/pull/995

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -11,6 +11,7 @@ extern "C" {
 }
 
 #include "display.h"
+#include "misc.h"
 #include "src/main.h"
 #include "src/video.h"
 
@@ -894,6 +895,12 @@ namespace platf::dxgi {
       return capture_e::timeout;
     }
 
+    std::optional<std::chrono::steady_clock::time_point> frame_timestamp;
+    if (auto qpc_displayed = std::max(frame_info.LastPresentTime.QuadPart, frame_info.LastMouseUpdateTime.QuadPart)) {
+      // Translate QueryPerformanceCounter() value to steady_clock time point
+      frame_timestamp = std::chrono::steady_clock::now() - qpc_time_difference(qpc_counter(), qpc_displayed);
+    }
+
     if (frame_info.PointerShapeBufferSize > 0) {
       DXGI_OUTDUPL_POINTER_SHAPE_INFO shape_info {};
 
@@ -1237,6 +1244,10 @@ namespace platf::dxgi {
     // Perform delayed destruction of the unused surface if the time is due.
     if (old_surface_delayed_destruction && old_surface_timestamp + 10s < std::chrono::steady_clock::now()) {
       old_surface_delayed_destruction.reset();
+    }
+
+    if (img_out) {
+      img_out->frame_timestamp = frame_timestamp;
     }
 
     return capture_e::ok;
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -20,6 +20,7 @@ extern "C" {
 #include "input.h"
 #include "main.h"
 #include "network.h"
+#include "stat_trackers.h"
 #include "stream.h"
 #include "sync.h"
 #include "thread_safe.h"
@@ -1003,6 +1004,8 @@ namespace stream {
     // Video traffic is sent on this thread
     platf::adjust_thread_priority(platf::thread_priority_e::high);
 
+    stat_trackers::min_max_avg_tracker<uint16_t> frame_processing_latency_tracker;
+
     while (auto packet = packets->pop()) {
       if (shutdown_event->peek()) {
         break;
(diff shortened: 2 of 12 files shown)


src/stream.cpp
```diff
@@ -74,7 +76,11 @@ namespace stream {
     }
 
     std::uint8_t headerType;  // Always 0x01 for short headers
-    std::uint8_t unknown[2];
+
+    // Sunshine extension
+    // Frame latency, in 1/10 ms units
+    //     zero when the frame is repeated or there is no backend implementation
+    std::uint16_t frame_latency;
```
ns6089, line 83: Is sunshine little-endian only, or should I use the conversion?

ns6089: Used `boost::endian::little_uint16_t`

src/stream.cpp
```diff
@@ -997,6 +1004,47 @@ namespace stream {
...
+
+      void
+      print_info(const std::chrono::steady_clock::time_point &frame_timestamp, std::chrono::seconds interval_in_seconds) {
+        if (steady_clock::now() > data.last_print_time + interval_in_seconds) {
+          double latency_min_ms = data.latency_min / 10.;
+          double latency_max_ms = data.latency_max / 10.;
+          double latency_avg_ms = data.latency_total / 10. / data.calls;
+
+          auto fp = boost::format("%1$.1f");
+          BOOST_LOG(info) << "Latency (min/max/avg): " << fp % latency_min_ms << "ms/" << fp % latency_max_ms << "ms/" << fp % latency_avg_ms << "ms";
```
cgutman, line 1031: If we're going to log this at the info level, we probably shouldn't do it so often, like maybe once a minute max.

I would also call it "Host Latency" or similar, like you did in the Moonlight Qt PR, just to avoid any possible confusion with end-to-end latency, network latency, etc.

src/platform/windows/display_vram.cpp
```diff
@@ -1239,6 +1246,10 @@ namespace platf::dxgi {
       old_surface_delayed_destruction.reset();
     }
 
+    if (img_out) {
+      img_out->frame_timestamp = frame_timestamp;
```
cgutman, line 1250: Can you also plumb this into `display_ram_t`?

src/stream.cpp
ns6089: I planned to remove it before merge, needed to test it somehow because moonlight side is not merged. Probably should start marking such commits to avoid possible confusion.

src/platform/windows/display_vram.cpp
ns6089: Yes, if you're ok with the overall structure.

src/stream.cpp
ns6089: And on the topic of "host latency". Maybe we can come with something more fitting? If a frame gets delayed because of frame pacing sleep, that latency doesn't get included here. Maybe something like "host processing latency", or "host processing time". Parsec even calls "encoding".

ns6089: > I planned to remove it before merge, needed to test it somehow because moonlight side is not merged.

But maybe it makes sense to keep it after all, just move to Debug log level (it stays quiet during streaming), and maybe move the tracking class declaration. Will try to do it nicely.

src/platform/windows/display_vram.cpp
cgutman: Yeah, the approach looks good to me.

src/stream.cpp
cgutman: "Host Processing Latency" sounds fine to me (though technically, FEC and packetization is "processing" that isn't included in your calculation).

ns6089: > "Host Processing Latency" sounds fine to me

Alright, will rename it. We can extends packet header with "capture latency" later on. or hopefully get rid of that latency altogether in most configurations.

src/platform/windows/display_vram.cpp
ns6089: Done.

src/stream.cpp
ns6089: Renamed "frame latency" to "frame processing latency". Will update moonlight pull requests right after we merge the sunshine part.

Moved stat tracking code to its own header so it can be reused. Frame processing latency is printed at debug log level every 20 seconds.

ns6089: > FEC encoding and packetization is "processing" that isn't included

I thought it was fast enough to neglect, and can't really inject stats into packet after encode? Alternatively we can delay it by one frame, and use the stats of the previous frame in packet header. But it would be getting quite convoluted.

cgutman: It mostly is, though we may have delays due to buffer exhaustion in the network stack if the outgoing link is slow. I don't think it's worth the effort to try to time those at this point.

Users should still see the indirect effects of the slow connection via the network latency and jitter stat even though we aren't directly measuring them here.

PR #1197, opened 2023-04-19
Investigate and mitigate possible stream freezes

ns6089:
## Description
Latest changes to the img queue may have caused occasional stream freezes.
This attempts to investigate and mitigate them.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display_base.cpp
+++ b/src/platform/windows/display_base.cpp
@@ -510,7 +510,11 @@ namespace platf::dxgi {
         PD3DKMTSetProcessSchedulingPriorityClass fn =
           (PD3DKMTSetProcessSchedulingPriorityClass) GetProcAddress(gdi32, "D3DKMTSetProcessSchedulingPriorityClass");
         if (fn) {
-          status = fn(GetCurrentProcess(), D3DKMT_SCHEDULINGPRIORITYCLASS_REALTIME);
+          // Set scheduling priority to "high" for NVIDIA, or to "realtime" for other gpu vendors
+          // As of 2023.07, NVIDIA driver has unfixed bug(s) where "realtime" can cause unrecoverable encoding freeze or outright driver crash
+          // This issue happens more frequently with HAGS, in DX12 games or when VRAM is filled close to max capacity
+          // Track OBS to see if they find better workaround or NVIDIA fixes it on their end, they seem to be in communication
+          status = fn(GetCurrentProcess(), adapter_desc.VendorId == 0x10DE ? D3DKMT_SCHEDULINGPRIORITYCLASS_HIGH : D3DKMT_SCHEDULINGPRIORITYCLASS_REALTIME);
           if (FAILED(status)) {
             BOOST_LOG(warning) << "Failed to set realtime GPU priority. Please run application as administrator for optimal performance.";
           }


ns6089:
Presumably it's duplication api that starts returning timeouts for no apparent reason.
Not sure what we can do here, maybe give it back its surface before sleeping and see if this changes anything?
```
[2023:04:19:20:10:33]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
[2023:04:19:20:10:34]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
[2023:04:19:20:10:35]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
[2023:04:19:20:10:36]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
[2023:04:19:20:10:37]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
[2023:04:19:20:10:38]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
[2023:04:19:20:10:39]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
[2023:04:19:20:10:40]: Fatal: Hang detected! Session failed to terminate in 10 seconds.
[2023:04:19:20:10:40]: Info: AcquireNextFrame() returned DXGI_ERROR_WAIT_TIMEOUT
```

Nonary:
If it helps any, when a freeze happens I can actually start a new session on my phone (using the channels feature) and it unfreezes the main stream. So the implication is it is possible for it to recover, whatever the stream is doing when someone connects with a second stream anyway.

ns6089:
> If it helps any, when a freeze happens I can actually start a new session on my phone (using the channels feature) and it unfreezes the main stream. So the implication is it is possible for it to recover, whatever the stream is doing when someone connects with a second stream anyway.

Plot thickens. This certainly means something. Glanced through the code, but couldn't find anything relevant.

Maybe someone else has any ideas what could it possibly mean?

ns6089:
Apparently OBS tried to fix something similar by lowering gpu scheduling priority https://github.com/obsproject/obs-studio/pull/8923
Can someone hit the CI so we can distribute the builds for testing?

ns6089:
Can somebody hit the CI again please? The nightly version has bugs that causing problems for testers, so I've rebased it again stable.

ns6089:
Thanks.

ns6089:
Also I think it can be rebased and merged. Feedback has been positive and the change is now limited to nvidia gpus and documented.

cgutman:
I'm going to run some performance tests just to make sure we don't see a significant high vs realtime performance difference in GPU-limited scenarios like I did on AMD in my testing. If perf looks good, I'll rebase and merge.

cgutman:
The driver update hang is our bug. It's because updating the driver forces renumeration of all displays, which means we loop forever trying to open the old displays by name (which never works because the old displays are gone)

MemeTao:
I wonder why Nvidia still hasn‘t fixed this bug after two years.

Yundi339:
> The driver update hang is our bug. It's because updating the driver forces renumeration of all displays, which means we loop forever trying to open the old displays by name (which never works because the old displays are gone)

In 2025, sunshine implements a feature about asynchronous encoder. After that, I hardly ever heard of encoder freeze problems due to GPU memory shortage (windows&nvidia). Thanks to all developers for sharing.

MemeTao:
> > The driver update hang is our bug. It's because updating the driver forces renumeration of all displays, which means we loop forever trying to open the old displays by name (which never works because the old displays are gone)
> 
> In 2025, sunshine implements a feature about asynchronous encoder. After that, I hardly ever heard of encoder freeze problems due to GPU memory shortage (windows&nvidia). Thanks to all developers for sharing.

I notice that sunshine avoid to set realtime schedule level on nvidia gpu. May be the real reason whey sunshine hardly seen freeze problems is the following codes:

```
if (adapter_desc.VendorId == 0x10DE) {
  // As of 2023.07, NVIDIA driver has unfixed bug(s) where "realtime" can cause unrecoverable encoding freeze or outright driver crash
  // This issue happens more frequently with HAGS, in DX12 games or when VRAM is filled close to max capacity
  // Track OBS to see if they find better workaround or NVIDIA fixes it on their end, they seem to be in communication
  if (hags_enabled && !config::video.nv_realtime_hags) {
    priority = D3DKMT_SCHEDULINGPRIORITYCLASS_HIGH;
  }
}
```

ns6089:
- Putting NVENC into async return mode didn't prevent freezes back when I was adding that feature to sunshine, but something might have changed on the driver side since then
- Sunshine uses "realtime" priority by default, but there's gui option for switching to "high"
- [D3DKMTSetProcessSchedulingPriorityClass](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/d3dkmthk/nf-d3dkmthk-d3dkmtsetprocessschedulingpriorityclass) is a "driver" function that we're not really supposed to use to begin with, but without it set to "realtime" near-zero consistent latency is impossible. All other programs (like OBS) don't care if it's 5ms or 25ms.

Yundi339:
> * Putting NVENC into async return mode didn't prevent freezes back when I was adding that feature to sunshine, but something might have changed on the driver side since then

I really like this feature, thank you~

I can be sure that after adding the asynchronous encoder, many users upgraded Sunshine, and there was no upgrade driver, and there was no more screen freeze causing moonlight showed "-1" to exit. This feature solves this problem for many users.

If users experience this problem again, it is almost always caused by network packet loss, such as the network operator's UDP QoS speed limit.

Maybe there are still people who have problems in places I can't see, but in the past six months, there has been no similar feedback on 'windows&nvidia'.

MemeTao:
> * Putting NVENC into async return mode didn't prevent freezes back when I was adding that feature to sunshine, but something might have changed on the driver side since then
> 
>     * Sunshine uses "realtime" priority by default, but there's gui option for switching to "high"
> 
>     * [D3DKMTSetProcessSchedulingPriorityClass](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/d3dkmthk/nf-d3dkmthk-d3dkmtsetprocessschedulingpriorityclass) is a "driver" function that we're not really supposed to use to begin with, but without it set to "realtime" near-zero consistent latency is impossible. All other programs (like OBS) don't care if it's 5ms or 25ms.

I found without D3DKMTSetProcessSchedulingPriorityClass(real time) Sunshine has bad fps performence when the game run out all of GPU-3D resources (windows 11 &&  HAG on)

PR #1208, opened 2023-04-23
build(linux): honor install prefix for tray icon

istori1:
## Description
Use install prefix for the tray icon location.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -944,7 +944,7 @@ elseif(UNIX)
 
         if(${SUNSHINE_TRAY} STREQUAL 1)
             install(FILES "${CMAKE_SOURCE_DIR}/sunshine.svg"
-                    DESTINATION "/usr/share/icons")
+                    DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons")
 
             set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
                     ${CPACK_DEBIAN_PACKAGE_DEPENDS}, \


ReenigneArcher:
I don't think gtk will look in any specified location, that's why I forced it to `/usr`. Can you give me a use case where this needs to be somewhere besides `/usr`?

hfiguiere:
> I don't think gtk will look in any specified location, that's why I forced it to `/usr`. Can you give me a use case where this needs to be somewhere besides `/usr`?

This is not true. It should be installed in the prefix. This is the standard.
With immutable systems like Fedora Silverblue, you can NOT write in `/usr`.

ReenigneArcher:
Okay, but our flatpak prefix is `/app`. I have not verified but I'm doubting it will find icons in this location. Please correct me if it will find them there.

https://github.com/LizardByte/Sunshine/blob/d70d084f9fbb4e0150977a89d94937418a3ccf9c/packaging/linux/flatpak/dev.lizardbyte.sunshine.yml#L218

hfiguiere:
Flatpak has nothing to do here. Installing icons have to follow the prefix. That's what prefix is for.

hfiguiere:
BTW it is wrong to set pass ` -DCMAKE_INSTALL_PREFIX=/app ` in a flatpak manifest. This is why buildsystem `cmake` or `cmake-ninja` exist, as they manage this kind of things.

PR #1231, opened 2023-04-28
Add RFC 6598 CGN CIDR block to LAN addresses list

### LINKED ISSUE #1159: Add RFC6598 CG-NAT IPv4 range to "LAN IPs range"

- [x] This issue is present in the nightly release

### Describe the Bug

Tailscale and other VPNs use the RFC6598 range of 100.64.0.0/10 for VPN devices. Currently, this range is not included in the "LAN IPs range" defined by Sunshine, so access to web UI, PIN etc. cannot be allowed for devices on the RFC6598 range. 

As I understand it, here is the code that defines the "LAN IPs range": https://github.com/LizardByte/Sunshine/blob/master/src/network.cpp#L16-L20

For reference, this is why Tailscale uses the RFC6598 range: https://tailscale.com/kb/1015/100.x-addresses/

Here is the RFC6598 document describing this range as the "Shared Address Space to accommodate the needs of Carrier-Grade NAT (CGN) devices": https://datatracker.ietf.org/doc/html/rfc6598

### Expected Behavior

_No response_

### Additional Context

_No response_

### Host Operating System

other, n/a

### Operating System Version

All OSes

### Architecture

other, n/a

0.19.1.d70d084f9fbb4e0150977a89d94937418a3ccf9c

### Package

Windows - Scoop (Third Party)

### GPU Type

none (software encoding)

### GPU Model

N/A

### GPU Driver/Mesa Version

N/A

_No response_

### Config

```Shell
N/A
```

### Apps

_No response_

### Relevant log output

```shell
N/A
```

cgutman:
## Description
Tailscale uses the Carrier-Grade NAT range assigned by RFC 6598 for its VPN interface. Since these addresses are intended for internal use in Carrier-Grade NAT systems, they will not be exposed externally (i.e. without a NAT/firewalll) so it should be safe to treat these as LAN addresses.

### Issues Fixed or Closed
Fixes #1159

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -16,7 +16,8 @@ namespace net {
   std::vector<std::tuple<std::uint32_t, std::uint32_t>> lan_ips {
     ip_block("192.168.0.0/16"sv),
     ip_block("172.16.0.0/12"sv),
-    ip_block("10.0.0.0/8"sv)
+    ip_block("10.0.0.0/8"sv),
+    ip_block("100.64.0.0/10"sv)
   };
 
   std::uint32_t


Nonary:
This should be fine to do, you are correct it should only be used for internal ISP usage. They're supposed to take the CGN IP and then use a public IP on behalf of customer. Hypothetically, if we were to assume that the ISP did not follow standards and instead allowed it to reach the customer without using NAT... the customer has more serious problems to worry about.

Nonary (approved):

PR #1237, opened 2023-04-29
Polish up the elevated command rework

cgutman:
## Description
This PR contains a set of fixes and improvements for the elevated command redesign that @Nonary completed in #1123.

- The first commit ensures that we create the credentials directory before trying to set ACLs on it. Otherwise first installs would not have ACLs set correctly on this directory.
- The second commit reverts a change from #1123 that was resetting ACLs recursively on our directory. We don't want to do that because it would temporarily wipe out our stricter credential directory ACL, creating a window where an ordinary user could read it.
- The third commit reintroduces the start menu shortcut, but it launches Sunshine.exe with a special parameter that causes it to run in launcher mode. If the service is already running, it will just launch the web UI and exit.  If the service isn't running, it will prompt the user for elevation, start the service, then launch the web UI and exit.
- The fourth commit adds a new `interactive` parameter to `platf::run_command()` which allows us to avoid showing an extra console window when launching the browser via the system tray and shortcut.
- The fifth commit fixes a bug in #1123 that was not properly merging the user environment variables into the processes we were launching. I ended up running into this bug while working on the other changes, because Chrome will crash if the user profile variables aren't present.
- The sixth commit resets ACLs on the config directory and our config files when we migrate them from the old directory where ACLs were less strict.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -739,9 +739,6 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
         IfSilent +2 0
         ExecShell 'open' 'https://sunshinestream.readthedocs.io/'
         nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\" /reset /T'
-        nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\\config\\\\credentials\\\" /inheritance:r'
-        nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\\config\\\\credentials\\\" \
-                                  /grant:r Administrators:\\\(OI\\\)\\\(CI\\\)\\\(F\\\)'
         nsExec::ExecToLog '\\\"$INSTDIR\\\\scripts\\\\migrate-config.bat\\\"'
         nsExec::ExecToLog '\\\"$INSTDIR\\\\scripts\\\\add-firewall-rule.bat\\\"'
         nsExec::ExecToLog '\\\"$INSTDIR\\\\scripts\\\\install-service.bat\\\"'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -738,7 +738,7 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
             "${CPACK_NSIS_EXTRA_INSTALL_COMMANDS}
         IfSilent +2 0
         ExecShell 'open' 'https://sunshinestream.readthedocs.io/'
-        nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\" /reset /T'
+        nsExec::ExecToLog 'icacls \\\"$INSTDIR\\\" /reset'
         nsExec::ExecToLog '\\\"$INSTDIR\\\\scripts\\\\migrate-config.bat\\\"'
         nsExec::ExecToLog '\\\"$INSTDIR\\\\scripts\\\\add-firewall-rule.bat\\\"'
         nsExec::ExecToLog '\\\"$INSTDIR\\\\scripts\\\\install-service.bat\\\"'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -764,6 +764,20 @@ if(WIN32) # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.h
         NoDelete:
         ")
 
+    # Adding an option for the start menu
+    set(CPACK_NSIS_MODIFY_PATH "OFF")
+    set(CPACK_NSIS_EXECUTABLES_DIRECTORY ".")
+    # This will be shown on the installed apps Windows settings
+    set(CPACK_NSIS_INSTALLED_ICON_NAME "${CMAKE_PROJECT_NAME}.exe")
+    set(CPACK_NSIS_CREATE_ICONS_EXTRA
+        "${CPACK_NSIS_CREATE_ICONS_EXTRA}
+        CreateShortCut '\$SMPROGRAMS\\\\$STARTMENU_FOLDER\\\\${CMAKE_PROJECT_NAME}.lnk' \
+            '\$INSTDIR\\\\${CMAKE_PROJECT_NAME}.exe' '--shortcut'
+        ")
+    set(CPACK_NSIS_DELETE_ICONS_EXTRA
+        "${CPACK_NSIS_DELETE_ICONS_EXTRA}
+        Delete '\$SMPROGRAMS\\\\$MUI_TEMP\\\\${CMAKE_PROJECT_NAME}.lnk'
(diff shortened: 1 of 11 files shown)


Nonary:
The window for it being open is very small, but that is a valid point.
We need to reset permissions on apps.json and sunshine.conf because both of those are capable of launching commands elevated.


I actually experienced a weird issue where apps.json was retaining permissions even after reset. It said it inherited it from Program Files which made no sense to me because Program File didn't have those permissions. I had to delete and recreate my apps.json file.

cgutman:
> We need to reset permissions on apps.json and sunshine.conf because both of those are capable of launching commands elevated.

Yeah, I think I'll add `icacls /reset` on each of the config files and config directory itself (which won't affect the credentials subfolder, since inheritance is disabled).

Nonary:
@cgutman How do we fix the installer to also add elevation required to sunshine.exe during the installation? We're going to have users try to launch Sunshine without admin rights and run into errors. Not sure how that would work for UAC disabled users though.

Tagging @ns6089 in case he has other suggestions for polishing this up.

ns6089:
I think we can use `highestAvailable` in sunshine.exe manifest. https://learn.microsoft.com/en-us/cpp/build/reference/manifestuac-embeds-uac-information-in-manifest?view=msvc-170#remarks

Admin-capable users will be met with elevation prompt. Non-admin users ideally should be routed to user-specific config folder, but don't know at the moment how to properly detect this.

cgutman:
Forcing elevation via manifest is unnecessary because the launcher already handles that. We don't want to have users see a UAC prompt if all the launcher needs to do is start a web browser. Doing so will not only annoy the user but also launch the browser as admin too. We request elevation only in the specific case where it's actually needed (service start).

If a limited user tries to launch Sunshine via the launcher shortcut, it will work as expected. If Sunshine is already running, it will launch the Web UI as expected. If Sunshine is not running, then will get a UAC prompt asking for the credentials of an administrator user.

cgutman:
> And was given the impression that running sunshine.exe in Program Files by hand (not service) without manual elevation is supposed to fail now.

Yep, that's correct but launching Sunshine.exe via the start menu shortcut without manual elevation should _not_ fail (and does not fail in my PR). If we put a manifest on Sunshine.exe that requires elevation, then it's going to elevate even when not needed (including in the portable build, where elevation is not needed ever). 

In the new post-https://github.com/LizardByte/Sunshine/pull/1123 world, the service is mandatory and the only supported way to run Sunshine when using the installer version. The installer has always required admin anyway to install to `Program Files`, so we know users have admin rights (directly or indirectly) already. For non-admin users, the portable version still fully supports that use case.

ns6089:
> If we put a manifest on Sunshine.exe that requires elevation, then it's going to elevate even when not needed (including in the portable build, where elevation is not needed ever).

Then we use sunshine with manifest in installer, and sunshine.exe without manifest in portable .zip?

> the service is mandatory and the only supported way to run Sunshine when using the installer version

Then we need portable and non-portable sunshine.exe even more so. Either we check for SYSTEM user and display clear error, or we're going to keep getting error reports forever. Everyone would expect sunshine.exe to either work or not work with clear indication of why.

ns6089:
Or I guess we can put some `service_required` file in the installation folder, and check for that. This way we can keep one sunshine.exe for everything.

Nonary:
I tried enabling the admin flag on sunshine.exe and it caused problems because suddenly every command required elevation even when it didn't need elevation. I have a suspicion this might not be an easy fix and its probably just best to just add an error message stating admin rights is required, but it kind of already does.
image

Nonary:
Yeah so I suspect users are going to be dumb and delete the credential file and then have it recreate automatically once they open sunshine again, only for it to now allow it to be read by users again.

There's also going to be users who just allow permissions to it by double clicking on it. Not really sure how we're going to fix that. Do we let them intentionally ruin their own security in this case?

cgutman:
> I tried enabling the admin flag on sunshine.exe and it caused problems because suddenly every command required elevation even when it didn't need elevation. I have a suspicion this might not be an easy fix and its probably just best to just add an error message stating admin rights is required, but it kind of already does.

Yeah, it's because of the way that we handle impersonation when running as System vs other users (and another reason why the service is mandatory now). I'm going to add a warning for the case where access is denied trying to read/write the credentials.

> Yeah so I suspect users are going to be dumb and delete the credential file and then have it recreate automatically once they open sunshine again, only for it to now allow it to be read by users again.

That wouldn't work since the credentials directory still doesn't allow non-admin writes (unless they modified the ACLs). Even if they fully deleted the credentials directory, non-admin Sunshine still wouldn't be able to create it again because the `%ProgramFiles\Sunshine` directory inherits the ACLs from `%ProgramFiles%` that only grant write access to admins.

> There's also going to be users who just allow permissions to it by double clicking on it. Not really sure how we're going to fix that. Do we let them intentionally ruin their own security in this case?

There's only so much we can do for users that tamper with our files. We should give them a secure system by default and guidance on how to keep it that way, but it's ultimately on them if they change things in ways that open up security holes.

There's basically no legitimate reason someone would ever need to modify the contents of their credentials folder anyway. The auto-generated key pair is fine. In the hypothetical case that someone wanted to get a Let's Encrypt cert issued for their Sunshine instance or something, I think those people are going to know to lock down their credential directory (just as `certbot` does for its key storage). If they don't, that's really on them because they exposed their private key by their own action.

Nonary:
@cgutman I have already seen 2 cases today where someone thought they "fixed" their issue in nightly by deleting the credential folder... 

It's because the error message currently says can't find/open and users troubleshoot it by deleting the folder I guess. Then they run it again, states permissions required, so they run it as admin and think they've fixed their problem.

I think adding an error message will stop that from happening so not a big deal most likely.

cgutman:
> It's because the error message currently says can't find/open and users troubleshoot it by deleting the folder I guess. Then they run it again, states permissions required, so they run it as admin and think they've fixed their problem.

I don't think the state of affairs now is a good indication of how users will behave once this PR is merged. We removed the Start Menu shortcut during the redesign, so there's no real user-friendly way to launch Sunshine in the current nightlies. That causes users to go searching in Program Files trying to figure out how to run it, leading to a cascade of confusion and bad decisions trying to "fix" things that aren't broken.

Once users can run Sunshine from the Start Menu again, I don't think we'll see nearly as many people messing up their installations trying to run it manually (and it will warn them now not to do that).

cgutman:
I'm going to go ahead and merge this, so we can see how things play out in nightlies with real users testing it. Nightlies are currently a bit broken launching certain apps due to lack of e67ca9a2df43873ea9611332aa74a90b4d238ec3.

If it turns out that we need more stern warnings or something else to further prevent confusion, we can add those in a separate PR.

PR #1246, opened 2023-05-04
Audio Capture Improvements

cgutman:
## Description
This PR contains several improvements to make Windows audio capture easier to configure and avoid breaking streaming entirely if the audio sink is invalid. It also improves the audio-info.exe tool to fix misleading output in several cases.

- The first commit adjusts the Audio Sink and Virtual Sink values to accept device names, device descriptions, and adapter names. This results in a more stable configuration (since names don't change across reconnection, but IDs do) and is also easier for users to set up without having to use audio-info.exe. The documentation and UI are also changed to reflect that sink names should be used on Windows.  The old format device IDs are still supported for backwards compatibility with existing configs.

- The second commit cleans up some misleading output from audio-info.exe that causes it to appear that certain channel configurations are not supported simply because the audio device is not currently using that format.

- The third commit adds 5.1 configurations with side speakers and quadraphonic configurations, so audio-info.exe displays those correctly instead of "Unknown".

- The fourth commit fixes inaccurate results caused by an incompatible sample rate when the audio device is using 192 KHz sample rate. In that scenario, audio-info.exe would incorrectly show all channel configurations as unsupported.

- The fifth commit adjusts the behavior when the audio sink is invalid, missing, or cannot be initialized. Rather than instantly terminating the stream right after it starts, which is incredibly confusing for users, it will print a log message and allow streaming to continue with no audio. The lack of audio will likely drive users to check the Audio/Video tab, which is what we want. This behavior, combined with the tendency for audio device IDs to not be stable, has been a major cause of user issues.

- The sixth commit fixes a bug in the audio reinit logic where `return` was accidentally used instead of `continue`. Instead of capturing from the newly reinitialized device, the audio thread would return instead. This fixes recovery of audio capture after a device disconnect, though we don't track changes to the default output yet. I also added a retry loop to reinit, because sometimes the audio device is temporarily lost only to come back shortly after (for example, a display while mode-setting).

### Screenshot

image

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -447,6 +447,8 @@ audio_sink
 
             tools\audio-info.exe
 
+         .. Tip:: If you have multiple audio devices with identical names, use the Device ID instead.
+
    .. Tip:: If you want to mute the host speakers, use `virtual_sink`_ instead.
 
 **Default**
@@ -466,7 +468,7 @@ audio_sink
    **Windows**
       .. code-block:: text
 
-         audio_sink = {0.0.0.00000000}.{FD47D9CC-4218-4135-9CE2-0C195C87405B}
+         audio_sink = Speakers (High Definition Audio Device)
 
 virtual_sink
 ^^^^^^^^^^^^
@@ -488,7 +490,7 @@ virtual_sink
 **Example**
    .. code-block:: text
 
-      virtual_sink = {0.0.0.00000000}.{8edba70c-1125-467c-b89c-15da389bc1d4}
+      virtual_sink = Steam Streaming Speakers
 
 Network
 -------
--- a/src_assets/common/assets/web/config.html
+++ b/src_assets/common/assets/web/config.html
@@ -448,13 +448,14 @@
           type="text"
           class="form-control"
           id="audio_sink"
-          placeholder="{0.0.0.00000000}.{FD47D9CC-4218-4135-9CE2-0C195C87405B}"
+          placeholder="Speakers (High Definition Audio Device)"
           v-model="config.audio_sink"
         />
         <div class="form-text">
-          The name of the audio sink used for Audio Loopback<br />
+          The name of the audio sink used for audio capture. If not set, the default audio device will be used.<br />
           You can find the name of the audio sink using the following command:<br />
           <pre>tools\audio-info.exe</pre>
+          If you have multiple audio devices with identical names, use the Device ID instead.
         </div>
       </div>
       <div class="mb-3" v-if="platform === 'linux'">
@@ -506,12 +507,12 @@
           type="text"
           class="form-control"
           id="virtual_sink"
-          placeholder="{0.0.0.00000000}.{8edba70c-1125-467c-b89c-15da389bc1d4}"
+          placeholder="Steam Streaming Speakers"
           v-model="config.virtual_sink"
         />
         <div class="form-text">
-          The virtual sink, is the audio device that's virtual (Like Steam Streaming Speakers), it allows Sunshine to
-          stream audio, while muting the speakers.
+          The virtual sink is an audio device that's virtual (like Steam Streaming Speakers). It allows Sunshine to
+          stream audio, while muting the host PC speakers.
         </div>
       </div>
       <!--Adapter Name -->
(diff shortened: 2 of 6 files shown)


src_assets/common/assets/web/config.html
```diff
@@ -448,11 +448,11 @@ <h1 class="my-4">Configuration</h1>
...
           id="audio_sink"
-          placeholder="{0.0.0.00000000}.{FD47D9CC-4218-4135-9CE2-0C195C87405B}"
+          placeholder="Speakers (High Definition Audio Device)"
           v-model="config.audio_sink"
         />
         <div class="form-text">
-          The name of the audio sink used for Audio Loopback<br />
+          The name of the audio sink used for audio capture. If not set, the default audio device will be used.<br />
           You can find the name of the audio sink using the following command:<br />
           <pre>tools\audio-info.exe</pre>
```
ReenigneArcher, line 457: Could we make the command example show up only when running on Windows? Or make this tool available on Unix platforms as well?

docs/source/about/advanced_usage.rst
```diff
@@ -466,7 +466,7 @@ audio_sink
    **Windows**
       .. code-block:: text
 
-         audio_sink = {0.0.0.00000000}.{FD47D9CC-4218-4135-9CE2-0C195C87405B}
+         audio_sink = Speakers (High Definition Audio Device)
```
ReenigneArcher, line 469: Should we have two examples? One with the old ID method, in the event a user has multiple of the same audio device when going by name (e.g. identical monitors)

ReenigneArcher (commented):
On this same topic... A question about the audio sinks...

Do we really need `audio_sink` and `virtual_sink`? This has always confused me. Couldn't a user just put a virtual sink in the `audio_sink` field? Without digging into the code, these two fields seem equivalent to me (based on the documentation that exists).

Also, how hard would it be to change audio device back to the original device once streaming ends? This has been a common complaint.

docs/source/about/advanced_usage.rst
cgutman: Identical monitors will actually have unique names. I really don't want to encourage use of the old IDs because they tend to break during normal things like GPU/audio driver updates or connecting a new monitor.

image

src_assets/common/assets/web/config.html
cgutman: Yep, it already does. It's in a `v-if="platform === 'windows'"` div.

cgutman:
> Do we really need audio_sink and virtual_sink? This has always confused me. Couldn't a user just put a virtual sink in the audio_sink field? Without digging into the code, these two fields seem equivalent to me (based on the documentation that exists).

Virtual sink is a device that we take full control over and reconfigure as needed based on the surround configuration requested by Moonlight. We will handle switching the default audio device to it and back to the normal device after we stop streaming.

Audio sink is meant for real devices that we shouldn't be reconfiguring because it's attached to real speakers. I believe we still switch the default audio device to the specified audio sink in the current code, but in theory it allows you to capture on a non-default audio device if you wanted to.

In almost all cases, there's no reason to specify either of these. If you just install Steam Streaming Speakers, we will automatically use them as a virtual sink, otherwise we'll use your default audio device which is probably what you want.

I have additional work planned to fix tracking changes in the default audio device and switching audio devices on the fly if the default changes. In that change, I can also stop changing the default device when audio sink is set.

> Also, how hard would it be to change audio device back to the original device once streaming ends? This has been a common complaint.

If we find a working virtual sink, we will actually do that.

docs/source/about/advanced_usage.rst
ReenigneArcher: That's interesting. I cannot test right now, but I was pretty sure I wasn't getting unique names with my old 3 HP monitor setup.

ReenigneArcher:
> If we find a working virtual sink, we will actually do that.

You're right, I just tested again and it does indeed switch back.

docs/source/about/advanced_usage.rst
cgutman: My initial test was on a Win11 PC with an Intel GPU. I tested on a Win10 PC with an AMD GPU and saw the same results with those `-2` and `-3` strings appended to make the name unique. However, this particular system has a duplicate SPDIF output device for some reason, so it does seem that duplication is possible in general even if it doesn't seem to happen for displays.

I'll add a note to the docs and the UI to prefer names but use IDs if the names are duplicated.

PR #1253, opened 2023-05-06
Archive Win32 binaries with debug info for crash analysis

cgutman:
## Description
It's very difficult to debug user-reported crashes on Windows, even with a crash dump, because we don't build our binaries with debug info. This PR changes the Windows build to build with debug info and archive the unstripped binaries in a separate zip file for crash analysis. CPack strips the binaries before packaging, so the actual release binaries are unaffected by this change.

With this change, we can download the associated `sunshine-debuginfo-win32.zip` for a given build and generate PDBs with [cv2pdb](https://github.com/rainers/cv2pdb) to debug crashes from any nightly or release build.

I gave the debug info zip file a name with a different format than other artifacts to hopefully reduce the chance someone downloads it by mistake.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -781,7 +781,7 @@ jobs:
         run: |
           mkdir build
           cd build
-          cmake -DCMAKE_BUILD_TYPE=Release \
+          cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo \
             -DSUNSHINE_ASSETS_DIR=assets \
             -G "MinGW Makefiles" \
             ..
@@ -800,6 +800,15 @@ jobs:
           # move
           mv ./cpack_artifacts/Sunshine.exe ../artifacts/sunshine-windows-installer.exe
           mv ./cpack_artifacts/Sunshine.zip ../artifacts/sunshine-windows-portable.zip
+
+      - name: Package Windows Debug Info
+        working-directory: build
+        run: |
+          # save the original binaries with debug info
+          7z -r `
+            "-xr!CMakeFiles" `
+            "-xr!cpack_artifacts" `
+            a "../artifacts/sunshine-debuginfo-win32.zip" "*.exe"
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v3


.github/workflows/CI.yml
```diff
@@ -793,6 +794,11 @@ jobs:
           mkdir -p artifacts
           cd build
 
+          # save the original binaries with debug info
+          for f in $(find . -name '*.exe' -not -path './CMakeFiles/*'); do
+            zip ../artifacts/sunshine-debuginfo-win32.zip "$f"
+          done
+
```
ReenigneArcher, line 801:
```suggestion
          # save the original binaries with debug info
          7z \
            "-xr!CMakeFiles" \
            a "../artifacts/sunshine-debuginfo-win32.zip" "*.exe"

```

This wouldn't require installing another dependency. I believe 7z is included in all GitHub runners by default.

ReenigneArcher: Also, FYI... there's just one artifact uploaded... all 3 artifacts will reside in the `sunshine-windows.zip` artifact.

- sunshine-windows-installer.exe
- sunshine-windows-portable.zip
- sunshine-debuginfo-win32.zip

cgutman: > This wouldn't require installing another dependency. I believe 7z is included in all GitHub runners by default.

It might be included in a normal shell, but it doesn't seem to be on MSYS2's `PATH`. https://github.com/cgutman/LB_Sunshine/actions/runs/4904317553/jobs/8757283657

> Also, FYI... there's just one artifact uploaded... all 3 artifacts will reside in the sunshine-windows.zip artifact.

Yep, that's true for branches, but nightly/release should publish them individually right? Or do I need to make another change to do that?

ReenigneArcher: > should publish them individually right

For releases, each file (all 3) in the artifacts directory will be a separate release asset.

In that case, I suggest to make a separate step, which will resolve both issues (assuming you don't want to upload the debug info as a release asset).

```yml
- name: Package Windows Debug
  if: ${{ needs.setup_release.outputs.create_release != 'true' }}
  working-directory: build
  run: |
    # save the original binaries with debug info
    7z \
      "-xr!CMakeFiles" \
      "-xr!cpack_artifacts" \
      a "../artifacts/sunshine-debuginfo-win32.zip" "*.exe"
```

If you do want it uploaded with releases, just remove the `if` line.

cgutman: We do want it for releases because we'll need it to debug crashes in those builds (which will probably be what most users run).

PR #1262, opened 2023-05-09
Add support for installing the Steam Streaming Speakers driver

cgutman:
## Description
This PR adds a feature to install the Steam Streaming Speakers driver if it's present on disk (Steam installed) but not already installed (if the user hasn't used Steam Remote Play). Since some users may have their own audio configuration that they don't want us messing with, I added a UI option to opt-out of this behavior.

The Steam Streaming Speakers driver allows us to support 5.1 and 7.1 surround sound on systems without a physical surround sound output. It also allows the "Play audio on host PC" option in Moonlight to correctly mute host audio.

### Screenshot
image

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -382,7 +382,11 @@ namespace config {
     true  // dwmflush
   };
 
-  audio_t audio {};
+  audio_t audio {
+    {},  // audio_sink
+    {},  // virtual_sink
+    true,  // install_steam_drivers
+  };
 
   stream_t stream {
     10s,  // ping_timeout
@@ -985,6 +989,7 @@ namespace config {
 
     string_f(vars, "audio_sink", audio.sink);
     string_f(vars, "virtual_sink", audio.virtual_sink);
+    bool_f(vars, "install_steam_audio_drivers", audio.install_steam_drivers);
 
     string_restricted_f(vars, "origin_pin_allowed", nvhttp.origin_pin_allowed, { "pc"sv, "lan"sv, "wan"sv });
     string_restricted_f(vars, "origin_web_ui_allowed", nvhttp.origin_web_ui_allowed, { "pc"sv, "lan"sv, "wan"sv });
--- a/src/platform/windows/audio.cpp
+++ b/src/platform/windows/audio.cpp
@@ -864,7 +864,7 @@ namespace platf::audio {
 
     /**
      * @brief Installs the Steam Streaming Speakers driver, if present.
-     * @return true if installation was successful
+     * @return `true` if installation was successful.
      */
     bool
     install_steam_audio_drivers() {
--- a/src/platform/windows/audio.cpp
+++ b/src/platform/windows/audio.cpp
@@ -10,6 +10,8 @@
 
 #include <synchapi.h>
 
+#include <newdev.h>
+
 #define INITGUID
 #include <propkeydef.h>
 #undef INITGUID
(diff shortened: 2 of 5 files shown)


src/config.cpp
```diff
@@ -382,7 +382,11 @@ namespace config {
     true  // dwmflush
   };
 
-  audio_t audio {};
+  audio_t audio {
+    {},  // audio_sink
+    {},  // virtual_sink
+    true,  // install_steam_drivers
```
ReenigneArcher, line 388:
```suggestion
    false,  // install_steam_drivers
```

Default should probably be false, so we don't mess with anything until they opt in?

src/platform/windows/audio.cpp
```diff
@@ -878,6 +862,78 @@ namespace platf::audio {
       return std::nullopt;
     }
 
+    /**
+     * @brief Installs the Steam Streaming Speakers driver, if present.
+     * @return true if installation was successful
```
ReenigneArcher, line 867:
```suggestion
     * @return `true` if installation was successful.
```

src_assets/common/assets/web/config.html
```diff
@@ -992,6 +1004,7 @@ <h1 class="my-4">Configuration</h1>
     "amd_vbaq": "enabled",
     "capture": "",
     "controller": "enabled",
+    "install_steam_audio_drivers": "enabled",
```
ReenigneArcher, line 1007:
```suggestion
    "install_steam_audio_drivers": "disabled",
```

ReenigneArcher (commented):
I'll trust your c++... my biggest observation is the default value. I'm thinking this should be disabled by default to avoid installing extra drivers in cases where a user might not want it.

cgutman:
> I'll trust your c++... my biggest observation is the default value. I'm thinking this should be disabled by default to avoid installing extra drivers in cases where a user might not want it.

I think it should be enabled by default for a few reasons:
- Steam installs it when you stream with Steam Remote Play, so there's existing precedent for installing audio drivers by default as part of game streaming.
- Host audio mute is a default enabled behavior in Moonlight, so it's likely to be used in the vast majority of streaming scenarios. Sunshine shouldn't require special configuration to work as expected (to the extent we can, of course).
- We take care to set their original audio device back to default after installing Steam Streaming Speakers to prevent changes to their default audio output as a result of the installation (Steam does this too)

For the scenario where the user doesn't want our default, I think it's much more likely that a user will recognize what's happening if they see Sunshine installing that driver and want to stop it vs. a user wondering why they pick 5.1 surround sound in Moonlight and get upmixed stereo or why Moonlight's host audio mute seems to be broken. I don't think users in the latter case would necessarily understand what's going on to try to fix it.

There seem to me to be far more users where this behavior will help than where it will hurt. If a user doesn't want it, they can simply remove the device and disable the option in Sunshine. We can also document it in the "Breaking Changes" section to ensure upgrading users adjust the option to their preference before streaming.

ReenigneArcher:
> adjust the option to their preference before streaming

So it doesn't install the driver until a stream starts? If that's the case, that will be better than installing it once sunshine starts up.

I trust your judgement here. If it becomes an issue I can add some instructions for manually removing the driver.

cgutman:
> So it doesn't install the driver until a stream starts? If that's the case, that will be better than installing it once sunshine starts up.

Yep, it only installs when a stream actually starts.

ReenigneArcher (dismissed):

cgutman:
Pushed a small fix to let us successfully install the driver in the case where there are no audio devices on the system (therefore no existing default device to restore).

ReenigneArcher (approved):

FallingSnow:
When using the steam audio driver and dragging the audio slider to 0, the audio gets progressively quieter. However once I hit 0 (and the windows icon changes to muted) the output volume is maxed again. So while windows says it is muted, the client plays volume at max.

Not sure if this is a stream driver issue.

Roshan-Localhost:
Hi @cgutman @ReenigneArcher 
For Some Reason for me the steam streaming drivers are not getting installed.
As per logs it does says that steam streaming drivers has been installed, but on device manager it doesn't show up anything. 
And the Audio is playing on both Moonlight side pc and on PC where sunshine installed.

` Info: Successfully installed Steam Streaming Speakers`

But, It does works if I Manually install the steam drivers, but it was a pain to install the steam drivers and there are now multiple copies of steam streaming speakers on my pc. 

btw I'm on windows 11.

Thanks

Volumie:
Hi @cgutman 

For me Steaming streaming speaker is not getting installed.
But the logs does say:
`Info: Successfully installed Steam Streaming Speakers`

So Sound is getting played on both the Moonlight PC and Sunshine PC.

Thanks

PR #1290, opened 2023-05-17
Print a warning if GameStream is using conflicting ports

cgutman:
## Description
We've had many instances of users that install Sunshine without turning off GameStream in GeForce Experience. This leads to all sorts of issues trying to stream that are difficult to debug, especially because users don't always mention that they were/are using GameStream.

This PR prints a log message if GameStream is enabled and Sunshine is configured to use the same set of ports as GFE. I printed them at the "Fatal" level in an attempt to draw the most attention to them.

It would be nice to have some prominent way of displaying major configuration warnings in the Web UI like specifying a missing display, having no working encoders, port conflicts, etc. Since we don't have that, a log message is the next best thing.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -333,6 +333,26 @@ namespace service_ctrl {
     return false;
   }
 }  // namespace service_ctrl
+
+/**
+ * @brief Checks if NVIDIA's GameStream software is running.
+ * @return `true` if GameStream is enabled.
+ */
+bool
+is_gamestream_enabled() {
+  DWORD enabled;
+  DWORD size = sizeof(enabled);
+  return RegGetValueW(
+           HKEY_LOCAL_MACHINE,
+           L"SOFTWARE\\NVIDIA Corporation\\NvStream",
+           L"EnableStreaming",
+           RRF_RT_REG_DWORD,
+           nullptr,
+           &enabled,
+           &size) == ERROR_SUCCESS &&
+         enabled != 0;
+}
+
 #endif
 
 /**
@@ -633,6 +653,14 @@ main(int argc, char *argv[]) {
 
   std::thread httpThread { nvhttp::start };
   std::thread configThread { confighttp::start };
+
+#ifdef _WIN32
+  // If we're using the default port and GameStream is enabled, warn the user
+  if (config::sunshine.port == 47989 && is_gamestream_enabled()) {
+    BOOST_LOG(fatal) << "GameStream is still enabled in GeForce Experience! This *will* cause streaming problems with Sunshine!"sv;
+    BOOST_LOG(fatal) << "Disable GameStream on the SHIELD tab in GeForce Experience or change the Port setting on the Advanced tab in the Sunshine Web UI."sv;
+  }
+#endif
 
   rtsp_stream::rtpThread();
 


src/main.cpp
```diff
@@ -333,6 +333,26 @@ namespace service_ctrl {
     return false;
   }
 }  // namespace service_ctrl
+
+/**
+ * @brief Checks if NVIDIA's GameStream software is running.
+ * @return `true` if GameStream is enabled.
+ */
+bool
+is_gamestream_enabled() {
```
ReenigneArcher, line 342: Should we put function somewhere under `src/platform/windows`?

src/main.cpp
```diff
@@ -634,6 +654,14 @@ main(int argc, char *argv[]) {
   std::thread httpThread { nvhttp::start };
   std::thread configThread { confighttp::start };
 
+#ifdef _WIN32
+  // If we're using the default port and GameStream is enabled, warn the user
+  if (config::sunshine.port == 47989 && is_gamestream_enabled()) {
+    BOOST_LOG(fatal) << "GameStream is still enabled in GeForce Experience! This *will* cause streaming problems with Sunshine!"sv;
+    BOOST_LOG(fatal) << "Disable GameStream on the SHIELD tab in GeForce Experience or change the Port setting on the Advanced tab in the Sunshine Web UI."sv;
```
ReenigneArcher, line 661: `fatal` will terminate Sunshine, right? I'm okay with that since Sunshine wouldn't work at all anyway (except the web ui), but just wondering.

If `fatal` terminates, won't the first `fatal` prevent the second one?

ReenigneArcher (commented):
Nice idea! No objections to keeping this as is.

Here is one idea though. What about a config option which when set to `enabled` would automatically disable gamestream in the registry (assuming running as a service or with elevated permissions)?

src/main.cpp
cgutman: `fatal` doesn't terminate. It's just another log level.

I made it `fatal` to attract attention, but maybe @TheElixZammuto 's change will make this unnecessary.

src/main.cpp
cgutman: I've gone back and forth on this one. The code in `src/platform` right now is actually code that implements common interfaces shared between platforms (the functions in `src/platform/common.h`). 

We could add a `src/platform/windows/win32.h` or similar and declare Windows-specific interfaces there then include it everywhere we need it, but we'd still end up with `#ifdef _WIN32` peppered around so it didn't feel like a big win.

cgutman:
> Here is one idea though. What about a config option which when set to enabled would automatically disable gamestream in the registry (assuming running as a service or with elevated permissions)?

It looks like it's not that easy. I tried changing `EnableStreaming` in the registry, but GameStream doesn't immediately quit when that is set to 0. It appears that GFE has some internal interface with the nvcontainer.exe processes hosting the streaming code that only triggers updates to the status when it starts or the toggle is changed in GFE.

src/main.cpp
TheElixZammuto: Actually the changes I made only shows FATAL level logs in the index.html of the Web UI, so it can be generalized for many error cases (e.g. ViGemBus not installed, Encoder Not Found, etc...)

PR #1305, opened 2023-05-24
Audio/Video config improvements

cgutman:
## Description
This PR includes a couple of cosmetic improvements to the Web UI to discourage use of manually specified audio sinks and GPUs on Windows. I reduced the opacity of the placeholder text to make it more visually distinct from populated text fields and field description text.

I also fixed an issue in the audio sink selection logic which caused a bug where specifying an Audio Sink manually would prevent the Play Audio on Host PC option in Moonlight from working (because we'd never use a virtual sink, even if one was available).

### Screenshot

image

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -153,14 +153,16 @@ namespace audio {
     }
 
     // Order of priority:
-    // 1. Config
-    // 2. Virtual if available
+    // 1. Virtual sink
+    // 2. Audio sink
     // 3. Host
     std::string *sink = &ref->sink.host;
     if (!config::audio.sink.empty()) {
       sink = &config::audio.sink;
     }
-    else if (ref->sink.null) {
+
+    // Prefer the virtual sink if host playback is disabled or there's no other sink
+    if (ref->sink.null && (!config.flags[config_t::HOST_AUDIO] || sink->empty())) {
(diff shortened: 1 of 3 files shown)


src/audio.cpp
```diff
@@ -153,14 +153,16 @@ namespace audio {
     }
 
     // Order of priority:
-    // 1. Config
-    // 2. Virtual if available
+    // 1. Virtual sink
```
ReenigneArcher, line 156: Wouldn't the config option still be the first priority?

src_assets/common/assets/web/header.html
```diff
@@ -68,6 +68,9 @@
   .nav-link.active {
     font-weight: 500;
   }
+  .form-control::placeholder {
+    opacity: 0.5;
+  }
```
ReenigneArcher, line 73: This is good enough for now. Eventually I'll clean up this css and separate it from the html code.

src/audio.cpp
cgutman: It was somewhat confusing as written before because Virtual Sink and Audio Sink are both "Config". This clarifies the prescedence between the two config options.

ReenigneArcher (approved):

PR #1308, opened 2023-05-24
Modify relevant NVIDIA control panel settings on Windows

ns6089:
## Description
This automates NVIDIA control panel settings modification, namely:
- sunshine.exe power profile, this can easily cut down 15ms of desktop latency, because default settings keep gpu clocks (particularly memory) at bare minimum
- global OpenGL/Vulkan DXGI swapchain, required for full frame rate capture with these apis, the setting is reverted to default value on sunshine.exe exit

## Type of Change

## Checklist

## Branch Updates


ns6089:
@ReenigneArcher So, regarding nvapi submodule. Could you list exact steps that must be taken there (and who should perform them).

tools/nvprefs/nvprefs.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+  NvAPI_Status
+  restore_global_profile(NvDRSSessionHandle session_handle, bool &session_modified) {
+    NvAPI_Status status;
+
+    NvDRSProfileHandle profile_handle = 0;
+    status = NvAPI_DRS_GetBaseProfile(session_handle, &profile_handle);
+    if (status != NVAPI_OK) return status;
+
+    status = NvAPI_DRS_RestoreProfileDefaultSetting(session_handle, profile_handle, OGL_CPL_PREFER_DXPRESENT_ID);
```
cgutman, line 138: We probably want to remember that we changed this and only restore the previous setting if we changed it. Otherwise we might revert this setting to default after a user changed it manually.

ns6089: Will work until the first crash or otherwise improper system shutdown. Pretty much all systems will normalize to "enabled" sooner or later.

ns6089: I guess we can save it to file?

cgutman: Yeah, a file or a registry key seems reasonable.

ns6089: Still, one little mistake and we pollute the settings with false "enabled", many ways it can go wrong. Resetting to default is simple, and should be reasonably safe to merge this close to release. I believe anything more complex needs more time in nightly.

We can reset to default this release, and leave the complex logic for the next. Or we can simply move the whole thing to the next. Thoughts?

ns6089: *Still*, we can similarly pollute the system during uninstallation, if it doesn't terminate sunshine.exe gracefully. Let's move it to the next release after all, too many potential problems.

ns6089: The things to keep track of (so far)
 - use registry, to persist across reinstallations, especially portable
 - call restore from the uninstaller, in case sunshine didn't exit gracefully
 - save "using default" as separate value, in case nvidia decides to change it

tools/nvprefs/nvprefs.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    return NVAPI_OK;
+  }
+
+  NvAPI_Status
+  restore_global_profile(NvDRSSessionHandle session_handle, bool &session_modified) {
+    NvAPI_Status status;
+
+    NvDRSProfileHandle profile_handle = 0;
+    status = NvAPI_DRS_GetBaseProfile(session_handle, &profile_handle);
```
Lucide, line 135: You might want to use [`NvAPI_DRS_GetCurrentGlobalProfile()`](https://docs.nvidia.com/gameworks/content/gameworkslibrary/coresdk/nvapi/group__drsapi.html#ga68dbe5987a8f530a0041ba94019b0ec0) instead. See [here](https://github.com/Lucide/vsync-tool#nvdrs_settingsettinglocation-values).

ns6089: From [NVIDIA Driver Settings Programming Guide](https://developer.download.nvidia.com/tools/docs/PG-5116-001_v02_public.pdf):
> A profile with one or more applications is called an Application Profile. Applications
that have a profile have their settings applied automatically when the application starts.
> 
> The Base Profile is a profile that always exists and is not associated with a specific
application. The settings in the Base Profile are applied to all processes on the system
automatically.
> 
> A profile with no applications is called a Global Profile. The settings from a Global
Profile are applied to all processes on the system, but only if that profile is selected to be
the Current Global Profile in the system.

...

> When an application starts, there might be three profiles contesting to provide settings to
the NVIDIA driver: the Application Profile, the Current Global Profile and the Base
Profile.
> 
> The settings from the Application Profile have the highest priority and override any
settings on the other two profiles. Next in priority order comes the Current Global
Profile, and finally the Base Profile.

No idea if anyone actually uses global profiles, but if somebody changes the profile we will lose the setting, even if the profile doesn't change that setting to begin with. There might be some sense in changing both base and global profiles at the same time, but it's going to complicate the undo logic, I will think about it.

tools/nvprefs/nvprefs.cpp
Lucide: Some potential useful things to keep in mind:

* nvapi drs api offers [`NvAPI_DRS_SaveSettingsToFile()`](https://docs.nvidia.com/gameworks/content/gameworkslibrary/coresdk/nvapi/group__drsapi.html#gaf2f69fee6df6510c8cb5a5468b0e0b05), this call makes a full copy of the settings db to a file. The backup file can be loaded and navigated the same way as the actual settings.
* I agree that restoring previous settings is important. Nvidia driver settings use inheritance so this is not so trivial. 
  * In our case, we can probably assume that sunshine will have no "bundled" app profile (an app profile bundled with the driver). This allows us to only use `.settingLocation != NVDRS_CURRENT_PROFILE_LOCATION` to determine if a setting has to be set to a previously user value or reset. (In a general case, we would also need to check `.isCurrentPredefined`).
  * _Getting_ the current setting value can fail if the setting is inherited. Just a reminder.
*  That said, restoring a custom setting for the sunshine profile is probably unnecessary. The global setting must be restored. With the assumption above, the restore process uses the same set-or-reset logic.
* As you wrote above, guaranteeing restoration is complicated. Avoiding global profile tampering altogether might simply be better? The user can use the tool as a do-undo command.

tools/nvprefs/nvprefs.cpp
Lucide: `NVDRS_GLOBAL_PROFILE_LOCATION` points to the active global profile. In most cases, yes, it will be the base profile. If the user uses something like [NVIDIA Profile Inspector](https://github.com/Orbmu2k/nvidiaProfileInspector), that might not be the case. You do not need to keep track of both, only the global profile will be sufficient.

ns6089: Global profile doesn't replace base profile, it extends it. It can't "point" to base profile. Similarly to application profile. Basically both global and application profiles work on opt-in basis and only override specific settings.

And regarding NVIDIA Profile Inspector, I don't see the ability to *change* global profile. It can change base profile, create named profile, and add applications to it.

Lucide: I think I remember that, in a "standard" situation, a global profile might not exist. In this case, `NvAPI_DRS_GetCurrentGlobalProfile()` returns the same as `NvAPI_DRS_GetBaseProfile()`, and `.settingLocation` is always `NVDRS_GLOBAL_PROFILE_LOCATION`. A different current global profile can be set with [`NvAPI_DRS_SetCurrentGlobalProfile()`](https://docs.nvidia.com/gameworks/content/gameworkslibrary/coresdk/nvapi/group__drsapi.html#ga35f52bf2ffa9d2a8f758912c41f4e936). You're right, in the profile inspector, you can't set the current global profile, but if you do programmatically, it will reflect the change.

.github/workflows/CI.yml
```diff
@@ -737,7 +737,7 @@ jobs:
 
   build_win:
     name: Windows
-    runs-on: windows-2019
+    runs-on: windows-2022
```
ReenigneArcher, line 740: Is this change necessary for your PR? I'd prefer if this change were in a separate PR.

third-party/nvapi-windows-amd64/include/NvApiDriverSettings.c
ReenigneArcher, line 1: What is the original source of these third-party files?

tools/nvprefs/nvprefs.cpp
ns6089: Sorry for seemingly ignoring you, just had to basically rewrite a lot of stuff and this question was hardly a priority at that moment.

After giving it more thought, I think the most logical way is to follow what NVIDIA control panel itself does. Whether after `NvAPI_DRS_SetCurrentGlobalProfile()` it switches to that profile, or sticks to the base one. Will test this and possibly make modifications accordingly.

Lucide: No worries, mine were just suggestions. If the code is sound, corrections will be trivial should issues arise.

ns6089: Just tested, NVIDIA control panel only displays and modifies base profile even after custom one is set, and that custom profile properly inherits from the base one. So unless custom global profile explicitly overrides the settings we change we should be fine, and I'd rather not change settings that are essentially hidden from the user. Sticking to the base profile for now, unless some particular use-case arises where it is not enough.

ns6089:
As far as I can tell, this should be ready to merge.

ReenigneArcher:
> As far as I can tell, this should be ready to merge.

I think we just need to deal with the third party files. I'm reviewing on my phone so I could be missing something.

https://github.com/LizardByte/Sunshine/pull/1308#discussion_r1240469153

ns6089:
> I think we just need to deal with the third party files. I'm reviewing on my phone so I could be missing something.
> 
> [#1308 (comment)](https://github.com/LizardByte/Sunshine/pull/1308#discussion_r1240469153)

That comment is now outdated, I found the opensource version NVIDIA distributes. To quote:
```
This release contains a version of nvapi.h, provided under MIT license, to enable open-source
re-implementations of NVAPI for Windows emulation environments.
```
So no need for binary libraries, standalone tools, and everything is getting called from sunshine.exe

`third-party/nvapi-opensource/readme.txt` contains the url for possible updates and everything.
But these updates won't be really needed unless we will need new functions, everything else is backwards-compatible.
It's only headers basically.

ReenigneArcher:
I'd prefer to add them as submodules if it's possible.

ns6089:
> I'd prefer to add them as submodules if it's possible.

Feel free place it anywhere you want, I will modify the PR. But I don't have the powuh to create repo/branch/whatever myself.

Lucide:
If the sunshine service is launched at startup and left idling in background, and the high performance setting is set on `main()`, does that mean the gpu will be in a constant high performance/power consumption state?

ns6089:
> If the sunshine service is launched at startup and left idling in background, and the high performance setting is set on `main()`, does that mean the gpu will be in a constant high performance/power consumption state?
> 
> Does the setting apply only on subsequent launches of sunshine?

A little bit of context. These settings are stored in `c:\ProgramData\NVIDIA Corporation\Drs\nvdrsdb0.bin`. After an application creates 3d context, it loads nvidia libraries and these libraries read that file in application memory space.

Since sunshine doesn't use any 3d contexts when not actively streaming, power states will not be constantly applied, and the setting will be applied even on the first launch (when streaming starts).

Lucide:
Oh, I see!
Does it also respect context "destruction"? i.e., does the driver unload the app settings when the stream ends?

ns6089:
> Oh, I see! Does it also respect context "destruction"? i.e., does the driver unload the app settings when the stream ends?

If it's power states no longer being applied, then yes. If it's settings being re-read on the next 3d context creation, I haven't tested it, but I don't see why they shouldn't. But since it tied to library load/unload, you might need to destroy every last 3d context for reload to happen.

Lucide:
It's the above, I was worrying about making the changes as stream-local as possible. So for the app profile this happens implicitly.

Does it make sense to move the base profile changes explicitly at stream start/end too?

promise I won't bother you no further😶‍🌫️

ns6089:
> Does it make sense to move the base profile changes explicitly at stream start/end too?

No worries, it's a valid question. We change the *base* profile because the api we use for capturing doesn't work well on opengl/vulkan games, and we need to change how their presentation works to sidestep it. And the game can be launched before streaming request comes.

third-party/nvapi-opensource/readme.txt
```diff
@@ -0,0 +1,3 @@
+Opensource version of nvapi
+Updates can be found at https://download.nvidia.com/XFree86/nvapi-open-source-sdk/
```
ReenigneArcher, line 2: Okay, I see it doesn't really seem to be available in any updated git repo (that I could find anyway). Could we just download and extract this with CMakeLists? I'm not really a fan of including other source code directly if it can be avoided.

https://cmake.org/cmake/help/latest/command/file.html#transfer
https://cmake.org/cmake/help/latest/command/file.html#archive-extract

ReenigneArcher (changes requested):

third-party/nvapi-opensource/readme.txt
ns6089: But let's look upon this objectively

What cons is this going to bring?
- We'll be effectively breaking repo history once they change the url, and it's going to happen sooner or later
- It can cause immense annoyance in places with spotty internet (hey, I like taking my laptop to the mountains)

What pros?
- I don't see any

ns6089: It's just, I think we're spending time trying to solve a problem that wasn't a problem to begin with. And if I start looking into that `archive-extract`, it will introduce actual problems that will require actual solving and more back and forth communication.

Lucide: If this acts as the regular nvapi drs then it's not even a real library at all, it's just a loader for the real library inside the driver.
It's made to be forward compatible forever, it's not something to keep updated.
The Nvidia drs dev guide is from 2011 and points this out.

src/platform/windows/nvprefs/undo_file.cpp
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  std::optional<undo_file_t>
+  undo_file_t::create_new_file(std::filesystem::path file_path) {
+    undo_file_t file;
+    file.file_handle.reset(CreateFileW(file_path.c_str(), GENERIC_WRITE | STANDARD_RIGHTS_ALL, 0, nullptr, CREATE_NEW, FILE_ATTRIBUTE_NORMAL, NULL));
+
+    if (file.file_handle) {
+      // give GENERIC_READ, GENERIC_WRITE and DELETE permissions to Users group
+      if (relax_permissions(file.file_handle.get()) != 0) {
```
cgutman, line 66: What's the thinking here for allowing the Users group to write this file (or really for any user/process besides us to write it)?

src/platform/windows/nvprefs/nvprefs_interface.cpp
```diff
@@ -0,0 +1,225 @@
...
+    }
+    unload();
+  }
+
+  bool
+  nvprefs_interface::load() {
+    if (!pimpl->loaded) {
+      // Check %ProgramData% variable, need it for storing undo file
+      wchar_t program_data_env[MAX_PATH];
+      auto get_env_result = GetEnvironmentVariableW(L"ProgramData", program_data_env, MAX_PATH);
```
cgutman, line 43: Maybe we should use `%LOCALAPPDATA%` instead of `%ProgramData%` to keep the undo state for installer and portable versions of Sunshine apart.

ns6089: The NVIDIA settings themselves are global and stored in %ProgramData% file. If we split out undo states, one can pollute the global state with no way for the other one to know about it.

src/platform/windows/nvprefs/undo_file.cpp
ns6089: It's really GENERIC_READ and DELETE that's is required logic-wise - read file, apply undo, delete file. But GENERIC_WRITE is needed for locking the file during the process.

"Us" is defined here by file path, not user or process. Since we have portable version that can be be run in parallel and as any user, and then can crash and leave the undo file with settings not reverted. If NVIDIA state is global and can be changed by anyone, then the undo file must be global too, I don't think there's a way around it.

third-party/nvapi-opensource/readme.txt
ns6089: That's the thing, the files in question are not code, but constants. And we don't really need to update them unless we would want to change some newly added parameters. Which we won't be needing for years, most likely ever. And if we can't find the parameter in question, that's the sure sign to update. Which takes 3 minutes since it's trivial and kept in one place. You will spend order of magnitudes more time on automation.

cgutman (approved):

cgutman:
It looks like we need to ignore the third-party files for clang-format. Can you try adding a `.clang-format-ignore` with `third-party/*` and see if that fixes it?

ReenigneArcher:
> 

Due to how the workflow finds the files, that probably won't work (assumption).

https://github.com/LizardByte/Sunshine/blob/50f353d183e2a6e22fd69f4320691c27d87c8b80/.github/workflows/cpp-lint.yml#L31

I like the idea though so will work on a way to implement that.

ReenigneArcher:
@cgutman This should work. https://github.com/LizardByte/Sunshine/pull/1449

PR #1364, opened 2023-06-11
fix(deps): replace libboost chrono and thread with standard chrono and thread

Era-Dorta:
## Description
Sunshine crashes on Ubuntu 22.04 due a missing lib boost dependency. This PR adds the missing dependency for the deb packages.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -285,7 +285,6 @@ jobs:
               libboost-locale1.71-dev \
               libboost-log1.71-dev \
               libboost-regex1.71-dev \
-              libboost-thread1.71-dev \
               libboost-program-options1.71-dev
 
             # Install cmake
@@ -311,7 +310,6 @@ jobs:
               libboost-filesystem-dev \
               libboost-locale-dev \
               libboost-log-dev \
-              libboost-thread-dev \
               libboost-program-options-dev
           fi
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -906,7 +906,6 @@ elseif(UNIX)
                 libboost-locale${Boost_VERSION}, \
                 libboost-log${Boost_VERSION}, \
                 libboost-program-options${Boost_VERSION}, \
-                libboost-thread${Boost_VERSION}, \
                 libcap2, \
                 libcurl4, \
                 libdrm2, \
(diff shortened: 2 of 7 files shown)


ReenigneArcher:
Thank you for the PR. I kind of think the solution is to switch to the standard chrono instead of boost... Unless I'm missing something.

Era-Dorta:
Thanks for lighting fast review. I didn't really look at the code much at first. I just got this error
`sunshine: error while loading shared libraries: libboost_chrono.so.1.74.0: cannot open shared object file: No such file or directory`
and it seemed like the obvious fix.

I now see that the code uses `std:chrono` and not `boost:chrono`. I haven't worked with C++ for a while, so I'm quite rusty at it. Does this mean that the binary somehow got built to link to `libboost` so the build environment for the deb packages is wrong?

ReenigneArcher:
I'm not quite sure how it linked to boost chrono as it's not even installed in the build environment.

https://github.com/LizardByte/Sunshine/blob/master/docker/ubuntu-20.04.dockerfile#L39

There might be a compile warning that doesn't fail the build and it is harder to detect.

But I think boost chrono was just used in 2 files. Probably the solution is to just switch those to standard chrono.

Era-Dorta:
Makes a lot more sense now, I've done as suggested in the last two commits.

Era-Dorta:
It looks like the boost threading doesn't like the std time. I've switched it to use std threading too. After a quick search, it seems like `libboost-thread` can be removed as a dependency, as this was the only place that was used.

Era-Dorta:
Sorry about the crashes with the CI, I've done a manual launch on my fork and it seems to work now https://github.com/Era-Dorta/Sunshine/actions/runs/5248057310

Era-Dorta:
Is there anything else I should do to move this forwards?

ReenigneArcher:
I just need to review it, which might take some time.

There were some ci failures (cancelled) for some reason.

Max-Might:
@ReenigneArcher Do you think this could be released as a hotfix? Version 0.20.0 is broken for a lot of people (me including) because of this issue.

ReenigneArcher:
We don't do hot fixes and this has not been merged yet. You can manually install the missing dependency and sunshine will work fine.

cgutman (approved):

PR #1384, opened 2023-06-21
Use X mouse even if /dev/uinput is present

john-peterson:
I can never get mouse to work in my cloud rig

In Debian it worked but no gamepad

Switched to ubuntu now gamepad work and mouse fail

If there was a program to monitor ALL INPUT LIVE and poll for new devices. Maybe I could know why

But alas this is what I have to do. Now the Mose and gamepad works from moonlight android

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -124,6 +124,18 @@ global_prep_cmd
 Controls
 --------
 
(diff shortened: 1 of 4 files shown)


ReenigneArcher:
I closed this PR due to an invalid template. If you would like it re-considered, please complete our PR template. Thank you.

PR #1427, opened 2023-07-07
Add standalone NVENC encoder

ns6089:
## Description
Add standalone NVENC encoder for reference frames invalidation right now. And possibly for VFR-like bitrate adjustment and more somewhere down the line. Windows version is fully functional . Linux cuda (and possibly opengl) support is out of scope for this PR, but can easily be done later (at least the encoder side),

### In later PRs
- Investigate and test the viability of "increased vbv" option. Ideally, P-frames should not steal bitrate budget from future frames, maybe this can be achieved with vbv offset in encoder.
- Investigate the viability of using multiple ref frames (L0 > 1) since we apply strict limits on the vbv. Maybe nvenc can intelligently pick blocks from previous frames if they have higher qp? Can't imagine this being free though.
- Run VMAF benchmark for const-qp mode (document the process in .md file), and pick default values for min-qp
- Check GFE default ref frame buffer values, particularly for h264 (level4 vs level5)
- New configuration page and documentation

## Type of Change

## Checklist

## Branch Updates

---
### Things to implement before merge (may be expanded)
- [x] Refactor colospace selection logic, it has too much duplication right now
- [x] Decide how to handle encoder ref frames buffer size, currently it's at 16 frames (dynamically based on resolution and framerate? through configuration?)
Result: use 5 ref frames **buffer**, which corresponds to DPB=5 in h264 terms, and DPB=6 in HEVC. Both are required minimums that must be supported by decoder profiles for given resolution.
Update: h265 level 4 only supports 4 ref frames, but in this case the client can set the limit, this level is very outdated nowadays
- [x] Investigate if NVENC can be told explicitly which ref frame to use, this can allow wider decoder support for ref frames invalidation.
Verdict: when L0 (forward prediction) list size is 1, it should always use last frame as reference,
- [x] Try to patch num_ref_frames in SPS header
Verdict: may be possible for h264, close to impossible for HEVC, either way too much hassle
- [x] Test encoder caps
NV_ENC_CAPS_SUPPORT_CABAC
NV_ENC_CAPS_WIDTH_MAX
NV_ENC_CAPS_HEIGHT_MAX
NV_ENC_CAPS_SUPPORT_CUSTOM_VBV_BUF_SIZE
NV_ENC_CAPS_SUPPORT_REF_PIC_INVALIDATION
NV_ENC_CAPS_SUPPORT_YUV444_ENCODE
NV_ENC_CAPS_SUPPORT_10BIT_ENCODE
NV_ENC_CAPS_SUPPORT_MULTIPLE_REF_FRAMES
NV_ENC_CAPS_SUPPORT_QPELMV
- [x] Switch to nv-codec-headers
- [x] Update Linux and MacOS platforms to new structures
- [x] ~~Use Peak-Signal-to-Noise-Ratio (Y-PSNR), Structural Similarity Index (Y-SSIM), and Video Multimethod Assessment Fusion (VMAF) for default min qp values~~ Not in this PR
- [x] ~~Add configuration page and documentation~~ Not in this PR

### From review
- [x] Look into entropyCodingMode encoding parameter
- [x] Proper cleanup in create_encoder()
- [x] Look into last_encoder_probe_supported_invalidate_ref_frames, if needs to be changed in multiple places
- [x] Lock h264 into High profile

### DIFF
--- a/src/nvenc/nvenc_base.cpp
+++ b/src/nvenc/nvenc_base.cpp
@@ -1,0 +1,542 @@
+#include "nvenc_base.h"
+
+#include "src/config.h"
+#include "src/utility.h"
+
+namespace {
+
+  GUID
+  quality_preset_guid_from_number(unsigned number) {
+    if (number > 7) number = 7;
+
+    switch (number) {
+      case 1:
+      default:
+        return NV_ENC_PRESET_P1_GUID;
+
+      case 2:
+        return NV_ENC_PRESET_P2_GUID;
+
+      case 3:
+        return NV_ENC_PRESET_P3_GUID;
+
+      case 4:
+        return NV_ENC_PRESET_P4_GUID;
+
+      case 5:
+        return NV_ENC_PRESET_P5_GUID;
+
+      case 6:
+        return NV_ENC_PRESET_P6_GUID;
+
+      case 7:
+        return NV_ENC_PRESET_P7_GUID;
+    }
+  };
+
+  bool
+  equal_guids(const GUID &guid1, const GUID &guid2) {
+    return std::memcmp(&guid1, &guid2, sizeof(GUID)) == 0;
+  }
+
+  auto
+  quality_preset_string_from_guid(const GUID &guid) {
+    if (equal_guids(guid, NV_ENC_PRESET_P1_GUID)) {
+      return "P1";
+    }
+    if (equal_guids(guid, NV_ENC_PRESET_P2_GUID)) {
+      return "P2";
+    }
+    if (equal_guids(guid, NV_ENC_PRESET_P3_GUID)) {
+      return "P3";
+    }
+    if (equal_guids(guid, NV_ENC_PRESET_P4_GUID)) {
+      return "P4";
+    }
+    if (equal_guids(guid, NV_ENC_PRESET_P5_GUID)) {
+      return "P5";
+    }
+    if (equal_guids(guid, NV_ENC_PRESET_P6_GUID)) {
+      return "P6";
+    }
+    if (equal_guids(guid, NV_ENC_PRESET_P7_GUID)) {
+      return "P7";
+    }
+    return "Unknown";
+  }
+
+}  // namespace
+
+namespace nvenc {
+
+  nvenc_base::nvenc_base(NV_ENC_DEVICE_TYPE device_type, void *device):
+      device_type(device_type),
+      device(device) {
+  }
+
+  nvenc_base::~nvenc_base() {
+    // Use destroy_encoder() instead
+  }
+
+  bool
+  nvenc_base::create_encoder(const nvenc_config &config, const video::config_t &client_config, const nvenc_colorspace_t &colorspace, NV_ENC_BUFFER_FORMAT buffer_format) {
+    if (!nvenc && !init_library()) return false;
+
+    if (encoder) destroy_encoder();
+    auto fail_guard = util::fail_guard([this] { destroy_encoder(); });
+
+    encoder_params.width = client_config.width;
+    encoder_params.height = client_config.height;
+    encoder_params.buffer_format = buffer_format;
+    encoder_params.rfi = true;
+
+    NV_ENC_OPEN_ENCODE_SESSION_EX_PARAMS session_params = { NV_ENC_OPEN_ENCODE_SESSION_EX_PARAMS_VER };
+    session_params.device = device;
+    session_params.deviceType = device_type;
+    session_params.apiVersion = NVENCAPI_VERSION;
+    if (nvenc_failed(nvenc->nvEncOpenEncodeSessionEx(&session_params, &encoder))) {
+      BOOST_LOG(error) << "NvEncOpenEncodeSessionEx failed";
+      return false;
+    }
+
+    uint32_t encode_guid_count = 0;
+    if (nvenc_failed(nvenc->nvEncGetEncodeGUIDCount(encoder, &encode_guid_count))) {
+      BOOST_LOG(error) << "NvEncGetEncodeGUIDCount failed: " << last_error_string;
+      return false;
+    };
+
+    std::vector<GUID> encode_guids(encode_guid_count);
+    if (nvenc_failed(nvenc->nvEncGetEncodeGUIDs(encoder, encode_guids.data(), encode_guids.size(), &encode_guid_count))) {
+      BOOST_LOG(error) << "NvEncGetEncodeGUIDs failed: " << last_error_string;
+      return false;
+    }
+
+    NV_ENC_INITIALIZE_PARAMS init_params = { NV_ENC_INITIALIZE_PARAMS_VER };
+
+    switch (client_config.videoFormat) {
+      case 0:
+        // H.264
+        init_params.encodeGUID = NV_ENC_CODEC_H264_GUID;
+        break;
+
+      case 1:
+        // HEVC
+        init_params.encodeGUID = NV_ENC_CODEC_HEVC_GUID;
+        break;
+
+      default:
+        BOOST_LOG(error) << "NvEnc: unknown video format " << client_config.videoFormat;
+        return false;
+    }
+
+    {
+      auto search_predicate = [&](const GUID &guid) {
+        return equal_guids(init_params.encodeGUID, guid);
+      };
+      if (std::find_if(encode_guids.begin(), encode_guids.end(), search_predicate) == encode_guids.end()) {
+        BOOST_LOG(error) << "NvEnc: encoding format is not supported by the gpu";
+        return false;
+      }
+    }
+
+    auto get_encoder_cap = [&](NV_ENC_CAPS cap) {
+      NV_ENC_CAPS_PARAM param = { NV_ENC_CAPS_PARAM_VER, cap };
+      int value = 0;
+      nvenc->nvEncGetEncodeCaps(encoder, init_params.encodeGUID, &param, &value);
+      return value;
+    };
+
+    auto buffer_is_10bit = [&]() {
+      return buffer_format == NV_ENC_BUFFER_FORMAT_YUV420_10BIT || buffer_format == NV_ENC_BUFFER_FORMAT_YUV444_10BIT;
+    };
+
+    auto buffer_is_yuv444 = [&]() {
+      return buffer_format == NV_ENC_BUFFER_FORMAT_YUV444 || buffer_format == NV_ENC_BUFFER_FORMAT_YUV444_10BIT;
+    };
+
+    {
+      auto supported_width = get_encoder_cap(NV_ENC_CAPS_WIDTH_MAX);
+      auto supported_height = get_encoder_cap(NV_ENC_CAPS_HEIGHT_MAX);
+      if (encoder_params.width > supported_width || encoder_params.height > supported_height) {
+        BOOST_LOG(error) << "NvEnc: gpu max encode resolution " << supported_width << "x" << supported_height << ", requested " << encoder_params.width << "x" << encoder_params.height;
+        return false;
+      }
+    }
+
+    if (buffer_is_10bit() && !get_encoder_cap(NV_ENC_CAPS_SUPPORT_10BIT_ENCODE)) {
+      BOOST_LOG(error) << "NvEnc: gpu doesn't support 10-bit encode";
+      return false;
+    }
+
+    if (buffer_is_yuv444() && !get_encoder_cap(NV_ENC_CAPS_SUPPORT_YUV444_ENCODE)) {
+      BOOST_LOG(error) << "NvEnc: gpu doesn't support YUV444 encode";
+      return false;
+    }
+
+    if (async_event_handle && !get_encoder_cap(NV_ENC_CAPS_ASYNC_ENCODE_SUPPORT)) {
+      BOOST_LOG(warning) << "NvEnc: gpu doesn't support async encode";
+      async_event_handle = nullptr;
+    }
+
+    encoder_params.rfi = get_encoder_cap(NV_ENC_CAPS_SUPPORT_REF_PIC_INVALIDATION);
+
+    init_params.presetGUID = quality_preset_guid_from_number(config.quality_preset);
+    init_params.tuningInfo = NV_ENC_TUNING_INFO_ULTRA_LOW_LATENCY;
+    init_params.enablePTD = 1;
+    init_params.enableEncodeAsync = async_event_handle ? 1 : 0;
+    init_params.enableWeightedPrediction = config.weighted_prediction && get_encoder_cap(NV_ENC_CAPS_SUPPORT_WEIGHTED_PREDICTION);
+
+    init_params.encodeWidth = encoder_params.width;
+    init_params.darWidth = encoder_params.width;
+    init_params.encodeHeight = encoder_params.height;
+    init_params.darHeight = encoder_params.height;
+    init_params.frameRateNum = client_config.framerate;
+    init_params.frameRateDen = 1;
+
+    NV_ENC_PRESET_CONFIG preset_config = { NV_ENC_PRESET_CONFIG_VER, { NV_ENC_CONFIG_VER } };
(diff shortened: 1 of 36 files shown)


src/nvenc/nvenc_base.cpp
```diff
@@ -0,0 +1,323 @@
...
+      BOOST_LOG(error) << "NvEncCreateBitstreamBuffer failed: " << nvenc->nvEncGetLastErrorString(encoder);
+      return false;
+    }
+    output_bitstream = create_bitstream_buffer.bitstreamBuffer;
+
+    if (!create_input_buffer()) {
+      return false;
+    }
+
+    BOOST_LOG(info) << "Created NvENC encoder at " << quality_preset_string_from_guid(init_params.presetGUID) << (config.vbv_multiplier > 1.0f ? " with modified VDV" : "");
```
cgutman, line 215: typo: `VDV` -> `VBV`

src/platform/windows/display_vram.cpp
```diff
@@ -830,6 +755,202 @@ namespace platf::dxgi {
...
+        }
+        set_colorspace(color_space, color_range);
+      }
+
+      nvenc = std::make_unique<nvenc::nvenc_d3d11>(base.device.get(), client_config.width, client_config.height, buffer_format);
+
+      auto nvenc_colorspace = nvenc::nvenc_colorspace_from_rtsp_csc(client_config.encoderCscMode, client_config.dynamicRange && display->is_hdr());
+
+      nvenc::nvenc_config nvenc_config;
+      if (config::video.nv.nv_preset) nvenc_config.quality_preset = *config::video.nv.nv_preset - 11;
```
cgutman, line 906: What's the `- 11` for?

src/platform/windows/display_vram.cpp
```diff
@@ -830,6 +755,202 @@ namespace platf::dxgi {
...
+    encoded_frame
+    encode_frame(uint64_t frame_index) override {
+      auto result = nvenc->encode_frame(frame_index, force_idr);
+      force_idr = false;
+      return result;
+    }
+
+    int
+    convert(platf::img_t &img_base) override {
+      return base.convert(img_base);
```
cgutman, line 922: Nice refactoring. This will make it pretty simple if we want to have NVENC directly consume RGB frames or if we need to go that route with AMF for AMD cards.

src/nvenc/nvenc_base.cpp
```diff
@@ -0,0 +1,323 @@
...
+        // H.264
+        auto &format_config = enc_config.encodeCodecConfig.h264Config;
+        set_common_format_config(format_config);
+        if (client_config.numRefFrames > 0) {
+          format_config.maxNumRefFrames = client_config.numRefFrames;
+        }
+        else {
+          format_config.maxNumRefFrames = 5;
+        }
+        format_config.numRefL0 = NV_ENC_NUM_REF_FRAMES_1;
```
cgutman, line 173: We should also populate `entropyCodingMode` here

src/nvenc/nvenc_base.cpp
```diff
@@ -0,0 +1,323 @@
...
+    switch (client_config.videoFormat) {
+      case 0: {
+        // H.264
+        auto &format_config = enc_config.encodeCodecConfig.h264Config;
+        set_common_format_config(format_config);
+        if (client_config.numRefFrames > 0) {
+          format_config.maxNumRefFrames = client_config.numRefFrames;
+        }
+        else {
+          format_config.maxNumRefFrames = 5;
```
cgutman, line 171: It looks like 0 is default. What does NVENC pick if you pass 0 here?

src/nvenc/nvenc_base.cpp
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    width = client_config.width;
+    height = client_config.height;
+
+    NV_ENC_INITIALIZE_PARAMS init_params = { NV_ENC_INITIALIZE_PARAMS_VER };
+    init_params.maxEncodeWidth = max_width;
+    init_params.maxEncodeHeight = max_height;
+    init_params.presetGUID = quality_preset_guid_from_number(config.quality_preset);
+    init_params.tuningInfo = NV_ENC_TUNING_INFO_ULTRA_LOW_LATENCY;
```
cgutman, line 96: Can you plumb in this `tuningInfo` value from the config? We expose this option to users as `config::video.nv.nv_tune`.

src/nvenc/nvenc_base.cpp
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    NV_ENC_PRESET_CONFIG preset_config = { NV_ENC_PRESET_CONFIG_VER, { NV_ENC_CONFIG_VER } };
+    if (nvenc->nvEncGetEncodePresetConfigEx(encoder, init_params.encodeGUID, init_params.presetGUID, init_params.tuningInfo, &preset_config) != NV_ENC_SUCCESS) {
+      BOOST_LOG(error) << "NvEncGetEncodePresetConfigEx failed: " << nvenc->nvEncGetLastErrorString(encoder);
+      return false;
+    }
+
+    NV_ENC_CONFIG enc_config = preset_config.presetCfg;
+    enc_config.profileGUID = NV_ENC_CODEC_PROFILE_AUTOSELECT_GUID;  // TODO: select specific profile?
```
cgutman, line 125: Yeah, we should select specific profiles. The specific values should be:
- `videoFormat == 0` -> `NV_ENC_H264_PROFILE_HIGH_GUID `
- `videoFormat == 1 && dynamicRange == 0` -> `NV_ENC_HEVC_PROFILE_MAIN_GUID`
- `videoFormat == 1 && dynamicRange == 1` -> `NV_ENC_HEVC_PROFILE_MAIN10_GUID`

src/nvenc/nvenc_base.cpp
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    BOOST_LOG(info) << "Created NvENC encoder at " << quality_preset_string_from_guid(init_params.presetGUID) << (config.vbv_multiplier > 1.0f ? " with modified VDV" : "");
+
+    return true;
+  }
+
+  nvenc_encoded_frame
+  nvenc_base::encode_frame(uint64_t frame_index, bool force_idr) {
+    if (!encoder || !output_bitstream) {
```
cgutman, line 222: Can this actually happen or are you just being cautious?

src/video.cpp
```diff
@@ -2003,6 +2172,8 @@ namespace video {
             active_hevc_mode = 0;
           }
 
+          last_encoder_probe_supported_invalidate_ref_frames = encoder->details->supports_invalidate_ref_frames;
```
cgutman, line 2175: We need to set this in the other paths that assign `chosen_encoder` too.

ns6089: Will look into it.

src/nvenc/nvenc_base.cpp
cgutman: H.264 Main or Baseline should work, but I think we should stick to H.264 High since that's what all the other encoders do. We manually specify the profile for the FFmpeg-based encoders to ensure we always get what we expect.

ns6089:
@cgutman, need your opinion on colorspace refactoring https://github.com/LizardByte/Sunshine/pull/1427/commits/a2f34ab3de3d8f18e13ae9960d76217b52ec0d8a. When you have time, it's not blocking me.

ns6089:
> @cgutman, need your opinion on colorspace refactoring [a2f34ab](https://github.com/LizardByte/Sunshine/commit/a2f34ab3de3d8f18e13ae9960d76217b52ec0d8a). When you have time, it's not blocking me.

Or better wait until I finish the Linux part, it approaches things slightly differently and might need adjustments.

src/nvenc/nvenc_base.cpp
ns6089: Done.

src/nvenc/nvenc_base.cpp
ns6089: Forced H264 to High, the rest should be fine on auto.

ns6089:
New options page (not yet pushed to this PR). Some things might still be slightly reworded, and need to run a few tests for better default QP values. But conceptually it should be done, and I think I will be able to backport it to ffmpeg nvenc backend so we don't end up with two pages.
2023-07-18 15_21_35-Sunshine — Mozilla Firefox

src/nvenc/nvenc_base.cpp
ns6089: Close to final concept image with the new options page https://github.com/LizardByte/Sunshine/pull/1427#issuecomment-1640116140

src/platform/linux/vaapi.cpp
```diff
@@ -398,8 +398,8 @@ namespace va {
     }
 
     void
-    set_colorspace(std::uint32_t colorspace, std::uint32_t color_range) override {
-      sws.set_colorspace(colorspace, color_range);
+    apply_colorspace() override {
```
cgutman, line 401: This function takes a parameter, right?

cgutman: Nevermind

cgutman:
LGTM. Let's try to get this refactoring in soon if possible. I've got some AV1 and client-side cursor changes that I'd like to work on, but I don't want to stomp all over your work.

If you can back out any debugging changes or WIP test changes (the scheduling priority revert?), I can test this across all my systems and we can get it merged.

ns6089:
> LGTM. Let's try to get this refactoring in soon if possible. I've got some AV1 and client-side cursor changes that I'd like to work on, but I don't want to stomp all over your work.
> 
> If you can back out any debugging changes or WIP test changes (the scheduling priority revert?), I can test this across all my systems and we can get it merged.

Understandable. Wired the encoder to use the existing configuration page, since some new options pretty much depend on whether I can make it work with realtime priority (and priority selection itself will add another option).

Haven't tested native HDR (only SDR in BT.2020) and CUDA path on Linux (didn't make drastic changes, but there's always a chance). Other than that, should be alright to merge in its current state. The rest of the features shouldn't produce conflicts, and can be done in later PRs.

third-party/nv-codec-headers
ReenigneArcher, line 1: This [commit](https://github.com/FFmpeg/nv-codec-headers/commit/855f8263d97bbdcaeabaaaa2997e1ccad7c52dc3) is not on the branch we are tracking -> https://github.com/FFmpeg/nv-codec-headers/tree/sdk/11.1.

Can we change the branch to sdk/12.0?

ns6089: I think simply doing this https://github.com/LizardByte/Sunshine/pull/1427/commits/eafa29788da80a2961833fc6d9edeb7ee4549beb is enough

ns6089: > changing the branch in the got modules file is not enough. You also must checkout the latest commit on that branch and commit it.

Alright. Rebased all of that into one commit https://github.com/LizardByte/Sunshine/pull/1427/commits/7a336ca3593a72a82535f7ff49fed4de7e7c964c

ns6089: And it now fails the build...

ns6089: Turns out I used one status constant from 12.1 in pretty debug print function. Should be fine now.

src/nvenc/nvenc_base.cpp
```diff
@@ -0,0 +1,541 @@
...
+    return encoded_frame;
+  }
+
+  bool
+  nvenc_base::invalidate_ref_frames(uint64_t first_frame, uint64_t last_frame) {
+    if (!encoder || !encoder_params.rfi) return false;
+
+    if (last_frame < first_frame ||
+        encoder_state.last_encoded_frame_index < first_frame ||
+        encoder_state.last_encoded_frame_index > first_frame + 100) {
```
cgutman, line 461: That last condition (`encoder_state.last_encoded_frame_index > first_frame + 100`) is actually possible if we have a period of severe packet loss. We will continue expanding our RFI range until we successfully receive a recovery frame from the host. We shouldn't log an error for that case.

src/nvenc/nvenc_utils.cpp
```diff
@@ -0,0 +1,74 @@
...
+        colorspace.matrix = NV_ENC_VUI_MATRIX_COEFFS_BT709;
+        break;
+
+      case video::colorspace_e::bt2020:
+        // Rec. 2020 with ST 2084 perceptual quantizer
+        colorspace.primaries = NV_ENC_VUI_COLOR_PRIMARIES_BT2020;
+        colorspace.tranfer_function = NV_ENC_VUI_TRANSFER_CHARACTERISTIC_BT2020_10;
+        colorspace.matrix = NV_ENC_VUI_MATRIX_COEFFS_BT2020_NCL;
+        break;
+    }
```
cgutman, line 67: These cases look mixed up.

src/video_colorspace.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    // When HDR is requested, that overrides the SDR colorspace selection
+    if (hdr_requested) {
+      if (hdr_display) {
+        // Rec. 2020 with ST 2084 perceptual quantizer
+        colorspace.colorspace = colorspace_e::bt2020;
+      }
+      else {
+        // Rec. 2020
+        colorspace.colorspace = colorspace_e::bt2020sdr;
```
cgutman, line 43: Moonlight expects the host to encode using its requested colorspace even when streaming in 10-bit modes, so the above case should be:
```
      if (hdr_requested && hdr_display) {
        // Rec. 2020 with ST 2084 perceptual quantizer
        colorspace.colorspace = colorspace_e::bt2020;
      }
      else {
      ...

src/nvenc/nvenc_utils.cpp
ns6089: Nice catch, this is why huge commits are evil :(

src/nvenc/nvenc_base.cpp
ns6089: The reason why it's there is because nvenc rfi api doesn't work with ranges, but individual frames. And these calls can get expensive really fast if we don't limit them to some reasonable number. I will add this to my todo, some more intelligent logic can be written based on previously requested range, but it shouldn't be a blocker since it's just a log message that happens in very specific circumstances.

src/video_colorspace.cpp
cgutman: Rec. 601 is probably correct. We generally request that as a default for historical reasons, except for some GPUs that only behave properly with Rec. 709. I'll take a look at the crash.

Btw, GFE defaults to Rec. 601 if the colorspace isn't recognized. It's just academic because Moonlight will not pass an invalid color space, but just mentioning it since you added a codepath for that.

src/nvenc/nvenc_utils.cpp
cgutman: I think the BT.2020 SDR and HDR cases are still reversed.

src/nvenc/nvenc_base.cpp
cgutman: That's fine with me. I just mentioned it because I saw the logs when I was performing RFI and packet loss tests with this branch.

src/nvenc/nvenc_utils.cpp
ns6089: Yep. and I don't have an excuse this time :(

src/video_colorspace.cpp
ns6089: There was some misunderstanding on how 10-bit encoding is activated.
Pushed a fix https://github.com/LizardByte/Sunshine/pull/1427/commits/93d4b27dfdb934bc6114a163c873a27cda16456f

cgutman:
The latest changes look good. Once you squash the fixup commits and rebase, I'll do a final testing pass on my local machines and we can get this in.

ns6089:
Squashed and rebased. Also added small "fix" for nvapi, it doesn't affect anything since we're strictly x64, so no point in opening full pull request for it.

cgutman:
Everything looks good in my tests:
- HDR with and without NVENC
- NvFBC on Linux
- All valid colorspace and color range combos

PR #1430, opened 2023-07-07
Expose Connection environment variables to the app start

TheElixZammuto:
## Description
When starting apps, some environment variables releated to the connection are passed to the commands (pre/detatched/do)
Instead of #1096 that relies on text replacement, this uses the environment variables system, which should be more robust and secure (Although I haven't tested cmd injection techinques). I've also documented inside the Web UI which vars are exposed and how to consume them in a standard cmd command on Windows (other platform suggestions are welcome!)

This should make scripts more elegant instead of relying on parsing log files. 

### Screenshot
env_vars

## Type of Change

## Checklist

## Branch Updates


src/process.cpp
```diff
@@ -120,6 +120,16 @@ namespace proc {
...
 
+    //Add Stream-specific environment variables
+    _env["SUNSHINE_APP_ID"] = std::to_string(_app_id);
+    _env["SUNSHINE_APP_NAME"] = _app.name;
+    _env["SUNSHINE_CLIENT_WIDTH"] = std::to_string(launch_session.width);
+    _env["SUNSHINE_CLIENT_HEIGHT"] = std::to_string(launch_session.height);
+    _env["SUNSHINE_CLIENT_FPS"] = std::to_string(launch_session.fps);
+    _env["SUNSHINE_CLIENT_HDR"] = launch_session.enable_hdr ? "true" : "false";
+    _env["SUNSHINE_CLIENT_GCMAP"] = std::to_string(launch_session.gcmap);
+    _env["SUNSHINE_CLIENT_HOST_AUDIO"] = launch_session.host_audio ? "true" : "false";
```
cgutman, line 131: Did you mean to expose `enable_sops` too?

src/nvhttp.cpp
```diff
@@ -17,6 +17,7 @@
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <boost/property_tree/xml_parser.hpp>
+#include<string>
```
cgutman, line 20: `#include <string>

src/nvhttp.cpp
```diff
@@ -267,12 +268,32 @@ namespace nvhttp {
...
     launch_session.gcm_key = util::from_hex<crypto::aes_t>(get_arg(args, "rikey"), true);
+    std::stringstream mode = std::stringstream(get_arg(args,"mode"));
+    //Split mode by the char "x", to populate width/height/fps
+    int x = 0;
+    std::string segment;
+    while(std::getline(mode, segment, 'x'))
+    {
+      if(x == 0)launch_session.width = stoi(segment);
+      if(x == 1)launch_session.height = stoi(segment);
+      if(x == 2)launch_session.fps = stoi(segment);
```
cgutman, line 279: Please adjust spacing around `if` and `while` to be consistent with the rest of the file.

src_assets/common/assets/web/apps.html
```diff
@@ -302,6 +302,21 @@ <h4 class="modal-title">Covers Found</h4>
...
+        <div class="form-text"><b>About Environment Variables: </b> All commands get these environment variables by default: </div>
+        <table>
+          <tr><td><b>Var Name</b></td><td><b></b></td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_APP_ID</td><td>App ID</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_APP_NAME</td><td>App Name</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_WIDTH</td><td>The Width requested by the client</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_HEIGHT</td><td>The Height requested by the client</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_FPS</td><td>The FPS requested by the client</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_HDR</td><td>(true/false) if HDR is enabled by the client</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_GCMAP</td><td>(int) the requested gamepad mask</td></tr>
```
cgutman, line 315: We probably want to give the user a little more than "the requested gamepad mask" for documentation of the GCMAP parameter. Maybe it makes sense to just expose a gamepad count instead?

src/nvhttp.cpp
```diff
@@ -267,12 +268,32 @@ namespace nvhttp {
...
+    }
+    launch_session.unique_id = (get_arg(args,"uniqueid"));
+    launch_session.uuid = (get_arg(args,"uuid"));
+    launch_session.appid = util::from_view(get_arg(args,"appid"));
+    launch_session.enable_sops = util::from_view(get_arg(args,"sops"));
+    launch_session.gcmap = util::from_view(get_arg(args,"gcmap"));
+    launch_session.enable_hdr = 0;
+    if(args.find("enableHdr"s) != std::end(args)){
+      launch_session.enable_hdr = util::from_view(get_arg(args,"enableHdr"));
+    }
```
cgutman, line 290: Should we also expose the audio configuration (maybe in a friendly form like `2.0`/`5.1`/7.1`)?

src/process.cpp
TheElixZammuto: Is sops equivalent to the option "Optimize game settings for streaming"? If so I guess there's only useful if exposed as an env, since Sunshine does not use this field

src/nvhttp.cpp
TheElixZammuto: Since the system uses the parameters passed by the "/launch" endpoint. I have a `surroundAudioInfo` param. How should the parameter be interpreted?

cgutman: The bottom 16-bits is a channel count, so we could just expose that. 

https://github.com/moonlight-stream/moonlight-common-c/blob/49fef0383020272941c4704d2ecf3e1e71174092/src/Limelight.h#L214-L217

src/process.cpp
```diff
@@ -120,6 +120,29 @@ namespace proc {
...
+    _env["SUNSHINE_CLIENT_ENABLE_SOPS"] = launch_session.enable_sops ? "true" : "false";
+    int channelCount = launch_session.surround_info & (65535);
+    switch (channelCount) {
+      case 2:
+        _env["SUNSHINE_CLIENT_AUDIO_CONFIGURATION"] = "2.1";
```
cgutman, line 136: This is regular Stereo (2.0). It doesn't have a LFE channel.

ReenigneArcher (changes requested):
Should we add something to the docs about environment variables as well?

We could also update these do command examples to use the environment variables: https://github.com/LizardByte/Sunshine/blob/nightly/docs/source/about/app_examples.rst?plain=1#L131-L183

Linux Example: https://docs.lizardbyte.dev/projects/sunshine/en/nightly/about/app_examples.html#changing-resolution-and-refresh-rate-linux-x11

macOS Example: https://docs.lizardbyte.dev/projects/sunshine/en/nightly/about/app_examples.html#changing-resolution-and-refresh-rate-macos

And we could remove this section (don't think it will be useful after this is merged): https://github.com/LizardByte/Sunshine/blob/nightly/docs/source/about/app_examples.rst?plain=1#L188-L189

src_assets/common/assets/web/apps.html
```diff
@@ -302,6 +302,23 @@ <h4 class="modal-title">Covers Found</h4>
...
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_FPS</td><td>The FPS requested by the client</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_HDR</td><td>(true/false) if HDR is enabled by the client</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_GCMAP</td><td>(int) the requested gamepad mask, in a bitset/bitfield format</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_HOST_AUDIO</td><td>(true/false) if the client has requested host audio</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_ENABLE_SOPS</td><td>(true/false) if the client has requested the option to optimize the game for optimal streaming</td></tr>
+          <tr><td style="font-family: monospace">SUNSHINE_CLIENT_AUDIO_CONFIGURATION</td><td>The Audio Configuration requested by the client (2.0/5.1/7.1)</td></tr>
+        </table>
+        <div class="form-text"><b>Example for Windows - QRes for Resolution Automation:</b> <pre>C:\Windows\System32\cmd.exe /v /c &lt;qres path&gt;\QRes.exe /X:%SUNSHINE_CLIENT_WIDTH% /Y:%SUNSHINE_CLIENT_HEIGHT%</pre></div>
```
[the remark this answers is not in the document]
TheElixZammuto: In my tests without the /v option using cmd the expansion wouldn't work

TheElixZammuto:
> Specifically there is a bunch of moonlight variables logged when the log level as set to debug, which gives a lot of useful information. I’ve been wanting to make those accessible but it appears those are logged after the stream starts, not beforehand.

Yup, the app starts when a /launch command is issued on the client side, so we can only access the variables sent using this endpoint. Do you have any specific variable you want to be passed?

src/nvhttp.cpp
```diff
@@ -267,12 +268,32 @@ namespace nvhttp {
 
     launch_session.host_audio = host_audio;
     launch_session.gcm_key = util::from_hex<crypto::aes_t>(get_arg(args, "rikey"), true);
+    std::stringstream mode = std::stringstream(get_arg(args, "mode"));
+    // Split mode by the char "x", to populate width/height/fps
+    int x = 0;
+    std::string segment;
+    while (std::getline(mode, segment, 'x')) {
+      if (x == 0) launch_session.width = stoi(segment);
```
ns6089, line 276: `std::stoi()` throws when the conversion can't be performed, but not sure if it can happen here. Maybe use `atoi(segment.c_str())` from C since it can't throw.

src_assets/common/assets/web/apps.html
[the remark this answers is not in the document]
ReenigneArcher: Interesting. The help mentions you need to use `!variable! instead of `%variable%` when using variable expansion. Fine with leaving the `/V` though.

Just for reference, here is the entire help message.

```txt
>cmd /?
Starts a new instance of the Windows command interpreter

CMD [/A | /U] [/Q] [/D] [/E:ON | /E:OFF] [/F:ON | /F:OFF] [/V:ON | /V:OFF]
    [[/S] [/C | /K] string]

/C      Carries out the command specified by string and then terminates
/K      Carries out the command specified by string but remains
/S      Modifies the treatment of string after /C or /K (see below)
/Q      Turns echo off
/D      Disable execution of AutoRun commands from registry (see below)
/A      Causes the output of internal commands to a pipe or file to be ANSI
/U      Causes the output of internal commands to a pipe or file to be
        Unicode
/T:fg   Sets the foreground/background colors (see COLOR /? for more info)
/E:ON   Enable command extensions (see below)
/E:OFF  Disable command extensions (see below)
/F:ON   Enable file and directory name completion characters (see below)
/F:OFF  Disable file and directory name completion characters (see below)
/V:ON   Enable delayed environment variable expansion using ! as the
        delimiter. For example, /V:ON would allow !var! to expand the
        variable var at execution time.  The var syntax expands variables
        at input time, which is quite a different thing when inside of a FOR
        loop.
/V:OFF  Disable delayed environment expansion.

Note that multiple commands separated by the command separator '&&'
are accepted for string if surrounded by quotes.  Also, for compatibility
reasons, /X is the same as /E:ON, /Y is the same as /E:OFF and /R is the
same as /C.  Any other switches are ignored.

If /C or /K is specified, then the remainder of the command line after
the switch is processed as a command line, where the following logic is
used to process quote (") characters:

    1.  If all of the following conditions are met, then quote characters
        on the command line are preserved:

        - no /S switch
        - exactly two quote characters
        - no special characters between the two quote characters,
          where special is one of: &<>()@^|
        - there are one or more whitespace characters between the
          two quote characters
        - the string between the two quote characters is the name
          of an executable file.

    2.  Otherwise, old behavior is to see if the first character is
        a quote character and if so, strip the leading character and
        remove the last quote character on the command line, preserving
        any text after the last quote character.

If /D was NOT specified on the command line, then when CMD.EXE starts, it
looks for the following REG_SZ/REG_EXPAND_SZ registry variables, and if
either or both are present, they are executed first.

    HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor\AutoRun

        and/or

    HKEY_CURRENT_USER\Software\Microsoft\Command Processor\AutoRun

Command Extensions are enabled by default.  You may also disable
extensions for a particular invocation by using the /E:OFF switch.  You
can enable or disable extensions for all invocations of CMD.EXE on a
machine and/or user logon session by setting either or both of the
following REG_DWORD values in the registry using REGEDIT.EXE:

    HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor\EnableExtensions

        and/or

    HKEY_CURRENT_USER\Software\Microsoft\Command Processor\EnableExtensions

to either 0x1 or 0x0.  The user specific setting takes precedence over
the machine setting.  The command line switches take precedence over the
registry settings.

In a batch file, the SETLOCAL ENABLEEXTENSIONS or DISABLEEXTENSIONS arguments
takes precedence over the /E:ON or /E:OFF switch. See SETLOCAL /? for details.

The command extensions involve changes and/or additions to the following
commands:

    DEL or ERASE
    COLOR
    CD or CHDIR
    MD or MKDIR
    PROMPT
    PUSHD
    POPD
    SET
    SETLOCAL
    ENDLOCAL
    IF
    FOR
    CALL
    SHIFT
    GOTO
    START (also includes changes to external command invocation)
    ASSOC
    FTYPE

To get specific details, type commandname /? to view the specifics.

Delayed environment variable expansion is NOT enabled by default.  You
can enable or disable delayed environment variable expansion for a
particular invocation of CMD.EXE with the /V:ON or /V:OFF switch.  You
can enable or disable delayed expansion for all invocations of CMD.EXE on a
machine and/or user logon session by setting either or both of the
following REG_DWORD values in the registry using REGEDIT.EXE:

    HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor\DelayedExpansion

        and/or

    HKEY_CURRENT_USER\Software\Microsoft\Command Processor\DelayedExpansion

to either 0x1 or 0x0.  The user specific setting takes precedence over
the machine setting.  The command line switches take precedence over the
registry settings.

In a batch file the SETLOCAL ENABLEDELAYEDEXPANSION or DISABLEDELAYEDEXPANSION
arguments takes precedence over the /V:ON or /V:OFF switch. See SETLOCAL /?
for details.

If delayed environment variable expansion is enabled, then the exclamation
character can be used to substitute the value of an environment variable
at execution time.

You can enable or disable file name completion for a particular
invocation of CMD.EXE with the /F:ON or /F:OFF switch.  You can enable
or disable completion for all invocations of CMD.EXE on a machine and/or
user logon session by setting either or both of the following REG_DWORD
values in the registry using REGEDIT.EXE:

    HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor\CompletionChar
    HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor\PathCompletionChar

        and/or

    HKEY_CURRENT_USER\Software\Microsoft\Command Processor\CompletionChar
    HKEY_CURRENT_USER\Software\Microsoft\Command Processor\PathCompletionChar

with the hex value of a control character to use for a particular
function (e.g.  0x4 is Ctrl-D and 0x6 is Ctrl-F).  The user specific
settings take precedence over the machine settings.  The command line
switches take precedence over the registry settings.

If completion is enabled with the /F:ON switch, the two control
characters used are Ctrl-D for directory name completion and Ctrl-F for
file name completion.  To disable a particular completion character in
the registry, use the value for space (0x20) as it is not a valid
control character.

Completion is invoked when you type either of the two control
characters.  The completion function takes the path string to the left
of the cursor appends a wild card character to it if none is already
present and builds up a list of paths that match.  It then displays the
first matching path.  If no paths match, it just beeps and leaves the
display alone.  Thereafter, repeated pressing of the same control
character will cycle through the list of matching paths.  Pressing the
Shift key with the control character will move through the list
backwards.  If you edit the line in any way and press the control
character again, the saved list of matching paths is discarded and a new
one generated.  The same occurs if you switch between file and directory
name completion.  The only difference between the two control characters
is the file completion character matches both file and directory names,
while the directory completion character only matches directory names.
If file completion is used on any of the built in directory commands
(CD, MD or RD) then directory completion is assumed.

The completion code deals correctly with file names that contain spaces
or other special characters by placing quotes around the matching path.
Also, if you back up, then invoke completion from within a line, the
text to the right of the cursor at the point completion was invoked is
discarded.

The special characters that require quotes are:
     <space>
     &()[]{}^=;!'+,`~

```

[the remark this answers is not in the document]
TheElixZammuto: I've tried again and I can assert that actually using `cmd /C <path>QRes.exe /X:%SUNSHINE_CLIENT_WIDTH% /Y:%SUNSHINE_CLIENT_HEIGHT%` works as expected

Nonary:
> > Specifically there is a bunch of moonlight variables logged when the log level as set to debug, which gives a lot of useful information. I’ve been wanting to make those accessible but it appears those are logged after the stream starts, not beforehand.
> 
> Yup, the app starts when a /launch command is issued on the client side, so we can only access the variables sent using this endpoint. Do you have any specific variable you want to be passed?

All of the parameters logged that start with a=nv-video would be great

```
a=x-nv-video[0].clientViewportWd:3840 
a=x-nv-video[0].clientViewportHt:2160 
a=x-nv-video[0].maxFPS:60 
a=x-nv-video[0].packetSize:1392 
a=x-nv-video[0].rateControlMode:4 
a=x-nv-video[0].timeoutLengthMs:7000 
a=x-nv-video[0].framesWithInvalidRefThreshold:0 
a=x-nv-video[0].initialBitrateKbps:80000 
a=x-nv-video[0].initialPeakBitrateKbps:80000 
a=x-nv-vqos[0].bw.minimumBitrateKbps:80000 
a=x-nv-vqos[0].bw.maximumBitrateKbps:80000 
a=x-nv-vqos[0].fec.enable:1 
a=x-nv-vqos[0].videoQualityScoreUpdateTime:5000 
a=x-nv-vqos[0].qosTrafficType:5 
a=x-nv-aqos.qosTrafficType:4 
a=x-nv-general.featureFlags:167 
a=x-nv-general.useReliableUdp:13 
a=x-nv-vqos[0].fec.minRequiredFecPackets:2 
a=x-nv-vqos[0].bllFec.enable:0 
a=x-nv-vqos[0].drc.enable:0 
a=x-nv-general.enableRecoveryMode:0 
a=x-nv-video[0].videoEncoderSlicesPerFrame:1 
a=x-nv-clientSupportHevc:1 
a=x-nv-vqos[0].bitStreamFormat:1 
a=x-nv-video[0].dynamicRangeMode:1 
a=x-nv-video[0].maxNumReferenceFrames:1 
a=x-nv-video[0].clientRefreshRateX100:0 
a=x-nv-audio.surround.numChannels:2 
a=x-nv-audio.surround.channelMask:3 
a=x-nv-audio.surround.enable:0 
a=x-nv-audio.surround.AudioQuality:0 
a=x-nv-aqos.packetDuration:5 
a=x-nv-video[0].encoderCscMode:0 
```

But I suppose just having HDR, fps and resolution is fine for my use cases though.

What's the option for figuring out if user has enabled/disabled the feature to close games on stream end? I could use that for my playnite watcher script.

TheElixZammuto:
> > > Specifically there is a bunch of moonlight variables logged when the log level as set to debug, which gives a lot of useful information. I’ve been wanting to make those accessible but it appears those are logged after the stream starts, not beforehand.
> > 
> > 
> > Yup, the app starts when a /launch command is issued on the client side, so we can only access the variables sent using this endpoint. Do you have any specific variable you want to be passed?
> 
> All of the parameters logged that start with a=nv-video would be great
> 
> ```
> a=x-nv-video[0].clientViewportWd:3840 
> a=x-nv-video[0].clientViewportHt:2160 
> a=x-nv-video[0].maxFPS:60 
> a=x-nv-video[0].packetSize:1392 
> a=x-nv-video[0].rateControlMode:4 
> a=x-nv-video[0].timeoutLengthMs:7000 
> a=x-nv-video[0].framesWithInvalidRefThreshold:0 
> a=x-nv-video[0].initialBitrateKbps:80000 
> a=x-nv-video[0].initialPeakBitrateKbps:80000 
> a=x-nv-vqos[0].bw.minimumBitrateKbps:80000 
> a=x-nv-vqos[0].bw.maximumBitrateKbps:80000 
> a=x-nv-vqos[0].fec.enable:1 
> a=x-nv-vqos[0].videoQualityScoreUpdateTime:5000 
> a=x-nv-vqos[0].qosTrafficType:5 
> a=x-nv-aqos.qosTrafficType:4 
> a=x-nv-general.featureFlags:167 
> a=x-nv-general.useReliableUdp:13 
> a=x-nv-vqos[0].fec.minRequiredFecPackets:2 
> a=x-nv-vqos[0].bllFec.enable:0 
> a=x-nv-vqos[0].drc.enable:0 
> a=x-nv-general.enableRecoveryMode:0 
> a=x-nv-video[0].videoEncoderSlicesPerFrame:1 
> a=x-nv-clientSupportHevc:1 
> a=x-nv-vqos[0].bitStreamFormat:1 
> a=x-nv-video[0].dynamicRangeMode:1 
> a=x-nv-video[0].maxNumReferenceFrames:1 
> a=x-nv-video[0].clientRefreshRateX100:0 
> a=x-nv-audio.surround.numChannels:2 
> a=x-nv-audio.surround.channelMask:3 
> a=x-nv-audio.surround.enable:0 
> a=x-nv-audio.surround.AudioQuality:0 
> a=x-nv-aqos.packetDuration:5 
> a=x-nv-video[0].encoderCscMode:0 
> ```
> 
> But I suppose just having HDR, fps and resolution is fine for my use cases though.
Unfortunately these are passed on the RTSP stage, so they cannot be passed on the app start

> What's the option for figuring out if user has enabled/disabled the feature to close games on stream end? I could use that for my playnite watcher script.

It seems that it's a client-only option

docs/source/about/app_examples.rst
```diff
@@ -125,24 +125,24 @@ Linux
...
-| **Field**            | **Value**                                                   |
-+----------------------+-------------------------------------------------------------+
-| Command Preparations | Do: ``wlr-xrandr --output HDMI-1 --mode 1920x1080@60Hz``    |
-|                      +-------------------------------------------------------------+
-|                      | Undo: ``wlr-xrandr --output HDMI-1 --mode 3840×2160@120Hz`` |
-+----------------------+-------------------------------------------------------------+
++----------------------+-----------------------------------------------------------------------------------------------------------------+
+| **Field**            | **Value**                                                                                                       |
++----------------------+-----------------------------------------------------------------------------------------------------------------+
+| Command Preparations | Do: ``sh -c "wlr-xrandr --output HDMI-1 --mode \"${SUNSHINE_CLIENT_WIDTH}x${SUNSHINE_CLIENT_HEIGHT}@60Hz\""``   |
```
ReenigneArcher, line 142: There should be one more of these Linux examples once you update your branch, from #1448

src/process.h
```diff
@@ -15,8 +15,10 @@
 
 #include "config.h"
 #include "platform/common.h"
+#include "rtsp.h"
 #include "utility.h"
 
+
```
ReenigneArcher, line 21:
```suggestion
```

ReenigneArcher (commented):
Can we add a macOS example?

Also, how about adding a section about the env variables to the docs? I could take care of this in a separate PR in order to get this moving along.

ReenigneArcher (approved):

Nonary:
I am not sure what specifically just yet, but Nightly is broken right now after this PR. It just refuses to stream anything and does not put anything in logs.

src/nvhttp.cpp
```diff
@@ -267,12 +268,32 @@ namespace nvhttp {
...
+    }
+    launch_session.unique_id = (get_arg(args, "uniqueid"));
+    launch_session.uuid = (get_arg(args, "uuid"));
+    launch_session.appid = util::from_view(get_arg(args, "appid"));
+    launch_session.enable_sops = util::from_view(get_arg(args, "sops"));
+    launch_session.surround_info = util::from_view(get_arg(args, "surroundAudioInfo"));
+    launch_session.gcmap = util::from_view(get_arg(args, "gcmap"));
+    launch_session.enable_hdr = 0;
+    if (args.find("enableHdr"s) != std::end(args)) {
+      launch_session.enable_hdr = util::from_view(get_arg(args, "enableHdr"));
```
ReenigneArcher, line 289: @TheElixZammuto 

Can we set a default value to empty/null if the client doesn't provide it? I'm thinking something similar to Python's `os.getenv` method. `os.getenv(key, "some default value")`

Apparently, Moonlight on Apple doesn't provide UUID (and who knows about every possible homebrew moonlight client). https://discord.com/channels/804382334370578482/1074377411069747250/1135639228613664858

Xythus: Perhaps something like

```cpp
  std::string
  get_arg(const args_t &args, const char *name, const char *default_value = nullptr) {
    auto it = args.find(name);
    if (it == std::end(args)) {
      if (default_value != NULL) {
        return std::string(default_value);
      }
      
      throw std::out_of_range(name);
    }
    return it->second;
  }
```

Used as such:
```cpp
   launch_session.gcm_key = util::from_hex<crypto::aes_t>(get_arg(args, "rikey", "0"), true);
    std::stringstream mode = std::stringstream(get_arg(args, "mode", "0x0x0"));
    // Split mode by the char "x", to populate width/height/fps
    int x = 0;
    std::string segment;
    while (std::getline(mode, segment, 'x')) {
      if (x == 0) launch_session.width = atoi(segment.c_str());
      if (x == 1) launch_session.height = atoi(segment.c_str());
      if (x == 2) launch_session.fps = atoi(segment.c_str());
      x++;
    }
    
    launch_session.unique_id = (get_arg(args, "uniqueid", "0123456789ABCDEF"));
    launch_session.uuid = (get_arg(args, "uuid", "00000000000000000000000000000000"));
    launch_session.appid = util::from_view(get_arg(args, "appid", "0123456789"));
    launch_session.enable_sops = util::from_view(get_arg(args, "sops", "0"));
    launch_session.surround_info = util::from_view(get_arg(args, "surroundAudioInfo", "0"));
    launch_session.gcmap = util::from_view(get_arg(args, "gcmap", "0"));
    launch_session.enable_hdr = 0;
    if (args.find("enableHdr"s) != std::end(args)) {
      launch_session.enable_hdr = util::from_view(get_arg(args, "enableHdr", "0"));
    }
    uint32_t prepend_iv = util::endian::big<uint32_t>(util::from_view(get_arg(args, "rikeyid", "0")));
```

PR #1466, opened 2023-07-25
Correct multi-FEC block size calculations

cgutman:
## Description
The old multi-FEC block calculations were way too conservative in determining our max data size before needing to split, which resulted in many frames being split into multiple FEC blocks unnecessarily. For example, a frame with 91 data shards would be split into 3 FEC blocks even though we could easily handle up to 212 data shards per block at the default FEC percentage of 20.

This issue is compounded by the fact that we always used a fixed split of 3 FEC blocks instead of calculating how many FEC blocks are actually required. Each FEC block requires separate calls to `reed_solomon_new()`, `reed_solomon_encode()`, and `WSASendMsg()/sendmsg()` which is costly in CPU time. Additionally, since the protocol supports a maximum of 4 FEC blocks per frame, our hardcoded split prevented us from sending frames as large as the protocol allows.

By fixing these calculations, this PR reduces CPU usage on the host, improves network utilization, and also enables up to 848 packets per frame at 20% FEC and 4096 packets per frame with 0% FEC (dynamically selected if data shard count is too high to send in 4 FEC blocks) up from 636 and 3172 respectively.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -560,7 +560,7 @@ namespace stream {
       auto parity_shards = (data_shards * fecpercentage + 99) / 100;
 
       // increase the FEC percentage for this frame if the parity shard minimum is not met
-      if (parity_shards < minparityshards) {
+      if (parity_shards < minparityshards && fecpercentage != 0) {
         parity_shards = minparityshards;
         fecpercentage = (100 * parity_shards) / data_shards;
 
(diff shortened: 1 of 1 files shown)


cgutman:
Heh, so now that the FEC block splitting logic is correct, it's exposing scalability issues in `platf::send_batch()`.

Rather than sending very loosly packed FEC blocks as we did before, we're now sending nearly full blocks of ~255 packets each. Because we send each FEC block as a single batch with `platf::send_batch()`, this means we're sending much larger batches than before, which is leading to packet loss. These large batches were possible before this change, but they were much less likely due to the fixed 3-way FEC block split.

So I'll have to finally write that batch throttling logic that I was putting off. For now, I can have it mimic the batching of the current code and we can optimize it using the FEC feedback messages from Moonlight later on.

PR #1486, opened 2023-07-31
Nearest Neighbour filtering

RomaRogov:
## Description
Hello!
I've implemented new lever in config to control resize algorithm for textures with captured screen. This allows to get high-resolution pixel perfect stream video for some retro games or just low resolution which can be doubled.
For example, I'm using this with my Abernic RG505 device which has PlayStation Vita screen (960x544px), which has artifacts when stream has native resolution. But if I double stream resolution with nearest neighbour scaling I'm getting pixel-perfect image without any compression artifacts. May be also useful for PS Vita users.
Also now I can play retro games with 1080p streaming on high resolution devices and they are not blurry at all. So, this lever, as I think, definitely has a few applications.
I'm not very used to C++, so please give feedback if my changes should be refactored before merge. Thanks in advance!

### Screenshot
Screenshot_20230731-150626_Moonlight (Debug)

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -434,6 +434,21 @@ dwmflush
 
       dwmflush = enabled
 
+point_filtering
+^^^^^^^^^^^^^^^
+
+**Description**
+   Enable nearest point filtering for scaling from captured screen to streamed video.
+   May be useful for retro games or pixel-perfect streaming to low-resolution devices (PlayStation Vita, for example).
+
+**Default**
+   ``disabled``
+
+**Example**
+   .. code-block:: text
+
+      point_filtering = enabled
+
 Audio
 -----
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -379,7 +379,8 @@ namespace config {
     {},  // encoder
     {},  // adapter_name
     {},  // output_name
-    true  // dwmflush
+    true,  // dwmflush
+    false //point_filtering
   };
 
   audio_t audio {
@@ -969,6 +970,7 @@ namespace config {
     string_f(vars, "adapter_name", video.adapter_name);
     string_f(vars, "output_name", video.output_name);
     bool_f(vars, "dwmflush", video.dwmflush);
+    bool_f(vars, "point_filtering", video.point_filtering);
 
     path_f(vars, "pkey", nvhttp.pkey);
     path_f(vars, "cert", nvhttp.cert);
--- a/src/config.h
+++ b/src/config.h
@@ -60,6 +60,7 @@ namespace config {
     std::string adapter_name;
     std::string output_name;
     bool dwmflush;
+    bool point_filtering;
   };
 
   struct audio_t {
--- a/src/platform/linux/graphics.cpp
+++ b/src/platform/linux/graphics.cpp
@@ -4,6 +4,7 @@
  */
 #include "graphics.h"
 #include "src/video.h"
+#include "src/config.h"
 
 #include <fcntl.h>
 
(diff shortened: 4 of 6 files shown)


ns6089:
So you want host-side nearest-neighbour upscaler? Then maybe it's better to simply use dedicated game upscaler like https://github.com/Blinue/Magpie. Adding it as processing option is basically a work multiplier, since we have to implement (and support) it across every platform and capture method. Just my two cents.

ns6089:
> just selection between linear and point in texture filter type

If we stick to texture sampler scaling, which already doesn't work that well (any amount of image downscaling produces chroma artifacts because paired with 4:2:0 subsampling it result in >2x downscaling ratio). And with kernel-base scaling, nearest-neighbour will require different kernel of different size. It's certainly possible, but it *will* require more work, probably sooner than later.

RomaRogov:
Oh, well, I see - things are actually more complicated and not so universal to be public change, though it works good for my case.
Maybe authors just can consider such kind of feature in the future, who knows :)

psyke83:
Perhaps this idea might have still have merit in the specific case where the client resolution is evenly divisible by the host resolution? Here's a quick proof of concept: https://github.com/LizardByte/Sunshine/compare/nightly...psyke83:nearest_scaling

This should improve clarity when the host and client resolution are matched, but there is additional benefit when the client resolution is a multiple of the host.

Consider my use case:
* Client machine is a laptop with a 1360x768 screen.
* Host machine is using a HDMI EDID emulator, so resolution is flexible (for the purpose of supersampling).

Streaming Moonlight at 1360x768 with current Sunshine nightly:
* Host at 1360x768 looks ok (no aliasing artifacts), but is slightly blurry.
* Standard host resolutions (1080p, 1440p, 4K) have aliasing artifacts and look somewhat blurry
* 2720x1536 (2x client panel res) has noticeable aliasing artifacts and still looks blurrier than 1360x768.

Streaming Moonlight at 2720x1536 (2x panel res) with current Sunshine nightly:
* Host at 1360x768 has no noticeable aliasing artifacts, but looks very blurry
* Standard host resolutions (1080p, 1440p, 4K) has noticeable (but less severe) aliasing artifacts, yet still looks quite blurry
* 2720x1536 has little to no aliasing artifacts, but still looks blurrier than expected.

Streaming Moonlight at 1360x768 with [nearest scaling](https://github.com/psyke83/SunshineStream/tree/nearest_scaling) branch:
* Host at 1360x768 looks much sharper.
* All other resolutions are unchanged compared to nightly.

Streaming Moonlight at 2720x1536 (2x panel res) with [nearest scaling](https://github.com/psyke83/SunshineStream/tree/nearest_scaling) branch:
* Host at 1360x768 looks much sharper.
* Host at 2720x1536 looks excellent (no aliasing artifacts or blurring) <- this is now my preferred streaming setup
* All other resolutions unchanged compared to nightly

Granted this may be an odd use case, but I can envison the same benefit for standard resolutions (1080p panel with Moonlight @ 1080p to 1080p host, as well as 1080p panel with Moonlight @ 4K to 4K or 1080p host, etc.).

psyke83:
Nearest neighbour filtering should involve less overhead than bilinear filtering, but it's unlikely to cause any real performance change. The important part to stress is that having a simple on/off toggle as seen in this PR is not optimal. 

To illustrate, looking at my setup, enabling "point filtering" from this PR would do the following:
* Moonlight @ 1360x768: host at 1360x768 looks good, all other resolutions will have scaling artifacts.
* Moonlight @ 2720x1536: host resolutions 1360x768 and 2720x1536 look good, but all others will have scaling artifacts.

I know this is a niche setup, but the behaviour will replicate for standard resolutions (1080P and 4K, 1440P and 720P resolutions pairs) as well as OP's 960x544 & doubled resolution. The only sensible user configuration (if at all needed) would be an Auto/Off toggle, where Auto will only use nearest filtering when the client resolution is divisible by the host (i.e. my example branch) and Off to preserve the old behaviour for users that prefer the softer look provided by bilinear filtering for all resolutions.

ns6089:
@psyke83 Our chroma shader relies on linear texture sampler for some math / texture fetch optimizations.
I've explained it here https://github.com/LizardByte/Sunshine/pull/1621#discussion_r1329723658.

ReenigneArcher:
@RomaRogov Thank you for the PR. Unfortunately at this time, we have decided not to accept this change. In order to re-consider this PR the comments by @ns6089 would need to be addressed.

PR #1530, opened 2023-08-13
AV1 encoding support

cgutman:
## Description
This PR implements support for encoding AV1 using FFmpeg and standalone NVENC. For the former, we require the changes in https://github.com/LizardByte/build-deps/pull/136, however we can't merge that yet due to the QSV memory leak introduced in FFmpeg 6.0. I tested the `av1_qsv` and `av1_amf` encoders using the WIP FFmpeg 6.0 build to ensure the encoder parameters and output bitstream were valid.

I added the `av1_vaapi` and `av1_videotoolbox` encoder names though they are not upstreamed in FFmpeg yet, so hopefully they will work when they get added.

The net effect of this PR alone is that it will turn on AV1 support for standalone NVENC (Windows). AV1 will be enabled for the other encoders when we update to FFmpeg 6.1.

Tested:
- Standalone NVENC + RFI
- QSV (requires FFmpeg 6.0+ build)
- AMF (requires FFmpeg 6.0+ build)
- libsvtav1 (broken and disabled due to inability to return IDR frames on demand)

cc: @ns6089 for the NVENC stuff and @psyke83 for the AMD stuff (and the change to disable `filler_data`)

### Issues Fixed or Closed
Partially resolves #1325 (NVENC Windows only)

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/nvenc/nvenc_base.cpp
+++ b/src/nvenc/nvenc_base.cpp
@@ -124,6 +124,11 @@ namespace nvenc {
         init_params.encodeGUID = NV_ENC_CODEC_HEVC_GUID;
         break;
 
+      case 2:
+        // AV1
+        init_params.encodeGUID = NV_ENC_CODEC_AV1_GUID;
+        break;
+
       default:
         BOOST_LOG(error) << "NvEnc: unknown video format " << client_config.videoFormat;
         return false;
@@ -291,6 +296,33 @@ namespace nvenc {
         set_ref_frames(format_config.maxNumRefFramesInDPB, format_config.numRefL0, 5);
         set_minqp_if_enabled(config.min_qp_hevc);
         fill_vui(format_config.hevcVUIParameters);
+        break;
+      }
+
+      case 2: {
+        // AV1
+        auto &format_config = enc_config.encodeCodecConfig.av1Config;
+        format_config.repeatSeqHdr = 1;
+        format_config.idrPeriod = NVENC_INFINITE_GOPLENGTH;
+        format_config.chromaFormatIDC = 1;  // YUV444 not supported by NVENC yet
+        format_config.enableBitstreamPadding = config.insert_filler_data;
+        if (buffer_is_10bit()) {
+          format_config.inputPixelBitDepthMinus8 = 2;
+          format_config.pixelBitDepthMinus8 = 2;
+        }
+        format_config.colorPrimaries = colorspace.primaries;
+        format_config.transferCharacteristics = colorspace.tranfer_function;
+        format_config.matrixCoefficients = colorspace.matrix;
+        format_config.colorRange = colorspace.full_range;
+        set_ref_frames(format_config.maxNumRefFramesInDPB, format_config.numFwdRefs, 8);
+        set_minqp_if_enabled(config.min_qp_av1);
+
+        if (client_config.slicesPerFrame > 1) {
+          // NVENC only supports slice counts that are powers of two, so we'll pick powers of two
+          // with bias to rows due to hopefully more similar macroblocks with a row vs a column.
+          format_config.numTileRows = std::pow(2, std::ceil(std::log2(client_config.slicesPerFrame) / 2));
+          format_config.numTileColumns = std::pow(2, std::floor(std::log2(client_config.slicesPerFrame) / 2));
+        }
         break;
       }
     }
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -810,6 +810,37 @@ hevc_mode
 
       hevc_mode = 2
 
+av1_mode
+^^^^^^^^^
+
+**Description**
+   Allows the client to request AV1 Main 8-bit or 10-bit video streams.
+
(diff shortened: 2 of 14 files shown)


src/nvenc/nvenc_base.cpp
```diff
@@ -293,6 +298,30 @@ namespace nvenc {
...
+        format_config.colorPrimaries = colorspace.primaries;
+        format_config.transferCharacteristics = colorspace.tranfer_function;
+        format_config.matrixCoefficients = colorspace.matrix;
+        format_config.colorRange = colorspace.full_range;
+        set_ref_frames(format_config.maxNumRefFramesInDPB, format_config.numBwdRefs, 8);
+        set_minqp_if_enabled(config.min_qp_av1);
+
+        if (client_config.slicesPerFrame > 1) {
+          // FIXME: Not supported yet
+          BOOST_LOG(warning) << "Client-controlled slicing is not yet supported by NVENC for AV1";
```
ns6089, line 321: Slices in AV1 are replaced with what they call tiles, and I think NvENC allows to control them with some limitations.
Have you looked into `enableCustomTileConfig, numTileColumns, numTileRows, tileWidths. tileHeights` parameters?

src/nvenc/nvenc_base.cpp
```diff
@@ -293,6 +298,30 @@ namespace nvenc {
...
+        format_config.chromaFormatIDC = 1;  // YUV444 not supported by NVENC yet
+        if (buffer_is_10bit()) {
+          format_config.inputPixelBitDepthMinus8 = 2;
+          format_config.pixelBitDepthMinus8 = 2;
+        }
+        format_config.colorPrimaries = colorspace.primaries;
+        format_config.transferCharacteristics = colorspace.tranfer_function;
+        format_config.matrixCoefficients = colorspace.matrix;
+        format_config.colorRange = colorspace.full_range;
+        set_ref_frames(format_config.maxNumRefFramesInDPB, format_config.numBwdRefs, 8);
```
ns6089, line 316: `numFwdRefs` here. It's not worded very well, but that's the correct one. Number of reference frames used in forward prediction is probably what they meant (and it mentions "golden" frames, while the other one mentions L1).

Max ref frames is 7. "The AV1 codec allows a maximum of 8 frames in its
decoded frame buffer. For a coding frame, it can choose any 7
frames from the decoded frame buffer as its reference frames".

ns6089: And I'm not sure AV1 allows to use more than 4 of those 7 for forward prediction, need to test this with RFI.

ns6089: Re-reading that paper again, 8 is probably the correct number. 7 is the limit one frame can use from the pool 8: 4 for forward prediction, and 3 for backwards.

ns6089: Or it's 7, with the intention of currently decoded frame to replace the 8-th. And god knows what NvENC means by their "maxNumRefFramesInDPB". 7 should be the safe number, I would go with 7 for now (while adding the comment to find a way to confirm or deny 8).

src/nvenc/nvenc_base.cpp
```diff
@@ -293,6 +298,30 @@ namespace nvenc {
...
+        set_minqp_if_enabled(config.min_qp_av1);
+
```
ns6089, line 318: Can add `format_config.enableBitstreamPadding = config.insert_filler_data;`

src/nvenc/nvenc_base.cpp
cgutman: I think your initial quote is correct - https://aomedia.org/av1/specification/symbols/

AFAIK, the DPB allows 8 reference frames (`NUM_REF_FRAMES`) but only 7 (`REFS_PER_FRAME`) can be referenced by any single frame (4 forward and 3 backward) plus an intra reference against the frame itself for 8 (`TOTAL_REFS_PER_FRAME`).

I've tested for several hours streaming with packet loss and didn't see any artifacts with RFI on AV1, so we either have the correct value of 8 or Nvidia is clamping it to something lower. I tried passing something obviously bogus like 900 and NVENC still initialized and worked, so I don't think we can distinguish these two. In either case though, it's doing the right thing at runtime.

src/nvenc/nvenc_base.cpp
cgutman: Done

src/nvenc/nvenc_base.cpp
cgutman: Yep, for some reason I was misreading the docs and thought I had to calculate all the tile sizes manually if I specified the tile counts, but that's only true if I set `enableCustomTileConfig`. Added this.

src/nvenc/nvenc_base.cpp
ns6089: I think we can test it by manually calling RFI to see when it produces an IDR frame. But it's not a blocker and can be done in the future.

ns6089 (dismissed):
Looks good to me.

psyke83 (dismissed):
I don't have AV1 hardware encoding capability on my card, so I can't help much on that end, but I'm fine with the change to filler data.

Issue #1456, opened 2023-07-19
Do/Undo Commands fail to work after executing a certain amount of times


Nonary:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

Sunshine will fail to execute do/undo commands after executing them a certain number of times, appears to be intermitent.

To replicate the bug:

Add an app with the following do/undo commands:

```
        {
            "name": "Debug",
            "output": "",
            "cmd": "",
            "exclude-global-prep-cmd": "true",
            "elevated": "false",
            "prep-cmd": [
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
                    "undo": "powershell.exe -Command \"Stop-Process -Name 'ping' -ErrorAction SilentlyContinue\"",
                    "elevated": "false"
                },
                {
                    "do": "",
[...]

PR #1538, opened 2023-08-15
Miscellaneous process fixes

cgutman:
## Description
The first commit contains a small fix for the process code to avoid a race condition where a process could spawn a child before we added it into our job for tracking purposes. This child would escape our control and wouldn't be terminated when the user quit the app. 

The second is cleanup related to #1456 to pass the environment as a const ref to avoid anyone being tempted to modify the passed in `env` parameter again.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/common.h
+++ b/src/platform/common.h
@@ -547,7 +547,7 @@ namespace platf {
   display_names(mem_type_e hwdevice_type);
 
   boost::process::child
-  run_command(bool elevated, bool interactive, const std::string &cmd, boost::filesystem::path &working_dir, boost::process::environment &env, FILE *file, std::error_code &ec, boost::process::group *group);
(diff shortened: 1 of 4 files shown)


cgutman:
> Boost documentation is pretty bare and I am not an experienced C++ developer, but one question I do have is that if native_handle() is thread safe?

It is. The job object is created by the `boost::process::group` constructor and closed by the destructor, so it's safe to access for the lifetime of the object. https://github.com/boostorg/process/blob/develop/include/boost/process/detail/windows/group_handle.hpp

> I am just not sure if it would fix the race condition itself, with this change it is possible it would just fail to create the child process and then give users an error during their connection attempt, as opposed to just working and being able to close it after ending the steam.

The most significant race condition here isn't a race between the direct child and Sunshine adding it to the job object (though that _is_ a race). We will add that process to our job within a few milliseconds, so nothing really bad will happen unless we crash or get terminated in that tiny window. The major race here is that the child process we just spawned could spawn a grand-child 
 before we add the child into our job and _that_ process will be lost forever from our job tracking (since job membership is inherited from the parent by default).

PR #1542, opened 2023-08-15
nvenc: improve rfi logic

ns6089:
## Description

Narrows down rfi arguments check to ref frames buffer size and removes log error message.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/nvenc/nvenc_base.cpp
+++ b/src/nvenc/nvenc_base.cpp
@@ -246,6 +246,7 @@ namespace nvenc {
         ref_frames_option = 1;
         encoder_params.rfi = false;
       }
+      encoder_params.ref_frames_in_dpb = ref_frames_option;
       // This limits ref frames any frame can use to 1, but allows larger buffer size for fallback if some frames are invalidated through rfi
       L0_option = NV_ENC_NUM_REF_FRAMES_1;
     };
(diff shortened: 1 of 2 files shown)


src/nvenc/nvenc_base.cpp
```diff
@@ -457,33 +458,31 @@ namespace nvenc {
...
+    if (first_frame == encoder_state.last_rfi_range.first &&
         last_frame <= encoder_state.last_rfi_range.second) {
       BOOST_LOG(debug) << "NvEnc: rfi request " << first_frame << "-" << last_frame << " already done";
       return true;
     }
 
     BOOST_LOG(debug) << "NvEnc: rfi request " << first_frame << "-" << last_frame << " expanding to last encoded frame " << encoder_state.last_encoded_frame_index;
+    last_frame = encoder_state.last_encoded_frame_index;
+
+    if (last_frame < first_frame || last_frame - first_frame + 1 >= encoder_params.ref_frames_in_dpb) {
```
cgutman, line 470: `last_frame < first_frame` is legitimately invalid, since it means Moonlight is invalidating a frame it hasn't possibly seen yet. You could log that as an error if you want.

src/nvenc/nvenc_base.cpp
```diff
@@ -457,33 +458,31 @@ namespace nvenc {
...
 
-    if (last_frame < first_frame ||
-        encoder_state.last_encoded_frame_index < first_frame ||
-        encoder_state.last_encoded_frame_index > first_frame + 100) {
-      BOOST_LOG(error) << "NvEnc: rfi request " << first_frame << "-" << last_frame << " invalid range (last encoded frame " << encoder_state.last_encoded_frame_index << ")";
-      return false;
-    }
-
-    if (first_frame >= encoder_state.last_rfi_range.first &&
+    if (first_frame == encoder_state.last_rfi_range.first &&
```
cgutman, line 461: Is the change from `>=` to `==` correct? If you receive an RFI range of [5, 8] then [6, 8], the latter is a subset of the former, so it should be a no-op.

ns6089: > If you receive an RFI range of [5, 8] then [6, 8]

This shouldn't happen. Until client receives uncorrupted rfi confirmation (or idr) frame, they will continue sending expanding rfi range. So I thought == reflects the behavior more closely, and it won't break even if the client starts sending narrowing range for whatever reason.

But reverted to >=, if the client wants to be stupid, it's on them.

src/nvenc/nvenc_base.cpp
ns6089: Yeah, moved to separate branch with error print.

PR #1548, opened 2023-08-16
fix: ensure user supplied framerates are used

jensaymoo:
## Description
The framerates list is not cleared before adding values from the file configuration.
If the list is not cleared, then the set parameters do not affect the behavior of sunshine server.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -836,6 +836,7 @@ namespace config {
     std::vector<std::string> list;
     list_string_f(vars, name, list);
 
+    input.clear();
     for (auto &el : list) {
       std::string_view val = el;
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -840,7 +840,6 @@ namespace config {
     // If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. 
     // That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
     input.clear();
-    
     for (auto &el : list) {
       std::string_view val = el;
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -836,9 +836,9 @@ namespace config {
     std::vector<std::string> list;
     list_string_f(vars, name, list);
 
-    //The framerate list must be cleared before adding values from the file configuration. 
-    //If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. 
-    //That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
+    // The framerate list must be cleared before adding values from the file configuration. 
+    // If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. 
+    // That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
     input.clear();
     
     for (auto &el : list) {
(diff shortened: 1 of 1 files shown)


src/config.cpp
```diff
@@ -835,6 +835,7 @@ namespace config {
     std::vector<std::string> list;
     list_string_f(vars, name, list);
 
+    input.clear();
```
ReenigneArcher, line 838: Please comment your code, so when someone looks at this in a year or 5 they understand why this is here.

jensaymoo: The frame rate list must be cleared before adding values from the file configuration. If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes [10, 30, 60, 90 and 120 fps](https://github.com/LizardByte/Sunshine/blob/31e8b798dabf47d5847a1b485f57cf850a15fcae/src/config.cpp#L423).

ReenigneArcher: Thank you for the explanation, but please add it to the code.

src/config.cpp
```diff
@@ -834,6 +834,11 @@ namespace config {
     std::vector<std::string> list;
     list_string_f(vars, name, list);
 
+    //The framerate list must be cleared before adding values from the file configuration. 
+    //If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. 
+    //That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
```
ReenigneArcher, line 839:
```suggestion
    // The framerate list must be cleared before adding values from the file configuration. 
    // If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. 
    // That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
```

ReenigneArcher (changes requested):

ReenigneArcher (approved):

src/config.cpp
```diff
@@ -836,6 +836,11 @@ namespace config {
     std::vector<std::string> list;
     list_string_f(vars, name, list);
 
+    // The framerate list must be cleared before adding values from the file configuration. 
+    // If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. 
+    // That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
+    input.clear();
+
```
ReenigneArcher, line 843:
```suggestion

```

src/config.cpp
```diff
@@ -836,6 +836,10 @@ namespace config {
     std::vector<std::string> list;
     list_string_f(vars, name, list);
 
+    // The framerate list must be cleared before adding values from the file configuration. 
+    // If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server. 
+    // That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
+    input.clear();
```
ReenigneArcher, line 842:
```suggestion
    // The framerate list must be cleared before adding values from the file configuration.
    // If the list is not cleared, then the specified parameters do not affect the behavior of the sunshine server.
    // That is, if you set only 30 fps in the configuration file, it will not work because by default, during initialization the list includes 10, 30, 60, 90 and 120 fps.
    input.clear();
```

PR #1561, opened 2023-08-20
Event Ballons and Tray Icon improvements

TheElixZammuto:
## Description
This PR improves the UX of Sunshine, adding an indicator state for the tray icon (Not streaming, when the stream is in progress and when Sunshine is not streaming but the app is still running).
The PR also adds Balloon notifications for each state change and when a user starts a pairing request. This particular balloon can also be clicked to open the Pairing page of the Web UI

This modification require [our fork of tray](https://github.com/LizardByte/tray/tree/balloons). I'm not sure about upstreaming yet, since macOS support for these changes is nil. 

### Screenshot
pin
starting
paused
stopped

## Fixes
- Closes #1313 

## Type of Change

## Checklist

## Branch Updates


src/system_tray.cpp
```diff
@@ -268,5 +276,74 @@ namespace system_tray {
...
+  void
+  update_tray_stopped(std::string app_name) {
+    tray.notification_title = NULL;
+    tray.notification_text = NULL;
+    tray.notification_cb = NULL;
+    tray.notification_icon = NULL;
+    tray.icon = TRAY_ICON;
+    tray_update(&tray);
+    char msg[256];
+    sprintf(msg, "Application %s successfuly stopped", app_name.c_str());
```
cgutman, line 322: `successfully`

src/system_tray.cpp
```diff
@@ -268,5 +276,74 @@ namespace system_tray {
...
+  void
+  update_tray_require_pin() {
+    tray.notification_title = NULL;
+    tray.notification_text = NULL;
+    tray.notification_cb = NULL;
+    tray.notification_icon = NULL;
+    tray.icon = TRAY_ICON;
+    tray_update(&tray);
+    tray.icon = TRAY_ICON;
+    tray.notification_title = "Request to Pair";
```
cgutman, line 339: I would use a title like "Incoming Pairing Request"

src/system_tray.cpp
```diff
@@ -268,5 +276,74 @@ namespace system_tray {
...
+    tray.notification_text = "Click here to insert the PIN and complete the pairing process";
```
cgutman, line 340: Maybe you could shorten to "Click here to complete the pairing process"

src/system_tray.cpp
```diff
@@ -268,5 +276,74 @@ namespace system_tray {
...
+    tray.notification_title = NULL;
+    tray.notification_text = NULL;
+    tray.notification_cb = NULL;
+    tray.notification_icon = NULL;
+    tray.icon = TRAY_ICON_PLAYING;
+    tray_update(&tray);
+    tray.icon = TRAY_ICON_PLAYING;
+    tray.notification_title = "Stream Started";
+    char msg[256];
+    sprintf(msg, "Streaming Started for App %s", app_name.c_str());
```
cgutman, line 290: The capitalization is somewhat inconsistent for these messages. For the title, let's capitalize each word and for the message, let's capitalize just the first word.

Maybe we could also drop "App" from the message. It might read funny for some app names.

cgutman:
This is pretty neat, especially the pairing notification.

My only concern is whether the streaming start/stop/pause notifications might interfere with some games.

ns6089:
> My only concern is whether the streaming start/stop/pause notifications might interfere with some games.

We can use [SHQueryUserNotificationState()](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shqueryusernotificationstate). Stack Overlfow [answer](https://stackoverflow.com/a/58755620).

src/system_tray.cpp
```diff
@@ -268,5 +276,74 @@ namespace system_tray {
...
+    tray.notification_cb = NULL;
+    tray.notification_icon = NULL;
+    tray.icon = TRAY_ICON_PLAYING;
+    tray_update(&tray);
+    tray.icon = TRAY_ICON_PLAYING;
+    tray.notification_title = "Stream Started";
+    char msg[256];
+    sprintf(msg, "Streaming Started for %s", app_name.c_str());
+    tray.notification_text = msg;
+    tray.notification_icon = TRAY_ICON_PLAYING;
```
cgutman, line 292: Can you update the tooltip for the playing and paused states, just so it's clear to users what the icon means?

src/system_tray.cpp
```diff
@@ -268,5 +276,74 @@ namespace system_tray {
...
+    tray.notification_title = NULL;
+    tray.notification_text = NULL;
+    tray.notification_cb = NULL;
+    tray.notification_icon = NULL;
+    tray.icon = TRAY_ICON_PLAYING;
+    tray_update(&tray);
+    tray.icon = TRAY_ICON_PLAYING;
+    tray.notification_title = "Stream Started";
+    char msg[256];
+    sprintf(msg, "Streaming Started for %s", app_name.c_str());
```
cgutman, line 290: "Started" should be lower-case to be consistent with the other messages.

.gitmodules
```diff
@@ -48,13 +48,13 @@
 	branch = master
 [submodule "third-party/tray"]
 	path = third-party/tray
-	url = https://github.com/dmikushin/tray
-	branch = master
+	url = https://github.com/LizardByte/tray
+	branch = balloons
```
cgutman, line 52: Should we merge to master rather than leaving the branch here?

ReenigneArcher: Additionally, would the changes made be compatible with the upstream project goals? The few interactions I've had with them, they seemed pretty receptive to improvements.

cgutman: I wasn't thinking in terms of upstreaming. I was thinking we probably want a protected branch that can't be directly committed to.

ReenigneArcher: That repo doesn't have branch protection rules as of yet... I can set it up whenever though.

TheElixZammuto: For upstreaming the only problem is feature parity, since I haven't touched the macOS side of things. As for the branch @ReenigneArcher let me know if I should make it myself or if you're planning to make it

ReenigneArcher: `master` branch is now protected and made another branch `master_upstream` if we eventually want to upstream the changes.

It will require a pull request to merge anything into `master`, but there is no CI.

MichaelMKenny:
It seems that with this pr the global undo scripts don't fire when an app closes. Example add an undo global script that runs winver, add notepad as an app, stream notepad, then click close button in the titlebar, winver doesn't trigger. But everything works when just on the latest nightly. 

I see that this pr changes process.cpp, and the problem is that in `proc_t::terminate()`, sunshine never enters the for loop `for (; _app_prep_it != _app_prep_begin; --_app_prep_it)` , which means it never gets to run the undo scripts.

.gitmodules
TheElixZammuto: I've updated our fork, merging our changes to the master branch

TheElixZammuto:
> It seems that with this pr the global undo scripts don't fire when an app closes. Example add an undo global script that runs winver, add notepad as an app, stream notepad, then click close button in the titlebar, winver doesn't trigger. But everything works when just on the latest nightly.
> 
> I see that this pr changes process.cpp, and the problem is that in `proc_t::terminate()`, sunshine never enters the for loop `for (; _app_prep_it != _app_prep_begin; --_app_prep_it)` , which means it never gets to run the undo scripts.

Should now be fixed, there was a stack overflow between `running` and `terminate`

MichaelMKenny:
Can confirmed it's fixed now 👍

CMakeLists.txt
```diff
@@ -440,7 +440,8 @@ ${CMAKE_BINARY_DIR}/generated-src/${filename}.h")
 
     # tray icon
     if(${SUNSHINE_ENABLE_TRAY})
-        pkg_check_modules(APPINDICATOR appindicator3-0.1)
+        pkg_check_modules(APPINDICATOR ayatana-appindicator3-0.1)
+        pkg_check_modules(LIBNOTIFY libnotify)
```
ReenigneArcher, line 444: This is now in `.\cmake\compile_definitions\linux.cmake`

.gitmodules
```diff
@@ -57,4 +57,4 @@
 [submodule "third-party/ffmpeg-linux-powerpc64le"]
 	path = third-party/ffmpeg-linux-powerpc64le
 	url = https://github.com/LizardByte/build-deps
-	branch = ffmpeg-linux-powerpc64le
+	branch = ffmpeg-linux-powerpc64le
```
ReenigneArcher, line 60:
```suggestion
	branch = ffmpeg-linux-powerpc64le

```

src/process.cpp
```diff
@@ -116,7 +117,7 @@ namespace proc {
 
     _app_id = app_id;
     _app = *iter;
-
+
```
ReenigneArcher, line 120:
```suggestion

```

CMakeLists.txt
ReenigneArcher: Also probably need to install all the icons for Linux?

That's in `.\cmake\packaging\linux.cmake`. Look for

`install(FILES "${CMAKE_SOURCE_DIR}/sunshine.svg" DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons")`

src/system_tray.cpp
```diff
@@ -10,10 +10,18 @@
     #include <accctrl.h>
     #include <aclapi.h>
     #define TRAY_ICON WEB_DIR "images/favicon.ico"
-  #elif defined(__linux__) || defined(linux) || defined(__linux)
-    #define TRAY_ICON "sunshine"
```
ReenigneArcher, line 14: I think on Linux it's required to pull the icons from the `.../share/icons` directory, unless something changed.

`svg` files were used before, as I don't know that `ico` is supported, and they scale better than `png`.

CMakeLists.txt
TheElixZammuto: I've fixed the ayatana build with conditional pkgconfig for both appindicator and ayatana-appindicator
I've also added the icons in the correct folder

src/stream.cpp
```diff
@@ -1732,6 +1733,11 @@ namespace stream {
 
       // If this is the last session, invoke the platform callbacks
       if (--running_sessions == 0) {
+#if defined SUNSHINE_TRAY && SUNSHINE_TRAY >= 1
+        if(proc::proc.running()){
```
ReenigneArcher, line 1737:
```suggestion
        if (proc::proc.running()) {
```

Fixes lint error

src/system_tray.cpp
```diff
@@ -10,10 +10,17 @@
...
+    #define TRAY_ICON_PAUSING WEB_DIR "images/sunshine-pausing.ico"
+    #define TRAY_ICON_LOCKED WEB_DIR "images/sunshine-locked.ico"
+  #else
     #define TRAY_ICON "sunshine"
-  #elif defined(__APPLE__) || defined(__MACH__)
-    #define TRAY_ICON WEB_DIR "images/logo-sunshine-16.png"
+    #define TRAY_ICON_PLAYING "sunshine-playing"
+    #define TRAY_ICON_PAUSING "sunshine-pausing"
+    #define TRAY_ICON_LOCKED "sunshine-locked"
+  #endif
```
ReenigneArcher, line 21: Should we leave the conditions for linux and macos? From my earlier testing the tray icon required png on macos.

Or maybe just a comment about it?

ReenigneArcher (changes requested):
Sorry for another review, I missed this the first time.

src/system_tray.cpp
TheElixZammuto: I can change it, but in theory the macOS tray icon does not work, am I right?

ReenigneArcher: It does not work with Sunshine, but the tray library does work with macOS.

I couldn't figure out how to make it not block the main thread of sunshine.

ReenigneArcher (approved):

PR #1681, opened 2023-09-29
feat(ui): add port mapping table

### LINKED ISSUE #1682: feat(ui): add port mapping table based on selected config port

ReenigneArcher:
## Description

Add a port configuration table to the UI.

Also validates the config port is is in the range of 1029 to 65514, and that all ports are in the range of 1024 to 65535.

Modifies alerts to include a font awesome icon.

### Screenshot

Default value:
image

Custom value:
image

Invalid values:
image
image

### Issues Fixed or Closed

- Resolves #1682 

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -836,7 +836,15 @@ write_file(const char *path, const std::string_view &contents) {
  */
 std::uint16_t
 map_port(int port) {
-  // TODO: Ensure port is in the range of 21-65535
+  // calculate the port from the config port
+  auto mapped_port = (std::uint16_t)((int) config::sunshine.port + port);
+
+  // Ensure port is in the range of 1024-65535
+  if (mapped_port < 1024 || mapped_port > 65535) {
+    BOOST_LOG(warning) << "Port out of range: "sv << mapped_port;
+  }
+
   // TODO: Ensure port is not already in use by another application
-  return (std::uint16_t)((int) config::sunshine.port + port);
+
+  return mapped_port;
 }
--- a/src_assets/common/assets/web/config.html
+++ b/src_assets/common/assets/web/config.html
@@ -72,6 +72,7 @@
           id="origin_web_ui_allowed"
           class="form-select"
           v-model="config.origin_web_ui_allowed"
+          @change="forceUpdate"
         >
           <option value="pc">Only localhost may access Web UI</option>
           <option value="lan">Only those in LAN may access Web UI</option>
@@ -79,6 +80,10 @@
         </select>
         <div class="form-text">
           The origin of the remote endpoint address that is not denied access to Web UI
+        </div>
+        <!-- add warning about exposing web ui to the internet -->
+        <div class="alert alert-danger" v-if="config.origin_web_ui_allowed === 'wan'">
+          <i class="fa-solid fa-xl fa-triangle-exclamation"></i> Exposing the Web UI to the internet is a security risk! Proceed at your own risk!
         </div>
       </div>
       <!--UPnP-->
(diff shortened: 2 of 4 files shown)


cgutman:
Can you add a description or some emphasis on the HTTP port entry that indicates that's the one that should be provided to Moonlight when adding a PC manually?

yungleballz:
I would like to suggest to include an info that the Web UI Port does not have to (and should not) be forwarded.

ReenigneArcher:
@cgutman yes, good idea.

@yungleballz that will depend on user preferences. Many users forward it so they can modify settings away from home, or pair a new client.

ReenigneArcher:
Idea: Add port checker -> https://www.npmjs.com/package/is-port-reachable

TheElixZammuto:
> @yungleballz that will depend on user preferences. Many users forward it so they can modify settings away from home, or pair a new client.

I agree with yungle here, we need a big red scary warning to avoid people forwarding the UI without knowing what they do

> Idea: Add port checker -> https://www.npmjs.com/package/is-port-reachable
> 
> This probably needs to be in a separate PR after webpack is implemented.

Due to NAT (and hairpin), having Sunshine itself make the test is not representitive, the Internet Streaming Tool uses a relay server IIRC

src/main.cpp
```diff
@@ -836,7 +836,13 @@ write_file(const char *path, const std::string_view &contents) {
  */
 std::uint16_t
 map_port(int port) {
-  // TODO: Ensure port is in the range of 21-65535
+  // Ensure port is in the range of 1024-65535
+  if (port < 1024 || port > 65535) {
+    BOOST_LOG(warning) << "Port out of range: "sv << port;
```
TheElixZammuto, line 841: While ports < 1024 are root-only, these are still valid ports. I think in this specific case making the warning is enough

src_assets/common/assets/web/config.html
```diff
@@ -625,14 +625,74 @@ <h1 class="my-4">Configuration</h1>
...
+              <td>UDP</td>
+              <td>{{+effectivePort + 11}}</td>
+            </tr>
+            <tr>
+              <td>Mic (unused)</td>
+              <td>UDP</td>
+              <td>{{+effectivePort + 13}}</td>
+            </tr>
+          </tbody>
+        </table>
```
TheElixZammuto, line 695: I would avoid to specify which port are what for. People would wrongly assume that they can choose which port might enable, while it's not the case

ReenigneArcher:
> Due to NAT (and hairpin), having Sunshine itself make the test is not representitive, the Internet Streaming Tool uses a relay server IIRC

I wanted to add two columns.

- Show if port is available on the localhost
- Show if port is available using public ip address (maybe we can use this site somehow? https://canyouseeme.org/)

src_assets/common/assets/web/config.html
```diff
@@ -625,14 +625,74 @@ <h1 class="my-4">Configuration</h1>
...
           v-model="config.port"
         />
         <div class="form-text">Set the family of ports used by Sunshine</div>
+        <!-- Add warning if any port is less than 1024 -->
+        <div class="alert alert-danger" v-if="(+effectivePort - 5) < 1024">
+          <b>Warning:</b> Ports below 1024 are not recommended!
+        </div>
+        <!-- Add warning if any port is above 65535 -->
+        <div class="alert alert-danger" v-if="(+effectivePort + 21) > 65535">
+          <b>Danger:</b> Ports above 65535 are not available!
```
TheElixZammuto, line 642:
```suggestion
          <b>Error:</b> Ports above 65535 are not available!
```

TheElixZammuto (changes requested):

src/main.cpp
ReenigneArcher: This condition should actually never happen, as I modified the range for config port 1029-65514 (to account for the difference from the config port (-5, +21)

ReenigneArcher:
@cgutman This will appear if the user specifies a non default port.

image

ReenigneArcher:
Modified UI to:

Default:
image

Non Default:
image

Too Low:
image

Too High:
image

ReenigneArcher:
Added these warnings if web ui is allowed on WAN.

image

image

TheElixZammuto (approved):

Issue #1668, opened 2023-09-23
[Arch] NVENC not being recognised


r0m5s:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

After updating arch to Linux 6.5 and Nvidia 535.104.05 drivers with the nvidia-utils-nvlax 535.104.05-1, sunshine stopped recognising the encoder and therefore starting. I'm running my desktop headless with a HDMI dongle.

I have a backup before the changes and If I get back to 6.4 kernel there's no issues starting the stream. Already tried building the nightly build and rebuilding my install and still nothing. I also get a error 503.

I know that the NvFBC patch is working since the output from ffmpeg test lines (from keylase/nvidia-patch) work without any issues.

### Expected Behavior

Stream should start without any errors

### Host Operating System

Linux

### Operating System Version

Linux 6.5.4-zen2-1-zen #1 ZEN SMP PREEMPT_DYNAMIC

### Architecture

64 bit

### Sunshine commit or version

0.20.0

### Package

Linux - AUR (Third Party)

### GPU Type

Nvidia

### GPU Model

GeForce GTX 1050 Ti

### GPU Driver/Mesa Version

535.104.05

### Capture Method (Linux Only)

NvFBC

### Config

```Shell
min_log_level = 1
nvenc_preset = 4
fps = [30,60,90,120]
resolutions = [
    352x240,
    480x360,
    858x480,
    1280x720,
    1920x1080,
    2560x1080,
    3440x1440,
    1920x1200,
    3840x2160,
    3840x1600,
    1280x800
]
capture = nvfbc
encoder = nvenc
output_name = 1
```

### Relevant log output

```shell
[2023:09:23:11:21:08]: Info: Sunshine version: 0.20.0
[2023:09:23:11:21:08]: Error: Unable to initialize capture method
[2023:09:23:11:21:08]: Error: Platform failed to initialize
[2023:09:23:11:21:08]: Info: Trying encoder [nvenc]
[2023:09:23:11:21:08]: Info: System tray created
[2023:09:23:11:21:09]: Info: Encoder [nvenc] failed
[2023:09:23:11:21:09]: Error: Couldn't find any working encoder matching [nvenc]
[2023:09:23:11:21:09]: Info: // Testing for available encoders, this may generate errors. You can safely ignore those errors. //
[...]

PR #1689, opened 2023-10-01
Add problems with nvcc PATH to linux troubleshooting

KuleRucket:
Add issue with nvcc path to troubleshooting

## Description
Some people have trouble using nvfbc cature even though everything appears to be set up correctly. This is because nvidia packages for cuda don't always set up the PATH correctly to nvcc. This adds an item to troubleshooting to document this problem.

### Screenshot
n/a

### Issues Fixed or Closed
Documents #1668

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/troubleshooting/linux.rst
+++ b/docs/source/troubleshooting/linux.rst
@@ -30,6 +30,15 @@ If screencasting fails with KMS, you may need to run the following to force unpr
 
       sudo setcap -r $(readlink -f $(which sunshine))
 
+NvFBC Capture fails
+-------------------
+Some users have had issues using nvfbc even though they have correctly patched nvidia drivers and cuda is installed. The problem occurs when nvcc is not in the PATH and cuda is not correctly compiled into Sunshine. This will not cause an error because not all users have nvidia cards. The solution is to ensure that nvcc is in your PATH before building sunshine. For example add the following to ~/.bashrc:
+
+   .. code-block:: bash
+
+      export PATH=/usr/local/cuda/bin:$PATH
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Should we just update the arch pkgbuild to export those paths?

KuleRucket:
> Should we just update the arch pkgbuild to export those paths?

Where could this be exported though? It's a;ready done for bash in /etc/profile by the cuda package but you would have to do it in different places for fish, dash, zsh etc. Also it still wouldn't solve the problem for other distros if they install cuda to a different location not in the path.

ReenigneArcher:
I was thinking here: https://github.com/LizardByte/Sunshine/blob/f76879eb0dc287f09ae3810bd61b191f21344f78/packaging/linux/Arch/PKGBUILD#L81-L82

KuleRucket:
That would be good for Arch people but I still think we should include something in troubleshooting as I don't think it would be an arch specific problem.

ReenigneArcher:
Out of release files, I believe this would only affect Arch since the PKGBUILD is just a build spec. Everything else is already compiled and not affected by these variables.

Otherwise it would only affect those who self compile.

ReenigneArcher:
Can you put this into the build docs, instead of troubleshooting? I believe in troubleshooting it will confuse users who are using pre-built binaries.

https://github.com/LizardByte/Sunshine/blob/a9f2679a491d6e73aa7ee361159d3963b1be35b3/docs/source/building/linux.rst?plain=1#L177

PR #1706, opened 2023-10-05
Initial linux touch implementation

ap4ss3rby:
I may or may not have missed a lot of details, but I hope someone else can fix my semi-working MT code to properly work.

Needs a properly patched Moonlight client for android testing with [this](https://github.com/moonlight-stream/moonlight-android/commit/394a57a26dcda8dd026592c79cf4981194b99c92) patch reverted

## Description

Currently, the code in this PR does the following:
- ~~Break Windows and macOS builds because I couldn't be bothered to fix their touch() functions for the touch device parameter~~
- Report that we infact have touch capability on linux hosts 
- Probably break absolute mouse mode even more
- Report native touch events using uinput. Making stuff like streaming waydroid to a phone for playing Android games at better quality than just playing directly on the phone possible
- ~~Not do multitouch because uinput is stupid and doesn't like how I'm feeding it multitouch data~~

## Type of Change

## Checklist

## Branch Updates


ap4ss3rby:
After fixing a bunch of mistakes with my understanding of uinput and implementing what I learned I was able to get at least 8 points recognized (couldn't test more because I ran out of space on my phone haha)

src/platform/linux/input.cpp
```diff
@@ -1903,6 +1963,8 @@ namespace platf {
    */
   platform_caps::caps_t
   get_capabilities() {
-    return 0;
+    platform_caps::caps_t caps = 0;
+    caps |= platform_caps::pen_touch;
```
cgutman, line 1967: You also have to implement `pen()` or Moonlight will start sending native pen events that just get ignored.

src/platform/linux/input.cpp
```diff
@@ -1756,6 +1812,10 @@ namespace platf {
     libevdev_enable_event_type(dev.get(), EV_ABS);
     libevdev_enable_event_code(dev.get(), EV_ABS, ABS_X, &absx);
     libevdev_enable_event_code(dev.get(), EV_ABS, ABS_Y, &absy);
+    libevdev_enable_event_code(dev.get(), EV_ABS, ABS_MT_SLOT, &mtslot);
+    libevdev_enable_event_code(dev.get(), EV_ABS, ABS_MT_TRACKING_ID, &trkid);
+    libevdev_enable_event_code(dev.get(), EV_ABS, ABS_MT_POSITION_X, &absx);
+    libevdev_enable_event_code(dev.get(), EV_ABS, ABS_MT_POSITION_Y, &absy);
```
cgutman, line 1818: You should use a separate uinput device for pen/touch input, create it on demand (on the first pen/touch event), and store it in the `client_input_t`. You can use the Windows implementation as reference.

src/platform/linux/input.cpp
```diff
@@ -1583,10 +1582,51 @@ namespace platf {
...
+    auto scaled_x = (int) std::lround((x + touch_port.offset_x) * ((float) target_touch_port.width / (float) touch_port.width));
+    auto scaled_y = (int) std::lround((y + touch_port.offset_y) * ((float) target_touch_port.height / (float) touch_port.height));
+
+
+    //This reports touch events as a type B uinput touchscreen/tablet
+    libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_SLOT, touch.pointerId);
+    libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_TRACKING_ID, id);
+    libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_POSITION_X, scaled_x);
+    libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_POSITION_Y, scaled_y);
+    libevdev_uinput_write_event(touchscreen, EV_SYN, SYN_REPORT, 0);
```
cgutman, line 1629: You should also implement ABS_MT_TOOL_TYPE, ABS_MT_PRESSURE, ABS_MT_ORIENTATION, ABS_MT_TOUCH_MAJOR, and ABS_MT_TOUCH_MINOR.

ABS_MT_TOOL_TYPE should always be MT_TOOL_FINGER in this codepath.

ABS_MT_PRESSURE, ABS_MT_TOUCH_MAJOR, and ABS_MT_TOUCH_MINOR should be pretty simple. Just scale the 0-1 float to some arbitrary integer range (touch major and minor have to be same range as X/Y) and be sure to handle the "unknown" cases. See the API docs here: https://github.com/moonlight-stream/moonlight-common-c/blob/d457fbb48714d4ce79de2594180dfd73bc5f070c/src/Limelight.h#L594-L629

Finally, ABS_MT_ORIENTATION might need some extra conversion logic to get to the evdev format.

You can reuse all this logic for the pen implementation too, in addition to adding ABS_MT_DISTANCE, BTN_STYLUS/BTN_STYLUS2, and ABS_MT_TOOL_TYPE. I _think_ ABS_TILT_X and ABS_TILT_Y is also valid for stylus devices.

src/platform/common.h
```diff
@@ -652,9 +652,10 @@ namespace platf {
    * @param input The client-specific input context.
    * @param touch_port The current viewport for translating to screen coordinates.
    * @param touch The touch event.
+   * @param input_ctx The global input context.
    */
   void
-  touch(client_input_t *input, const touch_port_t &touch_port, const touch_input_t &touch);
+  touch(client_input_t *input, const touch_port_t &touch_port, const touch_input_t &touch, input_t &input_ctx);
```
cgutman, line 658: If you need the global input context, you should store it inside your `client_input_t` like the Windows implmentation does

src/platform/linux/input.cpp
```diff
@@ -1583,10 +1582,51 @@ namespace platf {
...
+        break;
+    case LI_TOUCH_EVENT_CANCEL_ALL :
+        for (int i; i < 10; i++) {
+          libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_SLOT, i);
+          libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_TRACKING_ID, -1);
+        }
+        libevdev_uinput_write_event(touchscreen, EV_SYN, SYN_REPORT, 0);
+        return;
+    default:
+        id = touch.pointerId + 4;
```
cgutman, line 1613: Why the `+ 4`?

src/platform/linux/input.cpp
```diff
@@ -1583,10 +1582,51 @@ namespace platf {
...
+          id = -1;
+        break;
+    case LI_TOUCH_EVENT_CANCEL_ALL :
+        for (int i; i < 10; i++) {
+          libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_SLOT, i);
+          libevdev_uinput_write_event(touchscreen, EV_ABS, ABS_MT_TRACKING_ID, -1);
+        }
+        libevdev_uinput_write_event(touchscreen, EV_SYN, SYN_REPORT, 0);
+        return;
+    default:
```
cgutman, line 1612: You should explicitly check for LI_TOUCH_EVENT_DOWN and LI_TOUCH_EVENT_MOVE here and drop other events.

cgutman (changes requested):

src/platform/linux/input.cpp
ap4ss3rby: Do you mean adding an extra case to handle these 2 events and ? I need to know when the touch point is no longer active because it got released or cancelled as a type B uinput touchscreen doesn't care if points suddenly appear out of thin air but it wants to know when they end by signalling id -1 to the appropriate touch slot.

src/platform/linux/input.cpp
ap4ss3rby: Good catch, I forgot to remove the + 4 while I was cleaning up. I was testing a theory I had about why multitouch didn't work in the first version of this PR

src/platform/linux/input.cpp
ap4ss3rby: So we're going towards a full fat implementation instead of a basic one. Got it. Keep in mind I dropped some tablet related events since they seemed to cause problems with KDE Plasma and the easiest solution was to just drop tablet related events.

 If I can get around them I'll implement them into one device type. Otherwise I'll split touchscreen and tablet types if we're gonna create a new global context whenever we recieve a touch/tablet packet just to save myself a bit of headache

src/platform/linux/input.cpp
ap4ss3rby: Will implement but will require external testing since I don't have anything capable of doing pen events.

cgutman: I can test with my devices

src/platform/linux/input.cpp
cgutman: No, I didn't mean dropping everything other than down/move. I meant just dropping the events you don't explicitly handle (ie the `default:` case). Rather than having `default:` write an input event (assuming the event is a down or move), instead explicitly add cases for LI_TOUCH_EVENT_DOWN and LI_TOUCH_EVENT_MOVE, then make `default:` just return.

src/platform/linux/input.cpp
cgutman: > If I can get around them I'll implement them into one device type. Otherwise I'll split touchscreen and tablet types if we're gonna create a new global context whenever we recieve a touch/tablet packet just to save myself a bit of headache

I assume you mean splitting into separate uinput devices. That's totally fine if it's more convenient to expose separate pen and multitouch touchscreen devices.

src/platform/linux/input.cpp
ap4ss3rby: Done exactly as requested

src/platform/common.h
ap4ss3rby: Done

LegacyGamerHD:
@ap4ss3rby  do you know why on wayland i only get about 2 touch inputs

LegacyGamerHD:
on kde it works fine do the mouse cursor shows up

ReenigneArcher:
@ap4ss3rby are you able to implement the remaining changes requested by @cgutman?

PR #1712, opened 2023-10-07
Add missing dependencies for tray icon support

zackelia:
## Description

The deb from the Docker image does not have all of its dependencies listed when packaged with `SUNSHINE_TRAY=1`:
```sh
$ ldd $(which sunshine) | grep "not found"
	libayatana-appindicator3.so.1 => not found
	libnotify.so.4 => not found
```

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/cmake/packaging/linux.cmake
+++ b/cmake/packaging/linux.cmake
@@ -79,7 +79,8 @@ if(${SUNSHINE_TRAY} STREQUAL 1)
 
     set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
                     ${CPACK_DEBIAN_PACKAGE_DEPENDS}, \
-                    libappindicator3-1")
+                    libayatana-appindicator3-1, \
+                    libnotify4")
     set(CPACK_RPM_PACKAGE_REQUIRES "\
                     ${CPACK_RPM_PACKAGE_REQUIRES}, \
                     libappindicator-gtk3 >= 12.10.0")


cmake/packaging/linux.cmake
```diff
@@ -79,7 +79,8 @@ if(${SUNSHINE_TRAY} STREQUAL 1)
 
     set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
                     ${CPACK_DEBIAN_PACKAGE_DEPENDS}, \
-                    libappindicator3-1")
+                    libayatana-appindicator3-1, \
+                    libnotify4")
```
ReenigneArcher, line 83: This will be sufficient. `libnotify4` is a dependency of `libc6`, which is a dependency of `libayatana-appindicator3-1`.

```suggestion
                    libayatana-appindicator3-1)
```

ReenigneArcher (changes requested):
Thank you for the PR!

cmake/packaging/linux.cmake
cgutman: I think you've got the dependency direction flipped. libnotify4 depends on libc6, not the other way around.

zackelia: Thanks for reviewing! One clarification first though, I agree that `libc6` is a dependency of `libayatana-appindicator3-1`, but it seems to me that `libc` is also a dependency of `libnotify4` (not the other way around).

Using the docker image, if I install the Sunshine deb and `libayatana-appindicator3-1` I am still lacking `libnotify4.so.4`. It seems to me that `libnotify4` is necessary, do you agree?

zackelia: @cgutman Looks like you just beat me to it :)

ReenigneArcher: Okay, my mistake.

ReenigneArcher (approved):

cmake/packaging/linux.cmake
```diff
@@ -79,7 +79,8 @@ if(${SUNSHINE_TRAY} STREQUAL 1)
...
     set(CPACK_RPM_PACKAGE_REQUIRES "\
                     ${CPACK_RPM_PACKAGE_REQUIRES}, \
                     libappindicator-gtk3 >= 12.10.0")
```
cgutman, line 86: Do we need to fix this one too?

ReenigneArcher: Probably... https://packages.fedoraproject.org/pkgs/libnotify/libnotify/

zackelia: Using a vanilla Fedora 38 docker image, seems fine to me - no missing dependencies.

```
yum check-update
yum localinstall /sunshine.rpm
sunshine
```

I'm not exactly sure what in the tree is installing libnotify.so.4 but it's there after the `localinstall`.

Maybe in a future effort, the GitHub actions can start containers for the images they built, install the Sunshine package, and run a quick `sunshine --version` just to make sure that all the dependencies are in order.

ReenigneArcher: It's probably included in Fedora distros, but maybe we should add it for more minimal distros.

The sunshine package is actually installed in the docker images... but we don't run it right now. As far as I understand this would add an additional layer to the docker image which would not be desirable (although I don't think it would necessarily be the end of the world)

https://github.com/LizardByte/Sunshine/blob/56d13bc7e84db160dfbea1cbf94fd1722f5e8b32/docker/fedora-38.dockerfile#L135-L142

zackelia: According to `repoquery --requires /sunshine.rpm --tree`, libnotify is a top level dependency, just not sure where it's getting added in the CMake...

As for the Docker image, I agree we shouldn't do this test in the Dockerfile itself. My suggestion was in the GitHub Actions, after the Docker image is built, spin up a container using that image and make sure Sunshine can be executed. Not sure if this is possible as I'm not too familiar with GitHub actions.

ReenigneArcher: > just not sure where it's getting added in the CMake

My guess is `libappindicator-gtk3`, not 100% sure though. I can't see anyway to find dependencies on the fedora package site. This seems logical though as we kept this legacy appindicator on fedora builds.

ReenigneArcher: I will merge this and we can deal with rpm later if it becomes an issue. Thank you!

PR #1772, opened 2023-10-22
Manual DS4 mode: Allow Back/Select to trigger Touchpad click

ekchang:
## Description
When forcing manual DS4 emulation, there isn't a natural way to trigger Touchpad click for standard XInput. Current impl supports passing MISC_BUTTON (Xbox Series X Share button), but Steam Input doesn't expose this mapping, and most other XInput controllers on client don't have this button available.

Allow additional accessibility for clients wishing to emulate DS4 by having Back/Select also trigger Touchpad click. In most games supporting native DS4 input, Back/Share does not do anything anyway and typically should translate to a Touchpad click.

My specific usecase here is using Sunshine via Steam Deck Moonlight client. A DualShock 4 or DualSense controller configured via Steam Input cannot map the Touchpad click to Xbox 'Share' and the only way to pass through Touchpad clicks through to Sunshine is by disabling Steam Input altogether. 

### Screenshot
Screenshot 2023-10-21 194002

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -113,7 +113,7 @@ namespace config {
 
     std::string gamepad;
     bool ds4_back_as_touchpad_click;
-    
+
     bool keyboard;
     bool mouse;
     bool controller;
--- a/src/config.h
+++ b/src/config.h
@@ -112,7 +112,7 @@ namespace config {
     std::chrono::duration<double> key_repeat_period;
 
     std::string gamepad;
-    bool ds4_back_as_touchpad;
+    bool ds4_back_as_touchpad_click;
     
     bool keyboard;
     bool mouse;
--- a/src/config.h
+++ b/src/config.h
@@ -112,7 +112,8 @@ namespace config {
     std::chrono::duration<double> key_repeat_period;
 
     std::string gamepad;
-
+    bool ds4_back_as_touchpad;
+    
     bool keyboard;
     bool mouse;
     bool controller;
--- a/src/platform/windows/input.cpp
+++ b/src/platform/windows/input.cpp
@@ -1333,11 +1333,9 @@ namespace platf {
 
     // Allow either PS4/PS5 clickpad button or Xbox Series X share button to activate DS4 clickpad
     if (gamepad_state.buttonFlags & (TOUCHPAD_BUTTON | MISC_BUTTON)) buttons |= DS4_SPECIAL_BUTTON_TOUCHPAD;
-    
+
     // Manual DS4 emulation: check if BACK button should also trigger DS4 touchpad click
-    if (config::input.gamepad == "ds4"sv 
-      && config::input.ds4_back_as_touchpad_click
-      && (gamepad_state.buttonFlags & BACK)) buttons |= DS4_SPECIAL_BUTTON_TOUCHPAD;
+    if (config::input.gamepad == "ds4"sv && config::input.ds4_back_as_touchpad_click && (gamepad_state.buttonFlags & BACK)) buttons |= DS4_SPECIAL_BUTTON_TOUCHPAD;
 
     return (DS4_SPECIAL_BUTTONS) buttons;
   }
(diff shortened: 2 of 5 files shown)


src/platform/windows/input.cpp
```diff
@@ -1333,6 +1333,11 @@ namespace platf {
 
     // Allow either PS4/PS5 clickpad button or Xbox Series X share button to activate DS4 clickpad
     if (gamepad_state.buttonFlags & (TOUCHPAD_BUTTON | MISC_BUTTON)) buttons |= DS4_SPECIAL_BUTTON_TOUCHPAD;
+    
+    // Manual DS4 emulation: check if BACK button should also trigger DS4 touchpad click
+    if ((config::input.gamepad == "ps4"sv || config::input.gamepad == "ds4"sv)
```
ReenigneArcher, line 1338: Unless something changed recently, I don't think `ps4` is a valid option.

src_assets/common/assets/web/config.html
```diff
@@ -105,6 +105,29 @@ <h1 class="my-4">Configuration</h1>
         </select>
         <div class="form-text">Choose which type of gamepad to emulate on the host</div>
       </div>
+      <div class="accordion" v-if="config.gamepad === 'ds4'">
```
ReenigneArcher, line 108: Does this auto update when the gamepad option is changed? I had an issue recently where I did something similar and the ui didn't update upon change.

src_assets/common/assets/web/config.html
```diff
@@ -1140,6 +1163,7 @@ <h2 class="accordion-header">
     "encoder": "",
     "fps": "[10,30,60,90,120]",
     "gamepad": "auto",
+    "ds4_back_as_touchpad": "enabled",
```
ReenigneArcher, line 1166: Alphabetical please.

ReenigneArcher (changes requested):
Thank you for the PR!

In addition to the below comments. I would also request the docs are updated. All config options are in the "advanced usage" section.

Also, perhaps this option should be named "ds4_back_as_touchpad_click" or something along those lines?

src/platform/windows/input.cpp
ekchang: Yeah I wasn't sure if that was still valid so I kept it consistent with https://github.com/ekchang/Sunshine/blob/5b052d7c2c059aa26fdf79cc87885af2f461f544/src/platform/windows/input.cpp#L1177-L1179

But I'll go ahead and update if we're confident ds4 is the only supported config here

src_assets/common/assets/web/config.html
ekchang: It does! I can confirm that switching to auto and x360 will collapse this accordion from being visible in realtime.

src/platform/windows/input.cpp
ReenigneArcher: Thanks for the background... I think we should probably remove "ps4" from there as well. Would you take care of it in this PR?

src_assets/common/assets/web/config.html
ReenigneArcher: Great, thanks for confirming!

ekchang:
Updated docs:

Screenshot 2023-10-22 010310

ReenigneArcher (dismissed):

src_assets/common/assets/web/config.html
ReenigneArcher: Thanks, and I now see others that are out of order. I'll correct them in a separate PR.

src/config.h
```diff
@@ -112,7 +112,8 @@ namespace config {
     std::chrono::duration<double> key_repeat_period;
 
     std::string gamepad;
-
+    bool ds4_back_as_touchpad_click;
+
```
ReenigneArcher, line 116:
```suggestion
    bool ds4_back_as_touchpad_click;

```

PR #1798, opened 2023-10-29
Expose more nvenc and nvprefs options

ns6089:
### Screenshot

2023-10-29 13_42_26-Sunshine — Mozilla Firefox

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -1045,6 +1045,59 @@ nvenc_twopass
 
       nvenc_twopass = quarter_res
 
+nvenc_spatial_aq
+^^^^^^^^^^^^^^^^
+
+**Description**
+   Assign higher QP values to flat regions of the video.
+   Recommended to enable when streaming at lower bitrates.
+
+   .. Note:: This option only applies when using NVENC `encoder`_.
+
+**Choices**
+
+.. table::
+   :widths: auto
+
+   ========== ===========
+   Value      Description
+   ========== ===========
+   disabled   Don't enable Spatial AQ (faster)
+   enabled    Enable Spatial AQ (slower)
+   ========== ===========
+
+**Default**
+   ``disabled``
+
+**Example**
+   .. code-block:: text
+
+      nvenc_spatial_aq = disabled
+
+nvenc_vbv_increase
+^^^^^^^^^^^^^^^^^^
+
+**Description**
+   Single-frame VBV/HRD percentage increase.
+   By default sunshine uses single-frame VBV/HRD, which means any encoded video frame size is not expected to exceed requested bitrate divided by requested frame rate.
+   Relaxing this restriction can be beneficial and act as low-latency variable bitrate, but may also lead to packet loss if the network doesn't have buffer headroom to handle bitrate spikes.
+   Maximum accepted value is 400, which corresponds to 5x increased encoded video frame upper size limit.
+
+   .. Note:: This option only applies when using NVENC `encoder`_.
+
+   .. Warning:: Can lead to network packet loss.
+
+**Default**
+   ``0``
+
+**Range**
+   ``0-400``
+
+**Example**
+   .. code-block:: text
+
+      nvenc_vbv_increase = 0
+
 nvenc_realtime_hags
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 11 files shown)


src_assets/common/assets/web/config.html
```diff
@@ -1171,8 +1235,11 @@ <h2 class="accordion-header">
     "min_log_level": 2,
     "mouse": "enabled",
     "nvenc_h264_cavlc": "disabled",
+    "nvenc_latency_over_power": "enabled",
+    "nvenc_opengl_vulkan_on_dxgi": "enabled",
     "nvenc_preset": "1",
     "nvenc_realtime_hags": "enabled",
+    "nvenc_spatial_aq": "disabled",
```
ReenigneArcher, line 1242:
```suggestion
    "nvenc_spatial_aq": "disabled",
    "nvenc_vbv_increase": 0,
```

docs/source/about/advanced_usage.rst
```diff
@@ -1077,6 +1130,96 @@ nvenc_realtime_hags
...
+   .. Note:: This option only applies when using NVENC `encoder`_.
+
+   .. Note:: Doesn't apply to AV1 since it uses predetermined value as part of specification.
+
+   .. Warning:: Can lead to (or remedy) decoding artifacts.
+
+   .. Caution:: Applies to Windows only.
+
+**Default**
+   Variable is unset and sunshine will pick whatever it thinks is the best.
```
ReenigneArcher, line 1150: So the default is not 5?

Generally, placeholders in the config ui match the default.

src_assets/common/assets/web/config.html
```diff
@@ -1171,8 +1235,11 @@ <h2 class="accordion-header">
     "min_log_level": 2,
     "mouse": "enabled",
     "nvenc_h264_cavlc": "disabled",
+    "nvenc_latency_over_power": "enabled",
+    "nvenc_opengl_vulkan_on_dxgi": "enabled",
```
ReenigneArcher, line 1239:
```suggestion
    "nvenc_h264_hevc_ref_frames": "",
    "nvenc_latency_over_power": "enabled",
    "nvenc_opengl_vulkan_on_dxgi": "enabled",
```

src_assets/common/assets/web/config.html
```diff
@@ -1000,6 +1025,45 @@ <h2 class="accordion-header">
                   performance when the GPU is heavily loaded.
                 </div>
               </div>
+              <div class="mb-3" v-if="platform === 'windows'">
+                <label for="nvenc_h264_hevc_ref_frames" class="form-label">Default value for H.264 and HEVC reference frames
+                  in DPB</label>
+                <input type="number" min="1" max="15" class="form-control" id="nvenc_h264_hevc_ref_frames" placeholder="5"
```
ReenigneArcher, line 1031:
```suggestion
                <input type="number" min="1" max="15" class="form-control" id="nvenc_h264_hevc_ref_frames"
```

ReenigneArcher (changes requested):

docs/source/about/advanced_usage.rst
ns6089: Current default is constant 5, but I can't guarantee it will stay as single constant value in the future and not change dynamically depending on let's say video resolution.

ReenigneArcher: Okay, if it changes in the future to an auto mode this can be adjusted accordingly.

For now, I think this should be `5`. Also, as long as the config UI uses the correct defaults, then it allows us to modify the default when a user updates. As long as they are using our old default, they will also start using our new default.

ReenigneArcher, line 1150:
```suggestion
   5
```

docs/source/about/advanced_usage.rst
```diff
@@ -1077,6 +1130,96 @@ nvenc_realtime_hags
 
       nvenc_realtime_hags = enabled
 
+nvenc_h264_hevc_ref_frames
```
cgutman, line 1133: I'm not sure we should expose this option to users (at least in this fully-configurable form). The number of reference frames is limited by the codec specification and is resolution-dependent (as you mentioned below). 

It would be easy for a user to end up in a situation where they specify a value here that works for 720p then later switch to 4K and wonder why things are totally broken, or they specify a value that works for HEVC but not H.264.

I think if we want to expose this option, we should do so in a way that only exposes a few known good options. For now, maybe we expose "Auto" (our current behavior) and "1". If we later decide to increase reference frames up to the codec maximum for each resolution, we can either change the behavior of "Auto" and explicitly add "5" as an option or add an option for "Maximum", depending on how the decoder compatibility looks in our testing.

PR #1817, opened 2023-11-04
build(cmake): Use GNUInstallDirs to install data and lib directories

tgurr:
## Description
GNUInstallDirs (https://cmake.org/cmake/help/v3.18/module/GNUInstallDirs.html) is a CMake module which provides similar params as autotools allowing to adjust where files are installed.

On multiarch and/or cross layouts the prefix might be something like /usr/${arch} but arch independent files should still go into /usr/share.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/cmake/packaging/linux.cmake
+++ b/cmake/packaging/linux.cmake
@@ -9,9 +9,9 @@ if(${SUNSHINE_BUILD_APPIMAGE} OR ${SUNSHINE_BUILD_FLATPAK})
             DESTINATION "${SUNSHINE_ASSETS_DIR}/systemd/user")
 else()
     install(FILES "${SUNSHINE_SOURCE_ASSETS_DIR}/linux/misc/85-sunshine.rules"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/udev/rules.d")
+            DESTINATION "${CMAKE_INSTALL_LIBDIR}/udev/rules.d")
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sunshine.service"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/systemd/user")
+            DESTINATION "${CMAKE_INSTALL_LIBDIR}/systemd/user")
 endif()
 
 # Post install
@@ -63,19 +63,19 @@ set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS OFF)
 
 # application icon
 install(FILES "${CMAKE_SOURCE_DIR}/sunshine.svg"
-        DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/apps")
+        DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/apps")
 
 # tray icon
 if(${SUNSHINE_TRAY} STREQUAL 1)
     install(FILES "${CMAKE_SOURCE_DIR}/sunshine.svg"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/status"
+            DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/status"
             RENAME "sunshine-tray.svg")
     install(FILES "${SUNSHINE_SOURCE_ASSETS_DIR}/common/assets/web/public/images/sunshine-playing.svg"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/status")
+            DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/status")
     install(FILES "${SUNSHINE_SOURCE_ASSETS_DIR}/common/assets/web/public/images/sunshine-pausing.svg"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/status")
+            DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/status")
     install(FILES "${SUNSHINE_SOURCE_ASSETS_DIR}/common/assets/web/public/images/sunshine-locked.svg"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/status")
+            DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/scalable/status")
 
     set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
                     ${CPACK_DEBIAN_PACKAGE_DEPENDS}, \
@@ -89,17 +89,17 @@ endif()
 # desktop file
 # todo - validate desktop files with `desktop-file-validate`
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sunshine.desktop"
-        DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications")
+        DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications")
 if(NOT ${SUNSHINE_BUILD_APPIMAGE})
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sunshine_terminal.desktop"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications")
+            DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications")
 endif()
 if(${SUNSHINE_BUILD_FLATPAK})
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sunshine_kms.desktop"
-            DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications")
+            DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications")
 endif()
 
 # metadata file
 # todo - validate file with `appstream-util validate-relax`
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sunshine.appdata.xml"
-        DESTINATION "${CMAKE_INSTALL_PREFIX}/share/metainfo")
+        DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/metainfo")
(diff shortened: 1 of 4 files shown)


tgurr:
Please note I couldn't test building/running the flatpak package as I only do distribution packaging, please be sure to test the change there as well.

ReenigneArcher:
I don't see this as being better or worse than what we have right now. It's just different; therefore, I'm going to close this.

tgurr:
@ReenigneArcher Please reopen as this actually fixes being able to install this on distributions like Exherbo which have such a said layout and we need to manually patch this wrong behaviour downstream right now:

https://gitlab.exherbo.org/exherbo/net/-/blob/ee333903564a57dba5b1b88c4d7e29171dc5d458/packages/net-remote/sunshine/sunshine-0.21.0.exheres-0#L185

Also see this being fixed for many other projects in the same way (and not done by Exherbo devs so there is actually a need for this not only for Exherbo) just one quick recent example:

https://github.com/paullouisageneau/libdatachannel/commit/f1aa6c83a792af2003fa712494c942ee70639604

Conan-Kudo:
I fixed it in libdatachannel because otherwise it was broken on Fedora. That's also the case here.

packaging/linux/flatpak/dev.lizardbyte.sunshine.yml
```diff
@@ -317,6 +317,7 @@ modules:
       - cd ${FLATPAK_BUILDER_BUILDDIR} && npm install
     config-opts:
       - -DCMAKE_BUILD_TYPE=Release
+      - -DCMAKE_INSTALL_DATAROOTDIR=/app/share
```
ReenigneArcher, line 320: I think this needs to be added to the following locations, unless the default is okay. Please confirm.

- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/.github/workflows/CI.yml#L402-L412
- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/packaging/linux/Arch/PKGBUILD#L80-L86
- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/docker/debian-bookworm.dockerfile#L105-L115
- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/docker/debian-bullseye.dockerfile#L119-L129
- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/docker/fedora-38.dockerfile#L106-L115
- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/docker/fedora-39.dockerfile#L106-L115
- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/docker/ubuntu-20.04.dockerfile#L155-L165
- https://github.com/LizardByte/Sunshine/blob/11d472362cb1f5896c8a2d3de4b2184a2f0a37ad/docker/ubuntu-22.04.dockerfile#L120-L130

CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ and Nvidia GPUs for hardware encoding. Software encoding is also available. You
 Moonlight client on a variety of devices. A web UI is provided to allow configuration, and client pairing, from \
 your favorite web browser. Pair from the local server or any mobile device.")
 
+include(GNUInstallDirs)
```
ReenigneArcher, line 17: Can this be put into a linux specific cmake file?

ReenigneArcher (commented):
Please also fix conflicts.

packaging/linux/flatpak/dev.lizardbyte.sunshine.yml
tgurr: For regular Linux distributions the defaults which CMake with the use of GNUInstallDirs figures out **should** be okay, as thats main purpose of it.  (See also the comment from @Conan-Kudo above in regards to lib vs lib32/lib64 handling, which for example we don't have on Exherbo as we set different arch-specific prefixes instead.)

CMakeLists.txt
tgurr: Not an expert here but according to https://code-review.googlesource.com/c/re2/+/43990/5#message-36d7112c2c43973dcb63c5160937fb1981702010 which I found while googling for it that should be okay for other OS'es as well and I personally I didn't come across any special handling for GNUinstallDirs yet. Current re2 also has still has that in their main MakeLists.txt (https://github.com/google/re2/blob/main/CMakeLists.txt#L11) as well as cross-platform stuff like SDL2 (https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L39).

tgurr:
> Please also fix conflicts.

Rebased on current nightly.

packaging/linux/flatpak/dev.lizardbyte.sunshine.yml
tgurr: Thinking about it it's probably also not required for the flatpak as well as it sets  `-DCMAKE_INSTALL_PREFIX=/app` so stuff should properly end up being installed there.

CMakeLists.txt
ReenigneArcher: > should be okay for other OS'es as well

That wasn't the reason for my request. It's more to keep the code separated. I went through a great effort trying to separate out the former monster of a CMakeLists.txt and don't want to head back in that direction if it can be avoided.

tgurr: Please advice in which file you'd like to see it moved then.

ReenigneArcher: Probably at the top of `./cmake/packaging/unix.cmake`

packaging/linux/flatpak/dev.lizardbyte.sunshine.yml
detiam: > Probably at the top of `./cmake/packaging/unix.cmake`

@tgurr @ReenigneArcher `special_package_configuration.cmake` is include before `unix.cmake` been include in CMakeLists.txt, therefore `GNUInstallDirs` is not enabled when configure desktop file, the `@CMAKE_INSTALL_DATAROOTDIR@` in `.desktop` files is invalid and empty string, this broken both regular and flatpak version of the desktop file in v0.22.0 release. https://github.com/LizardByte/Sunshine/blob/faeeb7ec573590c798111024510b2d46c03beaeb/packaging/linux/sunshine.desktop#L15 but when I try to add `include(GNUInstallDirs)` to `special_package_configuration.cmake` and generate the desktop file again,
```
Exec=gio launch share/applications/sunshine_terminal.desktop
```
the `/usr/` in path is missing, a cmake bug? my distro is archlinux.
so how to fix this, revert the change for these desktop file, add `-DCMAKE_INSTALL_DATAROOTDIR="/usr/share"` in PKGBUILD and anywhere needed, or wait cmake fix `@CMAKE_INSTALL_DATAROOTDIR@` in `configure_file()`?

tgurr: Adding `-DCMAKE_INSTALL_DATAROOTDIR:PATH=/usr/share` where needed appears to be the correct approach to me, e.g. on Exherbo we have to do that anyways because we need to install arch-independent stuff into /usr and arch-dependent stuff into prefix e.g. /usr/x86_64-pc-linux-gnu like bin, lib and so on which was the reason for this MR in the first place. Unfortunate this doesn't work automatically for `configure_file()` but I can see no reason why `@CMAKE_INSTALL_PREFIX@` would work but `@CMAKE_INSTALL_DATAROOTDIR@` would not, according the the PKGBUILD `@CMAKE_INSTALL_PREFIX@` was and is also manually set: https://github.com/LizardByte/Sunshine/blob/nightly/packaging/linux/Arch/PKGBUILD#L65 which might explain why things worked in the first place and brings me back to my suggestion of passing `-DCMAKE_INSTALL_DATAROOTDIR:PATH=/usr/share` when/where needed. Hope you all agree.

detiam: @tgurr wait according to cmake doc you post above, variable `CMAKE_INSTALL_DATAROOT`'s value should really just `share`.
> `CMAKE_INSTALL_FULL_<dir>`
>
>    The absolute path generated from the corresponding `CMAKE_INSTALL_<dir>` value. If the value is not already an absolute path, an absolute path is constructed typically by prepending the value of the [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/v3.18/variable/CMAKE_INSTALL_PREFIX.html#variable:CMAKE_INSTALL_PREFIX) variable. However, there are some [special cases](https://cmake.org/cmake/help/v3.18/module/GNUInstallDirs.html#special-cases) as documented below.

 you need `CMAKE_INSTALL_FULL_DATAROOT` for value that auto prepend `CMAKE_INSTALL_PREFIX` in to `CMAKE_INSTALL_DATAROOT`, maybe you can make a new MR for update this and add `include(GNUInstallDirs)` in `special_package_configuration.cmake`?

tgurr: @detiam Do I get you right that you mean:

```
-Exec=gio launch @CMAKE_INSTALL_DATAROOTDIR@/applications/sunshine_terminal.desktop
+Exec=gio launch @CMAKE_INSTALL_FULL_DATAROOTDIR@/applications/sunshine_terminal.desktop
```
I lack the possibility to test your usecase and also don't want to steal the credits from you so if you'd prepare an MR instead I'd happiliy test my usecase to see if something breaks for me (which I don't expect, in the worst case we just set the different variable manually for our build). If you still want to me to prepare an MR instead please advise.

detiam: > @detiam Do I get you right that you mean:
> 
> ```
> -Exec=gio launch @CMAKE_INSTALL_DATAROOTDIR@/applications/sunshine_terminal.desktop
> +Exec=gio launch @CMAKE_INSTALL_FULL_DATAROOTDIR@/applications/sunshine_terminal.desktop
> ```
> 
> I lack the possibility to test your usecase and also don't want to steal the credits from you so if you'd prepare an MR instead I'd happiliy test my usecase to see if something breaks for me (which I don't expect, in the worst case we just set the different variable manually for our build). If you still want to me to prepare an MR instead please advise.

ok I opened a new PR here #2223 :)

Issue #1849, opened 2023-11-16
Provided device doesn't support required NVENC features


UrsusLvovich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

Install Sunshine v0.21.0
Open the UI, get this error
![image](https://github.com/LizardByte/Sunshine/assets/38516315/81cb1ddc-c00a-45f3-a423-6ffe4f5476c8)

### Expected Behavior

Be able to install Sunshine and run no  problems on modern hardware.

### Additional Context

Host OS: Pop OS 22.04LTS
CPU: AMD Ryzen 7 3800X
GPU: Nvidia RTX 3080
Sunshine Version v0.21.0

I did a little digging and it seems that the RTX 3080 doesn't support av1_nvenc encoding. Does this mean I can't use Sunshine with my 3080?
https://www.reddit.com/r/AV1/comments/160tsq1/av1_software_encoding_on_rtx_3080_efficiency_and/

### Host Operating System

Linux

### Operating System Version

Pop OS 22,04 LTS

### Architecture

64 bit

### Sunshine commit or version

0.21.0

### Package

Linux - flatpak

### GPU Type

Nvidia

### GPU Model

RTC 3080

### GPU Driver/Mesa Version

545.29.02

### Capture Method (Linux Only)

PipeWire

### Config

```Shell
my ~/.config/sunshine/sunshine.conf file is empty
```

### Relevant log output

```shell
[2023:11:16:17:33:46]: Warning: [av1_nvenc @ 0x5557c26f0b40] Codec not supported
[2023:11:16:17:33:46]: Fatal: [av1_nvenc @ 0x5557c26f0b40] Provided device doesn't support required NVENC features
[2023:11:16:17:33:46]: Error: Could not open codec [av1_nvenc]: Function not implemented
```

PR #1948, opened 2023-12-30
Fatal warning improvements

cgutman:
## Description
This PR has a few changes to improve the UX for "fatal" errors in the web UI.

- It demotes FFmpeg's `AV_LOG_FATAL` log level down to our `error` level to avoid things like #1849
- It adds an error for ViGEmBus initialization failure since it's common for portable users to fail to install it and wonder why their gamepad doesn't work
- Tweaked wording to be less strong regarding configuration errors, since some may be expected in esoteric configurations

### Screenshot
image

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -517,10 +517,9 @@ main(int argc, char *argv[]) {
     char buffer[1024];
 
     av_log_format_line(ptr, level, fmt, vl, buffer, sizeof(buffer), &print_prefix);
-    if (level <= AV_LOG_FATAL) {
-      BOOST_LOG(fatal) << buffer;
-    }
-    else if (level <= AV_LOG_ERROR) {
+    if (level <= AV_LOG_ERROR) {
+      // We print AV_LOG_FATAL at the error level. FFmpeg prints things as fatal that
+      // are expected in some cases, such as lack of codec support or similar things.
       BOOST_LOG(error) << buffer;
     }
     else if (level <= AV_LOG_WARNING) {
--- a/src/platform/windows/input.cpp
+++ b/src/platform/windows/input.cpp
@@ -200,6 +200,8 @@ namespace platf {
       if (!VIGEM_SUCCESS(status)) {
         BOOST_LOG(warning) << "Couldn't setup connection to ViGEm for gamepad support ["sv << util::hex(status).to_string_view() << ']';
 
+        // Log a special fatal message for this case to show the error in the web UI
+        BOOST_LOG(fatal) << "ViGEmBus is not installed or running. You must install ViGEmBus for gamepad support!"sv;
         return -1;
       }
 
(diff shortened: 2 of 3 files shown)


src/main.cpp
```diff
@@ -517,10 +517,9 @@ main(int argc, char *argv[]) {
     char buffer[1024];
 
     av_log_format_line(ptr, level, fmt, vl, buffer, sizeof(buffer), &print_prefix);
-    if (level <= AV_LOG_FATAL) {
-      BOOST_LOG(fatal) << buffer;
-    }
-    else if (level <= AV_LOG_ERROR) {
+    if (level <= AV_LOG_ERROR) {
```
ReenigneArcher, line 520:
```suggestion
    if (level <= AV_LOG_FATAL) {
```

Should this be "FATAL" so we still log the fatal messages, just at the lower error level?

src/platform/windows/input.cpp
```diff
@@ -200,6 +200,8 @@ namespace platf {
       if (!VIGEM_SUCCESS(status)) {
         BOOST_LOG(warning) << "Couldn't setup connection to ViGEm for gamepad support ["sv << util::hex(status).to_string_view() << ']';
 
+        // Log a special fatal message for this case to show the error in the web UI
+        BOOST_LOG(fatal) << "ViGEmBus is not installed or running. You must install ViGEmBus for gamepad support!"sv;
```
ReenigneArcher, line 204: Thank you! I think the final vigembus version has been very finicky to install.

src/main.cpp
cgutman: Nope, `AV_LOG_ERROR` is correct there because of how those log levels are defined.

```
#define AV_LOG_FATAL     8
#define AV_LOG_ERROR    16
```

So `level <= AV_LOG_ERROR` matches both `AV_LOG_FATAL` and `AV_LOG_ERROR`, but `level <= AV_LOG_FATAL` only matches `AV_LOG_FATAL`.

src/platform/windows/input.cpp
cgutman: Yeah, I've seen issues with it too (especially with those unexpected reboots when upgrading ViGEmBus). I'm planning to put up another PR to avoid installing it if it's already present on the system to hopefully avoid these issues.

src/main.cpp
ReenigneArcher: Perfect!

ReenigneArcher (approved):

src/platform/windows/input.cpp
ReenigneArcher: We should probably throw up a warning too, that it's known to unexpectedly reboot during install.

cgutman: The reason it's rebooting is that the driver is already installed and in use, so I think it should never reboot if we don't upgrade existing versions.

ReenigneArcher: That makes sense. Can we detect the installed version and put up a warning in the UI if it's not the latest?

We can get latest release info from https://api.github.com/repos/nefarius/ViGEmBus/releases/latest using the `tag_name` key for comparison.

PR #1955, opened 2023-12-31
Use dl handle for libva's vaTerminate rather than actual function

chewi:
## Description
libva is supposed to be optional. Building currently fails with an unresolved symbol if it is missing.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/vaapi.cpp
+++ b/src/platform/linux/vaapi.cpp
@@ -520,7 +520,7 @@ namespace va {
     auto hwctx = (AVVAAPIDeviceContext *) ctx->hwctx;
     auto priv = (VAAPIDevicePriv *) ctx->user_opaque;
 
-    vaTerminate(hwctx->display);
+    terminate(hwctx->display);
     close(priv->drm_fd);
     av_freep(&priv);
   }


ReenigneArcher:
> libva is supposed to be optional

How did you come to that conclusion?

chewi:
> > libva is supposed to be optional
> 
> How did you come to that conclusion?

If it was mandatory, you wouldn't load it with `dlopen`. :slightly_smiling_face: I've been packaging Sunshine for Gentoo Linux (more on than later) where do we our own ffmpeg build, and that can optionally be done without VAAPI support. I have successfully built and run Sunshine without any VAAPI support at all.

ReenigneArcher:
> our own ffmpeg build

Please include these patches, if you're not already. https://github.com/LizardByte/build-deps/tree/master/ffmpeg_patches

> without VAAPI support

If you're building without vaapi support... then cmake should probably just be modified to not include the vaapi.cpp source. That's what we do for cuda, drm, wayland, x11

cuda:
https://github.com/LizardByte/Sunshine/blob/nightly/cmake/compile_definitions/linux.cmake#L15-L99

drm:
https://github.com/LizardByte/Sunshine/blob/a6ffeb4b53a6d7a5089d7b4fc0479a9fb575ba5e/cmake/compile_definitions/linux.cmake#L101-L119

wayland:
https://github.com/LizardByte/Sunshine/blob/a6ffeb4b53a6d7a5089d7b4fc0479a9fb575ba5e/cmake/compile_definitions/linux.cmake#L121-L143

x11:
https://github.com/LizardByte/Sunshine/blob/a6ffeb4b53a6d7a5089d7b4fc0479a9fb575ba5e/cmake/compile_definitions/linux.cmake#L145-L156

cgutman:
I think this fix makes sense to take to allow builds that were compiled with VAAPI to be able to run without it present at runtime. It's a niche use case, but if the fix is just changing a single function call, I don't see any problem with it.

For builds without VAAPI, I agree with @ReenigneArcher that we should just not compile any of the VAAPI stuff in at all.

ReenigneArcher:
I'll merge this then, and work on a second PR to allow disabling vaapi with a cmake flag.

Issue #1873, opened 2023-11-30
Mousewheel or Middle mouse btn doesn't work in some games


cardil:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

In some games, the mouse wheel doesn't work (in my case: Witcher 3, Black Mesa, Fallout 3). In some of those, the middle mouse button also doesn't work (Witcher 3). The Left and Right buttons work as expected.

Also, in other games, the mouse wheel works as expected. I've tested: Age of Wonders 3, Cities Skylines, Cyberpunk 2077 (works although only when scrolling v. fast), Master of Orion, Two Point Hospital, War For The Overworld.

Evidence:
- https://youtu.be/mDSpNM0RsVE - [Wither 3](https://steamdb.info/sub/124923/). Mouse scroll and MMB doesn't work at all. The video shows a console, on the client, showing inputs that don't affect the game (Witcher should shoot the crossbow, and menus should react to scroll)
- https://youtu.be/O6ozGoWlFhQ - [Fallout 3](https://steamdb.info/app/22370/). The mouse scrolling seems to be delivered in an endless loop, and extremely fast, making the game unplayable.
- https://youtu.be/HRCoiPH5Zr4 - [Black Mesa](https://steamdb.info/app/362890/). The mouse scroll is extremely fast, making the weapon switching very hard.

### Expected Behavior

The mouse scroll and the middle mouse button should work in every game.

### Additional Context

Uninstalling the Sunshine, and enabling streaming via GeForce Experience, resolves the described issues in all affected games.

The issue doesn't affect the desktop or web browsers when accessed via Sunshine.

### Host Operating System

Windows

### Operating System Version

Windows 10 Pro (22H2, build 19045.3693)

### Architecture

64 bit

### Sunshine commit or version

0.21.0.336062d467d5f48ba56d05a523c212f791421108

### Package

Windows - installer

### GPU Type

Nvidia

### GPU Model

GeForce GTX 1060 6GB

### GPU Driver/Mesa Version

546.17

### Config

```Shell
min_log_level = 2
back_button_timeout = 2000
fps = [10,30,60]
nvenc_preset = 2
resolutions = [
    352x240,
    480x360,
    858x480,
    1280x720,
    1920x1080
]
```

### Apps

```json
{
    "env": {
[...]

PR #1707, opened 2023-10-06
Allow the Misc button to work as Guide on emulated Xbox 360 controllers

cgutman:
## Description
The Guide button is often reserved for OS functionality on the client OS like Game Bar, Home button, etc. which can leave users without a way to activate that functionality in game. Since a Misc button (Xbox One Share button, PS4 Mute button, etc) isn't physically present on Xbox 360 controllers, we might as well assign it to Guide to make it useful. Right now it's mapped to nothing.

I added the same thing in https://github.com/moonlight-stream/moonlight-common-c/commit/d055599608a9c50b9b5426790dad40fa3ab70687 on the client-side for GFE hosts.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I want maintainers to keep my branch updated

PR #1960, opened 2024-01-01
New input configuration options

cgutman:
### Description
This PR adds a few input configuration options based on user requests.
- Gamepad autoselection logic can now be configured to include/exclude touchpad and motion sensor support when deciding on a controller type
- High resolution scroll events can now be disabled for applications that don't handle them properly (#1873)
- Native pen/touch support can now be disabled for applications without proper pen/touch support

I've also moved a couple input-related options from the General tab to the Input tab where they belong.

### Screenshot
image

image

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -289,6 +289,24 @@ high_resolution_scrolling
 
       high_resolution_scrolling = enabled
 
+native_pen_touch
+^^^^^^^^^^^^^^^^
+
+**Description**
+   When enabled, Sunshine will pass through native pen/touch events from Moonlight clients. 
+
+   This can be useful to disable for older applications without native pen/touch support.
+
+   .. Caution:: Applies to Windows only.
+
+**Default**
+   ``enabled``
+
+**Example**
+   .. code-block:: text
+
+      native_pen_touch = enabled
+
 keybindings
 ^^^^^^^^^^^
 
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -170,6 +170,42 @@ ds4_back_as_touchpad_click
 
       ds4_back_as_touchpad_click = enabled
 
+motion_as_ds4
+^^^^^^^^^^^^^
+
+**Description**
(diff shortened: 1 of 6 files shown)


docs/source/about/advanced_usage.rst
```diff
@@ -237,6 +273,40 @@ always_send_scancodes
 
       always_send_scancodes = enabled
 
+high_resolution_scrolling
```
ReenigneArcher, line 276: I wonder if we should think about having app specific settings. This seems like one example where that could be beneficial.

src_assets/common/assets/web/config.html
```diff
@@ -818,6 +875,7 @@ <h2 class="accordion-header">
     "amd_vbaq": "enabled",
     "capture": "",
     "controller": "enabled",
+    "high_resolution_scrolling": "enabled",
     "install_steam_audio_drivers": "enabled",
     "ds4_back_as_touchpad_click": "enabled",
```
ReenigneArcher, line 880: Since you are touching this, can you alphabetize these keys?

ReenigneArcher (commented):
Can you review this real quick as well? https://discord.com/channels/804382334370578482/1166059339170197574/1166059339170197574

When I made the note, I thought the misc button setting was missing from the UI and docs.

But looking closer at https://github.com/LizardByte/Sunshine/pull/1707 I'm don't think any UI change is required. Maybe just a note in the docs about the behavior of the share button?

cgutman:
> But looking closer at https://github.com/LizardByte/Sunshine/pull/1707 I'm don't think any UI change is required. Maybe just a note in the docs about the behavior of the share button?

Yeah, there's no UI option to configure that.

docs/source/about/advanced_usage.rst
cgutman: We definitely should

src_assets/common/assets/web/config.html
cgutman: Yep

ReenigneArcher (approved):

PR #2000, opened 2024-01-07
docs(advanced-usage): cross reference config UI

ReenigneArcher:
## Description

This PR adds hyperlinks in the docs to take the user to the location in the UI where the config categories/settings are located.

Todo:
- [x] Add hash watcher event to toggle config UI tab -> https://developer.mozilla.org/en-US/docs/Web/API/Window/hashchange_event ... see #2002 

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [x] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I want maintainers to keep my branch updated

PR #2002, opened 2024-01-08
feat(ui): add hash handler for configuration ids

ReenigneArcher:
## Description

This PR adds a hash handler to allow navigating to ids which are not normally available unless the correct tab is selected. Originally I was going to dynamically determine the closest "tab" to the element in the hash, but the element is not available unless the tab is already loaded, so I had to add a mapping of each tab to the options that they contain.

In adding this new mapping, I decided that the `defaultConfig` mapping could be moved to this new mapping. A small bug was also discovered and fixed when moving the `defaultConfig`, which was that the values that were not string encoded previously would not be evaluate properly when comparing the default value to the value in the UI.

The hash watcher can be tested by visiting `https://localhost:47990/config/#encoder`. This should work when the page is loaded fresh, as well as if you already on the `config` page and then add the `#encoder` hash.

Moving forward, all config options, and tabs should have a matching `id`, which will be referenced in the docs. See #2000.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/about/advanced_usage.rst
+++ b/docs/source/about/advanced_usage.rst
@@ -320,9 +320,11 @@ keybindings
 **Default**
    .. code-block:: text
 
-      0x10, 0xA0,
-      0x11, 0xA2,
-      0x12, 0xA4
+      [
+        0x10, 0xA0,
+        0x11, 0xA2,
+        0x12, 0xA4
+      ]
 
 **Example**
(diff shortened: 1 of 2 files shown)


cgutman (approved):
Seems fine to me, but you might want @TheElixZammuto to take a look too.

PR #2018, opened 2024-01-13
Cross compiling

chewi:
You wanted to know how to cross-compile for faster CI builds. I have prepared this as a Dockerfile for Fedora 39. This can be run with `TARGETARCH=aarch64` or `TARGETARCH=ppc64le`. It produces an RPM. On my own machine, it takes less than 5 minutes.

I'm not suggesting you take this as-is but work it into your CI somehow. This was just an easy way for me to experiment and for you to try it out.

The `CXXFLAGS` and `LDFLAGS` variables work around some difficulty Fedora has in locating the C++ headers and libraries. Other than that though, it feels very clean, with it basically using what Fedora already provides.

Debian and Ubuntu could use a similar approach, but I think their "multiarch" support means you would normally do it differently.

### DIFF
--- a/cmake/compile_definitions/linux.cmake
+++ b/cmake/compile_definitions/linux.cmake
@@ -223,6 +223,11 @@ if (${SUNSHINE_TRAY} EQUAL 0 AND SUNSHINE_REQUIRE_TRAY)
     message(FATAL_ERROR "Tray icon is required")
 endif()
 
+pkg_check_modules(EVDEV libevdev REQUIRED)
+include_directories(SYSTEM ${EVDEV_INCLUDE_DIRS})
+link_directories(${EVDEV_LIBRARY_DIRS})
+list(APPEND SUNSHINE_EXTERNAL_LIBRARIES ${EVDEV_LIBRARIES})
+
 list(APPEND PLATFORM_TARGET_FILES
         src/platform/linux/publish.cpp
         src/platform/linux/graphics.h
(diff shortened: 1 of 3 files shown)


chewi:
I see that your existing Dockerfiles eventually produce an image for end users. I wasn't sure whether Docker would be able to cross-compile and then produce an image for each architecture at the end, at least within a single Dockerfile, but [apparently it can](https://docs.docker.com/build/building/multi-platform/) with some help from QEMU.

ReenigneArcher:
Thanks for this!

I'm working on this in #2020.

A few notable changed from the image you submitted here:

- the `--platform` is supplied in the FROM argument for the first image
- the `TARGETARCH` env is determined based on the `TARGETPLATFORM` which can be provided from the CI workflow... I don't think there is way to dynamically set this one time, so it's set in each step that needs it... similar for the `TUPLE` env

I'm not getting any `CMAKE_SYSTEM_PROCESSOR` value though.

```
2.350 CMake Error at cmake/dependencies/common.cmake:51 (message):
2.350   Unsupported system processor:
```

It should print the value after the error. Any idea?

Besides this error it seems to be running really quickly! Only 2-3 minutes to install dependencies is a huge win already!

chewi:
Yeah, I thought they might be set from CI. Hopefully you can work that out. I'm also looking at Debian/Ubuntu. I'll explain your current issue in that ticket.

PR #2042, opened 2024-01-20
Name and unpair individual clients

xanderfrangos:
## Description
Introduces the option to name clients during PIN pairing, along with a UI to remove individual client certificates. The format of `sunshine_state.json` was updated to include a `named_certs` array instead of the previous `certs` array. The old array is used for migration to the new format on startup.

Here's an example of the new format:

```json
{
    ...
    "root": {
        "uniqueid": "2258BFE6-698A-422C-5F0F-55B49A69C69E",
        "devices": [
            {
                "uniqueid": "0123456789ABCDEF",
                "named_certs": [
                    {
                        "name": "NVIDIA Shield",
                        "cert": "-----BEGIN CERTIFICATE-----\nCERT\n-----END CERTIFICATE-----\n",
                        "uniqueid": "41310952-60C8-1B01-A615-033CC8275E5F"
                    },
                    {
                        "name": "Steam Deck",
                        "cert": "-----BEGIN CERTIFICATE-----\nCERT\n-----END CERTIFICATE-----\n",
                        "uniqueid": "AE113204-32D9-6607-6285-7F3C200440EB"
                    }
                ]
            }
        ]
    }
}

```

The `named_certs` array includes a user-defined name and a generated UUID. The UUID allows for the same name to be used multiple times. It also helps target the correct client over the API without exposing the certificate.

The UI for managing clients has been added to the PIN page, under the PIN input. If a client wasn't named during the pairing process, or the certificate was imported from the old format, the client will be listed as "Unknown". After removing a paired client from the new UI, the user will be prompted to restart Sunshine in order to finish removing the certificate. 

While working on this, I was made aware of the [TheElixZammuto/Sunshine-1/state-revamp](https://github.com/TheElixZammuto/Sunshine-1/tree/state-revamp) fork, which has a similar goal. However, it looks like it's still missing the UI, API, and server code to deal with "unpairing". It also hasn't been updated since September, so it can't merge cleanly as-is. I understand if you prefer to use what that fork over mine, but I figured I would clean up my code and submit it, in case it's useful 

### Screenshot
Screenshot 2024-01-20 032346

### Issues Fixed or Closed
https://discord.com/channels/804382334370578482/1197647357030965328

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -693,7 +693,8 @@ namespace confighttp {
       // TODO: Input Validation
       pt::read_json(ss, inputTree);
       std::string pin = inputTree.get<std::string>("pin");
-      outputTree.put("status", nvhttp::pin(pin));
+      std::string name = inputTree.get<std::string>("name");
+      outputTree.put("status", nvhttp::pin(pin, name));
     }
     catch (std::exception &e) {
       BOOST_LOG(warning) << "SavePin: "sv << e.what();
@@ -717,6 +718,60 @@ namespace confighttp {
       response->write(data.str());
     });
     nvhttp::erase_all_clients();
+    proc::proc.terminate();
+    outputTree.put("status", true);
+  }
+
+  void
+  unpair(resp_https_t response, req_https_t request) {
+    if (!authenticate(response, request)) return;
+
+    print_req(request);
+
+    std::stringstream ss;
+    ss << request->content.rdbuf();
+
+    pt::ptree inputTree, outputTree;
+
+    auto g = util::fail_guard([&]() {
+      std::ostringstream data;
+      pt::write_json(data, outputTree);
+      response->write(data.str());
+    });
+
+    try {
+      // TODO: Input Validation
+      pt::read_json(ss, inputTree);
+      std::string uuid = inputTree.get<std::string>("uuid");
+      outputTree.put("status", nvhttp::unpair_client(uuid));
+    }
+    catch (std::exception &e) {
+      BOOST_LOG(warning) << "Unpair: "sv << e.what();
+      outputTree.put("status", false);
+      outputTree.put("error", e.what());
+      return;
+    }
+  }
+
+  void
+  listClients(resp_https_t response, req_https_t request) {
+    if (!authenticate(response, request)) return;
+
+    print_req(request);
+
+    pt::ptree named_certs = nvhttp::get_all_clients();
+
+    pt::ptree outputTree;
+
+    outputTree.put("status", false);
+
+    auto g = util::fail_guard([&]() {
+      std::ostringstream data;
+      pt::write_json(data, outputTree);
+      response->write(data.str());
+    });
+
+    outputTree.add_child("named_certs", named_certs);
     outputTree.put("status", true);
   }
 
@@ -765,7 +820,9 @@ namespace confighttp {
     server.resource["^/api/restart$"]["POST"] = restart;
     server.resource["^/api/password$"]["POST"] = savePassword;
     server.resource["^/api/apps/([0-9]+)$"]["DELETE"] = deleteApp;
-    server.resource["^/api/clients/unpair$"]["POST"] = unpairAll;
+    server.resource["^/api/clients/unpair-all$"]["POST"] = unpairAll;
+    server.resource["^/api/clients/list$"]["GET"] = listClients;
+    server.resource["^/api/clients/unpair$"]["POST"] = unpair;
     server.resource["^/api/apps/close$"]["POST"] = closeApp;
     server.resource["^/api/covers/upload$"]["POST"] = uploadCover;
     server.resource["^/images/sunshine.ico$"]["GET"] = getFaviconImage;
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -117,9 +117,15 @@ namespace nvhttp {
     std::string pkey;
   } conf_intern;
 
+  struct named_cert_t {
+    std::string name;
+    std::string uuid;
+    std::string cert;
+  };
+
   struct client_t {
-    std::string uniqueID;
     std::vector<std::string> certs;
+    std::vector<named_cert_t> named_devices;
   };
 
   struct pair_session_t {
@@ -145,7 +151,7 @@ namespace nvhttp {
 
   // uniqueID, session
   std::unordered_map<std::string, pair_session_t> map_id_sess;
-  std::unordered_map<std::string, client_t> map_id_client;
+  client_t client_root;
   std::atomic<uint32_t> session_id_counter;
 
   using args_t = SimpleWeb::CaseInsensitiveMultimap;
@@ -189,22 +195,18 @@ namespace nvhttp {
     root.erase("root"s);
 
     root.put("root.uniqueid", http::unique_id);
-    auto &nodes = root.add_child("root.devices", pt::ptree {});
-    for (auto &[_, client] : map_id_client) {
-      pt::ptree node;
+    client_t &client = client_root;
+    pt::ptree node;
 
-      node.put("uniqueid"s, client.uniqueID);
-
-      pt::ptree cert_nodes;
-      for (auto &cert : client.certs) {
-        pt::ptree cert_node;
-        cert_node.put_value(cert);
-        cert_nodes.push_back(std::make_pair(""s, cert_node));
-      }
-      node.add_child("certs"s, cert_nodes);
-
-      nodes.push_back(std::make_pair(""s, node));
+    pt::ptree named_cert_nodes;
+    for (auto &named_cert : client.named_devices) {
+      pt::ptree named_cert_node;
+      named_cert_node.put("name"s, named_cert.name);
+      named_cert_node.put("cert"s, named_cert.cert);
+      named_cert_node.put("uuid"s, named_cert.uuid);
+      named_cert_nodes.push_back(std::make_pair(""s, named_cert_node));
     }
+    root.add_child("root.named_devices"s, named_cert_nodes);
 
     try {
       pt::write_json(config::nvhttp.file_state, root);
@@ -223,9 +225,9 @@ namespace nvhttp {
       return;
     }
 
-    pt::ptree root;
+    pt::ptree tree;
     try {
-      pt::read_json(config::nvhttp.file_state, root);
+      pt::read_json(config::nvhttp.file_state, tree);
     }
     catch (std::exception &e) {
       BOOST_LOG(error) << "Couldn't read "sv << config::nvhttp.file_state << ": "sv << e.what();
@@ -233,7 +235,7 @@ namespace nvhttp {
       return;
     }
 
-    auto unique_id_p = root.get_optional<std::string>("root.uniqueid");
+    auto unique_id_p = tree.get_optional<std::string>("root.uniqueid");
     if (!unique_id_p) {
       // This file doesn't contain moonlight credentials
       http::unique_id = uuid_util::uuid_t::generate().string();
@@ -241,30 +243,61 @@ namespace nvhttp {
     }
     http::unique_id = std::move(*unique_id_p);
 
-    auto device_nodes = root.get_child("root.devices");
+    auto root = tree.get_child("root");
+    client_t client;
 
-    for (auto &[_, device_node] : device_nodes) {
-      auto uniqID = device_node.get<std::string>("uniqueid");
-      auto &client = map_id_client.emplace(uniqID, client_t {}).first->second;
+    // Import from old format
+    if (root.get_child_optional("devices")) {
+      auto device_nodes = root.get_child("devices");
+      for (auto &[_, device_node] : device_nodes) {
+        auto uniqID = device_node.get<std::string>("uniqueid");
 
-      client.uniqueID = uniqID;
-
-      for (auto &[_, el] : device_node.get_child("certs")) {
-        client.certs.emplace_back(el.get_value<std::string>());
+        if (device_node.count("certs")) {
+          for (auto &[_, el] : device_node.get_child("certs")) {
+            named_cert_t named_cert;
+            named_cert.name = ""s;
+            named_cert.cert = el.get_value<std::string>();
+            named_cert.uuid = uuid_util::uuid_t::generate().string();
+            client.named_devices.emplace_back(named_cert);
+            client.certs.emplace_back(named_cert.cert);
+          }
+        }
       }
     }
+
+    if (root.count("named_devices")) {
+      for (auto &[_, el] : root.get_child("named_devices")) {
+        named_cert_t named_cert;
+        named_cert.name = el.get_child("name").get_value<std::string>();
+        named_cert.cert = el.get_child("cert").get_value<std::string>();
+        named_cert.uuid = el.get_child("uuid").get_value<std::string>();
+        client.named_devices.emplace_back(named_cert);
+        client.certs.emplace_back(named_cert.cert);
+      }
+    }
+
+    // Empty certificate chain and import certs from file
+    cert_chain.clear();
+    for (auto &cert : client.certs) {
+      cert_chain.add(crypto::x509(cert));
+    }
+    for (auto &named_cert : client.named_devices) {
+      cert_chain.add(crypto::x509(named_cert.cert));
+    }
+
+    client_root = client;
   }
 
   void
   update_id_client(const std::string &uniqueID, std::string &&cert, op_e op) {
     switch (op) {
       case op_e::ADD: {
-        auto &client = map_id_client[uniqueID];
+        client_t &client = client_root;
         client.certs.emplace_back(std::move(cert));
-        client.uniqueID = uniqueID;
       } break;
       case op_e::REMOVE:
-        map_id_client.erase(uniqueID);
+        client_t client;
+        client_root = client;
         break;
     }
 
@@ -579,15 +612,16 @@ namespace nvhttp {
   /**
    * @brief Compare the user supplied pin to the Moonlight pin.
    * @param pin The user supplied pin.
+   * @param name The user supplied name.
    * @return `true` if the pin is correct, `false` otherwise.
    *
    * EXAMPLES:
    * ```cpp
-   * bool pin_status = nvhttp::pin("1234");
+   * bool pin_status = nvhttp::pin("1234", "laptop");
    * ```
    */
   bool
-  pin(std::string pin) {
+  pin(std::string pin, std::string name) {
     pt::ptree tree;
     if (map_id_sess.empty()) {
       return false;
@@ -612,6 +646,14 @@ namespace nvhttp {
 
     auto &sess = std::begin(map_id_sess)->second;
     getservercert(sess, tree, pin);
+
+    // set up named cert
+    client_t &client = client_root;
+    named_cert_t named_cert;
+    named_cert.name = name;
+    named_cert.cert = sess.client.cert;
+    named_cert.uuid = uuid_util::uuid_t::generate().string();
+    client.named_devices.emplace_back(named_cert);
 
     // response to the request for pin
     std::ostringstream data;
@@ -645,9 +687,7 @@ namespace nvhttp {
       auto clientID = args.find("uniqueid"s);
 
       if (clientID != std::end(args)) {
-        if (auto it = map_id_client.find(clientID->second); it != std::end(map_id_client)) {
-          pair_status = 1;
-        }
+        pair_status = 1;
       }
     }
 
@@ -740,6 +780,20 @@ namespace nvhttp {
     pt::write_xml(data, tree);
     response->write(data.str());
     response->close_connection_after_response = true;
+  }
+
+  pt::ptree
+  get_all_clients() {
+    pt::ptree named_cert_nodes;
+    client_t &client = client_root;
+    for (auto &named_cert : client.named_devices) {
+      pt::ptree named_cert_node;
+      named_cert_node.put("name"s, named_cert.name);
+      named_cert_node.put("uuid"s, named_cert.uuid);
+      named_cert_nodes.push_back(std::make_pair(""s, named_cert_node));
+    }
+
+    return named_cert_nodes;
   }
 
   void
@@ -1020,12 +1074,6 @@ namespace nvhttp {
     conf_intern.pkey = file_handler::read_file(config::nvhttp.pkey.c_str());
     conf_intern.servercert = file_handler::read_file(config::nvhttp.cert.c_str());
 
-    for (auto &[_, client] : map_id_client) {
-      for (auto &cert : client.certs) {
-        cert_chain.add(crypto::x509(cert));
-      }
-    }
-
     auto add_cert = std::make_shared<safe::queue_t<crypto::x509_t>>(30);
 
     // resume doesn't always get the parameter "localAudioPlayMode"
@@ -1149,8 +1197,48 @@ namespace nvhttp {
    */
   void
   erase_all_clients() {
-    map_id_client.clear();
+    client_t client;
+    client_root = client;
     cert_chain.clear();
     save_state();
+  }
+
+  /**
+   * @brief Remove single client.
+   *
+   * EXAMPLES:
+   * ```cpp
+   * nvhttp::unpair_client("4D7BB2DD-5704-A405-B41C-891A022932E1");
+   * ```
+   */
+  int
+  unpair_client(std::string uuid) {
+    int removed = 0;
+    client_t &client = client_root;
+    for (auto it = client.named_devices.begin(); it != client.named_devices.end();) {
+      if ((*it).uuid == uuid) {
+        // Find matching cert and remove it
+        for (auto cert = client.certs.begin(); cert != client.certs.end();) {
+          if ((*cert) == (*it).cert) {
+            cert = client.certs.erase(cert);
+            removed++;
+          }
+          else {
+            ++cert;
+          }
+        }
+
+        // And then remove the named cert
+        it = client.named_devices.erase(it);
+        removed++;
+      }
+      else {
+        ++it;
+      }
+    }
+
+    save_state();
+    load_state();
+    return removed;
   }
 }  // namespace nvhttp
--- a/src_assets/common/assets/web/pin.html
+++ b/src_assets/common/assets/web/pin.html
@@ -25,7 +25,7 @@
 
 <script type="module">
   import { createApp } from 'vue'
-  import i18n from './locale.js'
+  import i18nLocale from './locale.js'
   import Navbar from './Navbar.vue'
   import {initApp} from "./init";
 
@@ -35,8 +35,9 @@
     }
   });
 
-  initApp(app, (app => {
+  initApp(app, (async app => {
     // this must be after mounting the app
+    const i18n = await i18nLocale()
     document.querySelector("#form").addEventListener("submit", (e) => {
       e.preventDefault();
       let pin = document.querySelector("#pin-input").value;
--- a/src_assets/common/assets/web/pin.html
+++ b/src_assets/common/assets/web/pin.html
@@ -9,7 +9,7 @@
   <Navbar></Navbar>
   <div id="content" class="container">
     <h1 class="my-4 text-center">{{ $t('pin.pin_pairing') }}</h1>
-    <form action="" class="form d-flex flex-column align-items-center" id="form">
+    <form class="form d-flex flex-column align-items-center" id="form" @submit.prevent="registerDevice">
       <div class="card flex-column d-flex p-4 mb-4">
         <input type="text" pattern="\d*" :placeholder="`${$t('navbar.pin')}`" autofocus id="pin-input" class="form-control mt-2" required />
         <input type="text" :placeholder="`${$t('pin.device_name')}`" id="name-input" class="form-control my-4" required />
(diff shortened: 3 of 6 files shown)


ReenigneArcher:
@TheElixZammuto any thoughts on using this over what were working on?

TheElixZammuto:
> @TheElixZammuto any thoughts on using this over what were working on?

Just two things:
1. I would rename uniqueid to uuid or a similar name, since uniqueid is a protocol-specific name which can lead to confusion (it's not the same uniqueid as the one the moonlight client sends)
2. I would prefer to have named_certs as a root element instead of inside "devices", which is useless right now since Moonlight always send the same uniqueid

xanderfrangos:
@TheElixZammuto Easy enough. 👍 Regarding setting it as a "root element", do you mean the root of the JSON object or the `root` property (that currently contains `uniqueid` and `devices`)?

TheElixZammuto:
My example idea
```
{
    "root": {
        "uniqueid": "2258BFE6-698A-422C-5F0F-55B49A69C69E",
        "named_devices": [
            {
                "name": "NVIDIA Shield",
                "cert": "-----BEGIN CERTIFICATE-----\nCERT\n-----END CERTIFICATE-----\n",
                "uuid": "41310952-60C8-1B01-A615-033CC8275E5F"
            },
            {
                "name": "Steam Deck",
                "cert": "-----BEGIN CERTIFICATE-----\nCERT\n-----END CERTIFICATE-----\n",
                "uuid": "AE113204-32D9-6607-6285-7F3C200440EB"
            }
        ]
    }
}

```

xanderfrangos:
@TheElixZammuto I've updated `sunshine_state.json` to use the format you've outlined. I've also refactored `nvhttp.cpp` to no longer map `client_t`s to `uniqueid`s, since `root.devices` is gone (and it's always Moonlight's "0123456789ABCDEF" anyway). Only a single `client_t` is used now.

ReenigneArcher (changes requested):
In my opinion the client list and the ability to remove them should be in the troubleshooting section, near the unpair all clients option.

The pin/pair page should be lightweight.

xanderfrangos:
@ReenigneArcher Moved to Troubleshooting 👍

localhost_37990_troubleshooting

src_assets/common/assets/web/troubleshooting.html
```diff
@@ -82,23 +82,38 @@ <h2 id="restart">Restart Sunshine</h2>
...
-          </button>
+        <div class="p-2">
+          <div class="d-flex justify-content-end align-items-center">
+            <h2 class="text-center me-auto">Unpair Clients</h2>
+            <button class="btn btn-danger" :disabled="unpairAllPressed" @click="unpairAll">
+              Unpair All
+            </button>
+          </div>
+          <div id="apply-alert" class="alert alert-success d-flex align-items-center mt-3" :style="{ 'display': (showApplyMessage ? 'flex !important': 'none !important') }">
+            <div class="me-2"><b>Success!</b> Click 'Apply' to restart Sunshine and apply changes. This will terminate any running sessions.</div>
```
ReenigneArcher, line 95: Is it necessary to restart Sunshine? I'm guessing you're doing this to force disconnecting a client if it was one of those that were removed?

There's probably a way we can disconnect that single client instead of restarting Sunshine altogether.

src/confighttp.cpp
```diff
@@ -745,6 +799,8 @@ namespace confighttp {
     server.resource["^/api/password$"]["POST"] = savePassword;
     server.resource["^/api/apps/([0-9]+)$"]["DELETE"] = deleteApp;
     server.resource["^/api/clients/unpair$"]["POST"] = unpairAll;
+    server.resource["^/api/clients/list$"]["GET"] = listClients;
+    server.resource["^/api/clients/unpair-single$"]["POST"] = unpairClient;
```
ReenigneArcher, line 803:
```suggestion
    server.resource["^/api/clients/unpair$"]["POST"] = unpair;
```

src/confighttp.cpp
```diff
@@ -700,6 +701,59 @@ namespace confighttp {
     outputTree.put("status", true);
   }
 
+  void
+  unpairClient(resp_https_t response, req_https_t request) {
```
ReenigneArcher, line 705:
```suggestion
  unpair(resp_https_t response, req_https_t request) {
```

src/confighttp.cpp
```diff
@@ -700,6 +701,59 @@ namespace confighttp {
...
+    });
+
+    try {
+      // TODO: Input Validation
+      pt::read_json(ss, inputTree);
+      std::string uuid = inputTree.get<std::string>("uuid");
+      outputTree.put("status", nvhttp::unpair_client(uuid));
+    }
+    catch (std::exception &e) {
+      BOOST_LOG(warning) << "UnpairClient: "sv << e.what();
```
ReenigneArcher, line 728:
```suggestion
      BOOST_LOG(warning) << "Unpair: "sv << e.what();
```

src_assets/common/assets/web/troubleshooting.html
```diff
@@ -180,13 +198,38 @@ <h2 id="logs">Logs</h2>
           fetch("/api/clients/unpair", { method: "POST" })
```
ReenigneArcher, line 198:
```suggestion
          fetch("/api/clients/unpair-all", { method: "POST" })
```

src/nvhttp.cpp
```diff
@@ -217,48 +222,72 @@ namespace nvhttp {
...
-    auto device_nodes = root.get_child("root.devices");
-
-    for (auto &[_, device_node] : device_nodes) {
-      auto uniqID = device_node.get<std::string>("uniqueid");
-      auto &client = map_id_client.emplace(uniqID, client_t {}).first->second;
-
-      client.uniqueID = uniqID;
+    auto root = tree.get_child("root");
+    client_t client;
+    client.uniqueID = "0123456789ABCDEF"s;
```
ReenigneArcher, line 245: Why is this hard coded?

src/nvhttp.cpp
```diff
@@ -1096,7 +1150,47 @@ namespace nvhttp {
    */
   void
   erase_all_clients() {
-    map_id_client.clear();
+    client_t client;
+    client.uniqueID = "0123456789ABCDEF"s;
```
ReenigneArcher, line 1154: Same question... why hard coded?

ReenigneArcher (changes requested):

src_assets/common/assets/web/pin.html
```diff
@@ -8,10 +8,11 @@
 <body id="app">
   <Navbar></Navbar>
   <div id="content" class="container">
-    <h1 class="my-4">PIN Pairing</h1>
+    <h1 class="my-4 text-center">PIN Pairing</h1>
     <form action="" class="form d-flex flex-column align-items-center" id="form">
       <div class="card flex-column d-flex p-4 mb-4">
-        <input type="text" pattern="\d*"  placeholder="PIN" id="pin-input" class="form-control my-4" />
+        <input type="number" pattern="\d*"  placeholder="PIN" id="pin-input" class="form-control mt-2" />
+        <input type="text" placeholder="Name (optional)" id="name-input" class="form-control my-4" />
```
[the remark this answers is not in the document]
xanderfrangos: Your call. It wasn't required before, so I didn't want to enforce it. If it's going to be required, I think the field name should better indicate that they're coming up with a name for the device. Maybe "Name this device" or "Set device name"?

src_assets/common/assets/web/troubleshooting.html
xanderfrangos: Two reasons:
1. Yes, to disconnect the client.
2. The certificate is unfortunately is still valid after unpairing a single client, so it can simply reconnect without issue. I'm not sure why. If someone can fix this, then only the session would need to end.

If access can be fully revoked and the client is disconnected, I think that would be a lot better than requesting a restart.

src/nvhttp.cpp
xanderfrangos: I wasn't sure if the `uniqueID` is being used elsewhere in Sunshine. I had refactored out most of the code related to mapping certificates to `uniqueID`s, but left this (hard-coded to Moonlight's) just in case. If it's not needed, then I believe `uniqueID` can be removed from `client_t` entirely.

src/nvhttp.cpp
xanderfrangos: See previous response. It might be unnecessary.

src_assets/common/assets/web/pin.html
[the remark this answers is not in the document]
ReenigneArcher: I think it's important to have an identifiable name as user's aren't going to be able to differentiate based on a uuid.

Ideally we could also get the client type during the pair process, but I'm not sure that's possible at the moment.

I think "Device Name" is okay to use as the name of the field.

src_assets/common/assets/web/troubleshooting.html
ReenigneArcher: @cgutman do you what we can do to remove a single client without restarting sunshine?

src/nvhttp.cpp
ReenigneArcher: @cgutman pinging you since you're the most familiar with this.

src_assets/common/assets/web/pin.html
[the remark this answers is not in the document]
xanderfrangos: It might also be nice to track/display when the client last connected and if there's an active session. But that seems like another PR for another day.

For now I'll make the field required and labeled "Device Name".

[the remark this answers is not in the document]
ReenigneArcher: Agreed!

I'd like to see these items tracked in the future.

- Last known IP address
- Last known network type (LAN or WAN)
- Moonlight Client (e.g. Android, QT, iOS, etc.)
- Last known moonlight version
- Device Name (provided by Moonlight, not sure if feasible... but e.g. Nvidia Shield)
- Last session start and end times
- Last session application launched

src/nvhttp.cpp
cgutman: You can remove it. Nvidia used it to identify separate clients, but it caused more issues than it solved (preventing a game started on one device from being resumed on another). That's why Moonlight just hardcodes a single value everywhere.

src_assets/common/assets/web/pin.html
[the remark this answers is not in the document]
cgutman: I don't love the idea of making device name required. I'd much rather feed that info into the pairing process from the Moonlight side and allow the user to optionally adjust the name via this UI.

On the Moonlight side, it should be pretty easy to send the client's hostname across during the pairing process. We will run into some issues related to the privacy restrictions [on iOS](https://developer.apple.com/documentation/uikit/uidevice/1620015-name), but that's why we keep the optional field around as an override.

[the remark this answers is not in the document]
ReenigneArcher: If Moonlight can send over enough details, then the name is probably not even necessary.

This is what Plex does. None of these fields can be modified, they are all collected from the client, and everything is pretty easy to identify.

image

The only issue I see is some of these can change after the pairing process. i.e. the client version (and IP address if we're going to track that somehow).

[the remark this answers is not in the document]
xanderfrangos: It may still be helpful to optionally set a client name during pairing. 

I suspect most users don't name their devices at the OS level, which can make the Moonlight-provided name somewhat meaningless without more info. Tracking clients this way could be especially difficult if you're using Sunshine to share your PC with someone else (remote "local" co-op, for example). Personally, I would rather set the client name myself than explain to a friend that they need to set a name in Moonlight (or worse, their device's hostname).

Also, I'd expect (well, "hope" lol) that this PR will get approved before the required additions to Moonlight are implemented. So a name field is still needed (optional or not) for now.

src/nvhttp.cpp
xanderfrangos: Removed 👍 (https://github.com/LizardByte/Sunshine/pull/2042/commits/5ff80ada0b93dd8527c0c7e20e42e9f75065b0f3)

src/nvhttp.cpp
xanderfrangos: Removed 👍 (https://github.com/LizardByte/Sunshine/pull/2042/commits/5ff80ada0b93dd8527c0c7e20e42e9f75065b0f3)

src_assets/common/assets/web/troubleshooting.html
xanderfrangos: I think this is the only remaining concern.

ReenigneArcher:
@xanderfrangos sorry for the delays. Do you have time to resume this? I think it will fix #2305 which is probably a side effect of using a static uniqueID.

xanderfrangos:
@ReenigneArcher I unfortunately won't be able to work on this much until later this month. But to my understanding, the only outstanding concern (https://github.com/LizardByte/Sunshine/pull/2042#discussion_r1467967479) was something I wasn't sure how to help with anyway.

src_assets/common/assets/web/troubleshooting.html
ReenigneArcher: I think we would have to remove it from `cert_chain`. In https://github.com/LizardByte/Sunshine/pull/2365/files I am removing all of them for the unpair all.

xanderfrangos:
@ReenigneArcher I've updated this branch to use the new localization library and (more importantly) utilize https://github.com/LizardByte/Sunshine/pull/2365. Individually unpaired devices can no longer start or resume a session. This should resolve https://github.com/LizardByte/Sunshine/pull/2042#discussion_r1467967479.

Since active sessions aren't automatically disconnected, I've added language to make that clearer for users. Here's an updated screenshot from the Troubleshooting page (taken after unpairing a device):

Screenshot 2024-04-17 124543

Also, the unpairing UI now consistently uses the word "device". I think that's a little more user-friendly than "client", but either is fine as long as it's consistent.

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -365,9 +367,12 @@
     "restart_sunshine_success": "Sunshine is restarting",
     "troubleshooting": "Troubleshooting",
     "unpair_all": "Unpair All",
-    "unpair_all_desc": "Remove all your paired devices",
+    "unpair_all_desc": "Remove your paired devices. Unpaired devices with an active session will remain connected, but cannot start or resume a session.",
     "unpair_all_error": "Error while unpairing",
-    "unpair_all_success": "Unpair Successful!"
+    "unpair_all_success": "However, the device(s) may still be in an active session. Use the 'Force Close' button above to end any open sessions.",
```
ReenigneArcher, line 372:
```suggestion
    "unpair_all_success": "Unpair Successful! However, the device(s) may still be in an active session. Use the 'Force Close' button above to end any open sessions.",
```

src_assets/common/assets/web/troubleshooting.html
```diff
@@ -128,6 +140,8 @@ <h2 id="logs">{{ $t('troubleshooting.logs') }}</h2>
           unpairAllPressed: false,
           unpairAllStatus: null,
           restartPressed: false,
+          showApplyMessage: false,
+          clients: [],
           logs: 'Loading...',
           logFilter: null,
           logInterval: null,
```
ReenigneArcher, line 147: Can you alphabetize these? clients through logInterval

src_assets/common/assets/web/pin.html
```diff
@@ -8,10 +8,11 @@
 <body id="app" v-cloak>
   <Navbar></Navbar>
   <div id="content" class="container">
-    <h1 class="my-4">{{ $t('pin.pin_pairing') }}</h1>
+    <h1 class="my-4 text-center">{{ $t('pin.pin_pairing') }}</h1>
     <form action="" class="form d-flex flex-column align-items-center" id="form">
       <div class="card flex-column d-flex p-4 mb-4">
-        <input type="text" pattern="\d*" placeholder="PIN" id="pin-input" class="form-control my-4" />
+        <input type="text" pattern="\d*" placeholder="PIN" id="pin-input" class="form-control mt-2" required />
```
ReenigneArcher, line 14: Can you update this placeholder to use the localized version? `navbar.pin` ... or move that one to `common`.

src/nvhttp.cpp
```diff
@@ -1148,8 +1196,48 @@ namespace nvhttp {
    */
   void
   erase_all_clients() {
```
ReenigneArcher, line 1198: Maybe we should do a force close as part of the unpair all?

src_assets/common/assets/web/troubleshooting.html
xanderfrangos: Done 👍

I misunderstood what you meant initially, so the device list is now alphabetical too... which is better than being unsorted. 😅

src_assets/common/assets/web/pin.html
xanderfrangos: Done 👍

src/nvhttp.cpp
xanderfrangos: Makes sense. "Unpair All" now force closes the open app/session. I've also updated the language to specify that the user only might need to do it themselves if unpairing individual devices.

ReenigneArcher: Sorry, I am not seeing the change that would do the force close. Can you direct me to that part of the code?

xanderfrangos: Sure: https://github.com/LizardByte/Sunshine/pull/2042/commits/1b6a828b1cdcda2854097b2d5788ee1d48f9dff9

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -324,7 +325,7 @@
     "configuration": "Configuration",
     "home": "Home",
     "password": "Change Password",
-    "pin": "Pin",
+    "pin": "PIN",
```
ReenigneArcher, line 328:
```suggestion
    "pin": "Pin",
```

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -365,9 +367,13 @@
     "restart_sunshine_success": "Sunshine is restarting",
     "troubleshooting": "Troubleshooting",
     "unpair_all": "Unpair All",
-    "unpair_all_desc": "Remove all your paired devices",
+    "unpair_all_desc": "Remove your paired devices. Individually unpaired devices with an active session will remain connected, but cannot start or resume a session.",
```
ReenigneArcher, line 370:
```suggestion
    "unpair_desc": "Remove your paired devices. Individually unpaired devices with an active session will remain connected, but cannot start or resume a session.",
```

The key was throwing me off. If this is the generic description now for all of the unpairing options we should rename the key.

src_assets/common/assets/web/troubleshooting.html
```diff
@@ -75,24 +75,39 @@ <h2 id="restart">{{ $t('troubleshooting.restart_sunshine') }}</h2>
...
+              {{ $t('troubleshooting.unpair_all') }}
+            </button>
+          </div>
+          <br />
+          <p class="mb-0">{{ $t('troubleshooting.unpair_all_desc') }}</p>
+          <div id="apply-alert" class="alert alert-success d-flex align-items-center mt-3" :style="{ 'display': (showApplyMessage ? 'flex !important': 'none !important') }">
+            <div class="me-2"><b>{{ $t('_common.success') }}</b> {{ $t('troubleshooting.unpair_single_success') }}</div>
+            <button class="btn btn-success ms-auto apply" @click="clickedApplyBanner">{{ $t('_common.dismiss') }}</button>
+          </div>
+          <div class="alert alert-danger mt-3" v-if="unpairAllStatus === true">
```
ReenigneArcher, line 94:
```suggestion
          <div class="alert alert-success mt-3" v-if="unpairAllStatus === true">
```

src_assets/common/assets/web/public/assets/locale/en.json
xanderfrangos: Updated 👍 https://github.com/LizardByte/Sunshine/pull/2042/commits/52cb99611012dabdebb8caae51fb211bd8fdc58b

ReenigneArcher:
@xanderfrangos sorry for the delays. This is the next one on our radar. Could you rebase it?

xanderfrangos:
@ReenigneArcher I've rebased locally, but need to test/check a few things first. Should have it up over the weekend.

ReenigneArcher:
@xanderfrangos thank you! Also, FYI, in the last 24 hours we changed our default branch back to `master`. Basically `nightly` was renamed to `master`. If you have any problems rebasing, that could be the reason.

xanderfrangos:
@ReenigneArcher It got a little messy, so I squashed my changes and rebased onto `master`. The only new change is fixing localization strings on the PIN page. Everything is now up-to-date and working on my end.

xanderfrangos:
On a side note: I know you're intentionally using "Pin", but "PIN" is what it should be. They have two different meanings. A "pin" is a small, narrow, sharp object (or the act of "pinning" something in place). A "PIN" is a "personal identification number". 

If you're trying to avoid "PIN" in all caps for some reason, you could use "pairing code"/"pairing" instead. But Moonlight already refers to it as a "PIN", so using any other language could confuse users.

ReenigneArcher:
@xanderfrangos Thanks! This should get merged soon unless we find any last minute things.

I find it easier to try to just have 1 commit and amend it when there are changes. It makes it a lot easier to rebase.

Regarding the PIN, that sounds right. We should probably address in a separate PR.

ReenigneArcher:
I don't know if there's a way to handle this without changes to Moonlight, but I did face this small issue.

When starting my testing initially, there was one unnamed device paired. I removed the Sunshine host from the client, and then repaired it. In Sunshine this showed 2 devices, one named and one unnamed.

When I removed the named one, it still had access since it was also the unnamed one.

Once, I removed the unnamed device on Sunshine's side, it was not able to connect.

Otherwise this looks good. I guess we can just recommend that people unpair all if they have any unnamed devices.

src_assets/common/assets/web/pin.html
```diff
@@ -24,7 +25,7 @@ <h1 class="my-4">{{ $t('pin.pin_pairing') }}</h1>
 
 <script type="module">
   import { createApp } from 'vue'
-  import i18n from './locale.js'
+  import i18nLocale from './locale.js'
```
ReenigneArcher, line 28:
```suggestion
  import i18n from './locale.js'
```

I'm confused why this was changed? I'm not seeing an i18nLocale in locale.js. Maybe it was just a bad copy pasta?

xanderfrangos: Something changed with the localization library since I last worked on the PIN page months ago. The "success"/"failure" messages after pairing were throwing errors. 

The `i18n` import was previously an object with the function `i18n.global.t` (see [line 53](https://github.com/LizardByte/Sunshine/blob/bf8fdbb076207d69b7884491728f3c0f26fda1ac/src_assets/common/assets/web/pin.html#L53) for example). However, that `i18n` import is now itself an async function that returns the expected object (containing `global.t`). So I needed to await the import to get the object. This meant that I either needed to rename [the import](https://github.com/LizardByte/Sunshine/blob/bf8fdbb076207d69b7884491728f3c0f26fda1ac/src_assets/common/assets/web/pin.html#L28) or rename the [returned object](https://github.com/LizardByte/Sunshine/blob/bf8fdbb076207d69b7884491728f3c0f26fda1ac/src_assets/common/assets/web/pin.html#L40). I decided to rename the import (from `i18n` to `i18nLocale`) instead of having a new name for the returned object.

It technically works fine as-is. `locale.js` uses the async function as the default export, so it doesn't matter what we call the imported variable. If you prefer the import still be called `i18n`, we can rename [the returned object](https://github.com/LizardByte/Sunshine/blob/bf8fdbb076207d69b7884491728f3c0f26fda1ac/src_assets/common/assets/web/pin.html#L40) instead of the import. Or if there's a better way to access the localization library, we can do that.

xanderfrangos:
> I don't know if there's a way to handle this without changes to Moonlight, but I did face this small issue.
> 
> When starting my testing initially, there was one unnamed device paired. I removed the Sunshine host from the client, and then repaired it. In Sunshine this showed 2 devices, one named and one unnamed.
> 
> When I removed the named one, it still had access since it was also the unnamed one.
> 
> Once, I removed the unnamed device on Sunshine's side, it was not able to connect.
> 
> Otherwise this looks good. I guess we can just recommend that people unpair all if they have any unnamed devices.

Yeah, there's not much that can be done about unnamed devices. There's currently no way to figure out which clients/devices they're for and properly name them (or check if they were already paired). But I expect it to be a short-lived issue, since setting a device name for new pairings is now required.

src_assets/common/assets/web/pin.html
ReenigneArcher: @Hazer do you have any thoughts on this?

Hazer: This way you're loading the locale twice, it does some API request behind, so it's not ideal.

When I created the `initApp` method, I already provided the `i18n` for injection, so you can inject it there

Hazer: My suggestion rewriting like this:

```vue
<script type="module">
  import { createApp } from 'vue'
  import { initApp } from './init'
  import Navbar from './Navbar.vue'

  let app = createApp({
    components: {
      Navbar
    },
    inject: ['i18n'],
    methods: {
      registerDevice(e) {
        let pin = document.querySelector("#pin-input").value;
        let name = document.querySelector("#name-input").value;
        document.querySelector("#status").innerHTML = "";
        let b = JSON.stringify({pin: pin, name: name});
        fetch("/api/pin", {method: "POST", body: b})
          .then((response) => response.json())
          .then((response) => {
            if (response.status.toString().toLowerCase() === "true") {
              document.querySelector(
                "#status"
              ).innerHTML = `<div class="alert alert-success" role="alert">${this.i18n.t('pin.pair_success')}</div>`;
              document.querySelector("#pin-input").value = "";
              document.querySelector("#name-input").value = "";
            } else {
              document.querySelector(
                "#status"
              ).innerHTML = `<div class="alert alert-danger" role="alert">${this.i18n.t('pin.pair_failure')}</div>`;
            }
          });
      }
    }
  });

  initApp(app);
</script>
```

This way you can inject `i18n` easily, and also make a bit clearer what is this code.

src_assets/common/assets/web/pin.html
```diff
@@ -8,10 +8,11 @@
 <body id="app" v-cloak>
   <Navbar></Navbar>
   <div id="content" class="container">
-    <h1 class="my-4">{{ $t('pin.pin_pairing') }}</h1>
+    <h1 class="my-4 text-center">{{ $t('pin.pin_pairing') }}</h1>
     <form action="" class="form d-flex flex-column align-items-center" id="form">
```
Hazer, line 12: With the change proposed below, use this:

```suggestion
    <form class="form d-flex flex-column align-items-center" id="form" @submit.prevent="registerDevice">
```

src_assets/common/assets/web/pin.html
Hazer: Or, in case you're a bit adventurous, replace the whole `pin.html` with:

```vue
<!DOCTYPE html>
<html lang="en" data-bs-theme="auto">

<head>
  <%- header %>
</head>

<body id="app" v-cloak>
  <Navbar></Navbar>
  <div id="content" class="container">
    <h1 class="my-4 text-center">{{ $t('pin.pin_pairing') }}</h1>
    <pin-card></pin-card>
  </div>
</body>

<script type="module">
  import { createApp } from 'vue'
  import { initApp } from './init'
  import Navbar from './Navbar.vue'
  import PinCard from './PinCard.vue'

  let app = createApp({
    components: {
      Navbar,
      PinCard
    }
  });

  initApp(app);
</script>
```

and create a new file `PinCard.vue`:

```vue
<script setup>
import { ref } from 'vue'

const inputPIN = ref('')
const inputName = ref('')
const status = ref()

function registerDevice() {
  let b = JSON.stringify({pin: inputPIN.value, name: inputName.value});
  console.log(b)
  fetch("/api/pin", {method: "POST", body: b})
    .then((response) => response.json())
    .then((response) => {
      if (response.status.toString().toLowerCase() === "true") {
        status.value = true
        inputPIN.value = ""
        inputName.value = ""
      } else {
        status.value = false
      }
    });
}
</script>

<template>
  <form class="form d-flex flex-column align-items-center" id="form" @submit.prevent="registerDevice">
    <div class="card flex-column d-flex p-4 mb-4">
      <input
        type="text"
        pattern="\d*"
        v-model.trim="inputPIN"
        :placeholder="`${$t('navbar.pin')}`"
        id="pin-input"
        class="form-control mt-2"
        autofocus
        required />
      <input
        type="text"
        v-model.trim="inputName"
        :placeholder="`${$t('pin.device_name')}`"
        id="name-input"
        class="form-control my-4"
        required />
      <button class="btn btn-primary">{{ $t('pin.send') }}</button>
    </div>
    <div class="alert alert-warning">
      <b>{{ $t('_common.warning') }}</b> {{ $t('pin.warning_msg') }}
    </div>
    <div id="status" v-if="status !== undefined">
      <div class="alert alert-success" role="alert" v-if="status === true">{{ $t('pin.pair_success') }}</div>
      <div class="alert alert-danger" role="alert" v-else>{{ $t('pin.pair_failure') }}</div>
    </div>
  </form>
</template>

<style scoped>

</style>

```

xanderfrangos: @ReenigneArcher I've updated it to @Hazer's code from the [latest comment](https://github.com/LizardByte/Sunshine/pull/2042#discussion_r1615940931). 👍

Hazer (approved):

ReenigneArcher (approved):

moi952:
Hi,


Thanks for the dev :)
I tried development, and I only see unknown clients, couldn't we rename them as we connect with them?

Because I have a lot of unknown clients and I don't know which clients I'm using and which are old clients (potentially Android emulator, or old OS with the tests I've done)

Thanks



image

ReenigneArcher:
It's recommended to unpair all clients one time. Then you can name them as you pair them again.

PR #2059, opened 2024-01-23
Working Docker cross-compiling for Fedora 39

chewi:
Improvements to #2020 so that Fedora 39 now works.

* Does not require QEMU at all!
* Caches dnf metadata and downloads between rebuilds.
* Respects TARGETARCH as a Docker variable, adding DNF_ARCH instead.
* Reduces downloads and image size by ignoring weak dependencies.

~The image is still a little large at 591MB, twice what you have now. I'm not sure why, but I don't think you were doing `dnf update` before? That can bloat it quite a bit.~

I think you may need `setcap` to give Sunshine the correct permissions, but you didn't seem to have this before?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -17,3 +17,6 @@ venv/
 
 # ignore artifacts
 artifacts/
+
+# ignore Dockerfiles themselves
+/docker/
--- a/cmake/compile_definitions/linux.cmake
+++ b/cmake/compile_definitions/linux.cmake
@@ -223,6 +223,11 @@ if (${SUNSHINE_TRAY} EQUAL 0 AND SUNSHINE_REQUIRE_TRAY)
     message(FATAL_ERROR "Tray icon is required")
 endif()
 
+pkg_check_modules(EVDEV libevdev REQUIRED)
+include_directories(SYSTEM ${EVDEV_INCLUDE_DIRS})
+link_directories(${EVDEV_LIBRARY_DIRS})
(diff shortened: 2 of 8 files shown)


chewi:
Hah, turns out I was comparing with the compressed size shown on Docker Hub.

Before: 902MB
After: 591MB :partying_face:

chewi:
Okay, a lot of this is down to me adding `--setopt=install_weak_deps=False`. This avoid big dependencies like LLVM via Mesa. I'm not sure whether we need such dependencies. If we do, we can maybe be more selective about it.

ReenigneArcher:
Thanks for digging into this. I will try to pick work back up on this soon.

chewi:
I've tried to take the same approach for Ubuntu, but APT is not playing ball at all. This doesn't surprise me much, to be honest.

ReenigneArcher:
I don't think it's as critical for Ubuntu...

"Build and push" is the install sunshine step... you can see here it's rather quick even with qemu. https://github.com/LizardByte/Sunshine/actions/runs/7664452338/job/20888787801

chewi:
Okay, think I'm about done now.

ReenigneArcher:
Sorry, I just can't follow this diff. Half of items marked as new were already in the target branch.

We also can't cache all the dependencies... we only get 10GB of cache, and regularly far exceed this as is.

image

I think I would rather just discuss what needs to change in #2020, rather than having a PR into it.

PR #1603, opened 2023-09-05
ci(tests): add test framework

ReenigneArcher:
## Description

This PR adds a test framework and initial unit tests. Replaces #1323.

Tests are run in all CI builds (except flatpak right now as it's failing to build). Doc Tests are skipped on macOS (due to dependencies relying on homebrew boost), but are enabled in the Macports build.

Other updates:
- Doxyfile was updated to latest version
- License corrected in ArchLinux PKGBUILD spec
- Moved flatpak shared modules to a submodule

Todo:

- [x] Fix flatpak build
- [x] Determine how to handle cmake components
- [x] Add coverage report
- [x] Upload coverage to codecov
- [x] Disable optimizations for test executable
- [x] Make homebrew spec run tests
- [x] Generate coverage report for Portfile or Homebrew build, and upload
- [x] Add fixtures for nvidia, amd, etc... automatically skipping tests that rely on those fixtures when conditions are incorrect (depends on #2185)
- [x] Add more initial tests

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Dependency update (updates to dependencies)
- [x] Documentation update (changes to documentation)
- [x] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I want maintainers to keep my branch updated

PR #2091, opened 2024-02-02
build(cmake): use CMAKE_SOURCE_DIR property

ReenigneArcher:
## Description

- Use CMAKE_SOURCE_DIR property in cmake
- Add quotes in case some paths have spaces
- Replace CMAKE_CURRENT_SOURCE_DIR with CMAKE_SOURCE_DIR in some places

This was originally part of #1603, but decided to break it out as it's not necessarily a change that is only needed for unit testing.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
 endif()
 
 # set the module path, used for includes
-set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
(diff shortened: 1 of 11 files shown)


cgutman:
Do we want to be using `CMAKE_CURRENT_SOURCE_DIR` rather than `CMAKE_SOURCE_DIR`? The former can point to other subfolders within the tree (though it seemingly doesn't for our current usage or it would have broken the build), while the latter is defined to always point to the root of the source tree.

ReenigneArcher:
`CMAKE_SOURCE_DIR` is probably better, which is what I used originally. I was using `CMAKE_CURRENT_SOURCE_DIR` only to keep it consistent with some other recent PRs... but now that I think about it `CMAKE_SOURCE_DIR` is probably what we want if we ever implement unit testing.

cgutman (approved):

PR #2095, opened 2024-02-03
RTSP encryption support

cgutman:
## Description
This PR completes the full end-to-end encryption implementation by encrypting the RTSP traffic that is sent during session negotation. This completes the chain of trust from HTTPS -> RTSP -> Audio/Video/Control to prevent downgrade attacks on the other encryption schemes via manipulation of RTSP messages between Sunshine and Moonlight.

The first 2 commits are refactoring to get the code ready for the addition of encryption then https://github.com/LizardByte/Sunshine/pull/2095/commits/204489cac638a2eef2aff1c0a2a5fd13031d522a adds the actual encryption. cc: @ABeltramo 

### Negotiation

Since RTSP is typically the flexible mechanism that we use to negotiate new options, encrypting RTSP is somewhat tricky to avoid breaking compatibility on both ends while still not allowing downgrade attacks against new clients and servers.

To allow the client to advertise support for RTSP encryption, clients send a new `corever` parameter to the launch/resume request. Clients that support RTSP encryption will report `corever=1` (or later). This new parameter can be used for similar changes in the future where this type of backwards compatibility is required prior to RTSP negotiation.

If the server also supports RTSP encryption, it will use an alternate URL scheme (`rtspenc://`) in the `sessionUrl0` XML attribute included in the launch/resume response if (and only if) `corever >= 1`. This tells the client that server also supports RTSP encryption and thus RTSP encryption will be used.

Due to the risk of downgrade attacks, it is _critical_ that client and server both do not accept unencrypted RTSP when encryption is supported (ie when server saw `corever >= 1` and when client saw `rtspenc://`).

### Encrypted Messages

The general encryption scheme is very similar to the new control stream encryption from #2025 with the addition of a dedicated length field in the encrypted message header (since TCP doesn't provide message boundaries). It also has the most significant bit set in the first byte of the message to allow RTSP implementations to tell the RTSP encrypted messages from unencrypted messages by just reading the first byte, if they need that ability.

An additional benefit of this change is that hosts no longer have to parse the RTSP message itself to understand when the full RTSP request has been received. Now it's possible to simply read the entire message header which will give you the length of the payload with no message parsing required.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -3,6 +3,7 @@
  * @brief todo
  */
 #include "network.h"
+#include "config.h"
 #include "utility.h"
 #include <algorithm>
 
@@ -171,6 +172,22 @@ namespace net {
     }
     else {
       return address.to_string();
+    }
+  }
+
+  /**
+   * @brief Returns the encryption mode for the given remote endpoint address.
+   * @param address The address used to look up the desired encryption mode.
+   * @return The WAN or LAN encryption mode, based on the provided address.
+   */
+  int
+  encryption_mode_for_address(boost::asio::ip::address address) {
+    auto nettype = net::from_address(address.to_string());
+    if (nettype == net::net_e::PC || nettype == net::net_e::LAN) {
+      return config::stream.lan_encryption_mode;
+    }
+    else {
+      return config::stream.wan_encryption_mode;
     }
   }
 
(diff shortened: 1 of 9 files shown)


ABeltramo:
Thanks for the ping, I'll work on it on my end too!

PR #2105, opened 2024-02-05
Add support for launching URLs and non-executable files on Windows

### LINKED ISSUE #2085: Do not rely on Windows PATH.. because Steam in not here.. autodected steam folder..

- [X] This issue is present in the nightly release

### Describe the Bug

Issues is described here:
https://www.reddit.com/r/MoonlightStreaming/comments/182afkg/moonlightsunshine_wont_launch_big_picture_mode/

### Expected Behavior

Detect Steam.exe path from Windows registry. Hardcoded path its working.. Steam not adding anything to Windows Path by default.. so original command: steam steam://open/bigpicture could not work.. 
 If its too hard for you, you can add some icon for user to navigate to Steam folder to WebGUI..

### Additional Context

_No response_

### Host Operating System

Docker

### Operating System Version

Windows 10 22H2

### Architecture

64 bit

0.21

### Package

Windows - installer

### GPU Type

Nvidia

### GPU Model

Geforce 1070

### GPU Driver/Mesa Version

Its does not matter, problem is elsewhere..

Its does not matter, problem is elsewhere..

### Config

```Shell
audio_sink = 572b5350-8943-43ff-a3a1-851983754ecc
```

### Apps

```json
Its does not matter, problem is elsewhere..
```

### Relevant log output

```shell
Its does not matter, problem is elsewhere..
```

cgutman:
## Description
This PR introduces some limited ShellExecute-like behavior to process launch on Windows. This allows opening URLs and non-executable files that `CreateProcess()` wouldn't handle on its own. URL handling is particularly useful because it allows our Steam app entry to be independent of the install path of Steam, and it even allows app shortcuts like `steam://rungameid/NNNNN` to be used to directly launch games independent of the app binary name and path.

This PR also updates the Steam entry for Windows by moving from detached launch using a semi-hardcoded binary path to using the `steam://open/bigpicture` URL as a regular command. This allows our new job tracking code to do its magic to monitor Steam and any launched games, which allows the user to terminate running Steam games via Moonlight.

### Issues Fixed or Closed
Fixes #2085

## Type of Change

## Checklist

## Branch Updates


src_assets/windows/assets/apps.json
```diff
@@ -1,17 +1,15 @@
 {
-  "env": {
-    "PATH": "$(PATH);$(ProgramFiles(x86))\\Steam"
-  },
+  "env": {},
```
ReenigneArcher, line 2: I'll review this in more detail tomorrow... but shouldn't we leave `PATH": "$(PATH)` only removing the default Steam path?

cgutman: `"PATH": "$(PATH)` should be equivalent to having no `env` block. We take all the environment variables from the user then apply any additions or modifications from the `env` block, so if the `env` block is empty, the app just gets the user's unmodified environment variables.

src/platform/windows/misc.cpp
```diff
@@ -550,6 +556,246 @@ namespace platf {
...
+   * @return A command string suitable for use by CreateProcess().
+   */
+  std::wstring
+  resolve_command_string(const std::string &raw_cmd, const std::wstring &working_dir, HANDLE token) {
+    std::wstring raw_cmd_w = converter.from_bytes(raw_cmd);
+
+    // First, convert the given command into parts so we can get the executable/file/URL without parameters
+    auto raw_cmd_parts = boost::program_options::split_winmain(raw_cmd_w);
+    if (raw_cmd_parts.empty()) {
+      // This is highly expected, but we'll just return the raw string and hope for the best.
```
cgutman, line 631: heh, `unexpected`

src_assets/windows/assets/apps.json
ReenigneArcher: Perfect!

src/platform/windows/misc.cpp
```diff
@@ -550,6 +556,243 @@ namespace platf {
...
+    // We support common substitutions (%*, %1, %2, %L, %W, %V, etc), but there are other
+    // uncommon ones that are unsupported here.
+    //
+    // https://web.archive.org/web/20111002101214/http://msdn.microsoft.com/en-us/library/windows/desktop/cc144101(v=vs.85).aspx
+    std::wstring cmd_string { shell_command_string.data() };
+    size_t match_pos = 0;
+    while ((match_pos = cmd_string.find_first_of(L'%', match_pos)) != std::wstring::npos) {
+      std::wstring match_replacement;
+      auto next_char = std::tolower(match_pos + 1 < cmd_string.size() ? cmd_string.at(match_pos + 1) : 0);
+      switch (next_char) {
```
ReenigneArcher, line 706: I don't know how to do it in C++, but I'd think this would be a good case to use regex for the different substitutions.

https://regex101.com/r/LuFITp/2

src/platform/windows/misc.cpp
```diff
@@ -550,6 +556,243 @@ namespace platf {
...
+        case L'5':
+        case L'6':
+        case L'7':
+        case L'8':
+        case L'9': {
+          // Arguments numbers are 1-based, except for %0 which is equivalent to %1
+          int index = next_char - L'0';
+          if (next_char != L'0') {
+            index--;
+          }
```
ReenigneArcher, line 732: Is there a way to just check against any integer? It's probably rare, but there could be more than 10 args.

src/platform/windows/misc.cpp
cgutman: Due to the replacement rules and the simplicity of the the way those argument substitutions work, I don't think a regex here would really simplify things.

`std::regex_replace()` can't do multiple different replacements in a single call. Calling `std::regex_replace()` multiple times would be worse than the single iteration through the string that the code here does, and it would incorrectly handle replacements where the replacement itself contains another `%1` or whatever.

We could use `std::regex_search()` and process the matches in a loop, but that basically ends up looking like the loop we have here.

src/platform/windows/misc.cpp
cgutman: `%0`-`%9` is actually exhaustive. The argument replacements are only a single `%` followed by a single character.

When I tested using a shell command of `cmd.exe /k echo %0 %1 %2 %3 %4 %5 %6 %7 %8 %9 %10` and opened the test via the Run dialog with `C:\users\camer\foo.shelltest 1 2 3 4 5 6 7 8 9 10 11 12`, the output of the resulting command prompt was: `C:\Users\camer\foo.shelltest C:\Users\camer\foo.shelltest 1 2 3 4 5 6 7 8 C:\Users\camer\foo.shelltest0`. The last replacement of `%10` was parsed by Windows as `%1` then `0` was appended to the end.

PR #2110, opened 2024-02-06
refactor(logging): separate logging from main

ReenigneArcher:
## Description

In order to properly implement unit testing, `main.cpp` needs to be split out... so `src/main.cpp` can be excluded as a source for the test binary. Otherwise when the test is started, it just starts Sunshine, instead of starting the tests.

Also, it will help with the ability to test things currently in main. See https://cuhkszlib-xiaoxing.readthedocs.io/en/latest/external/gtest/googletest/docs/FAQ.html#how-do-i-test-a-file-that-defines-main

This PR pulls logging functions out of main. I started with logging because it is probably the most used out of anything.

In some cases `main.h` could be removed from the includes if it was only relying on logging. Somehow the code in the new `logging.h` provides `assert` and in cases where that became undefined I opted to use `<cassert>` instead if logging was not needed in that file.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -20,12 +20,27 @@ using namespace std::literals;
 
 namespace bl = boost::log;
 
+boost::shared_ptr<boost::log::sinks::asynchronous_sink<boost::log::sinks::text_ostream_backend>> sink;
+
 bl::sources::severity_logger<int> verbose(0);  // Dominating output
 bl::sources::severity_logger<int> debug(1);  // Follow what is happening
 bl::sources::severity_logger<int> info(2);  // Should be informed about
 bl::sources::severity_logger<int> warning(3);  // Strange events
 bl::sources::severity_logger<int> error(4);  // Recoverable errors
 bl::sources::severity_logger<int> fatal(5);  // Unrecoverable errors
+
+/**
+ * @brief Flush the log.
+ *
+ * EXAMPLES:
+ * ```cpp
+ * log_flush();
+ * ```
+ */
+void
+log_flush() {
+  sink->flush();
+}
 
 /**
  * @brief Print help to stdout.
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -3,6 +3,9 @@
  * @brief Logging implementation file for the Sunshine application.
  */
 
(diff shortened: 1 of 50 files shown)


src/logging.cpp
```diff
@@ -0,0 +1,90 @@
...
+  }
+}  // namespace help
+
+namespace version {
+  int
+  entry(const char *name, int argc, char *argv[]) {
+    // just exit, the version has already been printed
+    return 0;
+  }
+}  // namespace version
```
cgutman, line 36: I think the help and version stuff should stay in main.cpp, since it's essentially part of the entrypoint of Sunshine.exe. For tests, you'll want your own entrypoint with its own help and other arguments, so you don't want to pull that it via `logging.cpp` when you build your tests.

ReenigneArcher: Okay. I have to separate `print_help` because it's used in config.cpp.

With the unit test PR, I need to include all of the executable sources except `main.cpp` (since it contains `main()`, but if I drop `main.cpp` then anything that relies on it will fail. The suggestion from [here](https://cuhkszlib-xiaoxing.readthedocs.io/en/latest/external/gtest/googletest/docs/FAQ.html#how-do-i-test-a-file-that-defines-main) is to have a file with nothing but `main()`.

These two namespaces don't necessarily have to be in `logging` though... I could add something like `main_helper` or `arg_processor`... What do you think of something like that? (Actually there are a couple others that could be moved as well... `restore_nvprefs_undo` and `gen_creds`)

Side note: `restore_nvprefs_undo` seems to be an undocumented argument. I should probably update the `print_help` for that.

ReenigneArcher: I moved the `help` and `version` namespace back to `main.cpp` for now. Also added a commit to update the `print_help`.

cgutman: `restore_nvprefs_undo` is an internal argument that we pass to ourselves when relaunching to restore settings. I don't think we need to document it because it's only intended for internal use by Sunshine itself.

ReenigneArcher: Alright, I removed that.

PR #2148, opened 2024-02-17
Optionally allow the system installation of nlohmann_json to be used

chewi:
## Description
As before, this makes the Gentoo Linux package a little simpler. I did say that last time was my final unbundling request, but you've added this dependency since.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [X] I want maintainers to keep my branch updated

PR #2161, opened 2024-02-22
Use nlohmann_json package instead of submodule

chewi:
## Description
As requested in #2148.

## Type of Change
- [X] Dependency update (updates to dependencies)

## Checklist

## Branch Updates

### DIFF
--- a/packaging/macos/Portfile
+++ b/packaging/macos/Portfile
@@ -31,12 +31,13 @@ post-fetch {
     system -W ${worksrcpath} "${git.cmd} submodule update --init --recursive"
 }
 
+depends_build-append     port:npm9 \
+                         port:pkgconfig
+
 depends_lib              port:avahi \
                          port:curl \
                          port:libopus \
-                         port:miniupnpc \
-                         port:npm9 \
-                         port:pkgconfig
+                         port:miniupnpc
 
 boost.version            1.81
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -769,6 +769,7 @@ jobs:
             mingw-w64-x86_64-cmake
             mingw-w64-x86_64-curl
             mingw-w64-x86_64-miniupnpc
+            mingw-w64-x86_64-nlohmann-json
             mingw-w64-x86_64-nodejs
             mingw-w64-x86_64-nsis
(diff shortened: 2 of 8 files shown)


chewi:
Okay, fixed!

chewi:
If you really wanted, I could add a generic fallback for when the .pc file isn't found.

cgutman:
> The AppImage failed because sadly the Ubuntu 20.04 package lacks the .pc file. The 22.04 package has it, so we could bump the AppImage distro. That won't help people manually building on 20.04 though. I know you've said that 20.04 is EOL in 2030 but that is paid support. I doubt anyone is going to pay to use an old Ubuntu just to run this. Otherwise support ends in just over a year's time.

We can't switch the AppImage to Ubuntu 22.04 yet. AppImages are bound to run on a glibc equal or greater than the version they were built against. In addition to the AppImage tooling itself preventing building with a glibc version that new, it would break commonly used distros like Debian Bullseye which ship with an older glibc than Ubuntu 22.04.

chewi:
I thought AppImage might use its own glibc, but I know that's not straightforward. Steam doesn't for the same reason.

It doesn't matter though. I just realised you only use nlohmann_json on Windows. :joy:

ReenigneArcher:
> It doesn't matter though. I just realised you only use nlohmann_json on Windows. 😂

What? 😂

chewi:
If you have any doubts, trigger the pipeline and find out. :grin:

packaging/macos/Portfile
```diff
@@ -31,12 +31,13 @@ post-fetch {
     system -W ${worksrcpath} "${git.cmd} submodule update --init --recursive"
 }
 
+depends_build            port:npm9 \
```
ReenigneArcher, line 34: `depends_build` fails, and I don't know why.

https://github.com/LizardByte/Sunshine/blob/a7b0eacecd750ff96d6b526e6ff990a57e0657c6/packaging/macos/Portfile#L34

ReenigneArcher (approved):

packaging/macos/Portfile
ReenigneArcher: Yep, it worked! I guess macports is creating it's own list so we need to append to it?

PR #2173, opened 2024-02-26
Use a common function to abort for debugging purposes

cgutman:
## Description
`std::abort()` destroys important state that is useful for debugging (particularly on Windows), so let's use a different method that disturbs as little process state as possible before breaking into an attached debugger or generating a coredump. 

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/misc.cpp
+++ b/src/platform/windows/misc.cpp
@@ -489,7 +489,7 @@ namespace platf {
       auto winerror = GetLastError();
       // Log the failure of reverting to self and its error code
       BOOST_LOG(fatal) << "Failed to revert to self after impersonation: "sv << winerror;
-      std::abort();
+      DebugBreak();
     }
 
     return ec;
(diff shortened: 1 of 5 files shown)


src/platform/windows/misc.cpp
```diff
@@ -489,7 +489,7 @@ namespace platf {
       auto winerror = GetLastError();
       // Log the failure of reverting to self and its error code
       BOOST_LOG(fatal) << "Failed to revert to self after impersonation: "sv << winerror;
-      std::abort();
+      DebugBreak();
```
ReenigneArcher, line 492: Is this one intentionally different from the others that are `lifetime::debug_trap();`, with a `log_flush();` before it?

cgutman: Yeah, since it's in platform-specific code, I just call the Windows-only function directly. As for the log flush, I think we're okay there, since this codepath mainly exists just to ensure we abort immediately if we're not able to end impersonation. I doubt this codepath will ever be reached in the real world, but we need to have it just in case.

PR #2196, opened 2024-02-29
Improve the description for the adaptive P-state setting

trudnorx:
The description for the adaptive P-state setting was updated to be more accurate, relevant and balanced for the average user. It also warns the user of the worst potential harmful effects of the setting, which was not done by the original description.

### DIFF
--- a/src_assets/common/assets/web/config.html
+++ b/src_assets/common/assets/web/config.html
@@ -947,19 +947,18 @@
                   </div>
                 </div>
 
-                <!-- Prefer lower encoding latency over increased power consumption -->
+                <!-- Prevent GPU downclocking -->
                 <div class="mb-3" v-if="platform === 'windows'">
-                  <label for="nvenc_latency_over_power" class="form-label">Prefer lower encoding latency over increased power
-                    consumption</label>
+                  <label for="nvenc_latency_over_power" class="form-label">Prevent GPU downclocking</label>
                   <select id="nvenc_latency_over_power" class="form-select" v-model="config.nvenc_latency_over_power">
                     <option value="disabled">Disabled</option>
                     <option value="enabled">Enabled (default)</option>
                   </select>
                   <div class="form-text">
-                    Adaptive P-State algorithm which NVIDIA drivers employ doesn't work well with low latency streaming, so
-                    sunshine requests high power mode explicitly.<br>
-                    Disabling it is not recommended since this can lead to
-                    <strong>significantly increased encoding latency</strong>.
+                    When set to enabled, this option prevents NVIDIA GPUs from downclocking via adaptive P-State while a Sunshine stream is active.<br>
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Please edit your PR description and follow our template...

https://github.com/LizardByte/.github/blob/master/.github/pull_request_template.md?plain=1

src_assets/common/assets/web/config.html
```diff
@@ -947,19 +947,18 @@ <h2 class="accordion-header">
...
                   </select>
                   <div class="form-text">
-                    Adaptive P-State algorithm which NVIDIA drivers employ doesn't work well with low latency streaming, so
-                    sunshine requests high power mode explicitly.<br>
-                    Disabling it is not recommended since this can lead to
-                    <strong>significantly increased encoding latency</strong>.
+                    When set to enabled, this option prevents NVIDIA GPUs from downclocking via adaptive P-State while a Sunshine stream is active.<br>
+                    This may lead to decreased encoding latency, but it will also render a major GPU power efficiency feature inoperative.<br>
+                    As a consequence, the GPU's temperature as well as power and fan usage will greatly increase.<br>
+                    This may lower the GPU's durability and increase your energy bills and likelihood of computer issues due to overheating.
```
cgutman, line 961: > This may lower the GPU's durability and increase your energy bills and likelihood of computer issues due to overheating.

Nvidia makes no such claims on the option to "Prefer maximum performance" in the official Control Panel. They only state increased power consumption.

image

There's no reason a non-defective GPU should fail by simply idling at the nominal GPU clock, nor does any Nvidia documentation I have found indicate this would be the case. Nvidia provides the option to _globally_ configure maximum performance mode (which would run 24/7 regardless of application), and that option would be quite dangerous if hardware damage could occur.

And for our usage, the P-State change is only happening _while Sunshine is actively streaming_. If you're leaving Moonlight connected 24/7, well you're going to have plenty of other power being wasted by doing that (keeping the client and host displays awake all the time, for one). The difference in power consumption reported by GPU-Z on my RTX 2080 between idle at the desktop and Sunshine streaming with "Prefer maximum performance" while idle at the desktop is only 30W, which is not something that's going to stress the cooling system of a GPU or gaming system at all. The card has to handle 5-10x that while gaming. My card even still operates completely passively while Sunshine is connected in maximum performance mode.

Sure, it's using more power with Maximum Performance mode and maybe some cards use more power while idle than others, but that's why this option exists for the user to select whether they want that tradeoff.

cgutman (commented):
I had a pending change for to clarify the verbiage here in #2181 that has landed. If the new wording still seems problematic, I'm fine with changing it to clarify things but telling users that we're going to break their computer by using an officially supported control panel option provided by Nvidia is a bridge too far IMHO.

PR #2226, opened 2024-03-06
Make debuginfo artifacts harder to confuse with the Windows portable build

cgutman:
## Description
If a user accidentally downloads the debuginfo package, they will have a pretty confusing experience. They will see Sunshine.exe in there but then get a cryptic error when they try to run it (because a bunch of dependencies are missing). Based on our download stats, this confusion seems to happen quite regularly (probably because it's near the top in the list of artifacts today).

This change should move the debug info package down below the regular Windows binaries rather than being at the top above the "Show all assets" button. It also reduces the chance of confusion by renaming the files inside so they are not executable.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -843,11 +843,15 @@ jobs:
       - name: Package Windows Debug Info
         working-directory: build
         run: |
-          # save the original binaries with debug info
+          # use .dbg file extension for binaries to avoid confusion with real packages
+          Get-ChildItem -File -Recurse | `
+            % { Rename-Item -Path $_.PSPath -NewName $_.Name.Replace(".exe",".dbg") }
+
+          # save the binaries with debug info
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -843,11 +843,15 @@ jobs:
       - name: Package Windows Debug Info
         working-directory: build
```
ReenigneArcher, line 844: If it would be more convenient you could change the shell to `shell: msys2 {0}` and use bash... But I'm fine with this as is too.

ReenigneArcher (approved):

.github/workflows/CI.yml
cgutman: yeah, I was worried that might mess up the 7z.exe invocation, so I just stuck with PS.

PR #2240, opened 2024-03-10
fix(linux): automatically migrate config directory

ReenigneArcher:
## Description

This PR fixes an issue on Linux where the config directory may be changed without the user knowing, after #2034 was merged... and was most prevalent on flatpak installs.

@istori1 FYI

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/misc.cpp
+++ b/src/platform/linux/misc.cpp
@@ -100,22 +100,54 @@ namespace platf {
 
   fs::path
   appdata() {
+    bool found = false;
+    bool migrate_config = true;
     const char *dir;
+    const char *homedir;
+    fs::path config_path;
+
+    // Get the home directory
+    if ((homedir = getenv("HOME")) == nullptr || strlen(homedir) == 0) {
+      // If HOME is empty or not set, use the current user's home directory
+      homedir = getpwuid(geteuid())->pw_dir;
+    }
 
     // May be set if running under a systemd service with the ConfigurationDirectory= option set.
-    if ((dir = getenv("CONFIGURATION_DIRECTORY")) != nullptr) {
-      return fs::path { dir } / "sunshine"sv;
+    if ((dir = getenv("CONFIGURATION_DIRECTORY")) != nullptr && strlen(dir) > 0) {
+      found = true;
+      config_path = fs::path(dir) / "sunshine"sv;
     }
     // Otherwise, follow the XDG base directory specification:
     // https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
-    if ((dir = getenv("XDG_CONFIG_HOME")) != nullptr) {
-      return fs::path { dir } / "sunshine"sv;
+    if (!found && (dir = getenv("XDG_CONFIG_HOME")) != nullptr && strlen(dir) > 0) {
+      found = true;
+      config_path = fs::path(dir) / "sunshine"sv;
     }
-    if ((dir = getenv("HOME")) == nullptr) {
-      dir = getpwuid(geteuid())->pw_dir;
+    // As a last resort, use the home directory
+    if (!found) {
+      migrate_config = false;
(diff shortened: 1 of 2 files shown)


src/platform/linux/misc.cpp
```diff
@@ -100,22 +100,43 @@ namespace platf {
...
+      migrate_config = false;
+      config_path = fs::path(homedir) / ".config/sunshine"sv;
     }
-    if ((dir = getenv("HOME")) == nullptr) {
-      dir = getpwuid(geteuid())->pw_dir;
+
+    // migrate from the old config location if necessary
+    if (migrate_config && found) {
+      fs::path old_config_path = fs::path(homedir) / ".config/sunshine"sv;
+      if (old_config_path != config_path && fs::exists(old_config_path)) {
```
cgutman, line 134: We should also make sure the new directory doesn't exist because we don't want to wipe out a config that's already there.

src/platform/linux/misc.cpp
```diff
@@ -100,22 +100,43 @@ namespace platf {
...
+        fs::rename(old_config_path, config_path);
```
cgutman, line 135: let's use the non-throwing variant that takes a `std::error_code`, then if we get an error from that, let's return `old_config_path` instead.

So it would look something like (untested:
```
std::error_code ec;
fs::rename(old_config_path, config_path, ec);
if (ec) {
  return old_config_path;
}
```

src/platform/linux/misc.cpp
```diff
@@ -100,22 +100,43 @@ namespace platf {
...
+    }
+    // As a last resort, use the home directory
+    if (!found) {
+      migrate_config = false;
+      config_path = fs::path(homedir) / ".config/sunshine"sv;
     }
-    if ((dir = getenv("HOME")) == nullptr) {
-      dir = getpwuid(geteuid())->pw_dir;
+
+    // migrate from the old config location if necessary
```
cgutman, line 131: I'm a little worried about unconditional migration for a problem that only really affects Flatpak. I'd feel better if we gated this on an environment variable that we could set in our Flatpak.

src/platform/linux/misc.cpp
```diff
@@ -100,22 +100,43 @@ namespace platf {
...
   fs::path
   appdata() {
+    bool found = false;
+    bool migrate_config = true;
     const char *dir;
+    const char *homedir;
+    fs::path config_path;
+
+    // Get the home directory
+    if ((homedir = getenv("HOME")) == nullptr) {
```
cgutman, line 110: While you're here, we should also check `&& strlen(homedir) > 0` and similar for all the `getenv()` stuff below.

The XDG specification says "not set or empty" and we're only checking the former right now.

src/platform/linux/misc.cpp
ReenigneArcher: I'm made this block of code only run for flatpak builds using a compile definition. I think using env variables would require people changing their commands for the flatpak launch to something like this.

`flatpak override --env=VARIABLE_NAME=VARIABLE_VALUE full.application.Name`

src/platform/linux/misc.cpp
ReenigneArcher: Missed this one before... resolved now.

src/platform/linux/misc.cpp
cgutman: You can set environment variables in the Flatpak manifest like this: https://github.com/flathub/com.moonlight_stream.Moonlight/blob/master/com.moonlight_stream.Moonlight.json#L16-L21

Your approach is fine too.

ReenigneArcher: I think I like the env approach better. I will adjust it real quick.

ReenigneArcher: Updated

cgutman (approved):

PR #2242, opened 2024-03-11
Fix udev rule issues

cgutman:
## Description
Our udev rules need to be evaluated before the built-in `73-seat-late.rules` which enables the uaccess tag application for existing logged on users. Renaming them to `60-sunshine.rules` (in line with similar rules from Steam) will fix this issue.

Additionally, we want to trigger reevaluation of the udev rules immediately upon installation, which doesn't seem to happen consistently today.

Testing:
- [x] Arch
- [x] Fedora
- [x] Ubuntu/Debian
- [x] AppImage

## Type of Change

## Checklist

## Branch Updates


ReenigneArcher (dismissed):

packaging/linux/Arch/sunshine.install
```diff
@@ -2,11 +2,18 @@ do_setcap() {
...
+}
+
 post_install() {
   do_setcap
+  do_udev_reload
 }
 
 post_upgrade() {
   do_setcap
+  do_udev_reload
```
ReenigneArcher, line 17: can we put an echo here to tell user's they need to reboot?

cgutman: Turns out a `modprobe uinput` does the trick. See https://github.com/chrippa/ds4drv/issues/93

packaging/linux/Arch/sunshine.install
```diff
@@ -2,11 +2,19 @@ do_setcap() {
   setcap cap_sys_admin+p $(readlink -f $(which sunshine))
 }
 
+do_udev_reload() {
+  udevadm control --reload-rules
+  udevadm trigger --property-match=DEVNAME=/dev/uinput
+  modprobe uinput || true
```
ReenigneArcher, line 8:
```suggestion
  modprobe uinput || echo "Warning: Failed to reload udev rules, a system restart is required!"
```

I don't know if anything is output to the user, but if so maybe something like this would be better?

cgutman: It's possible for `uinput` to not be built as a module in which case the `modprobe uinput` is not necessary.

ReenigneArcher: Okay, I'm fine with what you had as well.

PR #2248, opened 2024-03-12
Improve KMS debuggability and avoid known broken cases

cgutman:
## Description
This PR adds some additional debugging which would be very helpful to debug the recent KMS-related bug reports. It also makes the `SYS_CAP_ADMIN` error much more obvious by upgrading it to fatal severity in cases where we have no available fallback capture mechanism.

The only functional change in this PR is to avoid using KMS capture with a non-atomic driver under X11. In this scenario, we will not be able to capture the mouse cursor because the properties we rely on are atomic-only. Falling back to X11 capture will allow the cursor to be captured properly in that scenario (at a performance cost).

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/kmsgrab.cpp
+++ b/src/platform/linux/kmsgrab.cpp
@@ -625,9 +625,12 @@ namespace platf {
           }
 
           // Skip non-Nvidia cards if we're looking for CUDA devices
+          // unless NVENC is selected manually by the user
           if (mem_type == mem_type_e::cuda && !card.is_nvidia()) {
             BOOST_LOG(debug) << file << " is not a CUDA device"sv;
-            continue;
+            if (config::video.encoder != "nvenc") {
+              continue;
+            }
           }
 
           auto end = std::end(card);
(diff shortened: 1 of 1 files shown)


src/platform/linux/kmsgrab.cpp
```diff
@@ -1647,8 +1660,9 @@ namespace platf {
         }
 
         if (!fb->handles[0]) {
-          BOOST_LOG(error)
-            << "Couldn't get handle for DRM Framebuffer ["sv << plane->fb_id << "]: Possibly not permitted: do [sudo setcap cap_sys_admin+p sunshine]"sv;
+          BOOST_LOG(error) << "Couldn't get handle for DRM Framebuffer ["sv << plane->fb_id << "]: Probably not permitted"sv;
+          BOOST_LOG((window_system != window_system_e::X11 || config::video.capture == "kms") ? fatal : error)
+            << "You must run [sudo setcap cap_sys_admin+p $(readlink -f sunshine)] for KMS display capture to work!"sv;
```
ReenigneArcher, line 1665: Should we include a link to this `https://docs.lizardbyte.dev/projects/sunshine/en/latest/about/setup.html#install` for more info, or the event they are using the flatpak/appimage?

There's also a cmake var "SUNSHINE_BUILD_FLATPAK" that could be exposed as a compile definition, if you want to provide a different message for flatpak. And "SUNSHINE_BUILD_APPIMAGE" for AppImage.

cgutman: Leaving this as-is per our discussion on Discord that the link doesn't take you directly to where you need to read.

ReenigneArcher (approved):

PR #2251, opened 2024-03-13
AMF: rate control improvements

### LINKED ISSUE #1040: AMF Encoder on Radeon 5000 and newer overshoots encoding target at higher constant bitrates

### Describe the Bug

At high bitrate settings, Sunshine's AMF hardware encoding usually overshoots the requested encoding and sends frames too large for error correction or a stable connection. This issue occurs regardless of whether h264 or HEVC is being used, and regardless of what the `AMD Quality` setting is.

When this happens, the following error appears repeatedly in the logging/console output, with varying numbers for the `DATA_SHARDS_MAX` figure before the `>` sign:
`Warning: Number of fragments for reed solomon exceeds DATA_SHARDS_MAX 483 > 255,  skipping error correction`

This issue will often result in severe stuttering or freezes and audio glitches at best, while at worst it will cause outright disconnects due to an error for the bitrate being "too high" for the connection, even on a local network where all devices are connected via gigabit ethernet with optimal network throughput.

This issue can be reproduced by setting the rate control to `cbr` on the host, then selecting the highest bitrate possible in Moonlight on a client device. (150 Mbps)

The following workarounds exist: 
- Use `vbr_latency` mode instead.
- Keep bitrates below certain amounts if using `cbr` mode (Below 60 mbps seems to be consistently ok)
- `cqp` mode can be enabled with quality levels of 18 or higher without seeing the error as much.

A few other users of RDNA GPUs have experienced this issue, notably: #222
I have noticed this happening on my host machine, which currently has an **RX 6650 XT** installed.

### Expected Behavior

This issue does not occur with Nvidia GPUs and NVENC with constant bitrate settings of 150 mbps, nor does it occur when using AMF on an AMD APU such as the Ryzen 5 5600G while it's passing through an Nvidia GPU and encoding its output. I was able to test a GTX 1070 this way and found no issues using cbr while encoding with either integrated graphics AMF encoder nor the dGPU's NVENC encoder. Encoding with AMF on RX 5000 series and later hardware, it seems, is the sole situation in which this issue occurs.

### Additional Context

This has been tested across clients on multiple platforms (macOS, Linux and Android) as well as at multiple resolutions (1080p as well as 1440p) and even with different encoders (both `amf_h264` and `amf_hevc`.) It will occur in both `cbr` and `vbr_peak` encoder modes, and to a lesser (but still significant) extent in `cqp` mode. It occurs regardless of whether the client device is receiving the stream via ethernet or wi-fi.

Enabling or disabling **AMF Preanalysis** or **AMF Variance Based Adaptive Quantization (VBAQ)** in the encoder settings does not affect the issue.

### Host Operating System

Windows

### Operating System Version

Windows 10 22H2

### Architecture

64 bit

0.18.4

### Package

Windows - installer

### GPU Type

AMD

### GPU Model

RX 6650 XT

### GPU Driver/Mesa Version

23.3.1

### Config

```Shell
upnp = enabled
amd_coder = cabac
gamepad = x360
qsv_coder = auto
resolutions = [
    352x240,
    480x360,
    858x480,
    1280x720,
    1920x1080,
    2560x1080,
    3440x1440,
    1920x1200,
    3860x2160,
    3840x1600,
    1600x900
]
amd_usage = ultralowlatency
origin_pin_allowed = pc
key_rightalt_to_key_win = disabled
vt_realtime = enabled
vt_software = auto
origin_web_ui_allowed = lan
amd_quality = quality
min_log_level = 2
amd_vbaq = enabled
fec_percentage = 20
qp = 20
vt_coder = auto
dwmflush = enabled
hevc_mode = 0
amd_rc = cbr
amd_preanalysis = enabled
fps = [10,30,60,90,"75","120"]
qsv_preset = medium
```

### Apps

_No response_

### Relevant log output

```shell
Warning: Number of fragments for reed solomon exceeds DATA_SHARDS_MAX 263 > 255
```

psyke83:
## Description

Changes to AMF encoder:
* Add user-adjustable `enforce_hrd` option and set by default, to help constrain the target bitrate for VBR/CBR rate control methods. Rationale for making user-selectable: I observed encoder throttling with `enforce_hrd` on RX 570 several years ago at 1080p or higher, but cannot re-test to ensure the issue is not still present. May also cause quality degradation for scene transitions that previously would trigger bitrate overshoots.
* Set CBR as default RC. User testing seems to suggest that CBR + HRD gives the best results for bitrate constraints.
* Add `lowlatency_high_quality` preset.
* Fix parsing of AMF Usage preset.

### Issues Fixed or Closed

Closes #1040

## Type of Change

## Checklist

## Branch Updates


ReenigneArcher:
Homebrew is failing because your default branch doesn't match what we have set, `nightly`

https://github.com/LizardByte/Sunshine/actions/runs/8274317116/job/22639486783?pr=2251#step:7:209

You can fix this by changing your default branch in your fork, and we can re-run the workflow.

src_assets/common/assets/web/config.html
```diff
@@ -1045,9 +1045,9 @@ <h2 class="accordion-header">
           <label for="amd_rc" class="form-label">AMF Rate Control</label>
           <select id="amd_rc" class="form-select" v-model="config.amd_rc">
             <option value="cqp">cqp -- constant qp mode</option>
-            <option value="vbr_latency">vbr_latency -- latency constrained variable bitrate (default)</option>
+            <option value="vbr_latency">vbr_latency -- latency constrained variable bitrate</option>
             <option value="vbr_peak">vbr_peak -- peak constrained variable bitrate</option>
-            <option value="cbr">cbr -- constant bitrate</option>
+            <option value="cbr">cbr -- constant bitrate (default)</option>
```
ReenigneArcher, line 1050: This should probably be alphabetical.

psyke83:
> Homebrew is failing because your default branch doesn't match what we have set, `nightly`
> 
> https://github.com/LizardByte/Sunshine/actions/runs/8274317116/job/22639486783?pr=2251#step:7:209
> 
> You can fix this by changing your default branch in your fork, and we can re-run the workflow.

Thanks - that sorted the errors. I had to refresh the PR again, as I overlooked the preset RC set in the [video](https://github.com/LizardByte/Sunshine/blob/5b564a8c6ac43c8c6c844c03b8e4a06651ebc025/src/config.cpp#L349-L351) struct. Everything should be OK now.

ReenigneArcher (dismissed):

psyke83:
I've been testing CBR + HRD on my RX 6600 for a few days, and haven't noticed any issues with reduced quality or peak bitrate spikes, but I'm wondering if I should use this PR to update some more features related to AMF?

Looking at the reference version of ffmpeg we're currently building against:
1. There's a new usage preset  - [lowlatency_high_quality](https://github.com/FFmpeg/FFmpeg/blob/41f32f0bc42490c007920cf7b844f7a84852b201/libavcodec/amfenc_h264.c#L39) - but this is only present for H264 and HEVC.
2. There are [new](https://github.com/FFmpeg/FFmpeg/blob/41f32f0bc42490c007920cf7b844f7a84852b201/libavcodec/amfenc_hevc.c#L70-L72) rate control methods that work across all three codecs. Looking at the description on the AMF [wiki](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/wiki/Rate%20Control%20Methods), the HQ-VBR methods will not be useful in Sunshine, as they use a numerical quality target instead of a target bitrate. HQ-CBR might be worth exposing, with the caveat that the wiki recommends that this rate control method should be used with HRD and filler data, so exposing filler data as an option might be a good idea.
3. Continuing from the last point, the wiki also suggests that HRD and filler data should be enabled together for the standard CBR rate control. Perhaps this is another justification to expose the filler data option?
4. Perhaps it might be a good idea to add some description text to at least some of the web UI entries to bring it to parity with NVENC?

psyke83:
Update on my previous notes:

* HQCBR seems to flat-out break Sunshine. Perhaps it's not worth troubleshooting the issue; the wiki says it works best with a large CPB, so it may not be useful for low latency streaming.
* The AMF usage presets were not being parsed correctly on my system (`Usage` / `HevcUsage` was always 0  - `transcoding` - regardless of what I selected), which turns out to be a bug in the `usage_from_view()` conversion. Added a fix for the issue. I wonder if this is related to the `ultralowlatency` issue that @cgutman recently worked around; i.e. it was really failing from the transcoding profile, or perhaps an undefined value received from `usage_from_view` was parsed differently?
* I've gone ahead and added the new usage profiles, as I don't see any reason to withold these choices from users (and one in particular is a new low latency profile).
* I still haven't added descriptions to the AMF web UI entries, but I can do so if anyone agrees that it's worthwhile.

docs/source/about/advanced_usage.rst
```diff
@@ -1558,6 +1559,38 @@ keybindings
...
+
+**Default**
+   ``enabled``
+
+**Example**
+   .. code-block:: text
+
+      amd_enforce_hrd = enabled
+
+`amd_filler_data <https://localhost:47990/config/#amd_filler_data>`__
```
cgutman, line 1578: I'm not sure that there's any reason that filler data would ever be desirable for our use case. My understanding of these two options is: HRD basically acts as a strict maximum bitrate cap, while filler data acts as a strict minimum bitrate cap (though there are numerous issues with it not working properly in the AMF bug tracker).

Filler data is literally just appending filler NALUs with all zero data. It doesn't increase the usable bitrate in any way. I don't see the point in wasting time, power, and bandwidth to take those zero bytes and compute FEC, encrypt, and send them across the wire, then have to do the reverse again on the client side. And if we don't receive enough of the frame on the client side, we might have to throw away an entirely good frame just because we dropped some packets full of zeros on the wire.

psyke83: I suppose the most obvious reason why the option is available is due to HRD compliance. The stream is not supposed to underflow or overflow, and filler data prevents the former case from occuring, whilst `enforce_hrd` mostly (exclusively?) handles overflow. 

Underflow shouldn't be an issue unless we need to support a picky decoder that requires full HRD compliance, but I'm not aware of any such cases, so I assume that won't ever be an issue for us. I'll back out the filler data commit.

cgutman: Okay, sounds good to me. We could also ask the AMF folks to get a more authoritative answer if you want. They tend to be pretty responsive.

psyke83:
Just a note on the new usage presets: I removed the `high_quality` preset, as it breaks Sunshine for the same reason as `hqcbr`; both enable the PreAnalysis (not `preencode`) feature.

From looking at the [HEVC API](https://raw.githubusercontent.com/GPUOpen-LibrariesAndSDKs/AMF/master/amf/doc/AMF_Video_Encode_HEVC_API.md), the usage presets set the following:

* AMF_VIDEO_ENCODER_HEVC_LOWLATENCY_MODE
* AMF_VIDEO_ENCODER_HEVC_PRE_ANALYSIS_ENABLE
* AMF_VIDEO_ENCODER_HEVC_PEAK_BITRATE
* AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD
* AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_SKIP_FRAME_ENABLE
* AMF_VIDEO_ENCODER_HEVC_VBV_BUFFER_SIZE
* AMF_VIDEO_ENCODER_HEVC_ENFORCE_HRD
* AMF_VIDEO_ENCODER_HEVC_PREENCODE_ENABLE
* AMF_VIDEO_ENCODER_HEVC_ENABLE_VBAQ
* AMF_VIDEO_ENCODER_HEVC_HIGH_MOTION_QUALITY_BOOST_ENABLE
* AMF_VIDEO_ENCODER_HEVC_GOP_SIZE
* AMF_VIDEO_ENCODER_HEVC_QUALITY_PRESET
* AMF_VIDEO_ENCODER_HEVC_QUERY_TIMEOUT

Many of these properties are overridden by our configuration, and others by ffmpeg -- even if the default is not changed. Example: the `quality` option sets `AMF_VIDEO_ENCODER_HEVC_QUALITY_PRESET` - and will always override the usage preset's value, even if we weren't currently passing the `quality` option via Sunshine. The `high_quality` usage preset is breaking specifically due to `AMF_VIDEO_ENCODER_HEVC_PRE_ANALYSIS_ENABLE` being enabled, and that's one of the few properties that isn't [explicitly set](https://github.com/FFmpeg/FFmpeg/blob/41f32f0bc42490c007920cf7b844f7a84852b201/libavcodec/amfenc_hevc.c#L303) via ffmpeg.

The `lowlatency_high_quality` usage preset changes properties that otherwise can't be set via Sunshine (and aren't overwritten by ffmpeg):
* AMF_VIDEO_ENCODER_HEVC_LOWLATENCY_MODE (enabled; `ultralowlatency` is the only other profile that sets this)
* AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_SKIP_FRAME_ENABLE (disabled; all other profiles except `transcoding` enable this)
* AMF_VIDEO_ENCODER_HEVC_HIGH_MOTION_QUALITY_BOOST_ENABLE (enabled; all other profiles set this to disabled)

We could probably fix the `high_quality` usage preset by hardcoding `preanalysis` to false, but I'm not sure if it's worthwhile. The `hqcbr` RC won't work at all without PreAnalysis enabled, so that's a non-starter.

ReenigneArcher (changes requested):
Only change would be to remove any changes to localization files other than `en.json`. I need to make this more clear in the docs, but it should be ONLY `en.json`.

psyke83:
> Only change would be to remove any changes to localization files other than `en.json`. I need to make this more clear in the docs, but it should be ONLY `en.json`.

I assume the rationale is so that missing strings from other languages use the `en` fallback, correct? 

I've made the requested change and tested a local build, and I can see that this will cause issues with the current state of the localization files - specifically `amd_rc_cbr` and `amd_rc_vbr_latency`. Setting the UI to any other language than "English" will show the outdated strings that are currently present in all other files.

The solution would be to delete the outdated strings from all other files besides `en.json`. Shall I do that?

ReenigneArcher:
Essentially, the `en.json` file is the source. The rest are updated by crowdin automatically -> https://github.com/LizardByte/Sunshine/pull/2290

If files other than `en.json` are modified, there will be a few issues.

- Crowdin won't detect these
- I would have rebase the automated PR, and that would be difficult given the different languages and the way it does the commits.

But yes, they may be outdated for a period of time, until I merge the automated updates from crowdin.

psyke83:
> Essentially, the `en.json` file is the source. The rest are updated by crowdin automatically -> #2290
> 
> If files other than `en.json` are modified, there will be a few issues.
> 
> * Crowdin won't detect these
> * I would have rebase the automated PR, and that would be difficult given the different languages and the way it does the commits.
> 
> But yes, they may be outdated for a period of time, until I merge the automated updates from crowdin.

I see, thanks for explaining. I've pushed the updated PR that only touches `en.json`, as requested.

ReenigneArcher (approved):

PR #2252, opened 2024-03-13
Replace Basic Authentication with JWT Tokens, Added Login Page

TheElixZammuto:
## Description
This PR replaces the current Login Page (which is based of the Basic Authentication) with a custom Login Page that implements Cookies + JWT to handle the session system.

This allows us to customize the UX of the login page, and it's more compatible with password managers. 

The JWT Key is generated on the fly by Sunshine on each boot and is kept in memory, this allows us to not fiddle with revocation lists and storing safely the encryption key. The only side effect is that the credentials will be invalidated on a Sunshine Reboot, but the Web UI is already capable to handle this edge case and show a login modal when the credentials expire without reloading the entiere page.

This breaks the current API Authentication, but nobody uses the Web UI API as far as we know. If so, let us know!

### Screenshot

2024-03-13 21_15_35-Sunshine e altre 4 pagine - Profilo 1 - Microsoft​ Edge
2024-03-13 21_26_13-Sunshine e altre 6 pagine - Profilo 1 - Microsoft​ Edge

### Issues Fixed or Closed
https://ideas.moonlight-stream.org/posts/329/sunshine-use-login-page-rather-than-login-prompt

## Type of Change

## Checklist

## Branch Updates


.gitmodules
```diff
@@ -42,3 +42,6 @@
 	path = third-party/wlr-protocols
 	url = https://gitlab.freedesktop.org/wlroots/wlr-protocols
 	branch = master
+[submodule "third-party/jwt-cpp"]
```
cgutman, line 45: Is this packaged for the distros we support? I think we'll want the option for downstream packagers to supply their own package instead of using our submodules (like we do with other unmodified dependencies).

ReenigneArcher: I'm not sure this one is popular enough yet. Looks like they're in vcpkg though.

https://github.com/Thalhammer/jwt-cpp/blob/master/docs/install.md

@TheElixZammuto Could you make make this alphabetical in the `.gitmodules` file?

src/confighttp.cpp
```diff
@@ -32,6 +32,7 @@
 #include "file_handler.h"
 #include "globals.h"
 #include "httpcommon.h"
+#include "jwt-cpp/jwt.h"
```
ReenigneArcher, line 35:
```suggestion
#include <jwt-cpp/jwt.h>
```

This should probably also be up with the simple web server includes.

third-party/jwt-cpp
ReenigneArcher, line 1: Can you use commit `08bcf77a687fb06e34138e9e9fa12a4ecbe12332` which is the latest release

src/confighttp.cpp
```diff
@@ -720,16 +750,112 @@ namespace confighttp {
...
+      print_req(request);
+
+      auto g = util::fail_guard([&]() {
+        std::ostringstream data;
+        pt::write_json(data, outputTree);
+        response->write(data.str());
+      });
+
+      const SimpleWeb::CaseInsensitiveMultimap headers {
+        { "Set-Cookie", "sunshine_session=redacted; expires=Thu, 01 Jan 1970 00:00:00 GMT; Secure; HttpOnly; SameSite=Strict; Path=/" }
```
ReenigneArcher, line 827: Is it possible for someone to change the system time to bypass this?

src_assets/common/assets/web/fetch.js
```diff
@@ -0,0 +1,9 @@
+export default async (url,config) => {
+    const response = await fetch(url, config);
+    console.log(response);
+    if(response.status == 401){
+        const event = new Event("sunshine:session_expire");
+        window.dispatchEvent(event);
+    }
+    return response;
+};
```
ReenigneArcher, line 9:
```suggestion
};

```

If you use VS Code, I think you can adjust this behavior so it does it automatically.

src_assets/common/assets/web/login.html
```diff
@@ -0,0 +1,52 @@
...
+    },
+    methods: {
+      onLogin() {
+        document.location.href = '/';
+      }
+    }
+  });
+  console.log("App", app);
+  app.mount("#app");
+</script>
```
ReenigneArcher, line 52:
```suggestion
</script>

```

src_assets/common/assets/web/login.html
```diff
@@ -0,0 +1,52 @@
...
+      'ResourceCard': ResourceCard,
+      'LoginForm': LoginForm
+    },
+    data() {
+      return {
+        a: 1
+      }
+    },
+    mounted() {
+      console.log("Ciao!", this.$el)
```
ReenigneArcher, line 42: 🤣

third-party/jwt-cpp
TheElixZammuto: Done. I've also changed some instructions due to missing methods in 0.7.0 of jwt-cpp

src/confighttp.cpp
TheElixZammuto: The expire date of the token (not the Cookie) is checked by Sunshine (the creation and expire is inside the token). So changing on the client does not deal any effect. 

Changing the server time would but given that it would require admin privileges anyway I don't see how it might be a security vulnerability per se

ReenigneArcher: Okay, thanks. I just wanted to double check.

Nonary:
Hey, just curious, do we have to replace the basic authentication? I'd rather have it added as an additional authentication method. The goal is to reduce risk, but I don't think its necessary to totally kill off basic auth as its still a secure method, just not really recommended for browser usage due to it exposing password every request.

TheElixZammuto:
> Hey, just curious, do we have to replace the basic authentication? I'd rather have it added as an additional authentication method. The goal is to reduce risk, but I don't think its necessary to totally kill off basic auth as its still a secure method, just not really recommended for browser usage due to it exposing password every request.

Simply to simplify the authentication methods and not supporting both of them. This could be useful in situations where we would like to add different/new types of authentication systems without having to deal with this. btw I'll let @ReenigneArcher and @cgutman decide on that, I don't have a very strong opinion on that

Nonary:
I don't think its a good idea to drop basic authentication as it is a breaking change and it would make it practically impossible to use the webAPI outside of the browser if we implemented it via proper security practices. As for authentication becoming more difficult to maintain, it shouldn't be that way... most auth is added in via decorator or composite patterns to where it just is added on top of something.

ReenigneArcher:
@Nonary are you using the API for anything? I thought you were parsing the logs files for your projects.

We did a search on GitHub and didn't really find anyone doing anything with our API.

src/confighttp.cpp
Hazer, line 1: Given that we are already refactoring stuff here, can we move out all authentication logic/implementation to another file like `http_authenticator` or something, and only call it here?

Would also make it easier to unit test, fix, replace or even support multiple auth styles in the future.

Hazer (changes requested):

Epirr:
Would love this to work with 1Password

ReenigneArcher:
PR replaced by #2995.

Nonary:
I'm overtaking this with #3999 

Submitting comment so that those who were previously following this PR are notified that this feature might be added to sunshine coming soon.

Issue #1720, opened 2023-10-08
Controller doesn't work in some games


rezo552:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

In some games the controller is not being recognized (linux host) while I dont have the same issue when using Steam link. The only difference I can see:
The controller appears like this when using Sunshine:

[193895.683930] input: Microsoft X-Box 360 pad as /devices/virtual/input/input38

The controller appears like this when using Steam:
[193918.271611] input: Microsoft X-Box 360 pad **0** as /devices/virtual/input/input40

### Host Operating System

Docker

### Operating System Version

Debian GNU/Linux 12 (bookworm)

### Architecture

64 bit

### Sunshine commit or version

Sunshine version: v0.20.0.1303defb67c586447a03185d60eb20bea91a8eff

### Package

Linux - flatpak

### GPU Type

AMD

### GPU Model

5700G

### GPU Driver/Mesa Version

Mesa 23.2.1-1

### Config

```Shell
# If no external IP address is given, Sunshine will attempt to automatically detect external ip-address
# external_ip = 123.456.789.12

# Set the familly of ports used by Sunshine
# port = 47989

# The private key must be 2048 bits
# pkey = /dir/pkey.pem

# The certificate must be signed with a 2048 bit key
# cert = /dir/cert.pem

# The name displayed by Moonlight
# If not specified, the PC's hostname is used
# sunshine_name = Sunshine

# The minimum log level printed to standard out
#
# none -> no logs are printed to standard out
#
# verbose = [0]
# debug   = [1]
# info    = [2]
# warning = [3]
# error   = [4]
# fatal   = [5]
# none    = [6]
#
min_log_level = info

# A list of commands to be run before/after all applications.
#   If any of the prep-commands fail, starting the application is aborted.
global_prep_cmd = [{"do":"/usr/bin/xfce4-minimise-all-windows","undo":"/usr/bin/sunshine-stop"}]

# The origin of the remote endpoint address that is not denied for HTTP method /pin
# Could be any of the following values:
#   pc|lan|wan
#     pc: Only localhost may access /pin
#     lan: Only those in LAN may access /pin
#     wan: Anyone may access /pin
#
# origin_pin_allowed = pc

# The origin of the remote endpoint address that is not denied for HTTPS Web UI
[...]

PR #2261, opened 2024-03-14
Linux: add support for more virtual input devices

ABeltramo:
## Description

**This is still a draft, please let me know if something is not working**

Added support for the followings new virtual devices:
 - Nintendo Switch pro joypad
 - DualSense (PS5) joypad, including:
   - Gyroscope
   - Accelerometer
   - Touchpad
   - LED
   - Battery status
   - Force feedback? (is this supported in Moonlight already? I have to check..)
   - Microphone and Headset jack? (we should be able to plug into that as well, to be checked..)

And generally completely refactored all virtual input code using [inputtino](https://github.com/games-on-whales/inputtino) as a library

### Requirements

The DualSense implementation requires access to `/dev/uhid`, (more on this later..) this is easily done by setting the right permissions, similar to how we deal with `udev` already:

First we’ll add our user to the input group:
```
sudo usermod -a -G input $USER
```
Then we add an udev rule to allow access to /dev/uhid:
```
echo 'KERNEL=="uhid", GROUP="input", MODE="0660"' | sudo tee /etc/udev/rules.d/99-uhid.rules
```
Finally, we have to make sure that the uhid kernel module is loaded at boot:
```
echo "uhid" | sudo tee /etc/modules-load.d/uhid.conf
```

### Code details

The main file is `src/platform/linux/inputtino.cpp` I've kept the original `input.cpp` so that we can easily compare the two implementations and possibly add what's missing but the idea would be to completely replace that implementation with this. To switch from one implementation to the other you can just set the Cmake variable `-DSUNSHINE_USE_INPUTTINO=OFF`.  

Unfortunately we have to deal with `uhid` in order to properly emulate gyro, acceleration and touchpad; there's a bit of a rationale into why this is not possible with `udev` [here](https://github.com/games-on-whales/inputtino/blob/uhid-ps5/src/uhid/README.adoc). This might be a deal breaker for you guys but I hope it's not; if that's the case, we can still emulate a PS5 joypad using `udev` without supporting gyro and acceleration.  

I think bringing in `uhid` has also other benefits, since it sits at a lower level than `uinput`, we can really **emulate any kind of usb device** and possibly even completely integrate **usb over IP** if this will ever be considered as an addition to the protocol. 

As for the code itself, Inputtino has got unit tests using `libinput` for mouse, keyboard, pen and touchscreen and `SDL2` for testing joypads. These aren't complete (yet) but should cover most of the codebase.

### Future plans

 - I'd like to give it a go at implementing gyro and acceleration support for a Nintendo joypad too.
 - Ideally, since we are emulating the real joystick USB messages, this should also be a good foundation for a Windows virtual driver implementation. I know nothing about this but I'd like to do more research on it and possibly add a multiplatform implementation to inputtino. 
   - I know of the EV license and all the requirements, for some reason it's hard to convince companies to invest in this without a functioning prototype first :sweat_smile: 
 - Add support for trackpads in Moonlight? Generally it seems to not be the same as a touchscreen: it moves relatively instead of absolutely and has got different gestures.

### What's missing?

Most notably the XTest fallback that is present in `input.cpp` if this is something that is used we can easily add it here as well.  
Everything else should be implemented, please let me know if something is missing or misbeaving.

I'm probably forgetting something else..

### Issues Fixed or Closed

Might help with #1720

## Type of Change

## Checklist

## Branch Updates


cmake/compile_definitions/linux.cmake
```diff
@@ -234,14 +234,22 @@ if (${SUNSHINE_TRAY} EQUAL 0 AND SUNSHINE_REQUIRE_TRAY)
     message(FATAL_ERROR "Tray icon is required")
 endif()
 
+option(SUNSHINE_USE_INPUTTINO "Use inputtino for virtual input" ON)
```
ReenigneArcher, line 237: If we're going to use an option, the option flag should go into https://github.com/LizardByte/Sunshine/blob/8316f44e10f3f6ef27fc27b3ef6e69f359b1b667/cmake/prep/options.cmake#L28

But I don't see much benefit it making this optional. I did see your comment:

> I've kept the original input.cpp so that we can easily compare the two implementations and possibly add what's missing

and

> the idea would be to completely replace that implementation with this

I'd like to get to that point before merging this, if we can get enough people testing.

Or, if we can't get enough testers, we could rename the option to `SUNSHINE_USE_LEGACY_INPUT` (OFF by default) which would clearly indicate that it's not maintained.

src/platform/linux/inputtino.cpp
```diff
@@ -0,0 +1,648 @@
...
+
+  std::vector<std::string_view> &
+  supported_gamepads() {
+    static std::vector<std::string_view> gps {
+      "auto"sv, "x360"sv, "ds4"sv, "ps4"sv, "switch"sv
+    };
+
+    return gps;
+  }
+}  // namespace platf
```
ReenigneArcher, line 648:
```suggestion
}  // namespace platf

```

src/platform/linux/inputtino.cpp
```diff
@@ -0,0 +1,648 @@
...
+  }
+
+  enum ControllerType {
+    XboxOneWired,
+    DualSenseWired,
+    SwitchProWired
+  };
+
+  int
+  alloc_gamepad(input_t &input, const gamepad_id_t &id, const gamepad_arrival_t &metadata, feedback_queue_t feedback_queue) {
```
ReenigneArcher, line 379: The gamepad options will need be exposed in the config ui.

https://github.com/LizardByte/Sunshine/blob/8316f44e10f3f6ef27fc27b3ef6e69f359b1b667/src_assets/common/assets/web/config.html#L141C1-L150C15

Can probably become (untested).

```html
        <!-- Emulated Gamepad Type -->
        <div class="mb-3" v-if="config.controller === 'enabled' && (platform === 'windows' || platform === 'linux')">
          <label for="gamepad" class="form-label">Emulated Gamepad Type</label>
          <select id="gamepad" class="form-select" v-model="config.gamepad">
            <option value="auto">Automatic</option>
            <option value="ds4">DS4 (PS4)</option>
            <option value="switch" v-if="platform === 'linux'">Nintendo Pro (Switch)</option>
            <option value="x360">X360 (Xbox 360)</option>
          </select>
          <div class="form-text">Choose which type of gamepad to emulate on the host</div>
        </div>
```

src/platform/linux/inputtino.cpp
```diff
@@ -0,0 +1,648 @@
...
+    ControllerType selectedGamepadType;
+
+    if (config::input.gamepad == "x360"sv) {
+      BOOST_LOG(info) << "Gamepad " << id.globalIndex << " will be Xbox 360 controller (manual selection)"sv;
+      selectedGamepadType = XboxOneWired;
```
ReenigneArcher, line 384: Should we add an option for xbox-one instead of using x360?

src/platform/linux/inputtino.cpp
ReenigneArcher, line 1: Can you add documentation blocks throughout this file?

ReenigneArcher (changes requested):

src/platform/linux/inputtino.cpp
ABeltramo: The docstrings are already present in the `.h` files where the functions are definited. For example, CLion automatically shows the right docstring for one of the functions in `inputtino.cpp`
Screenshot 2024-03-19 at 14 56 53

src/platform/linux/inputtino.cpp
ABeltramo: I wasn't sure how this was handled in the UI so I've re-used what was already defined there. Ideally we should also add the Nintendo joypad which is now present with this PR. 
I should look further into this..

cmake/compile_definitions/linux.cmake
ABeltramo: Thanks for pointing to `options.cmake` I must have missed it, makes sense.

.gitmodules
```diff
@@ -42,3 +42,7 @@
 	path = third-party/wlr-protocols
 	url = https://gitlab.freedesktop.org/wlroots/wlr-protocols
 	branch = master
+[submodule "third-party/inputtino"]
+	path = third-party/inputtino
+	url = https://github.com/games-on-whales/inputtino.git
+	branch = uhid-ps5
```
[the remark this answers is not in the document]
ABeltramo: Yep yep, I'd like to merge it and tag it once it's ~stable

src/platform/linux/inputtino.cpp
ReenigneArcher: I think there's a config option to do it automatically on save, also to remove trailing whitespace.

src/platform/linux/inputtino.cpp
ReenigneArcher: I think my comment [here](https://github.com/LizardByte/Sunshine/pull/2261#discussion_r1530351369) should be enough to help with the UI.

I also notice you have `ds4` and `ps4` under `supported_gamepads` (and we have that in our Windows input file as well)... I'm not sure what the difference is, but `ps4` doesn't appear to be exposed in the UI.

ReenigneArcher:
Homebrew is failing due to the name of your default branch. If you make your default `nightly` that error will go away.

ABeltramo:
Yep, libudev is used by inputtino, I thought it was already a dependency of libevdev, but it looks like that's not the case. I can easily make it optional since it's only used for an utility method that is not needed for Sunshine.

ReenigneArcher (commented):
A few documentation suggestions...

https://github.com/LizardByte/Sunshine/blob/8316f44e10f3f6ef27fc27b3ef6e69f359b1b667/gh-pages-template/index.html#L218-L227

Could be updated to:

```html
                                            <p class="mb-0">
                                                Sunshine emulates an Xbox, DS4, or Nintendo Switch controller. Use nearly any
                                                controller on your Moonlight client!<br>
                                                <small>
                                                  <ul>
                                                    <li>Switch emulation is only available on Linux.</li>
                                                    <li>Gamepad emulation is not currently supported on macOS.</li>
                                                  </ul>
                                                </small>
                                            </p>
```

https://github.com/LizardByte/Sunshine/blob/8316f44e10f3f6ef27fc27b3ef6e69f359b1b667/docs/source/about/advanced_usage.rst#L141-L168

Could be updated to:

```rst
`gamepad <https://localhost:47990/config/#gamepad>`__
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

**Description**
   The type of gamepad to emulate on the host.

   .. attention:: Gamepad is not supported on macOS.

**Choices**

.. table::
   :widths: auto

   =====     ===========
   Value     Description
   =====     ===========
   auto      Selected based on information from client
   x360      Xbox 360 controller
   ds4       DualShock 4 controller (PS4)
   switch    Nintendo Switch (Linux only)
   =====     ===========

**Default**
   ``auto``

**Example**
   .. code-block:: text

      gamepad = auto
```

ABeltramo:
I've removed the `libudev` dependency, will CI run automatically or can someone give it another go with the latest changes?

ReenigneArcher:
> will CI run automatically or can someone give it another go with the latest changes?

I can run it once conflicts are resolved (it won't allow me to run it with conflicts)

ABeltramo:
I've rebased on top of latest nightly and hopefully also updated to the latest stable of inputtino. :crossed_fingers:

src/platform/linux/inputtino.cpp
ABeltramo: Do you want me to copy paste the docstrings from the .h files?

ReenigneArcher: I won't require it, but it would probably be good to have some info here. It's for our sphinx/breathe docs, which I don't think it will pick up anything from your libraries `.h` files.

This is very much a work in progress, but the more documentation in the source the better. https://docs.lizardbyte.dev/projects/sunshine/en/latest/source_code/source_code.html

ReenigneArcher:
Changing the default branch of your fork to `nightly` will resolve the homebrew failures. https://github.com/LizardByte/Sunshine/actions/runs/8458201889/job/23174743604

src/platform/linux/inputtino.cpp
ABeltramo: Ha, I see that can be confusing; in here, most of the methods are already defined in Sunshine headers!  
Like: https://github.com/LizardByte/Sunshine/blob/nightly/src%2Fplatform%2Fcommon.h#L674-L742

ReenigneArcher: Okay, that's good enough.

ReenigneArcher:
Other failures seem to be because there are uninitialized variables. https://github.com/LizardByte/Sunshine/actions/runs/8458201886/job/23174863908?pr=2261#step:10:4540

ReenigneArcher:
What happened?

ABeltramo:
> What happened?

Changing the branch name to `nightly` automatically closed the PR (I guess because I had to delete the `inputtino` branch).. I've re-opened it here: <https://github.com/LizardByte/Sunshine/pull/2315>

ReenigneArcher:
Oh... that wasn't what I meant.

The "default" branch has to be nightly.
image

I'll try to adjust the homebrew spec so it just uses whatever branch the default is of the PR.

ReenigneArcher:
No worries... It's an unintended consequence of how the homebrew build is done, which basically checks out the forks branch... and the homebrew audit wants the default branch in the `head` field.

PR #2294, opened 2024-03-24
fix(linux/x11): use config value for output_name

ReenigneArcher:
## Description

When developing encoder fixtures for #1603, I noticed that X11 capture looks for the DISPLAY environment variable instead of using the config value.

I think if we got the value from the config it would make it so users don't have to set an ENV variable when trying to start Sunshine via SSH.

This is completely untested, so just leaving this here for anyone who wants to test.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/x11grab.cpp
+++ b/src/platform/linux/x11grab.cpp
@@ -797,7 +797,8 @@ namespace platf {
 
     BOOST_LOG(info) << "Detecting monitors"sv;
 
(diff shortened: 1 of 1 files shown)


cgutman:
I think the existing code is correct. The X11 "Display" is not really a display in true sense - it is more of a generic abstraction. We do use output name during the actual enumeration of the individual monitors to select the right one to stream.

PR #2315, opened 2024-03-28
Linux: add support for more virtual input devices

ABeltramo:
## Description

I had to rename the branch for CI to work, this is a continuation of #2261

Added support for the followings new virtual devices:
 - Nintendo Switch pro joypad
 - DualSense (PS5) joypad, including:
   - Gyroscope
   - Accelerometer
   - Touchpad
   - LED
   - Battery status
   - Force feedback? (is this supported in Moonlight already? I have to check..)
   - Microphone and Headset jack? (we should be able to plug into that as well, to be checked..)

And generally completely refactored all virtual input code using [inputtino](https://github.com/games-on-whales/inputtino) as a library

### Requirements

The DualSense implementation requires access to `/dev/uhid`, (more on this later..) this is easily done by setting the right permissions, similar to how we deal with `udev` already:

First we’ll add our user to the input group:
```
sudo usermod -a -G input $USER
```
Then we add an udev rule to allow access to /dev/uhid:
```
echo 'KERNEL=="uhid", GROUP="input", MODE="0660"' | sudo tee /etc/udev/rules.d/99-uhid.rules
```
Finally, we have to make sure that the uhid kernel module is loaded at boot:
```
echo "uhid" | sudo tee /etc/modules-load.d/uhid.conf
```

### Code details

The main file is `src/platform/linux/inputtino.cpp` I've kept the original `input.cpp` so that we can easily compare the two implementations and possibly add what's missing but the idea would be to completely replace that implementation with this. To switch from one implementation to the other you can just set the Cmake variable `-DSUNSHINE_USE_INPUTTINO=OFF`.  

Unfortunately we have to deal with `uhid` in order to properly emulate gyro, acceleration and touchpad; there's a bit of a rationale into why this is not possible with `udev` [here](https://github.com/games-on-whales/inputtino/blob/uhid-ps5/src/uhid/README.adoc). This might be a deal breaker for you guys but I hope it's not; if that's the case, we can still emulate a PS5 joypad using `udev` without supporting gyro and acceleration.  

I think bringing in `uhid` has also other benefits, since it sits at a lower level than `uinput`, we can really **emulate any kind of usb device** and possibly even completely integrate **usb over IP** if this will ever be considered as an addition to the protocol. 

As for the code itself, Inputtino has got unit tests using `libinput` for mouse, keyboard, pen and touchscreen and `SDL2` for testing joypads. These aren't complete (yet) but should cover most of the codebase.

### Future plans

 - I'd like to give it a go at implementing gyro and acceleration support for a Nintendo joypad too.
 - Ideally, since we are emulating the real joystick USB messages, this should also be a good foundation for a Windows virtual driver implementation. I know nothing about this but I'd like to do more research on it and possibly add a multiplatform implementation to inputtino. 
   - I know of the EV license and all the requirements, for some reason it's hard to convince companies to invest in this without a functioning prototype first :sweat_smile: 
 - Add support for trackpads in Moonlight? Generally it seems to not be the same as a touchscreen: it moves relatively instead of absolutely and has got different gestures.

### What's missing?

Most notably the XTest fallback that is present in `input.cpp` if this is something that is used we can easily add it here as well.  
Everything else should be implemented, please let me know if something is missing or misbeaving.

I'm probably forgetting something else..

### Issues Fixed or Closed

Might help with #1720

## Type of Change

## Checklist

## Branch Updates


ReenigneArcher:
I think for anyone to test this it will need the config ui updated to handle the options.

Should be basically this https://github.com/LizardByte/Sunshine/pull/2261#discussion_r1530351369 ... except slightly different now that localization was added (https://docs.lizardbyte.dev/projects/sunshine/en/nightly/contributing/localization.html#extraction)

Or, I could update it if you want?

ABeltramo:
Thanks for the offer!  
I have to look at how the UI works, but it shouldn't be a hard requirement since Moonlight will already tell which joypad type should be emulated. My understanding is that the UI option is only for forcing a specific type to be picked instead, but admittedly, I haven't looked at this too much.

I've been focusing on testing my real PS5 pad against the virtual one, and that's already being picked up without changing anything in the UI.

ReenigneArcher:
> My understanding is that the UI option is only for forcing a specific type to be picked instead

Yea, there's an automatic mode. I personally wouldn't be able to test anything other than Xbox using auto mode. Does Moonlight send anything other than xbox360 or ds4 though?

ABeltramo:
Yep, the 3 supported types are currently Xbox, PS and Nintendo. I've personally tested Xbox and PS whilst the Nintendo implementation has been tested by a user using Wolf (plus unit tests for all of them using SDL).  
With the force option, I should be able to test them all, I should probably work on it sooner rather than later..

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -158,8 +158,11 @@
     "gamepad_auto": "Automatic selection options",
     "gamepad_desc": "Choose which type of gamepad to emulate on the host",
     "gamepad_ds4": "DS4 (PS4)",
+    "gamepad_ds5": "DS5 (PS5)",
+    "gamepad_switch": "Nintendo Pro (Switch)",
     "gamepad_manual": "Manual DS4 options",
     "gamepad_x360": "X360 (Xbox 360)",
+    "gamepad_xone": "XOne (Xbox One)",
```
ABeltramo, line 165: Will the localization library automatically fallback to EN if missing or should I copy this keys also in the other json files?

src_assets/common/assets/web/config.html
```diff
@@ -147,7 +147,7 @@ <h1 class="my-4">{{ $t('config.configuration') }}</h1>
           <div class="form-text">{{ $t('config.controller_desc') }}</div>
         </div>
 
-        <!-- Emulated Gamepad Type -->
+        <!-- Emulated Gamepad Type (Windows only) -->
```
ReenigneArcher, line 150: We can have one emulated gamepad block that covers Windows and Linux.

```html
        <!-- Emulated Gamepad Type -->
        <div class="mb-3" v-if="config.controller === 'enabled' && platform !== 'macos'">
          <label for="gamepad" class="form-label">{{ $t('config.gamepad') }}</label>
          <select id="gamepad" class="form-select" v-model="config.gamepad">
            <option value="auto">{{ $t('_common.auto') }}</option>
            <option value="ds4" v-if="platform === 'windows'">{{ $t('config.gamepad_ds4') }}</option>
            <option value="ds5" v-if="platform === 'linux'">{{ $t('config.gamepad_ds5') }}</option>
            <option value="switch" v-if="platform === 'linux'">{{ $t('config.gamepad_switch') }}</option>
            <option value="x360" v-if="platform === 'windows'">{{ $t('config.gamepad_x360') }}</option>
            <option value="xone" v-if="platform === 'linux'">{{ $t('config.gamepad_xone') }}</option>
          </select>
          <div class="form-text">{{ $t('config.gamepad_desc') }}</div>
        </div>
```

ReenigneArcher (changes requested):

src_assets/common/assets/web/config.html
ABeltramo: Yeah, I've started with that and then refactored like it is now.  
I think it's cleaner with two separate blocks since they don't share any option, but I'd be happy to switch it back if you prefer it that way..

ReenigneArcher: I'd prefer the single block since it's less code to maintain, lol

ReenigneArcher: The config layout has been refactored. @Hazer can provide guidance if needed.

Hazer:
@ReenigneArcher @ABeltramo I'm taking over this PR to https://github.com/LizardByte/Sunshine/pull/2606 so I can update the PR to allow for merging. 

@ABeltramo Can you review my changes to your PR, just to be sure I didn't hurt your work?

ABeltramo:
Thanks for picking this up, I thought nobody was interested in this..  

I think we should automatically fallback in case some requirements are missing, for example:
- uhid not available -> uinput
- uinput not available -> xinput
- xinput not available -> no input I guess..

This would let us cover more use cases and allow people that don't want to expose udev (to tighten security/on unprivileged containers or VMs) to still be able to run a session with just mouse and keyboard over Xorg. I believe there's no equivalent solution for Wayland, but I haven't looked too much into it..

Apart from that, there's still some quirk over gyro and acceleration on my PS5 virtual joypad but everything else seems to work, have you had a chance to run it on your end too?

ReenigneArcher:
Closing, replaced by #2606

PR #2320, opened 2024-03-29
Revert "Add capture using WinRT Windows.Graphics.Capture API."

ReenigneArcher:
## Description

Reverts LizardByte/Sunshine#2149

Due to many bugs, such as:

- high cpu usage
- high ram usage
- UAC prompt not able to trigger (starting Sunshine from start menu icon)
- unable to pair new clients

These issues are likely caused by changing from MINGW64 to UCRT64, or possibly from changing from C++17 to C++20.

@tez011 FYI. I think we should break your PR into two parts, or maybe 3 parts.

1. Change from C++17 to C++20
2. Change from MINGW64 to UCRT64
3. New feature

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Dependency update (updates to dependencies)
- [x] Documentation update (changes to documentation)
- [x] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I want maintainers to keep my branch updated

PR #2149, opened 2024-02-18
Add capture using WinRT Windows.Graphics.Capture API.

tez011:
## Description
This adds a capture backend on Windows that uses the [Windows.Graphics.Capture API](https://learn.microsoft.com/en-us/windows/uwp/audio-video-camera/screen-capture). This is a more recent API that's capable of capturing the Xbox Game Bar, which the DXGI Desktop Duplication API seemingly cannot do.

To accomplish this, I used a slightly different build environment known as MSYS UCRT64. I also bumped the compilation standard to C++20 so that the GNU C++ compiler can handle the coroutine definitions present in WinRT headers. The UCRT64 and MinGW64 environment appear to be binary-compatible, and Sunshine's dependencies translate smoothly across this boundary. The build documentation in this change reflects these modifications.

This change can be enabled by setting `capture = wgc` in sunshine.conf on Windows servers. When this setting is off, no functionality change should be observable.
I've observed CPU and memory usage trends over a long streaming period and there don't appear to be any new leaks.

### Issues Fixed or Closed
Fixes #832 (manually verified)

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Dependency update (updates to dependencies)
- [x] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I want maintainers to keep my branch updated

PR #2322, opened 2024-03-29
Use C++20.

tez011:
## Description

#2149 was reverted in #2320 and needs to be split into three changes:

- C++20 (this one!)
- UCRT64
- WGC

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/cmake/targets/common.cmake
+++ b/cmake/targets/common.cmake
@@ -36,7 +36,7 @@ endif()
 
 target_link_libraries(sunshine ${SUNSHINE_EXTERNAL_LIBRARIES} ${EXTRA_LIBS})
 target_compile_definitions(sunshine PUBLIC ${SUNSHINE_DEFINITIONS})
-set_target_properties(sunshine PROPERTIES CXX_STANDARD 17
+set_target_properties(sunshine PROPERTIES CXX_STANDARD 20
         VERSION ${PROJECT_VERSION}
         SOVERSION ${PROJECT_VERSION_MAJOR})
 
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -7,6 +7,7 @@
 #include "logging.h"
 #include "utility.h"
 #include <algorithm>
+#include <sstream>
 
 using namespace std::literals;
 
@@ -169,7 +170,9 @@ namespace net {
   addr_to_url_escaped_string(boost::asio::ip::address address) {
     address = normalize_address(address);
     if (address.is_v6()) {
-      return "["s + address.to_string() + ']';
+      std::stringstream ss;
+      ss << '[' << address.to_string() << ']';
+      return ss.str();
     }
     else {
       return address.to_string();
--- a/src/platform/linux/input.cpp
+++ b/src/platform/linux/input.cpp
@@ -1510,7 +1510,7 @@ namespace platf {
     std::stringstream ss;
     ss << std::hex << std::setfill('0');
     for (const auto &ch : str) {
-      ss << ch;
+      ss << static_cast<uint_least32_t>(ch);
     }
 
     std::string hex_unicode(ss.str());
--- a/src/platform/linux/kmsgrab.cpp
+++ b/src/platform/linux/kmsgrab.cpp
@@ -614,7 +614,7 @@ namespace platf {
         for (auto &entry : fs::directory_iterator { card_dir }) {
           auto file = entry.path().filename();
 
-          auto filestring = file.generic_u8string();
+          auto filestring = file.generic_string();
           if (filestring.size() < 4 || std::string_view { filestring }.substr(0, 4) != "card"sv) {
             continue;
           }
@@ -1641,7 +1641,7 @@ namespace platf {
     for (auto &entry : fs::directory_iterator { card_dir }) {
       auto file = entry.path().filename();
 
-      auto filestring = file.generic_u8string();
+      auto filestring = file.generic_string();
       if (std::string_view { filestring }.substr(0, 4) != "card"sv) {
         continue;
       }
--- a/src/stream.h
+++ b/src/stream.h
@@ -3,6 +3,7 @@
  * @brief todo
  */
 #pragma once
+#include <utility>
 
 #include <boost/asio.hpp>
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -323,6 +323,9 @@ jobs:
           # allow newer gcc
           sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
 
+          # allow libfuse2 for appimage on 22.04
+          sudo add-apt-repository universe
+
           sudo apt-get install -y \
             build-essential \
             cmake \
@@ -338,6 +341,7 @@ jobs:
             libcurl4-openssl-dev \
             libdrm-dev \
             libevdev-dev \
+            libfuse2 \
             libminiupnpc-dev \
             libmfx-dev \
             libnotify-dev \
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -280,7 +280,7 @@ jobs:
         include:  # package these differently
           - type: AppImage
             EXTRA_ARGS: '-DSUNSHINE_BUILD_APPIMAGE=ON'
-            dist: 20.04
+            dist: 22.04
 
     steps:
       - name: Maximize build space
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -9,6 +9,7 @@
 #include <iostream>
 #include <thread>
 #include <unordered_map>
+#include <utility>
 
 #include <boost/asio.hpp>
 #include <boost/filesystem.hpp>
--- a/src/httpcommon.cpp
+++ b/src/httpcommon.cpp
@@ -7,6 +7,7 @@
 #include "process.h"
 
 #include <filesystem>
+#include <utility>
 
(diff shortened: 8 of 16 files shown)


tez011:
https://stackoverflow.com/questions/62723053/boost-log-expressions-v1-71-will-not-compile-under-c20

tez011:
I guess the linux builds are failing because boost 1.74 and c++20 don't play super well together. (1.74 *is* from 2020).

A glance at package versioning shows that our boost is pretty inconsistent between platforms and linux distributions. Windows uses 1.83 at the time of writing, Mac uses 1.81, Linux uses something between 1.74 and 1.81.

Where boost is versioned, could we bump it to 1.81?

tez011:
AppImage build fails because Ubuntu 20.04 provides only Boost 1.71, which is incompatible with C++20. https://stackoverflow.com/questions/62723053/boost-log-expressions-v1-71-will-not-compile-under-c20

There is a more recent LTS version. Could we drop 20.04? Even Debian 11 provides 1.74.

tez011:
What wonderful timing!
CI passes on my fork.

tez011:
depends on #2327

ReenigneArcher:
> depends on #2327

Merged.

src/platform/linux/input.cpp
```diff
@@ -1510,7 +1510,7 @@ namespace platf {
     std::stringstream ss;
     ss << std::hex << std::setfill('0');
     for (const auto &ch : str) {
-      ss << ch;
+      ss << static_cast<uint_least32_t>(ch);
```
tez011, line 1513: >char32_t — type for UTF-32 character representation, required to be large enough to represent any UTF-32 code unit (32 bits). It has the same size, signedness, and alignment as [std::uint_least32_t](https://en.cppreference.com/w/cpp/types/integer), but is a distinct type.
https://en.cppreference.com/w/cpp/language/types#char32_t

Therefore we should expect no data loss with a static_cast to uint_least32_t

src/platform/linux/kmsgrab.cpp
```diff
@@ -614,7 +614,7 @@ namespace platf {
         for (auto &entry : fs::directory_iterator { card_dir }) {
           auto file = entry.path().filename();
 
-          auto filestring = file.generic_u8string();
+          auto filestring = file.generic_string();
```
tez011, line 617: generic_u8string now returns `basic_string<char8_t>`. since char8_t is unsigned, this is now incompatible with `std::string_view = std::basic_string_view<char>`. Assuming this goes in with the UCRT64 change our locale is not exactly guaranteed to be utf-8 either, and the comparison we're immediately making is to an ANSI string so the u8-ness seems mostly irrelevant here. Therefore, making this change so compilation succeeds.

tez011:
Is there anything else in this pull request I should handle before it's able to be merged?

tez011:
Hi! Just waiting for testing/approval results for this pull request. This first change is fairly small so I'm hoping there aren't too many obstacles!

ReenigneArcher:
@tez011 sorry for the delays. We have a lot of open PRs. It might take a little bit before the other approvals come in. I think we are going to try to put out another patch to get a few things fixed before merging these bigger ones.

Can you rebase your branch on top of the latest nightly?

cmake/targets/common.cmake
```diff
@@ -36,7 +36,7 @@ endif()
 
 target_link_libraries(sunshine ${SUNSHINE_EXTERNAL_LIBRARIES} ${EXTRA_LIBS})
 target_compile_definitions(sunshine PUBLIC ${SUNSHINE_DEFINITIONS})
-set_target_properties(sunshine PROPERTIES CXX_STANDARD 17
+set_target_properties(sunshine PROPERTIES CXX_STANDARD 20
         VERSION ${PROJECT_VERSION}
         SOVERSION ${PROJECT_VERSION_MAJOR})
```
FrogTheFrog, line 42: Just below we have:
```
target_compile_options(sunshine PRIVATE $<$<COMPILE_LANGUAGE:CXX>:${SUNSHINE_COMPILE_OPTIONS}>;$<$<COMPILE_LANGUAGE:CUDA>:${SUNSHINE_COMPILE_OPTIONS_CUDA};-std=c++17>)
```
This either need to be changed to c++20 or removed? @ReenigneArcher do you know why we need to specify the standard explicitly again?

^ This includes here and in other files.

src/network.cpp
```diff
@@ -169,7 +170,9 @@ namespace net {
   addr_to_url_escaped_string(boost::asio::ip::address address) {
     address = normalize_address(address);
     if (address.is_v6()) {
-      return "["s + address.to_string() + ']';
+      std::stringstream ss;
```
FrogTheFrog, line 173: This change to me seems unnecessary? Why was this needed? Some reallocations might be avoided, but is it worth it?

src/platform/linux/input.cpp
FrogTheFrog, line 1513: Why are we casting? The `char32_t` is already unsigned.

src/platform/linux/kmsgrab.cpp
FrogTheFrog: Good thing is that we don't use `file.generic_string();` on Windows. Yet...

src/stream.h
```diff
@@ -3,6 +3,7 @@
  * @brief todo
  */
 #pragma once
+#include <utility>
```
FrogTheFrog, line 6: There are a lot of includes for this. Are they really needed? Especially here, in the header.

cmake/targets/common.cmake
tez011: I settled on this change after trying both options you describe. The cuda compiler doesn't support c++20.

src/stream.h
tez011: Yes. Compilation fails otherwise.

src/platform/linux/input.cpp
tez011: stringstreams don't approve of mixing type `char` and type `char32_t`. (You can try this.) This cast is done to force the stringstream to recognize that this value is a 32-bit integer and to hex it accordingly.

tez011:
```
/home/runner/work/Sunshine/Sunshine/src/platform/linux/input.cpp: In function ‘std::string platf::to_hex(const std::__cxx11::basic_string<char32_t>&)’:
/home/runner/work/Sunshine/Sunshine/src/platform/linux/input.cpp:1513:13: error: use of deleted function ‘std::basic_ostream<char, _Traits>& std::operator<<(std::basic_ostream<char, _Traits>&, char32_t) [with _Traits = std::char_traits<char>]’
 1513 |       ss << ch;
      |             ^~
In file included from /usr/include/c++/10/bits/unique_ptr.h:42,
                 from /usr/include/c++/10/bits/locale_conv.h:41,
                 from /usr/include/c++/10/locale:43,
                 from /usr/include/boost/locale/boundary/facets.hpp:17,
                 from /usr/include/boost/locale/boundary.hpp:12,
                 from /usr/include/boost/locale.hpp:11,
                 from /home/runner/work/Sunshine/Sunshine/src/platform/linux/input.cpp:21:
/usr/include/c++/10/ostream:553:5: note: declared here
  553 |     operator<<(basic_ostream<char, _Traits>&, char32_t) = delete;
      |     ^~~~~~~~
make[2]: *** [tests/CMakeFiles/test_sunshine.dir/build.make:665: tests/CMakeFiles/test_sunshine.dir/__/src/platform/linux/input.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:271: tests/CMakeFiles/test_sunshine.dir/all] Error 2
```

src/network.cpp
tez011: I seem to recall that under C++20, the return type of `boost::asio::ip::address::to_string()` changes which makes "summing" these illegal using the current syntax. I can't find the compilation error in my CI history though.

I might have been able to fix it by adding parentheses, but we use stringstreams elsewhere in sunshine (see another patch in this change), so I figured this would be fine.

src/stream.h
tez011: It seems they are needed before any boost includes, otherwise a lot of errors like this are observed:
https://github.com/tez011/Sunshine/actions/runs/8484576885/job/23247793427

If we can include a boost header in here, we can certainly include a system header above it, no?

FrogTheFrog: No problems with that, it's just that currently include state in Sunshine is a mess. For example, in this file even include `asio`, but it not used anywhere...

But that's out of scope for this PR.

FrogTheFrog:
@ReenigneArcher LGTM

cmake/targets/common.cmake
ReenigneArcher: I'm looking to bump to c++23 in https://github.com/LizardByte/Sunshine/pull/4096 and came here just to see what all was changed when bumping to 20. It looks like the bump to 23 will be quite a bit easier than the one to 20, as everything is just building right away.

It seems that with cuda we can specify the standard directly in the file name, though I'm not sure that's a good idea? https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#std-c-03-c-11-c-14-c-17-c-20-std

It also seems that at least recent versions of cuda now support c++20. Perhaps this can be set dynamically.

PR #2323, opened 2024-03-29
build(win): use UCRT64 environment instead of MinGW64.

tez011:
## Description
https://github.com/LizardByte/Sunshine/pull/2149 was reverted in https://github.com/LizardByte/Sunshine/pull/2320 and needs to be split into three changes:

  -  C++20 
   - UCRT64 (this one!)
  -  WGC

If https://github.com/msys2/MINGW-packages/pull/20477 is integrated, WGC can proceed without this change.

## Type of Change
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) 🤪

## Checklist

## Branch Updates

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -91,8 +91,7 @@ main(int argc, char *argv[]) {
   task_pool_util::TaskPool::task_id_t force_shutdown = nullptr;
 
 #ifdef _WIN32
-  // Switch default C standard library locale to UTF-8 on Windows 10 1803+
-  setlocale(LC_ALL, ".UTF-8");
+  setlocale(LC_ALL, "C");
 #endif
 
 #pragma GCC diagnostic push
--- a/src_assets/windows/misc/migration/migrate-config.bat
+++ b/src_assets/windows/misc/migration/migrate-config.bat
@@ -38,10 +38,11 @@ if exist "%OLD_DIR%\credentials\" (
 rem Create the credentials directory if it wasn't migrated or already existing
 if not exist "%NEW_DIR%\credentials\" mkdir "%NEW_DIR%\credentials"
 
-rem Disallow read access to the credentials directory for normal users
-rem Note: We must use the SID directly because "Administrators" is localized
+rem Disallow read access to the credentials directory contents for normal users
+rem Note: We must use the SIDs directly because "Users" and "Administrators" are localized
 icacls "%NEW_DIR%\credentials" /inheritance:r
 icacls "%NEW_DIR%\credentials" /grant:r *S-1-5-32-544:(OI)(CI)(F)
+icacls "%NEW_DIR%\credentials" /grant:r *S-1-5-32-545:(R)
 
 rem Migrate the covers directory
 if exist "%OLD_DIR%\covers\" (
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -935,29 +935,23 @@ jobs:
       - name: Setup Dependencies Windows
         uses: msys2/setup-msys2@v2
         with:
+          msystem: ucrt64
           update: true
           install: >-
-            base-devel
-            diffutils
             doxygen
             git
-            make
-            mingw-w64-x86_64-binutils
-            mingw-w64-x86_64-boost
-            mingw-w64-x86_64-cmake
-            mingw-w64-x86_64-curl
-            mingw-w64-x86_64-graphviz
-            mingw-w64-x86_64-miniupnpc
-            mingw-w64-x86_64-nlohmann-json
-            mingw-w64-x86_64-nodejs
-            mingw-w64-x86_64-nsis
-            mingw-w64-x86_64-onevpl
-            mingw-w64-x86_64-openssl
-            mingw-w64-x86_64-opus
-            mingw-w64-x86_64-toolchain
-            nasm
-            wget
-            yasm
+            mingw-w64-ucrt-x86_64-boost
+            mingw-w64-ucrt-x86_64-cmake
+            mingw-w64-ucrt-x86_64-curl
+            mingw-w64-ucrt-x86_64-graphviz
+            mingw-w64-ucrt-x86_64-miniupnpc
+            mingw-w64-ucrt-x86_64-nlohmann-json
+            mingw-w64-ucrt-x86_64-nodejs
+            mingw-w64-ucrt-x86_64-nsis
+            mingw-w64-ucrt-x86_64-onevpl
+            mingw-w64-ucrt-x86_64-openssl
+            mingw-w64-ucrt-x86_64-opus
+            mingw-w64-ucrt-x86_64-toolchain
 
       - name: Setup python
         # use this instead of msys2 python due to known issues using wheels, https://www.msys2.org/docs/python/
--- a/docs/source/building/windows.rst
+++ b/docs/source/building/windows.rst
@@ -3,7 +3,7 @@ Windows
 
 Requirements
 ------------
-First you need to install `MSYS2 <https://www.msys2.org>`__, then startup "MSYS2 MinGW 64-bit" and execute the following
+First you need to install `MSYS2 <https://www.msys2.org>`__, then startup "MSYS2 UCRT64" and execute the following
 codes.
 
 Update all packages:
@@ -15,26 +15,21 @@ Install dependencies:
    .. code-block:: bash
 
       pacman -S \
-        base-devel \
-        cmake \
-        diffutils \
         doxygen \
-        gcc \
         git \
-        make \
-        mingw-w64-x86_64-binutils \
-        mingw-w64-x86_64-boost \
-        mingw-w64-x86_64-cmake \
-        mingw-w64-x86_64-curl \
-        mingw-w64-x86_64-graphviz \
-        mingw-w64-x86_64-miniupnpc \
-        mingw-w64-x86_64-nlohmann-json \
-        mingw-w64-x86_64-nodejs \
-        mingw-w64-x86_64-onevpl \
-        mingw-w64-x86_64-openssl \
-        mingw-w64-x86_64-opus \
-        mingw-w64-x86_64-rust \
-        mingw-w64-x86_64-toolchain \
+        mingw-w64-ucrt-x86_64-boost \
+        mingw-w64-ucrt-x86_64-cmake \
+        mingw-w64-ucrt-x86_64-curl \
+        mingw-w64-ucrt-x86_64-graphviz \
+        mingw-w64-ucrt-x86_64-miniupnpc \
+        mingw-w64-ucrt-x86_64-nlohmann-json \
+        mingw-w64-ucrt-x86_64-nodejs \
+        mingw-w64-ucrt-x86_64-nsis \
+        mingw-w64-ucrt-x86_64-onevpl \
+        mingw-w64-ucrt-x86_64-openssl \
+        mingw-w64-ucrt-x86_64-opus \
+        mingw-w64-ucrt-x86_64-rust \
+        mingw-w64-ucrt-x86_64-toolchain \
         python \
         python-pip
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1272,10 +1272,16 @@ namespace config {
       BOOST_LOG(fatal) << "Failed to apply config: "sv << err.what();
     }
 
-    if (!config_loaded) {
 #ifdef _WIN32
+    // UCRT64 raises an access denied exception if launching from the shortcut
+    // as non-admin and the config folder is not yet present; we can defer
+    // so that service instance will do the work instead.
+
+    if (!config_loaded && !shortcut_launch) {
       BOOST_LOG(fatal) << "To relaunch Sunshine successfully, use the shortcut in the Start Menu. Do not run Sunshine.exe manually."sv;
       std::this_thread::sleep_for(10s);
+#else
+    if (!config_loaded) {
 #endif
       return -1;
     }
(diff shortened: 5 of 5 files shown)


psyke83:
This resolves the regex CPU/memory hog issue:

```
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -103,7 +103,7 @@ main(int argc, char *argv[]) {

 #ifdef _WIN32
   // Switch default C standard library locale to UTF-8 on Windows 10 1803+
-  setlocale(LC_ALL, ".UTF-8");
+  setlocale(LC_ALL, ".UTF-16");
 #endif

 #pragma GCC diagnostic push
```
Setting to UTF16 or UTF32 resolves the slowdown, but I'm not sure if that actually breaks our current regex parsing?

psyke83:
Relevant discussion (see also in comments): https://stackoverflow.com/questions/59067280/does-stdwregex-support-utf-16-unicode-or-only-ucs-2

Setting to UTF16 doesn't break my en-IE keyboard mappings, but it could have consequences for other languages?

psyke83:
I retested the #1477  by printing some Chinese characters with `.UTF-16" [set](https://github.com/LizardByte/Sunshine/blob/2af0ce364d6562cee9fa6215eefc36c031cb1f84/src/main.cpp#L106), but didn't change the [UTF-8 conversion](https://github.com/LizardByte/Sunshine/blob/2af0ce364d6562cee9fa6215eefc36c031cb1f84/src/main.cpp#L112). 

The result: Sunshine doesn't crash, regex doesn't consume CPU or memory, and the Chinese characters print in the web UI log correctly. This may be a viable fix, but @cgutman will probably want to review that change to make sure no other changes are required.

psyke83:
Apologies for the spam, but my observations were flawed. It appears that MINGW64 builds were not setting the UTF-8 codepage correctly, so the default `C` codepage was in use. UCRT64 builds *do* set a UTF-8 codepage, but cause the `std::regex` issue. Leaving the codepage to `C` (*not* UTF-16, as that was falling back to `C` on UCRT64 as well) seems to be the correct solution.

More context: https://github.com/msys2/MINGW-packages/pull/20477#issuecomment-2027821426

psyke83:
Two points on MSVCRT in the MSYS2 environments [documentation](https://www.msys2.org/docs/environments/):

> - It doesn't support the UTF-8 locale

This explains the behaviour we're seeing. I didn't mention in my prior comment, but setting the system default locale via  `std::setlocale(LC_ALL, "");` does apply correctly on both MINGW64 and UCRT64. For my system, it sets `English_Ireland.1252` for both runtimes, and regex doesn't exhibit any issues with either runtime.

> - Binaries linked with MSVCRT should not be mixed with UCRT ones because the internal structures and data types are different. (More strictly, object files or static libraries built for different targets shouldn't be mixed. DLLs built for different CRTs can be mixed as long as they don't share CRT objects, e.g. FILE*, across DLL boundaries.) Same rule is applied for MSVC compiled binaries because MSVC uses UCRT by default (if not changed).

This suggests to me that we need a complementary PR for [build-deps](https://github.com/LizardByte/build-deps) to also switch to UCRT64, so that the statically linked ffmpeg libraries don't cause unforseen issues.

tez011:
@psyke83, this is some impressive detective work, thank you so much! 

There's a block at the top of main that goes,

```
#ifdef _WIN32
  // Switch default C standard library locale to UTF-8 on Windows 10 1803+
  setlocale(LC_ALL, ".UTF-8");
#endif

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
  // Use UTF-8 conversion for the default C++ locale (used by boost::log)
  std::locale::global(std::locale(std::locale(), new std::codecvt_utf8<wchar_t>));
#pragma GCC diagnostic pop
```

I suppose the first declaration should be changed to remove UTF-8. Should the second be removed, too, do you think?

psyke83:
Regarding the UAC failing to trigger via shortcut launch: `std::filesystem::exists` returns false due to the restrictive permissions of the `credentials` folder. This may be a bug in UCRT64 or perhaps it is the more correct behaviour; I'm not sure.

A workaround (or perhaps the only solution) is the following:
```
diff --git a/src_assets/windows/misc/migration/migrate-config.bat b/src_assets/windows/misc/migration/migrate-config.bat
index 65d00d8f..dfce18ae 100644
--- a/src_assets/windows/misc/migration/migrate-config.bat
+++ b/src_assets/windows/misc/migration/migrate-config.bat
@@ -42,6 +42,7 @@ rem Disallow read access to the credentials directory for normal users
 rem Note: We must use the SID directly because "Administrators" is localized
 icacls "%NEW_DIR%\credentials" /inheritance:r
 icacls "%NEW_DIR%\credentials" /grant:r *S-1-5-32-544:(OI)(CI)(F)
+icacls "%NEW_DIR%\credentials" /grant:r Users:(R)

 rem Migrate the covers directory
```

This will grant the `Users` group read access for the folder (but *not* its contents). Technically it may reduce security because users can now list folder contents, but they still can't read or make any changes to the files contained within.

FrogTheFrog:
Whatever you end up doing, do not use `std::setlocale(LC_ALL, "");`. On Windows you can actually set global locale to be `UTF-8` and calling `std::setlocale(LC_ALL, "");` will then set it to `UTF-8` instead of the default `C` and we're back at the square 1.

tez011:
oof, I'm embarrassed by that. Should have read `setlocale` docs before blindly setting it to what I thought was the 'default' locale. Thanks.

src/main.cpp
```diff
@@ -102,8 +102,7 @@ main(int argc, char *argv[]) {
   task_pool_util::TaskPool::task_id_t force_shutdown = nullptr;
 
 #ifdef _WIN32
-  // Switch default C standard library locale to UTF-8 on Windows 10 1803+
-  setlocale(LC_ALL, ".UTF-8");
+  setlocale(LC_ALL, "C");
```
ReenigneArcher, line 105: Just linking the original PR for this code. Have we verified we no longer hit the issues linked to https://github.com/LizardByte/Sunshine/pull/1517 ?

psyke83: Current PR branch with Chinese characters printing to log is OK:
```diff
diff --git a/src/main.cpp b/src/main.cpp
index 7aaafbdd..07b6beda 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -200,6 +200,7 @@ main(int argc, char *argv[]) {
   // if anything is logged prior to this point, it will appear in stdout, but not in the log viewer in the UI
   // the version should be printed to the log before anything else
   BOOST_LOG(info) << PROJECT_NAME << " version: " << PROJECT_VER;
+  BOOST_LOG(info) << L"你好";

   if (!config::sunshine.cmd.name.empty()) {
     auto fn = cmd_to_func.find(config::sunshine.cmd.name);

```

PR with Chinese characters and Cameron's relevant change reverted crashes:
```diff
diff --git a/src/main.cpp b/src/main.cpp
index 7aaafbdd..bb6dbf68 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -108,7 +108,7 @@ main(int argc, char *argv[]) {
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
   // Use UTF-8 conversion for the default C++ locale (used by boost::log)
-  std::locale::global(std::locale(std::locale(), new std::codecvt_utf8<wchar_t>));
+  //std::locale::global(std::locale(std::locale(), new std::codecvt_utf8<wchar_t>));
 #pragma GCC diagnostic pop

   mail::man = std::make_shared<safe::mail_raw_t>();
@@ -200,6 +200,7 @@ main(int argc, char *argv[]) {
   // if anything is logged prior to this point, it will appear in stdout, but not in the log viewer in the UI
   // the version should be printed to the log before anything else
   BOOST_LOG(info) << PROJECT_NAME << " version: " << PROJECT_VER;
+  BOOST_LOG(info) << L"你好";

   if (!config::sunshine.cmd.name.empty()) {
     auto fn = cmd_to_func.find(config::sunshine.cmd.name);

```

Conclusion:
* The PR as-is will not break issues fixed by #1517
* #1517 never needed `setlocale` to be set to UTF-8, as it [wasn't](https://github.com/msys2/MINGW-packages/pull/20477#issuecomment-2027821426) [applying ](https://github.com/LizardByte/Sunshine/pull/2323#issuecomment-2027862071)on our MING64 builds.

FrogTheFrog:
https://github.com/msys2/MINGW-packages/pull/20477 has been merged. Maybe we don't need to migrate to ucrt anymore?

ReenigneArcher:
It seems that mingw is becoming legacy, so probably worth migrating anyway. The WGC just won't depend on this now.

ReenigneArcher:
> Regarding the UAC failing to trigger via shortcut launch: `std::filesystem::exists` returns false due to the restrictive permissions of the `credentials` folder. This may be a bug in UCRT64 or perhaps it is the more correct behaviour; I'm not sure.
> 
> A workaround (or perhaps the only solution) is the following:
> 
> ```diff
> diff --git a/src_assets/windows/misc/migration/migrate-config.bat b/src_assets/windows/misc/migration/migrate-config.bat
> index 65d00d8f..aa8d10aa 100644
> --- a/src_assets/windows/misc/migration/migrate-config.bat
> +++ b/src_assets/windows/misc/migration/migrate-config.bat
> @@ -39,9 +39,10 @@ rem Create the credentials directory if it wasn't migrated or already existing
>  if not exist "%NEW_DIR%\credentials\" mkdir "%NEW_DIR%\credentials"
> 
>  rem Disallow read access to the credentials directory for normal users
> -rem Note: We must use the SID directly because "Administrators" is localized
> +rem Note: We must use the SIDs directly because "Users" and "Administrators" are localized
>  icacls "%NEW_DIR%\credentials" /inheritance:r
>  icacls "%NEW_DIR%\credentials" /grant:r *S-1-5-32-544:(OI)(CI)(F)
> +icacls "%NEW_DIR%\credentials" /grant:r *S-1-5-32-545:(R)
> 
>  rem Migrate the covers directory
>  if exist "%OLD_DIR%\covers\" (
> ```
> 
> This will grant the `Users` aka `(S-1-5-32-545)` group read access for the folder (but _not_ its contents). Technically it may reduce security because users can now list folder contents, but they still can't read or make any changes to the files contained within.
> 
> Edit: amended to use the SID to avoid localization issues.

@tez011 cgutman mentioned he was fine with this approach on our Discord server, if you'd like to implement the change.

ReenigneArcher:
@tez011 did you see @psyke83 's PR into your branch? https://github.com/tez011/Sunshine/pull/1

tez011:
Thanks for the ping and for patience. I didn't get any email notifications about that pull request until yours...

I guess if the service launches sunshine as administrator, it'll have access to the configuration files specified and the patch will work good. I've merged that change. Thanks for the patch @psyke83 !

psyke83:
> Thanks for the ping and for patience. I didn't get any email notifications about that pull request until yours... though I did see it yesterday and thought it was closed.
> 
> I guess if the service launches sunshine as administrator, it'll have access to the configuration files specified and the patch will work good. I've merged that change. Thanks for the patch @psyke83 !

No problem, but it seems that my commit was lost after you rebased your `ucrt64` branch against nightly?

src_assets/windows/misc/migration/migrate-config.bat
```diff
@@ -38,10 +38,11 @@ if exist "%OLD_DIR%\credentials\" (
 rem Create the credentials directory if it wasn't migrated or already existing
 if not exist "%NEW_DIR%\credentials\" mkdir "%NEW_DIR%\credentials"
 
-rem Disallow read access to the credentials directory for normal users
-rem Note: We must use the SID directly because "Administrators" is localized
+rem Disallow read access to the credentials directory contents for normal users
+rem Note: We must use the SIDs directly because "Users" and "Administrators" are localized
 icacls "%NEW_DIR%\credentials" /inheritance:r
 icacls "%NEW_DIR%\credentials" /grant:r *S-1-5-32-544:(OI)(CI)(F)
+icacls "%NEW_DIR%\credentials" /grant:r *S-1-5-32-545:(R)
```
cgutman, line 45: Is this still needed with the change to config.cpp?

psyke83: I think so?

Test 1:
* Delete existing config directory
* Freshly install Sunshine built from current PR branch
* Set `port = 46989`
* Stop sunshine service and relaunch from shortcut as non-admin

Result: Sunshine launches with the correct custom web UI port.

Test 2:
* Repeat test 1 with build that omits `migrate-config.bat` OR keep PR build+config from test 1 and remove Users read folder privilege manually.

Result: several second delay on "Waiting for Web UI to be ready....", and UI opens to the (incorrect) default port.

tez011:
Now that C++20 is in, this can be considered.

What's the appetite for merging this environment change? If it's slim-to-none, I can try porting the WGC change to mingw64 now that https://github.com/msys2/MINGW-packages/pull/20477 is in. I have a suspicion that MSVCRT and the UWP runtime won't play together well, so hoping to get some confirmation here before I spend time in that rabbit hole.

ReenigneArcher:
> What's the appetite for merging this environment change?

We'd like to move to UCRT64 since mingw seems to be considered legacy by the packaging team.

ReenigneArcher:
Yes, I am trying to knock out as many of these open PRs as possible. Sorry for the delay.

ReenigneArcher:
PR updating ffmpeg to UCRT64: https://github.com/LizardByte/Sunshine/pull/2572

tez011:
Thanks for merging this!

I'll let this change stew for a bit while I ready the final WGC pull request. I hope to send that out soon!

PR #2325, opened 2024-03-29
fix(i18n): update localization docs and js strings

ReenigneArcher:
## Description

- Improve guidance for contributors around localization.
- Translate 2 js strings.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/docs/source/contributing/localization.rst
+++ b/docs/source/contributing/localization.rst
@@ -48,6 +48,11 @@ Extraction
       .. note:: The json keys should be sorted alphabetically. You can use `jsonabc <https://novicelab.org/jsonabc/>`__
          to sort the keys.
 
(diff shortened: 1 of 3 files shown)


ReenigneArcher:
@psyke83 do you think the comment I added to the docs is sufficient, or should I add more details?

psyke83:
> @psyke83 do you think the comment I added to the docs is sufficient, or should I add more details?

There may be some confusion for certain scenarios, but it's plausible for this scenario to happen if the contributor is specifically looking at the localization documentation for reference.

Consider if a someone sends a PR for a feature that requires new strings, but wants to include their native language as well as English; will the guideline to only modify `en.json` still apply? If so, it might be worth explaining that it's necessary to wait for the PR to be merged with English strings only and then add their translation via https://translate.lizardbyte.dev/ after merge (if my understanding of the correct process is accurate).

PR #2358, opened 2024-04-03
Add the ability to force an IDR from the troubleshooting menu.

20kdc:
## Description

Adds the ability to force an IDR from the troubleshooting menu.
    
Sometimes, devices are simply receiving video, and are not in locations which are amiable to retrieval.
    
For example, a device mounted on one's head, even if within an openable shell, is awkward to retrieve.
    
In these situations, higher resolutions and higher bitrates are often used, which leaves little room for any error correction or recovery; and the current mechanisms used for these can fail anyway. (Usual symptoms are the entire stream turning green, or inconsistent, blocky brightness issues.)
    
In this case, it's required that there is an easy way to manually recover the stream.
    
While using the web UI is theoretically suboptimal for this, the exposure of a way to do this over HTTP allows hooking up whatever controls a more technical user feels are necessary. (A mechanism to automatically send IDRs at a fixed rate was considered, but ultimately I decided this would likely be rejected due to the continuous hitching this would produce in the output for anyone who actually used the feature.)
    
It is also possible that in some situations, a user may simply not wish to interrupt the flow of a stream; for example, when using Sunshine/Moonlight as part of a home theatre setup.
    
In this case, they can press the button on the web UI and playback can continue more-or-less uninterrupted.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -40,6 +40,7 @@
 #include "utility.h"
 #include "uuid.h"
 #include "version.h"
+#include "stream.h"
 
 using namespace std::literals;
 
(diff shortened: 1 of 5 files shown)


20kdc:
I'd need a legally standing commitment that LizardByte won't take the project proprietary or pay-for-builds to commit to signing a CLA.

ReenigneArcher:
The CLA is the agreement. You can either agree, or not contribute.

PR #2385, opened 2024-04-07
fix(linux): Arch requires library at runtime

pyamsoft:
## Description
On my Archlinux system, the package builds correctly, but without including the libappindicator-gtk3 dependency, it fails to launch at runtime with the following error message:

```/usr/bin/sunshine: error while loading shared libraries: libappindicator3.so.1: cannot open shared object file: No such file or directory```

This patch adds the required library as a dependency for runtime.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/packaging/linux/Arch/PKGBUILD
+++ b/packaging/linux/Arch/PKGBUILD
@@ -13,6 +13,7 @@ install=sunshine.install
 depends=('avahi'
          'boost-libs'
          'curl'
+         'libappindicator-gtk3'
          'libayatana-appindicator'
          'libcap'
          'libdrm'

### CODE REVIEW

**Review by psyke83 — APPROVED** [review]
This library is probably installed Arch users' systems if using GNOME and Network Manager, so it's probably an oversight.

From v0.23.0 release binary:
```
$ ldd /usr/bin/sunshine | grep app
        libayatana-appindicator3.so.1 => /usr/lib/libayatana-appindicator3.so.1 (0x000079cdb7c2e000)
$ sudo pacman -R libappindicator-gtk3
checking dependencies...
error: failed to prepare transaction (could not satisfy dependencies)
:: removing libappindicator-gtk3 breaks dependency 'libappindicator-gtk3' required by network-manager-applet
```

pyamsoft:
Hi, thanks for the approval!

As a fun aside:
Ironically, I am an Arch Gnome user. The appindicator library is not installed as a requirement unless you have network-manager-applet, which is only a required dep of gnome-flashback session. The reason I encountered this in the first place was because I do not have the gnome flashback package installed on my machine (or any other of the various packages that would otherwise require it)

ReenigneArcher:
@psyke83 Sunshine can use either appindicator version, 'libayatana-appindicator' or 'libappindicator-gtk3'...

I guess what happened is they may have had the gtk3 version installed when installing sunshine.

https://github.com/LizardByte/Sunshine/blob/7e26d2fd3064c6adf3e7a52a275046dd0cf194be/cmake/compile_definitions/linux.cmake#L202-L210

I'm thinking we should just reverse the order when checking for this. Ayatana version should be primary.

pyamsoft:
Sounds like this PR should be closed in favor of updating the cmake file then?

psyke83:
> @psyke83 Sunshine can use either appindicator version, 'libayatana-appindicator' or 'libappindicator-gtk3'...
> 
> I guess what happened is they may have had the gtk3 version installed when installing sunshine.
> 
> https://github.com/LizardByte/Sunshine/blob/7e26d2fd3064c6adf3e7a52a275046dd0cf194be/cmake/compile_definitions/linux.cmake#L202-L210
> 
> I'm thinking we should just reverse the order when checking for this. Ayatana version should be primary.

I think so? The PKGBUILD will install the ayatana library first, so it can't really conflict if the legacy library is also installed. I supposed we could also remove the ayatana library from the PKGBUILD and instead inject the library selected via cmake via an environment variable instead, but that might be too much hassle.

ReenigneArcher:
> Sounds like this PR should be closed in favor of updating the cmake file then?

Yes

pyamsoft:
I will close unless you would like to continue discussion here - or please close from your end when done. Thanks!

PR #2461, opened 2024-04-22
fix(macos/input): incorrect mouse input for non-main display

TimmyOVO:
## Description

fix: abs mouse input not working in macOS
fix: wrong abs mouse coordinates when capturing non-main display in macOS
fix: wrong mouse coordinates correction when capturing non-main display in macOS

## Type of Change

## Checklist

## Branch Updates


ReenigneArcher:
I don't have the ability to test these changes. Ideally unit tests would be added around the changed code, but I understand that's not always the easiest thing to do.

Can you explain how you tested your changes?

src/platform/macos/display.mm
TimmyOVO, line 1: In this [commit](https://github.com/LizardByte/Sunshine/commit/a4d9ee3fa4630bc0e7fdff230e923829c1d06015#diff-1518590f5c5028e6a34dd695918cfc679e97e1452dc0dc814718f250d93795deR481), there is a change that would require both env_width and env_height to have non-zero values, but these values are not properly set in that commit, which could break absolute mouse input on macOS.

src/platform/macos/input.cpp
TimmyOVO, line 1: Before we support change the display we need to capture in macOS, the location we get here according to [Apple Document](https://developer.apple.com/documentation/coregraphics/1455788-cgeventgetlocation?language=objc) should be `The current location of the specified mouse event in global display coordinates`, we can not just simply assuming the display's bounds  to `[0,display_width]` and `[0,display_height]`. So we need to get the correspond display bounds for display we are capturing using [`CGDisplayBounds`](https://developer.apple.com/documentation/coregraphics/1456395-cgdisplaybounds?language=objc) which return `The bounds of the display, expressed as a rectangle in the global display coordinate space (relative to the upper-left corner of the main display).`.

src/platform/macos/input.cpp
```diff
@@ -379,10 +383,15 @@ const KeyCodeMap kKeyCodesMap[] = {
...
   abs_mouse(input_t &input, const touch_port_t &touch_port, float x, float y) {
-    auto scaling = ((macos_input_t *) input.get())->displayScaling;
+    auto macos_input = static_cast<macos_input_t *>(input.get());
+    auto scaling = macos_input->displayScaling;
+    auto display = macos_input->display;
 
     CGPoint location = CGPointMake(x * scaling, y * scaling);
-
+    CGRect display_bounds = CGDisplayBounds(display);
+    // in order to get the correct mouse location for capturing display , we need to add the display bounds to the location
```
TimmyOVO, line 392: The location we received are relative to the display we are capturing ,so we get the display bounds using [CGDisplayBounds](https://developer.apple.com/documentation/coregraphics/1456395-cgdisplaybounds?language=objc) then add them up to get the right input coordinates.

TimmyOVO (commented):
That's all i done here in order to get correct input coordinates after we support specific display capture.

src/platform/macos/input.cpp
```diff
@@ -328,15 +328,19 @@ const KeyCodeMap kKeyCodesMap[] = {
...
 
-    if (location.y < 0)
-      location.y = 0;
-    if (location.y >= (double) CGDisplayPixelsHigh(display))
-      location.y = (double) CGDisplayPixelsHigh(display) - 1;
+    // limit mouse to current display bounds
+    if (location.x < display_bounds.origin.x)
+      location.x = display_bounds.origin.x;
+    if (location.x >= display_bounds.origin.x + display_bounds.size.width)
+      location.x = display_bounds.origin.x + display_bounds.size.width - 1;
```
cgutman, line 338: You should use `std::clamp()` instead.

Example:
```
location.x = std::clamp(location.x,  display_bounds.origin.x, display_bounds.origin.x + display_bounds.size.width - 1);
```

BayLee4:
Hi, just wanted to report that I tested the changes (by forking the homebrew tap [repository](https://github.com/LizardByte/homebrew-homebrew) and then pointing it to the right branch/commit) and confirm the issue is now gone, everything[1] works perfectly 👍

[1]: Well except the CTRL + ALT + SHIFT + N shortcut which still doesn't hide the cursor in anything but the main display, but that behavior is already present upstream so that would be a different issue

PR #2471, opened 2024-04-26
Remove the need for a patched nvidia library for NvFBC.

hgaiser:
## Description
This PR is based on https://github.com/hgaiser/nvfbc-rs/pull/5 , which in turn is based on https://github.com/keylase/nvidia-patch/blob/master/win/nvfbcwrp/nvfbcwrp_main.cpp#L23-L25 .

With this change you don't need a patched NVIDIA library to use NvFBC. My best guess is that this is what GeForce Experience uses to allow it to work on GTX / RTX cards.

### Screenshot
N/A

### Issues Fixed or Closed
N/A

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/cuda.cpp
+++ b/src/platform/linux/cuda.cpp
@@ -616,7 +616,7 @@ namespace cuda {
 
         // Set privateData to allow NvFBC on consumer NVIDIA GPUs.
         // Based on https://github.com/keylase/nvidia-patch/blob/3193b4b1cea91527bf09ea9b8db5aade6a3f3c0a/win/nvfbcwrp/nvfbcwrp_main.cpp#L23-L25 .
-        const unsigned int MAGIC_PRIVATE_DATA[4] = {0xAEF57AC5, 0x401D1A39, 0x1B856BBE, 0x9ED0CEBA};
+        const unsigned int MAGIC_PRIVATE_DATA[4] = { 0xAEF57AC5, 0x401D1A39, 0x1B856BBE, 0x9ED0CEBA };
         params.privateData = MAGIC_PRIVATE_DATA;
         params.privateDataSize = sizeof(MAGIC_PRIVATE_DATA);
 
--- a/src/platform/linux/cuda.cpp
+++ b/src/platform/linux/cuda.cpp
@@ -614,6 +614,12 @@ namespace cuda {
       make() {
         NVFBC_CREATE_HANDLE_PARAMS params { NVFBC_CREATE_HANDLE_PARAMS_VER };
 
+        // Set privateData to allow NvFBC on consumer NVIDIA GPUs.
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Thank you for the PR! I also didn't know we could handle it like this.

If this affects NVENC as well, I think there might need to be an additional change somewhere in https://github.com/LizardByte/Sunshine/blob/nightly/src/nvenc/nvenc_base.cpp

hgaiser:
Yeah so I'm a bit confused by NVENC. It almost seems like enabling this patch also allows NVENC.. I never thought about it but using [Moonshine](https://github.com/hgaiser/moonshine) I never use a patched NVIDIA library and NVENC encoding works fine there.

I feel like this might trigger something in the NVIDIA library that also allows encoding, but I'm not sure. I tried rebooting, but I still see that I can encode, even if I use `nightly` and not this PR.

```
[2024:04:27:00:38:36]: Error: Failed to create session: This hardware does not support NvFBC
[2024:04:27:00:38:37]: Info: Found H.264 encoder: h264_nvenc [nvenc]
[2024:04:27:00:38:37]: Info: Found HEVC encoder: hevc_nvenc [nvenc]
```

It might also mean that if you use NVENC, but not NvFBC, that you still need a patched driver :shrug: . I haven't been able to trigger this case at least.

Can you test this to see how this works for you? Whether you can capture & encode using this PR but can't capture nor encode without this PR?

ReenigneArcher:
There might be some confusion. The `./src/nvenc` code is only used on Windows (as far as I know), since nvfbc is deprecated on Windows. For reference, it was introduced here: https://github.com/LizardByte/Sunshine/pull/1427

Currently I don't have a quick way to test your PR, but I could ask for testers on our Discord server?

hgaiser:
> There might be some confusion. The `./src/nvenc` code is only used on Windows (as far as I know), since nvfbc is deprecated on Windows. For reference, it was introduced here: #1427

Ah I see. I'm curious, why isn't ffmpeg used on Windows? I never looked into it much but I thought it was cross platform?

> Currently I don't have a quick way to test your PR, but I could ask for testers on our Discord server?

Yeah I think that would be nice 👍

ReenigneArcher:
We do use ffmpeg on Windows as well. I'm not up to speed on the nvenc encoder, but I know the other capture methods do use ffmpeg.

ReenigneArcher:
I'm not sure anyone has been able to test this. I do have an extra Quadro M4000 card (I guess it has native NvFBC, but I'm not 100% sure), but I will have to deploy a linux environment to use it in so might take me a little bit.

If anyone else already has the os/hardware combination to test this build it would be greatly appreciated.

snow-nene:
I have tried the latest sunshine, but it seems not work. 
My nvidia card is RTX4090D and the driver version is 575.57.08.

Issue #221, opened 2022-06-22
Sunshine Reads Incorrect Monitor Index


JustPlainGarak:
### Describe the Bug

When running Sunshine, even though my primary display shows as index 0, if I leave the Monitor Number blank or input index number 0, Sunshine always picks my HDMI connected display, which is not my primary. I have to cycle through index numbers and restart the service until I find the correct monitor even though the index number for my primary DisplayPort connected display still shows as monitor 0 in xrandr.

### Expected Behavior

Sunshine should select the monitor at index 0 by default.

### Additional Context

I am running 4 connected displays, 1 HDMI and 3 DisplayPort. The index number for my primary display always shows index 0 even though Sunshine picks the HDMI display, which can be any index number from 1-3 in xrandr.

### Sunshine Host Operating System and Version

Fedora 36 KDE Plasma

### Architecture

x86_64

### Sunshine Version

0.14.0

### GPU Type

AMD

### GPU Model

Radeon RX 6700XT

### GPU Driver/Mesa Version

Mesa 23.0.0-git (using gloriousegroll copr repo)

### Capture Method (Linux Only)

Not sure, whatever the default for a Fedora/Wayland install would be

ReenigneArcher:
What is the output of `xrandr --listmonitors`?

And please paste your config file as well.

JustPlainGarak:
Hey, thanks for the reply! I will include that info for you now. I took some logs as well (one left at default which picks the wrong monitor and one with the "wrong" monitor index selected, 3, outputting to the correct monitor).

`randr --listmonitors
Monitors: 4
 0: +*XWAYLAND0 2560/700x1440/400+2560+1080  XWAYLAND0
 1: +XWAYLAND1 2560/600x1440/340+0+1080  XWAYLAND1
 2: +XWAYLAND2 2560/600x1440/340+5120+1080  XWAYLAND2
 3: +XWAYLAND3 1920/600x1080/340+640+0  XWAYLAND3
`
`cat /usr/local/sunshine/config/sunshine.conf 
sunshine_name = EmpokNor
nv_preset = default
amd_coder = auto
upnp = disabled
gamepad = x360
qp = 22
nv_rc = auto
hevc_mode = 0
dwmflush = enabled
min_log_level = 2
amd_quality = default
origin_web_ui_allowed = lan
vt_software = auto
vt_realtime = enabled
key_rightalt_to_key_win = disabled
origin_pin_allowed = pc
resolutions = [
    352x240,
[...]

PR #2490, opened 2024-05-01
feat(ui): list available displays with `select`

Hazer:
## Description
This is a PoC, and a proposal, for allowing the screen to be selected via a list, instead of having to search for an ID and write it down. Also, this PR doesn't cover it, but we could in the future, further improve it, and allow the client to request it and maybe remotely select the preferred display.

For this PR I took basically the same code from every `display_names()` and stripped it to just the basics, but keeping the check, fail-overs, and the same counting for IDs as the other methods use, so we have a 1:1 mapping. I took a bit more effort in some targets than others, but I focused in the targets I could build and run myself first.

All targets are implemented, but some probably will need some more work done, testers are really welcome.

I don't consider this as a breaking change, because:
 1. The select uses the same value as the text input used, so it won't break even if it had been set in the past;
 2. If for some reason, we fail to get a list, we fallback to the old text input, so if something breaks, the user can still use the old UI.
 
 
 ### Depends on
 - #2491 

#### Built and run so far
- [x] macOS Intel
- [ ] macOS Apple Silicon 
- [ ] x11 
- [ ] Windows
- [ ] kms
- [ ] Cuda
- [ ] Wl

#### macOS
Screenshot 2024-04-30 at 10 13 49
#### Windows
Screenshot 2024-04-30 at 20 03 48
#### Linux AppImage x11
Screenshot 2024-04-30 at 22 51 04

## Type of Change
- [x] Repository update (changes to repository files, e.g. .github/...)

## Checklist

## Branch Updates

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1036,7 +1036,6 @@ jobs:
             mingw-w64-x86_64-curl
             mingw-w64-x86_64-graphviz
             mingw-w64-x86_64-miniupnpc
-            mingw-w64-x86_64-nlohmann-json
             mingw-w64-x86_64-nodejs
             mingw-w64-x86_64-nsis
             mingw-w64-x86_64-onevpl
--- a/.gitmodules
+++ b/.gitmodules
@@ -54,3 +54,6 @@
 	path = third-party/wlr-protocols
 	url = https://gitlab.freedesktop.org/wlroots/wlr-protocols
 	branch = master
+[submodule "third-party/nlohmann_json"]
+	path = third-party/nlohmann_json
+	url = https://github.com/nlohmann/json
--- a/cmake/compile_definitions/common.cmake
+++ b/cmake/compile_definitions/common.cmake
@@ -56,6 +56,7 @@ set(SUNSHINE_TARGET_FILES
         "${CMAKE_SOURCE_DIR}/third-party/moonlight-common-c/src/RtspParser.c"
         "${CMAKE_SOURCE_DIR}/third-party/moonlight-common-c/src/Video.h"
         "${CMAKE_SOURCE_DIR}/third-party/tray/tray.h"
+        "${CMAKE_SOURCE_DIR}/src/display_device/dd.h"
         "${CMAKE_SOURCE_DIR}/src/upnp.cpp"
         "${CMAKE_SOURCE_DIR}/src/upnp.h"
         "${CMAKE_SOURCE_DIR}/src/cbs.cpp"
@@ -138,4 +139,5 @@ list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
         ${Boost_LIBRARIES}
         ${OPENSSL_LIBRARIES}
         ${CURL_LIBRARIES}
+        ${JSON_LIBRARIES}
         ${PLATFORM_LIBRARIES})
--- a/cmake/compile_definitions/macos.cmake
+++ b/cmake/compile_definitions/macos.cmake
@@ -39,6 +39,7 @@ set(PLATFORM_TARGET_FILES
         "${CMAKE_SOURCE_DIR}/src/platform/macos/av_video.h"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/av_video.m"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/display.mm"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/display_options.mm"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/input.cpp"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/microphone.mm"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/misc.mm"
--- a/cmake/compile_definitions/windows.cmake
+++ b/cmake/compile_definitions/windows.cmake
@@ -74,7 +74,6 @@ list(PREPEND PLATFORM_LIBRARIES
         avrt
         iphlpapi
         shlwapi
-        PkgConfig::NLOHMANN_JSON
         ${CURL_STATIC_LIBRARIES})
 
 if(SUNSHINE_ENABLE_TRAY)
--- a/cmake/dependencies/common.cmake
+++ b/cmake/dependencies/common.cmake
@@ -19,6 +19,15 @@ pkg_check_modules(CURL REQUIRED libcurl)
 pkg_check_modules(MINIUPNP miniupnpc REQUIRED)
 include_directories(SYSTEM ${MINIUPNP_INCLUDE_DIRS})
 
+# nlohmann_json
+if(SUNSHINE_SYSTEM_NLOHMANN_JSON)
+    pkg_check_modules(NLOHMANN_JSON nlohmann_json>=3.9.0 REQUIRED IMPORTED_TARGET)
+    set(JSON_LIBRARIES PkgConfig::NLOHMANN_JSON)
+else()
+    add_subdirectory("${CMAKE_SOURCE_DIR}/third-party/nlohmann_json")
+    set(JSON_LIBRARIES nlohmann_json::nlohmann_json)
+endif()
+
 # ffmpeg pre-compiled binaries
 if(WIN32)
     if(NOT CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
--- a/cmake/dependencies/windows.cmake
+++ b/cmake/dependencies/windows.cmake
@@ -2,6 +2,3 @@
 
 set(Boost_USE_STATIC_LIBS ON)  # cmake-lint: disable=C0103
 find_package(Boost 1.71.0 COMPONENTS locale log filesystem program_options REQUIRED)
-
-# nlohmann_json
-pkg_check_modules(NLOHMANN_JSON nlohmann_json REQUIRED IMPORTED_TARGET)
--- a/cmake/prep/options.cmake
+++ b/cmake/prep/options.cmake
@@ -13,6 +13,7 @@ option(SUNSHINE_CONFIGURE_ONLY "Configure special files only, then exit." OFF)
 option(SUNSHINE_ENABLE_TRAY "Enable system tray icon. This option will be ignored on macOS." ON)
 option(SUNSHINE_REQUIRE_TRAY "Require system tray icon. Fail the build if tray requirements are not met." ON)
 
+option(SUNSHINE_SYSTEM_NLOHMANN_JSON "Use system installation of nlohmann_json rather than the submodule." OFF)
 option(SUNSHINE_SYSTEM_WAYLAND_PROTOCOLS "Use system installation of wayland-protocols rather than the submodule." OFF)
 
 option(CUDA_INHERIT_COMPILE_OPTIONS
--- a/docs/source/building/windows.rst
+++ b/docs/source/building/windows.rst
@@ -28,7 +28,6 @@ Install dependencies:
         mingw-w64-x86_64-curl \
         mingw-w64-x86_64-graphviz \
         mingw-w64-x86_64-miniupnpc \
-        mingw-w64-x86_64-nlohmann-json \
         mingw-w64-x86_64-nodejs \
         mingw-w64-x86_64-onevpl \
         mingw-w64-x86_64-openssl \
--- a/docs/source/source_code/source_code.rst
+++ b/docs/source/source_code/source_code.rst
@@ -63,6 +63,13 @@ Source
    src/*
 
 .. toctree::
+   :caption: src/display_device
+   :maxdepth: 1
+   :glob:
+
+   src/display_device/*
+
+.. toctree::
    :caption: src/platform
    :maxdepth: 1
    :glob:
--- a/src/config.h
+++ b/src/config.h
@@ -12,6 +12,7 @@
 #include <vector>
 
 #include "nvenc/nvenc_config.h"
+#include "display_device/dd.h"
 
 namespace config {
   struct video_t {
@@ -82,11 +83,7 @@ namespace config {
     bool install_steam_drivers;
   };
 
-  struct display_options_t {
-    int id;
-    std::string name;
-    bool is_primary_display;
-  };
+
 
   constexpr int ENCRYPTION_MODE_NEVER = 0;  // Never use video encryption, even if the client supports it
   constexpr int ENCRYPTION_MODE_OPPORTUNISTIC = 1;  // Use video encryption if available, but stream without it if not supported
--- a/src/config.h
+++ b/src/config.h
@@ -82,6 +82,12 @@ namespace config {
     bool install_steam_drivers;
   };
 
+  struct display_options_t {
+    int id;
+    std::string name;
+    bool is_primary_display;
+  };
+
   constexpr int ENCRYPTION_MODE_NEVER = 0;  // Never use video encryption, even if the client supports it
   constexpr int ENCRYPTION_MODE_OPPORTUNISTIC = 1;  // Use video encryption if available, but stream without it if not supported
   constexpr int ENCRYPTION_MODE_MANDATORY = 2;  // Always use video encryption and refuse clients that can't encrypt
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -543,6 +543,16 @@ namespace confighttp {
     outputTree.put("platform", SUNSHINE_PLATFORM);
     outputTree.put("version", PROJECT_VER);
 
+    pt::ptree displays;
+    for (const auto &[id, name, is_primary_display] : platf::display_options()) {
+      pt::ptree display_value;
+      display_value.put("id", id);
+      display_value.put("name", name);
+      display_value.put("is_primary", is_primary_display);
+      displays.push_front(std::make_pair(std::to_string(id), display_value));
+    }
+    outputTree.push_back(std::make_pair("displays", displays));
+
     auto vars = config::parse_config(file_handler::read_file(config::sunshine.config_file.c_str()));
 
     for (auto &[name, value] : vars) {
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -26,6 +26,8 @@
 #include <Simple-Web-Server/server_https.hpp>
 #include <boost/asio/ssl/context_base.hpp>
 
+#include <nlohmann/json.hpp>
+
 #include "config.h"
 #include "confighttp.h"
 #include "crypto.h"
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -533,34 +533,24 @@ namespace confighttp {
 
     print_req(request);
 
-    pt::ptree outputTree;
+    nlohmann::json outputJson;
     auto g = util::fail_guard([&]() {
-      std::ostringstream data;
-
-      pt::write_json(data, outputTree);
-      response->write(data.str());
+      response->write(outputJson.dump());
     });
 
-    outputTree.put("status", "true");
-    outputTree.put("platform", SUNSHINE_PLATFORM);
-    outputTree.put("version", PROJECT_VER);
+    outputJson.emplace("status", "true");
+    outputJson.emplace("platform", SUNSHINE_PLATFORM);
+    outputJson.emplace("version", PROJECT_VER);
 
-    pt::ptree displays;
-    for (const auto &[id, name, is_primary_display] : platf::display_options()) {
-      pt::ptree display_value;
-      display_value.put("id", id);
-      display_value.put("name", name);
-      display_value.put("is_primary", is_primary_display);
-      displays.push_front(std::make_pair(std::to_string(id), display_value));
-    }
-    outputTree.push_back(std::make_pair("displays", displays));
+    nlohmann::json displays = platf::display_options();
+    outputJson.emplace("displays", displays);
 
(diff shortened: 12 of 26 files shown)


ReenigneArcher:
This may cause conflicts with https://github.com/LizardByte/Sunshine/pull/2032

Hazer:
> This may cause conflicts with #2032

@ReenigneArcher That PR is amazing, maybe I'll try writing some of the macOS code needed in there, will that be a library with another repo or will be merged in the main code?

ReenigneArcher:
@Hazer it will mostly be in a separate library... in order to not grow the Sunshine code by 20%. @FrogTheFrog is doing the heavy lifting. They are slowly porting things over. https://github.com/LizardByte/libdisplaydevice

Hazer:
> ```
>   struct display_options_t {
>     std::string id; //< An actual value for "output_name"
>     std::string friendly_name; //< Some name for the end-user to identify this display
>     bool is_primary_display;
>   };
> ```
Sure, I was already discontent with the id as it was and wanted to change it, was currently reading https://github.com/LizardByte/Sunshine/issues/221 to gather ideas for it, string is the way to go.

> Also, on Windows primary display is a display that has origin coordinate of (0, 0). There can be multiple primary displays.

As I was writing this PR, "primary display" became less and less something simply achievable, it seems if we don't use x11, any pure wayland compositor, has no specific API for that, and people relies on asking the DE (KDE, GDM, etc) directly, so it's not doable, I was already thinking of removing it, or making it has any display that has origin coordinate of (0, 0), but I was trying to find out if it made sense for linux.

So meanwhile, you gave me the idea to rewrite it something like this:
```cpp
   struct display_options_t {
     std::string id;
     std::string friendly_name;
     current_display_settings_t current_settings;
   };

   struct current_display_settings_t {
     display_device::display_origin_t origin; //< This doesn't exists yet, but could be a nice addition to the API
     display_device::display_mode_t mode;
     
     bool current_display_settings_t::is_primary() const {
         return origin.x == 0 && origin.y == 0;
         // or maybe even:
         return origin.is_primary();
     }
   };
```

@FrogTheFrog what do you think? Meanwhile your PR is not merged, I just create a similar struct with the same name, we just replace it when it arrives.

Also @FrogTheFrog, could your `display_device::display_mode_t` include a `scaling_factor: float` or something? This way we can differentiate Retina displays from others, and will be good for determining Linux HiDPI scaling too. How does Windows handle HiDPI scaling?

FrogTheFrog:
> @FrogTheFrog what do you think? Meanwhile your PR is not merged, I just create a similar struct with the same name, we just replace it when it arrives.

After some deliberation, I have decided not to expose anything more from the inner API (including types apart from the needed ones) than :
```
device_info_map_t
enum_available_devices();

std::string
get_display_name(const std::string &device_id);
```
as we just don't need this information exposed to Sunshine at the moment (at least for Windows). Everything is done by passing the configuration anyway. So, the mode struct will not be available.

I would recommend keeping the boolean (as you're now exposing Windows specific origin point logic to Sunshine) or just dropping it altogether as you've wanted.

> Also @FrogTheFrog, could your `display_device::display_mode_t` include a `scaling_factor: float` or something? This way we can differentiate Retina displays from others, and will be good for determining Linux HiDPI scaling too. How does Windows handle HiDPI scaling?

HiDPI is a standalone setting hidden somewhere like the HDR and is not influenced by the display mode.


That aside, please don't use the same namespace as I'm only working on Windows code and for other OS'es it will have `noop` implementation. Since I will not be able to replace it completely, it will really mess everything up for me.

Hazer:
> After some deliberation, I have decided not to expose anything more from the inner API [...]
> as we just don't need this information exposed to Sunshine at the moment. [...] So, the mode struct will not be available.
> That aside, please don't use the same namespace as I'm only working on Windows code and for other OS'es it will have `noop` implementation. Since I will not be able to replace it completely, it will really mess everything up for me.

@FrogTheFrog That kinda sucks for me, I was expecting to just import the `display_device.h`, replace my struct with yours, populate the struct myself and be done with it. This feels like unnecessary duplication here, as this information is quite generic, and I don't need any of the other implementation details, just the same structs, but oh well. I will duplicate.

> I would recommend keeping the boolean (as you're now exposing Windows specific origin point logic to Sunshine) or just dropping it altogether as you've wanted.

Actually, I don't know, I think it's not a good idea for other platforms as a boolean either, it seems wayland can use the same origin logic, but via offset, and macOS can use the same origin logic but will ever only have one (0, 0) that will always be the primary, it's not windows related, I'm just not sure about x11 yet, but probably yes or will need a origin coordinate conversion, but same stuff. Also, with more information, I can show this in the list for user selection, for example, if the monitor friendly name is something I don't recognize, or maybe like my linux sample with "VirtualX" all over the place, the user will be able to see the current resolution, something like:
```
CABBAGE - 1920x1080 60hz scale 2.0 (-1920, 0) // HiDPI/Retina, left monitor
FQCADD - 1920x1080 120hz (-960, -1080) // top monitor
23MP55 - 1920x1080 60hz primary (0, 0)
```

And if I feel fancy, I could draw into Canvas all displays layed out using origin/resolution, so I'm not sure I want just the boolean anymore... But I'm still discovering what I'll really do.

FrogTheFrog:
I would suggest to keep it simple for now (don't make the mistake I did...) and add new functionality little by little.

My other problem is that I have too much code in the PR already and while I'm not against your suggestions, I don't want it to become a scope creep where I keep adding and adding stuff non-stop. I just want to merge my stuff already :').

That being said I could add current display mode to `device_info_map_t` and position, just no scaling info yet, though.

FrogTheFrog:
Actually please find out what data you really want and then let's discuss it.

Also I think I've misunderstood it. If you want to reuse the "extended" `device_info_t`, then please do. I though about something else...

Hazer:
> (I really should go to sleep...)

Hope you got some rest, now I'm the one way past the bedtime

FrogTheFrog:
> > (I really should go to sleep...)
> 
> Hope you got some rest, now I'm the one way past the bedtime

I got slightly sick :D

Got a few points/questions:

1. Should the scale factor be represented in a `rational_t`? Like we have for refresh rate (would have to rename the type into a generic one). I copied what Windows was doing (they hate floats for some reason) to save the settings and just wondering whether we should apply the same pattern.
2. Should the `scale_factor` really be part of the `resolution_t` and not exist on it's own in the `current_settings`? On Windows it is not considered a part of the display mode and persists between mode changes (like the HDR setting).
3. One pitfall of the `rational_t` is that the `denominator` always has to be non-zero. Otherwise you create a black hole when converting it to float (which you already did)...

FrogTheFrog:
Stop commiting and go to sleep!

Hazer:
@FrogTheFrog Hope your feeling better! I got some rest, and some food poisoning later on... So I got even more rest!

Those are great points, that I'm not sure how to answer myself, and I probably should do a closer look into windows and your work so far, it will be easier if I understand from the same PoV, but at first glance:

#### 1. 
Not sure, macOS handles as Double, Linux seems like a mess, x11 and wayland have different implementations, but it also seems to change between DEs, it seems Gnome uses Integer, while KDE uses Float and supports Fractional Scaling, while every other tool may have it's own standards, in my mind initially, double/float was fine. I think we can use the same pattern you're using with windows, but can you show me how to convert between `double` and `rational_t`? I was lazy to find out how do you do it properly :D

#### 2.
I see what you mean, but I'm not sure yet, let's take my MacBook screen for example, `1280x800`, `@2x` scaling, which means, my actual resolution is `2560x1600`, that's my actual pixel count, but both "resolutions" are valid, what changes it's the scale, but if you see the macbook settings it says 1280x800, it's the default Retina (HiDPI) resolution. 

So when streaming, that information is useful, as I may want to stream in `1280x800@1x` or in native `1280x800@2x`==`2560x1600`, and it seems deeply tied to resolution to me, but as you pointed out, it doesn't for windows, kinda backwards even, windows mostly always just renders `1:1` with the resolution and that's it. In my mind, they are tied, but in reality it may be an external like HDR.

The only issue is when changing "resolutions" in macOS, at least before Sonoma, we are only changing the scale actually, the native pixel count will still be 2560x1600 when I'm rendering my screen in "1440x900" (it's called 'scaled looking like 1440x900' in the settings), it just changes the scale to `1.??x`.

Sooo, if we think in math terms it's represented as `resolution * scale_factor == native_resolution`, so I guess, part of the `mode` changes because macOS and Linux may use, but not part of `resolution` itself? Meanwhile windows will always be `1x` scale, and may even hardcode this value? What do you think?

#### 3.
ahhahaha yeah, I was sure I wasn't using the right values here, I just wanted to test it and be done with it for the day tbh, there's another APIs to get this value, so I wasn't even sure I should use this one. Actually, I need this info, given 60hz, which is my monitor refresh rate, what are the correct values for `numerator` and `denominator` the way it's currently planned?

====

I probably should try implementing the actual mode change for macOS as a PR for the libdisplay and actually map the available APIs for mode change in macOS, I already have some idea, it's mostly around the crusty `CGDisplayMode` CoreGraphics API, but I don't really have used it myself. It's good to refine the public API if I actually try to use the API in practice, here I mostly just reused your structs without much brain power in place.

FrogTheFrog:
1.
Hmmm, conversion could look like this (maybe we should add a static method `static rational_t fromFloat(float value)` so that it can be improved later on):
`numerator = static_cast<int>(value * 1000); denominator= 1000;`
as for converting back, it's simply `numerator/denominator` (again, maybe in a const method to check for `denominator!= 0`).

2.
We can hardcode it `1x` for Windows (I will do the same until I figure how to read it), but I would still like to keep it out of the `resolution_t` and maybe even `display_mode_t`. It's not about whether it's part of the resolution (as a whole) or not, but whether if it's part of the OS API. For example, it needs to be changed separately on and Linux KDE (if I remember correctly) and Windows:
image
so I would like to keep it separated to that no unnecessary stuff is passed to functions (separation of concerns). Unless it's part of the MacOS API and then we need to think of a better structure.

3.
It's up to you really. It can be `numerator = 60; denominator= 1;` or `numerator = 60000; denominator= 1000;`.

====

I'm still working on the `libdisplaydevice` so plz wait a little (or a lot if you can since I'm busy IRL) :D.

The plan is still to make it platform agnostic more or less and expose only the `session_t` (will be renamed) class with some additional methods (namely the `enum_available_devices` and `get_display_name` equivalent).
`settings_t` (will also be renamed) will have a shared interface between OS'es (kinda like it does right now) and will be responsible for applying the `parsed_config_t`.
Everything behind `settings_t` is up to you to decide how to implement or how it will look like. The methods in `displaydevice.h` will not be exposed directly and will be private.

With that said, I would recommend to start working on branch based on my PR since it will not change by a lot and you'll find some BS most likely that you will have to hack around (like the audio context capture I have to do on Windows).

FrogTheFrog:
I think I am (or we) approaching this from a wrong point. We don't need to expose the mode structure at all. How about we have the following:

```
  enum class device_state_e {
    inactive,
    active,
    primary
  };

  enum class hdr_state_e {
    unknown,
    disabled,
    enabled
  };

  struct resolution_t {
    unsigned int width;
    unsigned int height;
  };

  struct rational_t {
    unsigned int numerator;
    unsigned int denominator;
  };

  struct position_t {
    int x;
    int y;
  };

  struct device_info_t {
    std::string display_name;
    std::string friendly_name;
    device_state_e device_state;
    hdr_state_e hdr_state;
    resolution_t resolution;
    rational_t refresh_rate;
    rational_t scaling; // resolution_scale, scale or what's the best name?
    position_t origin_point;
  };
  
  using device_info_map_t = std::map<std::string, device_info_t>;
  ```

**Comment by Hazer** [comment]
#### 1/3
Ahh, thanks! Got it now, I had another thing in mind. In my case it will be `numerator = screen.maximumFramesPerSecond; denominator= 1;` then, as macOS already handles this as Integer in the decimal range. It has another way of doing it with fractionally values but this way is even less work.

#### 2
I did investigate how `CGDisplayMode` works, and you're completely right. There's no scale there in macOS either, we cannot get neither the scale nor the native pixel resolution at this point, the scale is a property of Display/Screen and all the modes are already calculated with scale in mind. We could even say that the Scale changes WHEN the mode changes, not the other way.

Regarding the UI, here are KDE, Ubuntu(Gnome or unity I guess) and macOS as example, and probably why my mind got stuck with the idea that it was correlated with resolution or mode:
kde:
kde
gnome:
ubuntu_display
macOS 1280x800@2x (2560x1600):
Screen Shot 2024-05-06 at 21 23 19
macOS 1440x900@1.??x
Screen Shot 2024-05-06 at 21 23 29

> How about we have the following:

Looks perfect to me. I'd use `resolution_scale` to make it explicit, and probably I'd add to `device_info_t` the `std::string id`, instead of only relying on it as the key for `device_info_map_t`, this way I could just pass this struct along and have everything I need, but it's not a necessary redundancy, it's just my preference.

> I would recommend to start working on branch based on my PR

I can do that, but I can wait too, if we use the above structs as protocol, so `libdisplaydevice` exposes `device_info_t` and children, I can easily move forward without giving you trouble, be it here or actually starting a PR pointing to your PR.

Also, thanks for the patience and your time so far!

**Comment by FrogTheFrog** [comment]
> Looks perfect to me. I'd use resolution_scale to make it explicit, and probably I'd add to device_info_t the std::string id, instead of only relying on it as the key for device_info_map_t, this way I could just pass this struct along and have everything I need, but it's not a necessary redundancy, it's just my preference.

This is a completely valid pattern and I don't mind, but can we please use `device_id` instead of `id`? I want to be explicit that this id belongs to a device and not a display. There is a difference on Windows...
bVrNg7G

Anyway, that's all. Best of luck to us both!

**Comment by Hazer** [comment]
I'll split this PR next month or so, I think it will be better to postpone, now that I have a better understanding of the whole stack, and I have other improvements planned to be merged first.

**Comment by FrogTheFrog** [comment]
Heyo, I have reached a point where I am implementing the last function for the OS abstraction layer - the enumeration function. After thinking for a while, I would like to have the following structure in the end:

```
namespace display_device {
  /**
   * @brief The device's HDR state in the operating system.
   */
  enum class HdrState {
    Disabled,
    Enabled
  };

  /**
   * @brief Display's resolution.
   */
  struct Resolution {
    unsigned int m_width {};
    unsigned int m_height {};
  };

  /**
   * @brief An arbitrary point object.
   */
  struct Point {
    int m_x {};
    int m_y {};
  };

  /**
   * @brief Enumerated display device information.
   */
  struct EnumeratedDevice {
    /**
     * @brief Available information for the active display only.
     */
    struct Info {
      Resolution m_resolution {}; /**< Resolution of an active device. */
      float m_resolution_scale {}; /**< Resolution scaling of an active device. */
      float m_refresh_rate {}; /**< Refresh rate of an active device. */
      bool m_primary {}; /**< Indicates whether the device is a primary display. */
      Point m_origin_point {}; /**< A starting point of the display. */
      std::optional<HdrState> m_hdr_state {}; /**< HDR of an active device. */
    };

    std::string m_device_id {}; /**< A unique device ID used by this API to identify the device. */
    std::string m_display_name {}; /**< A logical name representing given by the OS for a display. */
    std::string m_friendly_name {}; /**< A human-readable name for the device. */
    std::optional<Info> m_info {}; /**< Additional information about an active display device. */
  };

  /**
   * @brief A list of EnumeratedDevice objects.
   */
  using EnumeratedDeviceList = std::vector<EnumeratedDevice>;
}  // namespace display_device
```

I have trimmed the fat and kept it simple. Since it will now live in the library, I will no longer expose it as a JSON via `NLOHMANN` API, instead all the types will be serializable via the public `toJson(...)` and `fromJson(...)` (to hide the dependency).

As you can see, I am using the `m_` with a snake case, which is not common in web development. I was planning on removing the prefix when converting to json string, but maybe it should also be converted to camel case? I would like to hear your opinion on this :D .

**Comment by ns6089** [comment]
Sorry for seemingly out-of-nowhere comment
```cpp
    struct Info {
      Resolution m_resolution {}; /**< Resolution of an active device. */
      float m_resolution_scale {}; /**< Resolution scaling of an active device. */
      float m_refresh_rate {}; /**< Refresh rate of an active device. */
      bool m_primary {}; /**< Indicates whether the device is a primary display. */
      Point m_origin_point {}; /**< A starting point of the display. */
      std::optional<HdrState> m_hdr_state {}; /**< HDR of an active device. */
    };
```
@FrogTheFrog maybe use `std::variant<float, rational_t> m_refresh_rate;` if it's platform specific?

FrogTheFrog:
> Sorry for seemingly out-of-nowhere comment
> 
> ```c++
>     struct Info {
>       Resolution m_resolution {}; /**< Resolution of an active device. */
>       float m_resolution_scale {}; /**< Resolution scaling of an active device. */
>       float m_refresh_rate {}; /**< Refresh rate of an active device. */
>       bool m_primary {}; /**< Indicates whether the device is a primary display. */
>       Point m_origin_point {}; /**< A starting point of the display. */
>       std::optional<HdrState> m_hdr_state {}; /**< HDR of an active device. */
>     };
> ```
> 
> @FrogTheFrog maybe use `std::variant<float, rational_t> m_refresh_rate;` if it's platform specific?

I thought about that initially, but I don't really need to use the `rational_t` type outside of the platform specific stuff and I can convert `float -> rational_t` without much of hassle.

FrogTheFrog:
FYI, I have already finalized the "public" types and am happy about them (for now):
https://github.com/LizardByte/libdisplaydevice/blob/master/src/common/include/displaydevice/types.h

ns6089:
> I can convert float -> rational_t without much of hassle

Wait, it's this `rational_t` type is just numerator paired with denominator? Then the conversion to float should be destructive and not perfectly reversible.

FrogTheFrog:
In any case the user would input the refresh rate as a string/number that would have to be converted to the rational anyway, so I've decided to store the intermediate value as a floating type as it makes sense.

ns6089:
> In my specific case, the denominator is always 10 to the power of x

That's rather dangerous assumption, unless all APIs state that explicitly. And consequences are multiplied because you're relying on it in a public interface.
At the end of the day it's your decision to make, but I would think twice in this case.

FrogTheFrog:
> > In my specific case, the denominator is always 10 to the power of x
> 
> That's rather dangerous assumption, unless all APIs state that explicitly. And consequences are multiplied because you're relying on it in a public interface. At the end of the day it's your decision to make, but I would think twice in this case.

Sorry, I'm bad at explaining stuff. I make no such assumptions. The types I have provided a link for are for humans/HMI (for the lack of better words). When I'm converting human input to the rational, the denominator will always be `10^x` since I don't know how else to convert it otherwise and Windows API is happy with it (I have tested that it treats `60/1` and `60000/1000` the same).

Internally I do not create new types and just use the Windows' ones. There are a few exceptions where I need to types to be saved to a file for persistence, so I've just made my own types (like `rational_t`) that are 1:1 to the Windows' ones. I do not (for example) convert the Windows' rational type to float and pass it around to another function where it would be converted back to rational type as it would loose the information during conversion.

FrogTheFrog:
That aside if you think that we should use rational type instead of float in general, just please say so. I have no preference and would gladly use it instead of floating point @ns6089

ns6089:
I think it's either you're overcomplicating things, or I'm oversimplifying them.
This is how I understand it:
1. Windows lists available refresh rates as pairs of integers
2. Windows uses the same pairs of integers when switching refresh rates (here you're currently making the assumption it will switch to closest listed mode, instead of doing some VRR magic on some exotic display)
3. There is no point in creating some extra "HMI" representation of these pairs of integers in the public interface, clients can do it themselves

FrogTheFrog:
> I think it's either you're overcomplicating things, or I'm oversimplifying them. This is how I understand it:
> 
>     1. Windows lists available refresh rates as pairs of integers
> 
>     2. Windows uses the same pairs of integers when switching refresh rates (here you're currently making the assumption it will switch to closest listed mode, instead of doing some VRR magic on some exotic display)
> 
>     3. There is no point in creating some extra "HMI" representation of these pairs of integers in the public interface, clients can do it themselves

1. Yup, I call it Rational.
2. Yes, but I also explicitly ask it to select the closest "best" (according to Windows) refresh rate and if I'm not happy with it I try to set exactly as provided.
3. The problem for me is that I wanted to make the interface platform agnostic and I don't know other OS expose the refresh rate, so I thought `double` would make sense, however now I've realized that it might not be true for Windows :/

FrogTheFrog:
So, in short @ns6089. After thinking about it, I would like to use `rational_t` for anything floating point related in the public API.

Would you still say that I should use something like:
```
using FloatingType = std::variant<double, rational_t>
```

ns6089:
Yes, I think this is the best. `float` is guaranteed to be a subset of `double`, so two-type variant should be enough.

Hazer:
@FrogTheFrog sorry I missed the whole discussion, as it started while I was moving to live in another city and I got behind the amount of notifications I was receiving while doing so.

The only thing I didn't quite like, but I still think it's a good decision, is hiding the `NLOHMANN` dependency, because I plan on making a PR to use it on Sunshine directly for all APIs, as the current json manipulation we are doing isn't great. But I do understand that hiding it is a good idea. 

I'm not a fan of `m_` and other prefixes, but I come from mobile development and we fought during years to get rid of it on Android style guides, even the guy who came with the idea on AOSP apologized, so I'm not exactly the right person to ask about it, yeah I'm fine with using it like this in C++, but serializing without that would be great, lol.

Again, amazing work, and I'll be revisiting this very soon. Also, don't be afraid to ping me on Discord if I get AWOL or if you need faster input from me. I sent you a message and added you as friendo.

PR #2550, opened 2024-05-21
fix(macos): Mouse input broken in-game

Hazer:
## Description
This issue was initially reported [on Discord](https://discord.com/channels/804382334370578482/1238526773940191303/1238526773940191303)

Basically, we have 2 main issues:
- The cursor sometimes keeps showing on the screen even when it should not appear;
- While the cursor moves on the screen, some 3D engines can't capture the movements and cannot move mechanisms bound to the mouse, such as Camera, Object Rotation, etc.

The behavior can be reproduced on Minecraft Java Edition and Tomb Raider. FPS games are mostly prone, some games may not really have issues, but I bet most 3D FPS are failing.

### Screenshot
bug:
bug gif

fixed:
fixed gif

## Type of Change

## Checklist

## Branch Updates


tests/unit/test_mouse.cpp
```diff
@@ -0,0 +1,103 @@
...
+class MouseTest: public virtual BaseTest, public ::testing::WithParamInterface<util::point_t> {
+protected:
+  void
+  SetUp() override {
+    BaseTest::SetUp();
+#ifdef _WIN32
+    // TODO: Windows tests are failing, `get_mouse_loc` seems broken and `platf::abs_mouse` too
+    //       the alternative `platf::abs_mouse` method seem to work better during tests,
+    //       but I'm not sure about real work
+    GTEST_SKIP_((std::string("MouseTest:: skipped for now. TODO Windows").c_str());
```
ReenigneArcher, line 19: I think just use a regular string here, like: https://github.com/LizardByte/Sunshine/blob/26ceec5f3c97b0130b82cb28ba53c3b4dafc8419/tests/conftest.cpp#L199

packaging/macos/sunshine_pr.rb
```diff
@@ -0,0 +1,68 @@
+require "language/node"
+
+class @PROJECT_NAME@ < Formula
+  desc "@PROJECT_DESCRIPTION@"
+  homepage "@PROJECT_HOMEPAGE_URL@"
+  url "@GITHUB_CLONE_URL@", branch: "@GITHUB_BRANCH@"
```
ReenigneArcher, line 6: I don't really understand the purpose of a separate file if this is the only change? I'd also really like to avoid a separate file if possible. Ideally we validate the "production" file on PRs, with the only difference being what get's checked out.

Even though the other one is using "tag" it should still checkout the right commit, unless I'm missing something.

.github/workflows/CI.yml
```diff
@@ -823,31 +838,35 @@ jobs:
...
+              sudo sqlite3 "$dbPath" "SELECT * FROM access WHERE service='kTCCServiceScreenCapture';"
+              execute_sql_query "$value" "$dbPath"
+              echo "Updated permissions for $dbPath"
+              echo "-------------------"
+              sudo sqlite3 "$dbPath" "SELECT * FROM access WHERE service='kTCCServiceScreenCapture';"
+            done
           done
 
-      - name: Run tests
+      - name: Run tests directly
```
ReenigneArcher, line 856:
```suggestion
      - name: Run tests
```

.github/workflows/CI.yml
```diff
@@ -823,31 +838,35 @@ jobs:
...
         id: test
         timeout-minutes: 10
+        working-directory:
+          /opt/local/var/macports/build/_Users_runner_work_Sunshine_Sunshine_ports_multimedia_Sunshine/Sunshine/work/build/tests
         run: |
-          sudo port test "Sunshine"
+          sudo port install doxygen
+          sudo port install graphviz
+          sudo ./test_sunshine --gtest_color=yes
```
ReenigneArcher, line 864: Is this to get around the threading issue? If so, could we miss that issue when the user is running Sunshine from macports?

Is sudo required to run the test?

.github/workflows/CI.yml
```diff
@@ -823,31 +838,35 @@ jobs:
...
 
       - name: Test Logs
         if: always()
         run: |
-          logfile="/opt/local/var/macports/logs/_Users_runner_work_Sunshine_Sunshine_ports_multimedia_Sunshine/Sunshine/main.log"
+          logfile="/opt/local/var/macports/build/_Users_runner_work_Sunshine_Sunshine_ports_multimedia_Sunshine/Sunshine/work/build/tests/test.log"
           cat "$logfile"
```
ReenigneArcher, line 870: This step probably isn't even necessary if we stick with running the test binary directly.

packaging/macos/sunshine_pr.rb
Hazer: I agree on not having another file, actually, this commit shouldn't be in this PR, I was testing and forgot to remove.

But regarding why, it's so I can download the file from the open PR and run it without changing anything, to install locally with brew, but when giving the PR branch name to the tag, it can't find and fails to proceed with installation. I'll remove this change to another place

.github/workflows/CI.yml
Hazer: @ReenigneArcher so basically, I don't even know if it can be said is a threading issue, and not a macports issue. This `test_sunshine` runs perfectly in any other scenario, via brew, via IDE, via terminal, anything, but when executed from anywhere inside port, it thread locks. I've tried everything I had in mind running port test locally, couldn't fix it.

I tried removing it from the tests section in ports, and running it in `post-build`, it does the same, locks on any input API, probably because of `CFRunLoop` getting locked, but only in ports.

So instead, I'd rather run it manually. And it's running faster too. Given that the tests are built using ports anyway, seems fine to me.

Regarding the user facing the issue later, I created a filter in the ports file, to avoid tests with HID in the name, we may also do it for other scenarios we want to avoid testing on the user device end. Actually, do we need to force run tests on installation on user devices? I was thinking, this test should not run on user device anyway, it moves the user mouse, and if the user is using the mouse during the tests, the tests will randomly break anyway. Other tests may have similar issues, I'd rather just run tests on the CI, and allow users to run tests manually later.

Yep, sudo required, I want to see if I can remove this restriction, but it seems the tests that stuff write to files break when running without sudo.

.github/workflows/CI.yml
Hazer: I think it's still useful, because I was looking at this file, and it seems to create the BOOST logs from the test run, only the boost stdout without all the other test output, even without the results, but it may be good for debugging.

.github/workflows/CI.yml
ReenigneArcher: > Actually, do we need to force run tests on installation on user devices? I was thinking, this test should not run on user device anyway, it moves the user mouse, and if the user is using the mouse during the tests, the tests will randomly break anyway. Other tests may have similar issues, I'd rather just run tests on the CI, and allow users to run tests manually later.

We don't need to force it, and I don't think using the other method does that. It would be good if users could run the tests on their specific machine though, once we get more tests it might be helpful to use as a debugging tool. There would probably need to be a message to not move the mouse during the test or other notifications depending on what tests we end up adding... but user running tests would be completely optional. CI is the main focus.

Hazer:
Rebase with #2606 before moving from draft

.github/workflows/CI.yml
```diff
@@ -737,31 +749,35 @@ jobs:
...
```
ReenigneArcher, line 781: We don't need test logs anymore if running the test binary manually.

.github/workflows/CI.yml
```diff
@@ -737,31 +749,35 @@ jobs:
...
       - name: Run tests
         id: test
         timeout-minutes: 10
+        working-directory:
+          /opt/local/var/macports/build/_Users_runner_work_Sunshine_Sunshine_ports_multimedia_Sunshine/Sunshine/work/build/tests
         run: |
-          sudo port test "Sunshine"
+          sudo port install doxygen
+          sudo port install graphviz
+          sudo ./test_sunshine --gtest_color=yes
```
ReenigneArcher, line 775:
```suggestion
          sudo port install \
            doxygen \
            graphviz
          sudo ./test_sunshine --gtest_color=yes
```

.github/workflows/CI.yml
```diff
@@ -716,19 +716,31 @@ jobs:
           name: sunshine-macports
           path: artifacts/
 
-      - name: Fix screen capture permissions
-        if: ${{ matrix.os_version != 12 }}  # macOS-12 is okay
-        # can be removed if the following is fixed in the runner image
-        # https://github.com/actions/runner-images/issues/9529
-        # https://github.com/actions/runner-images/pull/9530
+      - name: Fix permissions
```
ReenigneArcher, line 719: Did the changes made here actually help?

src/platform/windows/input.cpp
```diff
@@ -536,12 +551,29 @@ namespace platf {
     auto &mi = i.mi;
 
     mi.dwFlags = MOUSEEVENTF_MOVE;
+    // mi.dwFlags = MOUSEEVENTF_MOVE | MOUSEEVENTF_MOVE_NOCOALESCE;  // TODO: Not sure if needed
```
ReenigneArcher, line 554: Before merging we should probably remove the commented out code.

Comment applies to the block above and a line in the `get_mouse_loc` function as well.

How were you testing this on Windows? Only via the CI workflows?

.github/workflows/CI.yml
Hazer: Here I'm not sure, but they will be useful to write tests in https://github.com/LizardByte/Sunshine/pull/2508

.github/workflows/CI.yml
Hazer: These logs are the boost_log outputs of the code run during the tests, without the Test logs, just pure execution logs, felt useful to me, but I can remove.

src/platform/windows/input.cpp
Hazer: At first yeah, but later I got it running on windows, but I can't continue implementing this on windows this month, as we discussed. I'll remove the comments and keep them in another branch to continue when I'm available again.

PR #2594, opened 2024-05-27
refactor: split server and inputs into their own groups

Hazer:
## Description
Breaking the project in smaller pieces will make the project easier to contribute and maintain in the future, it's not a major rewrite, just moving files or splitting a single file into many without major refactors.

Moves the following files to the `server` folder:
```
confighttp.cpp
confighttp.h
crypto.cpp
crypto.h
httpcommon.cpp
httpcommon.h
network.cpp
network.h
nvhttp.cpp
nvhttp.h
rtsp.cpp
rtsp.h
upnp.cpp
upnp.h
```

Moves the following input to the `input` folder and split it into:
```
common.cpp
common.h
gamepad.cpp
gamepad.h
init.h
keyboard.cpp
keyboard.h
mouse.cpp
mouse.h
pen.cpp
pen.h
processor.cpp
processor.h
touch.cpp
touch.h
```

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [x] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I want maintainers to keep my branch updated

PR #2606, opened 2024-05-29
feat(input/linux): add support for more virtual input devices

Hazer:
## Description
This is a direct continuation of https://github.com/LizardByte/Sunshine/pull/2315 @ABeltramo 's amazing work.

- Rebased to master (f0a00ae35683dfd5742952fb4cf99edf268b57fd)
- Adopted new vue components
- Broken the `inputtino.cpp` into namespaces by kind
- Renamed `input.cpp` to `legacy_input.cpp`

> I had to rename the branch for CI to work, this is a continuation of #2261
> 
> Added support for the followings new virtual devices:
> 
> * Nintendo Switch pro joypad
> * DualSense (PS5) joypad, including:
>   
>   * Gyroscope
>   * Accelerometer
>   * Touchpad
>   * LED
>   * Battery status
>   * Force feedback? (is this supported in Moonlight already? I have to check..)
>   * Microphone and Headset jack? (we should be able to plug into that as well, to be checked..)
> 
> And generally completely refactored all virtual input code using [inputtino](https://github.com/games-on-whales/inputtino) as a library
> 
> ### Requirements
> The DualSense implementation requires access to `/dev/uhid`, (more on this later..) this is easily done by setting the right permissions, similar to how we deal with `udev` already:
> 
> First we’ll add our user to the input group:
> 
> ```
> sudo usermod -a -G input $USER
> ```
> 
> Then we add an udev rule to allow access to /dev/uhid:
> 
> ```
> echo 'KERNEL=="uhid", GROUP="input", MODE="0660"' | sudo tee /etc/udev/rules.d/99-uhid.rules
> ```
> 
> Finally, we have to make sure that the uhid kernel module is loaded at boot:
> 
> ```
> echo "uhid" | sudo tee /etc/modules-load.d/uhid.conf
> ```
> 
> ### Code details
> The main file is `src/platform/linux/inputtino.cpp` I've kept the original `input.cpp` so that we can easily compare the two implementations and possibly add what's missing but the idea would be to completely replace that implementation with this. To switch from one implementation to the other you can just set the Cmake variable `-DSUNSHINE_USE_INPUTTINO=OFF`.
> 
> Unfortunately we have to deal with `uhid` in order to properly emulate gyro, acceleration and touchpad; there's a bit of a rationale into why this is not possible with `udev` [here](https://github.com/games-on-whales/inputtino/blob/uhid-ps5/src/uhid/README.adoc). This might be a deal breaker for you guys but I hope it's not; if that's the case, we can still emulate a PS5 joypad using `udev` without supporting gyro and acceleration.
> 
> I think bringing in `uhid` has also other benefits, since it sits at a lower level than `uinput`, we can really **emulate any kind of usb device** and possibly even completely integrate **usb over IP** if this will ever be considered as an addition to the protocol.
> 
> As for the code itself, Inputtino has got unit tests using `libinput` for mouse, keyboard, pen and touchscreen and `SDL2` for testing joypads. These aren't complete (yet) but should cover most of the codebase.
> 
> ### Future plans
> * I'd like to give it a go at implementing gyro and acceleration support for a Nintendo joypad too.
> * Ideally, since we are emulating the real joystick USB messages, this should also be a good foundation for a Windows virtual driver implementation. I know nothing about this but I'd like to do more research on it and possibly add a multiplatform implementation to inputtino.
>   
>   * I know of the EV license and all the requirements, for some reason it's hard to convince companies to invest in this without a functioning prototype first 😅
> * Add support for trackpads in Moonlight? Generally it seems to not be the same as a touchscreen: it moves relatively instead of absolutely and has got different gestures.
> 
> ### What's missing?
> Most notably the XTest fallback that is present in `input.cpp` if this is something that is used we can easily add it here as well. Everything else should be implemented, please let me know if something is missing or misbeaving.
> 
> I'm probably forgetting something else..

### Issues Fixed or Closed
Might help with #1720

## Type of Change

## Checklist

## Branch Updates


src_assets/common/assets/web/configs/tabs/Inputs.vue
```diff
@@ -22,15 +23,27 @@ const config = ref(props.config)
...
+          <template #windows>
+            <option value="ds4">{{ $t('config.gamepad_ds4') }}</option>
+            <option value="x360">{{ $t('config.gamepad_x360') }}</option>
+          </template>
+
+          <template #linux>
+            <option value="ds5">{{ $t("config.gamepad_ds5") }}</option>
+            <option value="switch">{{ $t("config.gamepad_switch") }}</option>
+            <option value="xone">{{ $t("config.gamepad_xone") }}</option>
+          </template>
```
ReenigneArcher, line 41:
```suggestion
          <template #linux>
            <option value="ds5">{{ $t("config.gamepad_ds5") }}</option>
            <option value="switch">{{ $t("config.gamepad_switch") }}</option>
            <option value="xone">{{ $t("config.gamepad_xone") }}</option>
          </template>

          <template #windows>
            <option value="ds4">{{ $t('config.gamepad_ds4') }}</option>
            <option value="x360">{{ $t('config.gamepad_x360') }}</option>
          </template>
```

Alphabetize platforms.

cmake/compile_definitions/linux.cmake
```diff
@@ -230,14 +230,33 @@ if (${SUNSHINE_TRAY} EQUAL 0 AND SUNSHINE_REQUIRE_TRAY)
     message(FATAL_ERROR "Tray icon is required")
 endif()
 
+if(${SUNSHINE_USE_LEGACY_INPUT})
+    list(APPEND PLATFORM_TARGET_FILES "${CMAKE_SOURCE_DIR}/src/platform/linux/input/legacy_input.cpp")
```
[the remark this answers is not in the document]
ReenigneArcher: I agree with the TODO

ReenigneArcher:
@ABeltramo any idea how to solve these errors on Fedora 40? https://github.com/LizardByte/Sunshine/actions/runs/9295282809/job/25581974710?pr=2606#step:10:3266

We're using gcc-14 on that distro, I'm guessing that's part of the problem.

ABeltramo:
I've quickly skimmed thru the logs:

> error: 'clamp' is not a member of 'std'

sounds like it's missing `#include <algorithm>` I've just pushed a potential fix, can you upgrade the submodule to [bc6afda](https://github.com/games-on-whales/inputtino/commit/bc6afda52b2a67abb26a7134fe63c4063a655edf) ?

Hazer:
> sounds like it's missing `#include <algorithm>` I've just pushed a potential fix, can you upgrade the submodule to [bc6afda](https://github.com/games-on-whales/inputtino/commit/bc6afda52b2a67abb26a7134fe63c4063a655edf) ?

@ABeltramo Seems to work! But now it needs the same fix in `uinput/pentablet.cpp` too.

ABeltramo:
This is probably the most in-depth bug report I've ever received, thanks @azuwis !  
I know there's something wrong with the actual values of gyro and accelerometer as I was mentioning [here](https://github.com/LizardByte/Sunshine/pull/2315#issuecomment-2138359017) but thanks for confirming it, I have to check where the error is between Moonlight reported values and the virtual pad implementation

ABeltramo:
> > sounds like it's missing `#include <algorithm>` I've just pushed a potential fix, can you upgrade the submodule to [bc6afda](https://github.com/games-on-whales/inputtino/commit/bc6afda52b2a67abb26a7134fe63c4063a655edf) ?
> 
> @ABeltramo Seems to work! But now it needs the same fix in `uinput/pentablet.cpp` too.

My bad, I've added that too!

Hazer:
> I have to check where the error is between Moonlight reported values and the virtual pad implementation

@azuwis thanks for your report!

@ABeltramo I'm splitting in smaller parts the upper input layer too, so it will be easier to understand, here is the PR https://github.com/LizardByte/Sunshine/pull/2594 if you want to use it

And the (hopefully) last changes missing to build on all platforms: https://github.com/games-on-whales/inputtino/pull/4

ABeltramo:
@Hazer I've pushed a fix for Gyro and Acceleration, it should now be very close to the original:

https://github.com/LizardByte/Sunshine/assets/1122217/5ac48259-4322-484b-bea0-4c42ffb77d38

You have to update inputtino to the latest and change the following (I'm afraid I can't push to this PR, right?):

https://github.com/LizardByte/Sunshine/blob/6cf7f3dc61985f22d08211c50b9ee2cd07659096/src/platform/linux/input/inputtino_gamepad.cpp#L210

to:
```cpp
std::get<inputtino::PS5Joypad>(*gamepad->joypad).set_motion(inputtino::PS5Joypad::GYROSCOPE, deg2rad(motion.x), deg2rad(motion.y), deg2rad(motion.z));
```

(add the `deg2rad` conversion).

It would be great if @azuwis can give it another go after this change is in!

Hazer:
@ABeltramo sent you an invite for my fork, if you want to change directly.
Another fix for fedora: https://github.com/games-on-whales/inputtino/pull/5

Updated with latest ps5 changes, can you give it another go @azuwis?

ABeltramo:
> @ABeltramo sent you an invite for my fork, if you want to change directly. Another fix for fedora: [games-on-whales/inputtino#5](https://github.com/games-on-whales/inputtino/pull/5)
> 
> Updated with latest ps5 changes, can you give it another go @azuwis?

Damn you Fedora! I'm going to install it in a VM to quickly test this compilation issues..  
I don't see the `deg2rad` change that I've mentioned before in the commit that you pushed, could it be because I'm on mobile?  
Thanks for the invite, I'll double-check everything again tomorrow!

Hazer:
@ABeltramo a few developers VS a hat, who would win?
fedora hat
So far the hat is beating both of us 🤣 

I don't know, maybe it's because it's on mobile. I've updated now to https://github.com/games-on-whales/inputtino/commit/99c984068408380bcc13766ff124a2e45f27dbd3, so I guess everything should be there.

ABeltramo:
Thanks @azuwis for confirming! @Hazer I've pushed the missing bit to your branch, I think there are only a few things left for this PR:

- [ ] Fix the Fedora build
- [ ] Fix the PS5 trackpad integration (it seems to not move properly)
- [ ] Decide if we want to add the automatic failover (mentioned [here](https://github.com/LizardByte/Sunshine/pull/2315#issuecomment-2138359017)) when uhid/uinput aren't available and implement it

Anything else missing?

ReenigneArcher:
For your third checklist, I don't know that it's worth it. I don't think we'll be keeping both code paths for long.

ABeltramo:
> For your third checklist, I don't know that it's worth it. I don't think we'll be keeping both code paths for long.

That sounds good to me! Shall we clean it up then? I mean, it's always going to be  present in git history if anyone ever is going to need it..

ReenigneArcher:
I'm fine with that, or handling it in a separate PR right after the next stable release.

I'm just a little worried about a big change like this with so few testers.

ABeltramo:
That's a valid concern; whilst the DualSense stuff is definitely experimental and lightly tested users can always force an xbox joypad from the config page and that's been in use for more than a year in Wolf. I'd assume that any big issue would have been reported by now..

ABeltramo:
I've tried building the Fedora image locally and it seems to work with the latest code, can anyone start the CI build just to confirm?

Hazer:
@ABeltramo @ReenigneArcher should we create a method to check the all the input status on startup/session start? This way at least we can log/show on the webui something to the user in an easy to digest way, I'm thinking we may avoid some reports.

Something like:
```cpp
enum DeviceStatus {
    UHID_NOT_AVAILABLE = 0,
    UINPUT_NOT_AVAILABLE,
    XINPUT_NOT_AVAILABLE,
    DEVICE_STATUS // Helper to indicate the number of status
};

std::bitset<DEVICE_STATUS> checkDeviceStatus() {
    std::bitset<DEVICE_STATUS> status;
    // TODO: Implement
    return status;
}

void printStatus() {
    auto status = checkDeviceStatus();
    if (status.test(UHID_NOT_AVAILABLE)) std::cout << "UHID not available [LINK_TO_DOCS]\n";
    if (status.test(UINPUT_NOT_AVAILABLE)) std::cout << "UINPUT not available [LINK_TO_DOCS]\n";
    if (status.test(XINPUT_NOT_AVAILABLE)) std::cout << "XINPUT not available [LINK_TO_DOCS]\n";
    if (!status.any()) std::cout << "All input devices available\n";
}
```

I think we don't have an easy way to do it right now, or I'm missing something and it's already implemented somewhere?

ABeltramo:
> @ABeltramo @ReenigneArcher should we create a method to check all the input status on startup/session start? This way at least we can log/show on the webui something to the user in an easy to digest way, I'm thinking we may avoid some reports.
> 
> I think we don't have an easy way to do it right now, or I'm missing something and it's already implemented somewhere?
> 

That should be implemented in the `supported_gamepads` method, we should try creating each joypad type and only fill the returned array with the ones that work; possibly logging the error message too.   
When that's properly implemented, everything else should flow automatically from the web UI to the Moonlight session; I can help you with checking and implementing that if you want!

ReenigneArcher:
All your CI are belong to us
- GitLab ... probabaly

https://gitlab.freedesktop.org/1

Hazer:
@ABeltramo That would be great, but I'm not entirely sure it should be checked right there, at least without refactoring something, as it is run during the creation/parsing of the config file, the config file parsing/creation should not care if the gamepad is actually enabled or not, and probably should allow all options that platform theoretically supports.

I'm more inclined to just log it in `platf::input()` at first, just for logging, then create a new API to better handle the web ui.

How about we improve it to something like the following before making it dynamic?
```cpp
struct supported_gamepad_t {
  std::string option_name;
  bool is_enabled;
  std::string reason_disabled_key; // or something else
}

std::vector<supported_gamepad_t> &supported_gamepads(bool check_status);
```

This way we can show the options in the UI, and if not available, we can make it disabled as a selection, but show a description of probable reasons it's not available to the user. While being able to avoid checking status on the config parsing step.

I can take the refactor, and provide the web ui, while you help me out with the `supported_gamepads` status check, if that's fine to you

ABeltramo:
Might be worth doing it in a separate PR?  Changing that interface will likely affect Windows/OSX as well.  
Happy to help nevertheless.

Hazer:
@ABeltramo can you review my changes in https://github.com/LizardByte/Sunshine/pull/2606/commits/eaa08e99a78e850c848c2e44d1da9315b8103b37?

Do they make sense to you or I'm completely off on it?

@ReenigneArcher can you test on windows for me?

src/platform/linux/input/inputtino_gamepad.cpp
```diff
@@ -237,12 +245,78 @@ namespace platf::gamepad {
...
-    static std::vector<std::string_view> gps {
-      "auto"sv, "xone"sv, "5"sv, "switch"sv
+  std::bitset<GamepadStatus::GAMEPAD_STATUS>
+  checkGamepadStatus() {
+    std::bitset<GamepadStatus::GAMEPAD_STATUS> status;
+    // Check for uhid device file
+    std::ifstream uhid("/dev/uhid");
+    if (!uhid.good()) {
+      status.set(GamepadStatus::UHID_NOT_AVAILABLE);
+    }
```
ABeltramo, line 255: This part is better handled by inputtino; I think it would be better to try to create a virtual device and then pass back (or log) the actual error:

```c++
auto ds5 = inputtino::PS5Joypad::create({ .name = "Sunshine DualSense (virtual) pad", .vendor_id = 0x054C, .product_id = 0x0CE6, .version = 0x8111 });
if(!ds5){
  ds5.getErrorMessage(); // TODO: do something with this?
}
```

It's basically the same code that is used in `alloc()`.

ABeltramo (commented):
I haven't checked the config page but I assume that everything will flow correctly down there: if a joypad is not available it shouldn't be showed as an option in the UI, right?

src/platform/linux/input/inputtino_gamepad.cpp
```diff
@@ -237,12 +245,78 @@ namespace platf::gamepad {
...
+      reason_disabled = "gamepads.uinput-not-available";
+    }
+    else {
+      reason_disabled = "gamepads.xinput-not-available";
+    }
+
+    static std::vector gps {
+      supported_gamepad_t { "auto", true, "" },
+      supported_gamepad_t { "xone", uxinput_available, reason_disabled },
+      supported_gamepad_t { "5", uhid_available, "gamepads.uhid-not-available" },
```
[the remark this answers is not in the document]
Hazer: I just used what was there before, but I'd like to change to something else, which one you think it's the best? ps5, ds5, DualSense?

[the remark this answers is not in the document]
ReenigneArcher: `ds5` is what's int he config UI, so it should be that.

src/platform/windows/input.cpp
```diff
@@ -1719,17 +1719,33 @@ namespace platf {
...
+    auto reason = enabled ? "" : "gamepads.vigem-not-available";
+
     // ds4 == ps4
-    static std::vector<std::string_view> gps {
-      "auto"sv, "x360"sv, "ds4"sv, "ps4"sv
+    static std::vector gps {
+      supported_gamepad_t { "auto", true, reason },
+      supported_gamepad_t { "x360", enabled, reason },
+      supported_gamepad_t { "ds4", enabled, reason },
+      supported_gamepad_t { "ps4", enabled, reason }
```
ReenigneArcher, line 1740:
```suggestion
```

Unless I'm misunderstanding something, `ps4` shouldn't be here. It was probably a left over from when the config was more relaxed and not managed by the web UI.

src/platform/linux/input/inputtino_gamepad.cpp
Hazer: @ABeltramo Even better! I don't need to manually reset after or do some cleanup, right? I'll update.

src/platform/linux/input/inputtino_gamepad.cpp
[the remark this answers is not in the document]
Hazer: @ABeltramo never mind, @ReenigneArcher answered already.

Hazer:
> I haven't checked the config page but I assume that everything will flow correctly down there: if a joypad is not available it shouldn't be showed as an option in the UI, right?

Not yet, this won't change the default behavior yet, mostly because of the config file, it will show all theoretically available options, but it will create logs with the errors, if any.

I'll open a new PR with the specific changes to the UI, using the changes I made here, and linking to the docs too. I felt like the extra logs where necessary, but the UI changes are better suited for another PR, as you pointed out.

src/platform/linux/input/inputtino_gamepad.cpp
Hazer: Updated.

ABeltramo:
@Hazer I've done a quick bunch of tests on the latest code and it looks good to me! I'd like to fixup the PS5 trackpad before merging this, just bear with me on that..  

Anyone that has got a drawing tablet to test that implementation as well? I have only tested it with unit tests using `libinput` but it would be nice if someone with a real device can give some feedback.

gschintgen:
> @Hazer I've done a quick bunch of tests on the latest code and it looks good to me! I'd like to fixup the PS5 trackpad before merging this, just bear with me on that..  
> 
> Anyone that has got a drawing tablet to test that implementation as well? I have only tested it with unit tests using `libinput` but it would be nice if someone with a real device can give some feedback.

I've got an old Thinkpad Yoga with built-in Wacom digitizer. That would be the Moonlight client device I suppose?
(I'm just not sure I can test it before next weekend.)

ABeltramo:
> I've got an old Thinkpad Yoga with built-in Wacom digitizer. That would be the Moonlight client device I suppose? (I'm just not sure I can test it before next weekend.)

I'm not sure if that's actually supported by Moonlight, but it would be great to test it!

gschintgen:
I was wondering about that didn't want to ask a stupid question 😇

So if there's a way to actually do a meaningful test (with the mentioned hardware), I'll be glad to help out.

gschintgen:
I'm now running your PR on my host (Ubuntu 22.04).

With moonlight 5.0.1 (6.0 is not yet available via snap; client is Ubuntu 22.04), pen input doesn't seem to work correctly:

Locally on the client everything is fine: I can hover with the pen over the display to move the pointer, I can tap the screen with the pen, in Gnome settings there's a Wacom tablet section where I can configure the stylus, I can draw, etc.
But on the host system, accessed via Moonlight, the pen simply works as a relative mouse input instead of absolute: if I initially position the pen on the client's screen, this position is transmitted correctly to the host and movement is fine too, but if I move it away from the display (so that the capacitive connection is lost) and reposition it somewhere else, then the relative movement of the pen is executed on the host system, but at the wrong absolute position.

Sunshine's logs confirm that the incoming packets are for "relative mouse move" events. So I guess this is not yet implemented in Moonlight. (Release notes for 6.0 don't mention pen support either.)

Anyway, if something else comes up (like issues with controllers) I'll let you know.

(BTW, with this PR the logs part of the Troubleshooting tab in the Web UI seems to be stuck at "Loading...". I never had that previously.)

src/main.cpp
```diff
@@ -272,6 +272,11 @@ main(int argc, char *argv[]) {
 
   reed_solomon_init();
   auto input_deinit_guard = input::init();
+
+  if (input::probe_gamepads()) {
```
ReenigneArcher, line 276: Do we re-probe before starting a session like we do for encoders?

ReenigneArcher (commented):
Are there any new run time deps we need to add here?

https://github.com/LizardByte/Sunshine/blob/master/cmake/packaging/linux.cmake#L29-L67

Or in our flatpak or arch linux manifests?

ABeltramo:
> Are there any new run time deps we need to add here?
> 
> https://github.com/LizardByte/Sunshine/blob/master/cmake/packaging/linux.cmake#L29-L67
> 
> Or in our flatpak or arch linux manifests?

No additional runtime library needed. In order to use the PS5 joypad it also needs access to `/dev/uhid` whilst `/dev/uinput` is required for all the other devices (this was already needed by Sunshine before this PR).

ABeltramo:
> I'm now running your PR on my host (Ubuntu 22.04).
> 
> With moonlight 5.0.1 (6.0 is not yet available via snap; client is Ubuntu 22.04), pen input doesn't seem to work correctly:
> 
> Locally on the client everything is fine: I can hover with the pen over the display to move the pointer, I can tap the screen with the pen, in Gnome settings there's a Wacom tablet section where I can configure the stylus, I can draw, etc. But on the host system, accessed via Moonlight, the pen simply works as a relative mouse input instead of absolute: if I initially position the pen on the client's screen, this position is transmitted correctly to the host and movement is fine too, but if I move it away from the display (so that the capacitive connection is lost) and reposition it somewhere else, then the relative movement of the pen is executed on the host system, but at the wrong absolute position.
> 
> Sunshine's logs confirm that the incoming packets are for "relative mouse move" events. So I guess this is not yet implemented in Moonlight. (Release notes for 6.0 don't mention pen support either.)
> 
> Anyway, if something else comes up (like issues with controllers) I'll let you know.
> 
> (BTW, with this PR the logs part of the Troubleshooting tab in the Web UI seems to be stuck at "Loading...". I never had that previously.)

Thanks for the very detailed report, `relative mouse move` means that the remote Moonlight instance picked it up as a mouse, unfortunately.. Maybe @cgutman can shed some light on which HW is currently supported as a drawing tablet?

ReenigneArcher:
@ABeltramo follow up question. Any changes (other than adjusted `udev` rules) needed to the setup process or post install scripts for `uhid`? Just want to make sure all bases are covered.

ABeltramo:
> @ABeltramo follow up question. Any changes (other than adjusted `udev` rules) needed to the setup process or post install scripts for `uhid`? Just want to make sure all bases are covered.

Nope, the user that runs Sunshine needs to be able to access the `uhid` device; so that single additional rule is really all we need.

ABeltramo:
Latest commit fixed PS5 touchpad support that now works the same as the local joypad.  
I think this is good to be merged! :+1:

ABeltramo (approved):

src/main.cpp
Hazer: @ReenigneArcher Not right now, just on start, but in another PR I want dynamically fetch this on the Web UI and show the current status to the user, I could probe on session too. Do you want the session probe in this PR or I could postpone?

ReenigneArcher: I think it can be later as I don't think it's any worse than what we're doing now (and probably not as important as encoder probing).

Hazer:
> (Release notes for 6.0 don't mention pen support either.)

There is a fix for pen in 6.0, that maybe is related: https://github.com/moonlight-stream/moonlight-qt/commit/b01a83ff3949c9ae42d75a1ad5c80c4fb9a529f8

The commit says Windows, but the code doesn't look platform specific. Worth a shot testing it when available.

VoxelTek:
Hi there,

Just tested out this branch, figured I might as well mention my experiences:

Switch Pro mode pretends to have a gyro, but doesn't pass through gyro input. Otherwise, it works fine.
DS5 mode worked fine after enabling the various settings, although my computer did complain that the virtual controller's battery was low. Aside from that, it worked fine, and so did gyro.

Not sure if it's specific to this branch, but if controller rumble occurs, the controller will then continue to rumble ever-so-slightly until client disconnect. I believe this happened in both Switch Pro and DS5 mode.

ABeltramo:
@VoxelTek thanks for checking this out, this is exactly the kind of report that I'm looking for!

> Switch Pro mode pretends to have a gyro, but doesn't pass through gyro input. 

This is to be expected, gyro and acc are currently only implemented for the DualSense.

> DS5 mode worked fine after enabling the various settings, although my computer did complain that the virtual controller's battery was low. Aside from that, it worked fine, and so did gyro.

I've just pushed a fix for the battery values and at least on my Plasma desktop I can see that it's reporting both charging/discharging with the value that Moonlight sends. I do have some discrepancy between the client (OSX btw) and what Sunshine receives so I'm not sure how reliable it is..


> Not sure if it's specific to this branch, but if controller rumble occurs, the controller will then continue to rumble ever-so-slightly until client disconnect. I believe this happened in both Switch Pro and DS5 mode.

I've pushed a fix for how rumble works on Xbox and Nintendo joypads, DualSense seems correct but I don't have a whole lot of testing material for this. I've tried both checking the values with `SDL` and with `fftest`, how are you testing rumble?

gschintgen:
I can confirm the issue with rumble keeping rumbling for quite some time. That's with a physical Xbox One controller emulated host-side as Xbox One controller. (Ubuntu on both sides; moonlight 5.0.1)

Tonight I'll have another go with the latest commits.

ABeltramo:
> I can confirm the issue with rumble keeping rumbling for quite some time. That's with a physical Xbox One controller emulated host-side as Xbox One controller. (Ubuntu on both sides; moonlight 5.0.1) Tonight I'll have another go with the latest commits.
> […](#)
> On June 13, 2024 6:53:25 PM GMT+02:00, ABeltramo ***@***.***> wrote: @VoxelTek thanks for checking this out, this is exactly the kind of report that I'm looking for! > Switch Pro mode pretends to have a gyro, but doesn't pass through gyro input. This is to be expected, gyro and acc are currently only implemented for the DualSense. > DS5 mode worked fine after enabling the various settings, although my computer did complain that the virtual controller's battery was low. Aside from that, it worked fine, and so did gyro. I've just pushed a fix for the battery values and at least on my Plasma desktop I can see that it's reporting both charging/discharging with the value that Moonlight sends. I do have some discrepancy between the client (OSX btw) and what Sunshine receives so I'm not sure how reliable it is.. > Not sure if it's specific to this branch, but if controller rumble occurs, the controller will then continue to rumble ever-so-slightly until client disconnect. I believe this happened in both Switch Pro and DS5 mode. I've pushed a fix for how rumble works on Xbox and Nintendo joypads, DualSense seems correct but I don't have a whole lot of testing material for this. I've tried both checking the values with `SDL` and with `fftest`, how are you testing rumble? -- Reply to this email directly or view it on GitHub: [#2606 (comment)](https://github.com/LizardByte/Sunshine/pull/2606#issuecomment-2166240707) You are receiving this because you were mentioned. Message ID: ***@***.***>

Thanks, let me know if latest commit fixes it! How are you testing it btw?

gschintgen:
I'm testing by just playing DiRT 3...

I don't think rumble was working previously (i.e. without this PR). At least I was always a bit surprised when rumble actually worked on my Xbox 360 controller on another client. I didn't care much though.

With this PR the controller became not only active but almost hyperactive. 😄

But the annoying thing is that it never really stopped. Let's say I drive over a dirt road. The controller (Xbox One) rumbles heavily and even after the course it will stay in that same rumble pattern. Even in the menus...

ABeltramo:
I can reproduce all rumble issues when playing a game, something that unit tests couldn't capture. I think I had the wrong mental model into how this [force feedback](https://www.kernel.org/doc/html/v4.13/input/ff.html) works in Linux..

VoxelTek:
> how are you testing rumble?

I've been loading up Half-Life 2 and playing from Route Kanal, and shortly after entering the game, there's a rumble event. After that, the rumble quiets down, but the controller continues to vibrate ever-so-slightly. I've tried with a few controllers on a few client platforms, and have experienced this issue on both.

ABeltramo:
Thank you ever so much for testing and reporting your issues! I've pushed a fix for rumble on all devices. I've tested both XOne and DS5 and they both seem correct now, please let me know if this fixes it for you too!

I have the feeling that the rumble effect is a bit too low, I've checked the raw values and they look correct so I'm not sure if it's specific to the Moonlight client, the client OS or just my imagination :sweat_smile: .   
I'd like to tune it up a little, but please let me know how it feels for your platform too!

gschintgen:
> please let me know if this fixes it for you too!

Unfortunately this commit kills rumble completely. (I also noticed that my previous comment was with a commit that was a few days old, so I can't say for sure that the culprit is the *very* latest commit, i.e. a66dd26, but it goes from strong and lingering rumble to no rumble.)

I double checked the in-game settings and even rebooted the client device. (Not the host though.)

There's nothing unusual in the logs at the default 'info' level. Would it be helpful to have it at 'debug'?

Let's see what VoxelTek has to say.

Unless specified otherwise my setup will be: Host Ubuntu Desktop 22.04, Client Moonlight 5.0.1 on Ubuntu Server 24.04 (to keep it minimal), physical and emulated controller: Xbox One, connected via Bluetooth dongle.

I also have a slight suspicion that this PR might cause weird performance issues. Previously my moonlight log was perfectly clean (everything wired, mini pc dedicated to moonlight without noteworthy background tasks), and now there are moderately frequent log entries for dropped frames, audio packet queue overflow, network dropped audio. I already checked with master and the logs were clean. I'll keep an eye on it over the next few days.

ABeltramo:
I'm going to run more tests tomorrow, just to make sure; I must say this feels very odd.. I'm not following what's going on master, but unless we are missing some important rebase, there's no way this PR will cause all those issues.  

Could it be that you aren't updating the git submodules? Are you manually compiling and running the binary as is or making some package out of this like flatpak?

gschintgen:
I'm a bit at a loss here. I can no longer get any rumble out of the controller (via Moonlight; I suppose the hardware is fine).

I rebooted host and client. I enabled uhid (even though I'm not emulating DS5.). I recompiled various commits of this PR. Nothing.

*Maybe* I'm doing something wrong when it comes to the build process (with all the various libraries and submodules it's pulling in.) I'll check again over the next few days. A working build from the CI would appreciated :-)

gschintgen:
> I'm going to run more tests tomorrow, just to make sure; I must say this feels very odd.. I'm not following what's going on master, but unless we are missing some important rebase, there's no way this PR will cause all those issues.
> 
> Could it be that you aren't updating the git submodules? Are you manually compiling and running the binary as is or making some package out of this like flatpak?

I configured git to always fetch recursively, but I'm not a git expert by any means and it is my first contact with a project that makes extensive use of submodules. I'm building locally:
```
Sunshine/build$ cmake .. && make -j12 && cpack -G DEB
```
Over the last few months I did quite a few builds, so I'm fairly confident with it, except when it comes to submodules. (E.g. I didn't know that fetching recursively isn't enough in case a commit or a branch pulls in an entirely new submodule. There's some init to do in that case.)

ABeltramo:
I'm not an expert myself on submodules but I think it's fairly easy to mistakenly automatically update them all to the latest commit instead to what the repo actually points at.   
Usually, `git status` should tell you if you have any difference in commit hash between what you have pulled locally and what the current commit is pointing at.

It might be better to hold off until a CI run is built so that we can both test the same exact binary, @ReenigneArcher or @Hazer could you guys start a run with the latest?

gschintgen:
I think it's the controller that is messed up. I had that already after my first try with this PR. It also rambled on and on and afterwards it no longer rumbled, but it was late and the next day it was fine again, so I shrugged it off.

I'm now trying to get it to rumble (at all) with a CLI tool (fftest) on the client device but without moonlight being involved. Since that didn't work out (evtest showed gamepad input, but fftest couldn't make it rumble), I un-paired it from the bluetooth dongle. And now the pesky controller won't pair again... Anyway, I'll figure it out.

I'm still very much looking forward to having all those nice features provided by this PR :-)

ReenigneArcher:
@ABeltramo can you provide a way to skip looking for libevdev?

In one of our builds we need to use libevdev as an external project from source.

https://github.com/LizardByte/Sunshine/actions/runs/9524228182/job/26256861152?pr=2606#step:6:5332

We use the following to find or get libevdev: https://github.com/LizardByte/Sunshine/blob/master/cmake/dependencies/libevdev_Sunshine.cmake

Then if we are using it from an external project, I tried to override `pkg_check_module` in later calls with this macro: https://github.com/LizardByte/Sunshine/blob/7e4c59444e340e4b28e151999de376361f01c1e4/cmake/macros/common.cmake#L32

But that doesn't seem to work for inputtino and libevdev. The `find_package` override works for simple-web-server and boost so I'm not sure what the issue is.

gschintgen:
I can confirm that the lingering rumble issue is fixed!

Tested with 19d11aebc9806 (.deb build artifact downloaded from the CI).

I also felt the effects were a bit weak, but that may be game specific. The rumble effects feel similar when comparing `fftest` executed directly on the client PC where the controller is paired and `fftest` executed in a terminal on the host while streaming the desktop.

(To fix my controller issue I had to repair after purging bluez and .... change the batteries!)

ABeltramo:
> I can confirm that the lingering rumble issue is fixed!
> 
> Tested with 19d11aebc9806 (.deb build artifact downloaded from the CI).
> 
> I also felt the effects were a bit weak, but that may be game specific. The rumble effects feel similar when comparing `fftest` executed directly on the client PC where the controller is paired and `fftest` executed in a terminal on the host while streaming the desktop.
> 
> (To fix my controller issue I had to repair after purging bluez and .... change the batteries!)

Thanks for confirming, I'm glad it works for you too now!

VoxelTek:
Hey! Just did a test of the latest version. I can confirm, the rumble issue is indeed fixed, however, the "low battery" message still appears.
image

ABeltramo:
> Hey! Just did a test of the latest version. I can confirm, the rumble issue is indeed fixed, however, the "low battery" message still appears. image

Thanks for testing it out! How far off is the battery from what is reported in the client? I'm not sure how reliable is Moonlight (SDL I guess?) in getting that value.. It would be nice to see if it's the Moonlight control packet reporting that or something wrong in the virtual device in Sunshine.

ReenigneArcher:
@ABeltramo it looks like the percentage we give is divided by 10. Is that the issue? https://github.com/games-on-whales/inputtino/blob/2739465690c7bbd1a27cb4e285ff08f486a208e3/src/uhid/joypad_ps5.cpp#L383

VoxelTek:
> Thanks for testing it out! How far off is the battery from what is reported in the client? I'm not sure how reliable is Moonlight (SDL I guess?) in getting that value.. It would be nice to see if it's the Moonlight control packet reporting that or something wrong in the virtual device in Sunshine.
@ABeltramo 

I'll have to do more investigation, I've been primarily testing using a modded Switch running a Moonlight client. I have tried with other proper controllers, however, and I do believe that those also presented the "low battery" message. I'll double-check in a bit. 

When I received those errors, the Switch was most certainly on somewhere around 80-90%, or more.

cgutman:
> Maybe @cgutman can shed some light on which HW is currently supported as a drawing tablet?

SDL2 doesn't have proper pen support, so you'd need to test on Android or iOS using a stylus. Those should support all of the fancy pen features (pressure, tilt, hover, buttons) assuming your hardware has them.

PR #2715, opened 2024-06-17
fix(linux/input): battery values conversion

ABeltramo:
## Description
A quick fix after this report https://github.com/LizardByte/Sunshine/pull/2606#issuecomment-2171961026

What is happening there is that the Moonlight Switch client doesn't support battery, in absence of a value the virtual pad state defaults to `0` which causes the pad to be seen with 0% battery.  

I've done two things:
 - set the default in inputtino to be `CHARGED`/`100%` so that we avoid annoying notifications for clients that don't report that value
 - adjusted the battery raw values, Moonlight sends them in the range `[0, 255]` whilst inputtino expects a percentage value `[0, 100]`

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/input/inputtino_gamepad.cpp
+++ b/src/platform/linux/input/inputtino_gamepad.cpp
@@ -258,7 +258,9 @@ namespace platf::gamepad {
           state = inputtino::PS5Joypad::BATTERY_FULL;
           break;
       }
-      std::get<inputtino::PS5Joypad>(*gamepad->joypad).set_battery(state, battery.percentage);
+      // Battery values in Moonlight are in the range [0, 0xFF (255)]
+      // Inputtino expects them as a percentage [0, 100]
+      std::get<inputtino::PS5Joypad>(*gamepad->joypad).set_battery(state, battery.percentage / 2.55);
     }
   }
 


src/platform/linux/input/inputtino_gamepad.cpp
```diff
@@ -258,7 +258,9 @@ namespace platf::gamepad {
           state = inputtino::PS5Joypad::BATTERY_FULL;
           break;
       }
-      std::get<inputtino::PS5Joypad>(*gamepad->joypad).set_battery(state, battery.percentage);
+      // Battery values in Moonlight are in the range [0, 0xFF (255)]
+      // Inputtino expects them as a percentage [0, 100]
+      std::get<inputtino::PS5Joypad>(*gamepad->joypad).set_battery(state, battery.percentage / 2.55);
```
ReenigneArcher, line 263: I think this could provide a double as the percentage... would that be an issue?

ABeltramo: Yep, it should then automatically cast it back to the nearest int since it's the type of the param of the method. Not sure if there's a better way, we should be fine since we don't really care for the decimals and there's no need for a precise measure..

PR #2733, opened 2024-06-20
fix(macos): fix boost on macos

Hazer:
## Description
macOS has some issues with Boost linking, this PR fixes the runtime error when statically linking boost, fixes the dynamic macOS build missing -licudata (icu4c) and improves brew installation options.

## Checklist
- [ ] Test macports installation
- [ ] Test brew installation
- [x] Test Clion build and run

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/cmake/compile_definitions/common.cmake
+++ b/cmake/compile_definitions/common.cmake
@@ -125,8 +125,8 @@ include_directories(
         "${CMAKE_SOURCE_DIR}/third-party/moonlight-common-c/enet/include"
         "${CMAKE_SOURCE_DIR}/third-party/nanors"
         "${CMAKE_SOURCE_DIR}/third-party/nanors/deps/obl"
-        ${Boost_INCLUDE_DIRS}
         ${FFMPEG_INCLUDE_DIRS}
+        ${Boost_INCLUDE_DIRS}  # has to be the last one or we runtime error on macOS ffmpeg encoder
 )
 
 list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
--- a/cmake/compile_definitions/common.cmake
+++ b/cmake/compile_definitions/common.cmake
@@ -126,7 +126,7 @@ include_directories(
         "${CMAKE_SOURCE_DIR}/third-party/nanors"
         "${CMAKE_SOURCE_DIR}/third-party/nanors/deps/obl"
         ${FFMPEG_INCLUDE_DIRS}
-        ${Boost_INCLUDE_DIRS}  # has to be the last one or we runtime error on macOS ffmpeg encoder
+        ${Boost_INCLUDE_DIRS}  # has to be the last, or we get runtime error on macOS ffmpeg encoder
 )
 
 list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
--- a/cmake/prep/options.cmake
+++ b/cmake/prep/options.cmake
@@ -15,7 +15,11 @@ option(SUNSHINE_REQUIRE_TRAY "Require system tray icon. Fail the build if tray r
 
 option(SUNSHINE_SYSTEM_WAYLAND_PROTOCOLS "Use system installation of wayland-protocols rather than the submodule." OFF)
 
-option(BOOST_USE_STATIC "Use static boost libraries." ON)
+if(APPLE)
+    option(BOOST_USE_STATIC "Use static boost libraries." OFF)
+else()
+    option(BOOST_USE_STATIC "Use static boost libraries." ON)
+endif()
 
 option(CUDA_INHERIT_COMPILE_OPTIONS
         "When building CUDA code, inherit compile options from the the main project. You may want to disable this if
--- a/packaging/macos/Portfile
+++ b/packaging/macos/Portfile
@@ -46,7 +46,8 @@ depends_test             port:doxygen \
 
 configure.args           -DBUILD_WERROR=ON \
                          -DCMAKE_INSTALL_PREFIX=${prefix} \
-                         -DSUNSHINE_ASSETS_DIR=etc/sunshine/assets
+                         -DSUNSHINE_ASSETS_DIR=etc/sunshine/assets \
+                         -DBOOST_USE_STATIC=ON
 
 configure.env-append     BRANCH=@GITHUB_BRANCH@
 configure.env-append     BUILD_VERSION=@BUILD_VERSION@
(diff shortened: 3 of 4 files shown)


cmake/compile_definitions/common.cmake
```diff
@@ -119,14 +119,20 @@ list(APPEND SUNSHINE_DEFINITIONS SUNSHINE_TRAY=${SUNSHINE_TRAY})
 
 include_directories("${CMAKE_SOURCE_DIR}")
 
+if(APPLE AND NOT BOOST_USE_STATIC AND NOT FETCH_CONTENT_BOOST_USED)
+    list(APPEND BOOST_INCLUDE ${PLATFORM_INCLUDE_DIRS})
```
ReenigneArcher, line 123:
```suggestion
    list(APPEND BOOST_INCLUDE ${Boost_INCLUDE_DIR})
```

cmake/compile_definitions/macos.cmake
```diff
@@ -15,6 +15,9 @@ endforeach()
 
 if(NOT BOOST_USE_STATIC AND NOT FETCH_CONTENT_BOOST_USED)
     ADD_DEFINITIONS(-DBOOST_LOG_DYN_LINK)
+
+    set(PLATFORM_INCLUDE_DIRS
+            ${Boost_INCLUDE_DIR})
```
ReenigneArcher, line 20:
```suggestion
```

packaging/sunshine.rb
```diff
@@ -53,6 +56,7 @@ def install
     ENV["BRANCH"] = "@GITHUB_BRANCH@"
     ENV["BUILD_VERSION"] = "@BUILD_VERSION@"
     ENV["COMMIT"] = "@GITHUB_COMMIT@"
+    ENV["LIBRARY_PATH"] = "/usr/local/opt/icu4c/lib"
```
ReenigneArcher, line 59: This location changes depending if it's arm or x86_64.

cmake/compile_definitions/common.cmake
```diff
@@ -119,14 +119,20 @@ list(APPEND SUNSHINE_DEFINITIONS SUNSHINE_TRAY=${SUNSHINE_TRAY})
...
+
 include_directories(
         SYSTEM
         "${CMAKE_SOURCE_DIR}/third-party"
         "${CMAKE_SOURCE_DIR}/third-party/moonlight-common-c/enet/include"
         "${CMAKE_SOURCE_DIR}/third-party/nanors"
         "${CMAKE_SOURCE_DIR}/third-party/nanors/deps/obl"
-        ${Boost_INCLUDE_DIRS}
         ${FFMPEG_INCLUDE_DIRS}
+        ${BOOST_INCLUDE}
```
ReenigneArcher, line 135:
```suggestion
        ${BOOST_INCLUDE}
        ${FFMPEG_INCLUDE_DIRS}
```

Alphabetize

packaging/sunshine.rb
Hazer: As far as my understanding goes, it does not. This is the path brew install icu4c as a keg-only library, it's not the system installed version. This should use the same architecture as the brew architecture installed.

I did not test this assumption, do you have any resource I could verify?

packaging/macos/Portfile
```diff
@@ -46,7 +46,8 @@ depends_test             port:doxygen \
 
 configure.args           -DBUILD_WERROR=ON \
                          -DCMAKE_INSTALL_PREFIX=${prefix} \
-                         -DSUNSHINE_ASSETS_DIR=etc/sunshine/assets
+                         -DSUNSHINE_ASSETS_DIR=etc/sunshine/assets \
+                         -DBOOST_USE_STATIC=ON
```
ReenigneArcher, line 50: Can you make these alphabetical?

packaging/sunshine.rb
ReenigneArcher: I mean the homebrew install location is different depending on the architecture.

These are the two install paths for homebrew... maybe you can work from that?

```
/opt/homebrew
/opt/local
```

Hazer: Oh, I see, I improved on that, now brew itself tells me the path

cmake/prep/options.cmake
```diff
@@ -15,7 +15,11 @@ option(SUNSHINE_REQUIRE_TRAY "Require system tray icon. Fail the build if tray r
 
 option(SUNSHINE_SYSTEM_WAYLAND_PROTOCOLS "Use system installation of wayland-protocols rather than the submodule." OFF)
 
-option(BOOST_USE_STATIC "Use static boost libraries." ON)
+if(APPLE)
+    option(BOOST_USE_STATIC "Use static boost libraries." OFF)
+else()
+    option(BOOST_USE_STATIC "Use static boost libraries." ON)
+endif()
```
ReenigneArcher, line 22: Are you still getting issues with `BOOST_USE_STATIC=ON`? If not we should revert this.

cmake/compile_definitions/common.cmake
```diff
@@ -125,8 +125,8 @@ include_directories(
         "${CMAKE_SOURCE_DIR}/third-party/moonlight-common-c/enet/include"
         "${CMAKE_SOURCE_DIR}/third-party/nanors"
         "${CMAKE_SOURCE_DIR}/third-party/nanors/deps/obl"
-        ${Boost_INCLUDE_DIRS}
         ${FFMPEG_INCLUDE_DIRS}
+        ${Boost_INCLUDE_DIRS}  # has to be the last one or we runtime error on macOS ffmpeg encoder
```
ReenigneArcher, line 129:
```suggestion
        ${Boost_INCLUDE_DIRS}  # has to be the last, or we get runtime error on macOS ffmpeg encoder
```

cmake/prep/options.cmake
Hazer: Without manually setting icu path, it won't build with ON, so to make easier to build locally for development, I think OFF is a good default for Clion starting setup. We could let be ON by default, but then we need to update the build docs with a step to set icu path and flags manually.

ReenigneArcher: Okay, I'm fine with it as is if we must.

Before we have different defaults though, could you investigate finding the icu path with CMake functions, such as `find_package`?

Hazer: I've tried a bit, but had no success, as icu4c is a keg-only formula, it's not widely available in the system, I plan on finding a way to do that, but I'd rather fix the build first for most users, then improve it, since I don't know how long it will take me to do it and I won't be able to build the project for a few weeks starting today.

ReenigneArcher (approved):

PR #2746, opened 2024-06-23
fix(upnp): Support miniupnpc API version 18 (release 2.2.8)

chewi:
## Description
The miniupnpc API has changed slightly. This allows both older and newer versions. The additional arguments are for returning data that Sunshine doesn't need to care about, so `nullptr, 0` is sufficient. I have already made the same change to a couple of other projects.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/upnp.cpp
+++ b/src/upnp.cpp
@@ -109,7 +109,11 @@ namespace upnp {
       IGDdatas data;
       urls_t urls;
       std::array<char, INET6_ADDRESS_STRLEN> lan_addr;
+#if (MINIUPNPC_API_VERSION >= 18)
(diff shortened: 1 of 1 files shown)


src/upnp.cpp
```diff
@@ -109,7 +109,11 @@ namespace upnp {
       IGDdatas data;
       urls_t urls;
       std::array<char, INET6_ADDRESS_STRLEN> lan_addr;
+#if (MINIUPNPC_API_VERSION >= 18)
```
Hazer, line 112: Can we create a function that returns the status, and does this if inside? I really don't like the duplication.

Hazer (changes requested):

PR #2760, opened 2024-06-26
Support recombined YUV 4:4:4 encoding (Prototype, Windows-only for now) 

ns6089:
## Description
The continuation of https://github.com/LizardByte/Sunshine/pull/2533. It's possible to emulate YUV 4:4:4 on gpus that don't support it natively by doubling the YUV 4:2:0 pixel count and running custom recombination shaders on both encoding and decoding side. Like Microsoft did it in [MS-RDPEGFX](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/8131c1bc-1af8-4907-a05a-f72f4581160f).

Early prototype stage, currently non-functional. Requires changes on moonlight's side. The recombination on sunshine's side can be done with pixel shaders, but moonlight more or less requires compute shaders. I currently have custom libplacebo shader implemented for plvk backend, in the future it should be possible to add Direct3D11 compute shader for d3d11va backend, and maybe some OpenGL shader too (not sure about this one).

---

## Type of Change

## Checklist

## Branch Updates


mirh:
Awesomely crazy. 
Could there be anything worth doing with an emulated 4:2:2 stream then? Like, I don't know, slightly lower recombination overhead, or lower bandwidth requirements?

Or perhaps not hitting encoding limits at higher resolutions. Like, is my understanding correct that this pixel doubling would not allow for 1440p on (say) older VCE versions that [max out](https://github.com/obsproject/obs-amd-encoder/wiki/Hardware-Support) at 4K?

ns6089:
I don't think anyone but Intel supports 4:2:2. About 4K limit, 1440p might still work depending on how exactly said limit is implemented, the overall pixel count stays within 4K range.

---
amd

mirh:
> I don't think anyone but Intel supports 4:2:2. 

To be honest, I was more thinking of TVs than computers here. It's a [mixed bag](https://forum.videohelp.com/threads/406912-4-2-2-channels-H-265-HEVC-samsung-tv-screen-black-%21/page2) even there, but still it's not so rare. 
But now that you mention pcs, decoding is much lighter on the cpu than encoding. I don't think that would usually be a deal breaker. Or nevertheless, couldn't the client-side recombination just fake to be 4:4:4 then? Or would whatever empty padding you add ruin the image more than the results you could get with just plain 4:2:0?

> About 4K limit, 1440p might still work depending on how exactly said limit is implemented, the overall pixel count stays within 4K range.

You mean if the limit is actually implemented like 4096x2160 (usual old amd) vs 4096x4096 (usual old nvidia)?
Or can you really call it a day just as long as the supported total pixel count, whatever the "shape", is 7.372.800  (2560x1440x2) or more?

ns6089:
You can't *encode* 4:2:2 on nvidia gpus, implementing a path exclusively for intel will be too expensive.
> Or can you really call it a day just as long as the supported total pixel count, whatever the "shape", is 7.372.800 (2560x1440x2) or more?

I'm already calling it a day :sunglasses:
Doubling one dimension allows to minimize discontinuities in motion estimation, in contrast to tiling. Current half-naive implementation for example has single motion estimation vertical "seam" in U and V planes.
```cpp
    //     Y       U     V
    // +-------+ +---+ +---+
    // |       | |   | |   |
    // |   Y   | |UR | |VR |
    // |       | |   | |   |
    // +---+---+ +---+ +---+
    // |   |   |
    // |UL |VL |
    // |   |   |
    // +---+---+
```

mirh:
> You can't encode 4:2:2 on nvidia gpus, implementing a path exclusively for intel will be too expensive.

You can't encode 4:4:4 on amd gpus either, and yet this is what this PR is about isn't it?

ns6089:
Personally, I don't see a point in supporting recombination into 4:2:2
It will still have visible artifacts while having computational overhead close to 4:4:4 and significant amount of additional development time. And this development time will be multiplied by the amount of distinct clients,

mirh:
I mean, sure, of course this is already miraculous. 
I was just trying to think outside the box (4:2:2 is still subpar, but even the worst case scenario starts to be [bearable](https://www.rtings.com/tv/learn/chroma-subsampling#when-does-it-matter) instead). 

If any I guess the improvement isn't that clear cut, because unlike with a direct cable connection it's not like there aren't already compression artifacts anyway. So if 4:4:4 couldn't fit in some whatever doubled 4:2:0 4K scenario, just lowering the resolution could also be a possible (and if not any easily immediate) alternative?

ns6089:
The code in this pull request is Not a Contribution under LizardByte Individual Contributor License Agreement.
The code in this pull request is shared under GNU GENERAL PUBLIC LICENSE Version 3.

The feature itself is completed on sunshine side.

PR #2786, opened 2024-07-01
macOS: New implementation of service publication

cathyjf:
The current implementation of service publication on macOS uses
`avahi-client`, but the majority of macOS machines do not have Avahi
installed because macOS provides a native alternative (`mDNSresponder`),
meaning that there is no reason to install Avahi.

The current implementation also attempts to load the Avahi client
libraries using `dlopen(3)`, which has a variety of restrictions on
macOS, such as only being willing to load from certain directories.
Depending on where the Avahi binaries are installed, they might not
be loadable through the current invocation of `dlopen(3)`.

Instead of using an Avahi client on macOS, it makes more sense to use
the native macOS API for publishing services via `mDNSresponder`. This
commit supplies such an implementation that uses the macOS native API.
It also has the advantage of being much simpler than the previous
implementation. Furthermore, this new implementation works on all
macOS machines, because it relies only on native APIs, rather than on
third-party software that is not commonly installed on macOS.

## Type of Change
- [X] Dependency update - *Removes a dependency on macOS (Avahi)*

## Checklist
  - My code uses the same style as the old implementation that I have replaced.
  - N/A

## Branch Updates

### DIFF
--- a/docs/source/building/macos.rst
+++ b/docs/source/building/macos.rst
@@ -12,7 +12,7 @@ MacPorts
 Install Requirements
    .. code-block:: bash
 
-      sudo port install avahi cmake curl doxygen graphviz libopus miniupnpc npm9 pkgconfig python311 py311-pip
+      sudo port install cmake curl doxygen graphviz libopus miniupnpc npm9 pkgconfig python311 py311-pip
 
 Homebrew
 """"""""
--- a/packaging/macos/Portfile
+++ b/packaging/macos/Portfile
@@ -38,8 +38,7 @@ depends_build-append     port:doxygen \
                          port:python311 \
                          port:py311-pip
 
-depends_lib              port:avahi \
-                         port:curl \
+depends_lib              port:curl \
                          port:libopus \
                          port:miniupnpc
 
--- a/packaging/sunshine.rb
+++ b/packaging/sunshine.rb
@@ -37,6 +37,7 @@ class @PROJECT_NAME@ < Formula
   depends_on "icu4c" => :recommended
 
   on_linux do
+    depends_on "avahi"
     depends_on "libcap"
     depends_on "libdrm"
     depends_on "libnotify"
--- a/src/platform/linux/publish.cpp
+++ b/src/platform/linux/publish.cpp
@@ -2,7 +2,6 @@
  * @file src/platform/linux/publish.cpp
  * @brief Definitions for publishing services on Linux.
  * @note Adapted from https://www.avahi.org/doxygen/html/client-publish-service_8c-example.html
- * @todo Use a common file for this and src/platform/macos/publish.cpp
  */
 #include <thread>
 
--- a/src/platform/macos/publish.cpp
+++ b/src/platform/macos/publish.cpp
@@ -1,442 +1,123 @@
 /**
  * @file src/platform/macos/publish.cpp
  * @brief Definitions for publishing services on macOS.
- * @note Adapted from https://www.avahi.org/doxygen/html/client-publish-service_8c-example.html
- * @todo Use a common file for this and src/platform/linux/publish.cpp
  */
+#include <dns_sd.h>
 #include <thread>
 
-#include "misc.h"
 #include "src/logging.h"
 #include "src/network.h"
 #include "src/nvhttp.h"
 #include "src/platform/common.h"
-#include "src/utility.h"
 
 using namespace std::literals;
 
-namespace avahi {
-
-  /**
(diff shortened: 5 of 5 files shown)


Hazer:
@cathyjf that's something I wanted to investigate in the future. Thank you for your PR, this great.

Can you remove avahi from the Portfile too?

ReenigneArcher:
- Can you add `avahi` here?

  https://github.com/LizardByte/Sunshine/blob/a940cdb394055139ca6a964289f414da562452e3/packaging/sunshine.rb#L39-L57

- > I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

  Why did you mark it n/a? Anything you can add documentation blocks to would be immensely helpful as I have a goal to get to a point where everything is documented. Reference: https://docs.lizardbyte.dev/projects/sunshine/en/master/source_code/source_code.html

- Can you name the anonymous namespace?

cathyjf:
> * Can you add `avahi` here?

Done.

> * Anything you can add documentation blocks to would be immensely helpful as I have a goal to get to a point where everything is documented.

I've now added a health additional amount of documentation.

> * Can you name the anonymous namespace?

The C++ Core Guidelines ("a collaborative effort led by Bjarne Stroustrup") state in [section SF.22](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities): "Use an unnamed (anonymous) namespace for all internal/non-exported entities". I agree with this guideline. The use of the anonymous namespace makes it clear that the items inside it are not part of the public interface of the file. Without using this idiom, it's hard to tell what is part of the interface and what is not. For example, for the GNU/Linux and Windows implementations of service publication, there are a ton of top-level functions, and it's not at all clear at a glance what is part of the interface and what is part of the implementation. In my new implementation, it's very clear, because everything is inside an anonymous namespace except for one function (`start`), which is the public interface.

I agree with you that something similar could be achieved by naming the anonymous namespace something like `_internal` or `_private`. However, then we would be relying on the name of the namespace to give a clue as to its purpose. By contrast, when we use an anonymous namespace, the compiler enforces that the items inside it have internal linkage and cannot be accessed outside of the translation unit (file). Therefore, I believe we should follow the established C++ idiom and leave this namespace unnamed. However, if you want to deviate from the common practice, I'm happy to change it -- just please confirm this in a follow-up reply.

ReenigneArcher:
Thank you for the updates and the explanation!

cathyjf:
I have fixed the lint error.

cathyjf:
I have now fixed the clang-format error, which unfortunately did not show up until after I fixed the cmake-lint error with the previous modification. That's why I didn't fix both at once.

ReenigneArcher:
Last question, do you have any ideas on how to unit test this?

Hazer:
I haven't got the time to actually code review, but running this branch locally I got the DNS working just fine with Android, Switch and QT clients, I'll do a proper code review later.

Hazer (approved):

cathyjf:
> Last question, do you have any ideas on how to unit test this?

It would be complicated to craft programmatic unit tests for this code. First, we would need to restructure the code so that where we currently directly call the macOS API functions (`DNSServiceRegister`, etc.), we would instead call some function objects that we define, so that, for testing purposes, the function objects can be replaced with fake versions. This restructuring would make the code more complicated. Next, we would need to actually implement fake versions of `DNSServiceRegister` and friends. The fake versions would not need to do any DNS operations, but they would need to interact with each other correctly, i.e., `DNSServiceRegister` would need to store the callback passed to it in some data structure; `DNSServiceRefSockFD` would need to return an actual file descriptor so that `select` can find that it's ready for reading (or alternatively we would need a fake version of `select` too); `DNSServiceProcessResult` would need to invoke the callback that was stored by `DNSServiceRegister`; and so on. Finally, we would then write some tests that store our fake versions of the functions inside the relevant function objects, before invoking the `start` function to test it.

After all that, we wouldn't really be able to test very much. We could write tests to verify that calling `start` causes the callback to eventually be called with the correct port. We could also write tests to ensure that failure tests don't crash and fail gracefully. The amount of functionality that we could verify with the tests is pretty limited.

The goal of writing codes for code is laudable but this isn't really a piece of code where it makes sense in my opinion to write unit tests, because the unit testing would be quite a bit more complicated than the code being tested, but not actually test very much.

ReenigneArcher:
This unit test failures are not related to this PR. I need to investigate what's going on.

PR #2803, opened 2024-07-04
Implement basic network flow control

ns6089:
## Description

Combat RX/TX buffer overflows and improve multi-FEC on large frames.

Adopted from https://github.com/LizardByte/Sunshine/pull/1466
Should supersede https://github.com/LizardByte/Sunshine/pull/2787

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -29,6 +29,8 @@ extern "C" {
 #include "thread_safe.h"
 #include "utility.h"
 
+#include "platform/common.h"
+
 #define IDX_START_A 0
 #define IDX_START_B 1
 #define IDX_INVALIDATE_REF_FRAMES 2
@@ -1246,12 +1248,28 @@ namespace stream {
     platf::adjust_thread_priority(platf::thread_priority_e::high);
 
     stat_trackers::min_max_avg_tracker<uint16_t> frame_processing_latency_tracker;
+
+    stat_trackers::min_max_avg_tracker<double> frame_send_batch_latency_tracker;
+    stat_trackers::min_max_avg_tracker<double> frame_fec_latency_tracker;
+    stat_trackers::min_max_avg_tracker<double> frame_network_latency_tracker;
+
     crypto::aes_t iv(12);
+
+    auto timer = platf::create_high_precision_timer();
+    if (!timer || !*timer) {
+      BOOST_LOG(error) << "Failed to create timer, aborting video broadcast thread";
+      return;
+    }
+
+    auto ratecontrol_next_frame_start = std::chrono::steady_clock::now();
 
     while (auto packet = packets->pop()) {
       if (shutdown_event->peek()) {
         break;
       }
+
+      auto frame_packet_start_time = std::chrono::steady_clock::now();
+      std::chrono::nanoseconds fec_time = 0ns;
 
       auto session = (session_t *) packet->channel_data;
       auto lowseq = session->video.lowseq;
@@ -1381,6 +1399,21 @@ namespace stream {
       }
 
       try {
+        // Use around 80% of 1Gbps          1Gbps            percent    ms     packet      byte
+        size_t ratecontrol_packets_in_1ms = std::giga::num * 80 / 100 / 1000 / blocksize / 8;
+
+        // Send less than 64K in a single batch.
+        // On Windows, batches above 64K seem to bypass SO_SNDBUF regardless of its size,
+        // appear in "Other I/O" and begin waiting for interrupts.
+        // This gives inconsistent performance so we'd rather avoid it.
+        size_t send_batch_size = 64 * 1024 / blocksize;
+
+        // Don't ignore the last ratecontrol group of the previous frame
+        auto ratecontrol_frame_start = std::max(ratecontrol_next_frame_start, std::chrono::steady_clock::now());
+
+        size_t ratecontrol_frame_packets_sent = 0;
+        size_t ratecontrol_group_packets_sent = 0;
+
         auto blockIndex = 0;
         std::for_each(fec_blocks_begin, fec_blocks_end, [&](std::string_view &current_payload) {
           auto packets = (current_payload.size() + (blocksize - 1)) / blocksize;
@@ -1404,9 +1437,26 @@ namespace stream {
             }
           }
 
+          auto fec_start = std::chrono::steady_clock::now();
+
           // If video encryption is enabled, we allocate space for the encryption header before each shard
           auto shards = fec::encode(current_payload, blocksize, fecPercentage, session->config.minRequiredFecPackets,
             session->video.cipher ? sizeof(video_packet_enc_prefix_t) : 0);
+
+          fec_time += std::chrono::steady_clock::now() - fec_start;
+
+          auto peer_address = session->video.peer.address();
+          auto batch_info = platf::batched_send_info_t {
+            nullptr,
+            shards.prefixsize + shards.blocksize,
+            0,
+            (uintptr_t) sock.native_handle(),
+            peer_address,
+            session->video.peer.port(),
+            session->localAddress,
+          };
+
+          size_t next_shard_to_send = 0;
 
           // set FEC info now that we know for sure what our percentage will be for this frame
           for (auto x = 0; x < shards.size(); ++x) {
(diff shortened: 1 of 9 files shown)


ns6089:
Basic implementation wasn't that hard, but it uncovered a few problems.
1. I will try to implement hybrid sleep/spin wait after all, 0.5ms timer precision should be barely enough.
2. On large bitrates our Reed-Solomon implementation takes significant amount of time. I will try moving it to separate thread, but in the long run it won't be enough.

@cgutman Have you considered alternative RS implementations that make use of SIMD instructions?
Like this one https://github.com/catid/leopard

cgutman:
> On large bitrates our Reed-Solomon implementation takes significant amount of time. I will try moving it to separate thread, but in the long run it won't be enough.

I don't think this needs to block merging this PR, unless you think these changes are going to make it worse somehow. 

> @cgutman Have you considered alternative RS implementations that make use of SIMD instructions?

`nanors` has SIMD optimizations too, but we don't really get much on x86_64 because it only guarantees SSE2. Bumping it to SSSE3 would enable the hand-optimized codepaths that should be much faster. Assuming we're willing to drop support for CPUs without SSSE3, it should just be as simple as adding `-mssse3` to https://github.com/LizardByte/Sunshine/blob/c92e4f27cd1ac3412f1a270e2208e87114995c7a/cmake/targets/common.cmake#L88-L90

If SSSE3 resolves the performance issues you saw at the bitrates we're targeting, that's probably enough. If we want to go further, we can use GCC/Clang function multiversioning to have the compiler build AVX2, SSSE3, and SSE2 variants, but that requires more extensive modifications. We could also switch to totally new library if necessary.

ns6089:
Yeah, enabling SSSE3 for nanors gives RS encode around 6 times speedup, bringing it to sub millisecond range. Don't need to bother with separate thread then, one less thing to worry about. We should absolutely enabled it.

cgutman:
Excellent, that's a pretty substantial win.

In terms of hardware support we're dropping with SSSE3, it looks like Intel CPUs prior to Core 2/Atom (~2006), AMD CPUs prior to Bulldozer (~2011), and VIA CPUs prior to Nano (~2008). It looks like all common x64 emulation layers like Rosetta 2 and XTA/Prism support SSSE3, so there should be no issues there either.

The only non-SSSE3 CPUs there that might otherwise be performant enough for _some_ modern games/applications would be AMD K10-based processors (Phenom II). However, those also lack AES-NI, so they're going to get punished by encryption too with ~20x more cycles/byte vs AES-NI.

Overall, I think bumping up the CPU requirement to SSSE3 is probably reasonable. Users with very old CPUs can use an older version of Sunshine.

ns6089:
@cgutman This generic low latency flow control doesn't seem to be working well, at least on Windows. `WSASendMsg()` in particular can take more than 5ms even when using extremely low batches (4 packets each), and that's with ethernet level (IEEE 802.3x) flow control disabled at link level. With a lot of calls the combined latency grows. Maybe we can outline all possible congestion scenarios and combat them on a case by case basis? I'm interested in what the aforementioned IEEE 802.3x can't cover. The first thing that comes to mind is streaming over the internet where WAN link is slower than LAN link, but we should be able to detect such cases and maybe apply higher latency throttling with 1ms packet batches.

ns6089:
> The first thing that comes to mind is streaming over the internet where WAN link is slower than LAN link, but we should be able to detect such cases and maybe apply higher latency throttling with 1ms packet batches.

Or try using QoS shaping https://learn.microsoft.com/en-us/windows/win32/api/qos2/ns-qos2-qos_flowrate_outgoing
In theory this exactly what we want in this scenario, if it works of course.

ns6089:
Oh, turns out my travel wifi router has atrocious ethernet switch and more or less requires 0.1ms batches. I guess this is the kind of devices we have problems with. Will try to optimize for it I guess, even though `WSASendMsg()` occasionally shoots up to 10ms even on such abysmal batches,

ns6089:
> Oh, turns out my travel wifi router has atrocious ethernet switch and more or less requires 0.1ms batches. I guess this is the kind of devices we have problems with. Will try to optimize for it I guess

Two things were necessary when dealing with this switch
1. Increasing udp socket send buffer size with `SO_SENDBUF`
2. Calling `WSASendMsg()` with small batches, 10 packets worked best

No sleep or pacing was needed, socket buffer took care of the congestion.
Streams easily at constant 300+Mbps now when before it couldn't even reach 100Mbps.

I guess I can try adding some throttling logic on top of this larger buffer. In theory it should be enough to appease slow clients without resorting to costly spin waits with periodic latency spikes.

ns6089:
Should be ready. Works really well on Windows from my tests, haven't tested Linux or MacOS.

@cgutman Maybe we can distinguish WAN streaming from LAN streaming based on the packet size? And apply more aggressive throttling in this case, based on requested bitrate. Something like 2 or 3 times the bitrate, will give 1/2 or 1/3 frame time latency.
https://github.com/LizardByte/Sunshine/pull/2803/commits/ea2b9a6bb41896f3fcc361381fe81509661d4475#diff-a8b463faa2b4901d7f646c0a3a856d705fd293dc7b99cee500a375dae91ef6bfR1402-R1403

ns6089:
Fixed the typo in linux/macos code, builds shouldn't fail now.

ns6089:
I also have the follow-up refactoring to periodic loggers that remove most of the syntactic bloat from them, improving the readability.
https://github.com/ns6089/Sunshine/commit/60e2c2f98e405e6912bcf009ad4e7fe3c6a2b836#diff-a8b463faa2b4901d7f646c0a3a856d705fd293dc7b99cee500a375dae91ef6bf
I can add it to this PR, if anyone prefers.

cmake/targets/common.cmake
```diff
@@ -87,7 +87,7 @@ set_source_files_properties("${CMAKE_SOURCE_DIR}/src/upnp.cpp"
 # third-party/nanors
 set_source_files_properties("${CMAKE_SOURCE_DIR}/third-party/nanors/rs.c"
         DIRECTORY "${CMAKE_SOURCE_DIR}" "${TEST_DIR}"
-        PROPERTIES COMPILE_FLAGS "-include deps/obl/autoshim.h -ftree-vectorize")
+        PROPERTIES COMPILE_FLAGS "-include deps/obl/autoshim.h -ftree-vectorize -mssse3 -funroll-loops")
```
cgutman, line 90: We should double-check that passing `-mssse3` won't break non-x86 builds (or only do it on x86).

@ReenigneArcher do we have an non-x86 builds in pre-merge CI?

src/stream.cpp
```diff
@@ -1438,37 +1498,89 @@ namespace stream {
...
+                  };
+
+                  platf::send(send_info);
+                }
+              }
+              if (config::sunshine.min_log_level <= 1) {
+                // Print send_batch() latency stats to debug log every 20 seconds
+                auto print_info = [&](double min_latency, double max_latency, double avg_latency) {
+                  auto f = stat_trackers::one_digit_after_decimal();
+                  BOOST_LOG(debug) << "Network: individial send_batch() latency (min/max/avg): " << f % min_latency << "ms/" << f % max_latency << "ms/" << f % avg_latency << "ms";
```
cgutman, line 1547: nit: `individual`

src/stream.cpp
```diff
@@ -1438,37 +1498,89 @@ namespace stream {
...
+                auto now = std::chrono::steady_clock::now();
+                if (now < due) {
+                  timer->sleep_for(due - now);
+                }
+
+                ratecontrol_group_packets_sent = 0;
+              }
+
+              size_t current_batch_size = x - next_shard_to_send + 1;
+              batch_info.buffer = shards.data(next_shard_to_send);
```
cgutman, line 1522: I was seeing corrupt packets when video encryption was enabled and I traced it to this line where we're forgetting the encryption header.

This should be:
```suggestion
              batch_info.buffer = shards.prefix(next_shard_to_send);
```

cgutman (changes requested):
Pacing behavior seems great in my testing on Windows and Linux. I'm seeing the expected behavior of densely packed FEC blocks without the nasty packet loss I had in my previous attempt.

Once these last few issues are resolved, this looks good to merge.

src/stream.cpp
ns6089: Fixed.

src/stream.cpp
ns6089: Fixed (that structure is confusing)

cmake/targets/common.cmake
ReenigneArcher: All arm builds are after merge due to the time required with qemu. I can build them locally though if needed.

ReenigneArcher: I'm having issues building on the ubuntu 22.04 and debian image, but the failure is not related to this as I cannot even dependencies.

```txt
Processing triggers for libc-bin (2.35-0ubuntu3.8) ...
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
Segmentation fault
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
Segmentation fault
dpkg: error processing package libc-bin (--configure):
installed libc-bin package post-installation script subprocess returned error exit status 139
Processing triggers for ca-certificates (20230311ubuntu0.22.04.1) ...
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
Errors were encountered while processing:
libc-bin
E:Sub-process /usr/bin/dpkg returned an error code (1)
```

I'm not sure what's going on with that.

ns6089: @cgutman Sounds good. SSSE3 change has its own commit https://github.com/LizardByte/Sunshine/pull/2803/commits/374485e244e2cee39494ce0dbc6d102aeb862fc2, so just exclude it during rebase. I can force-push it myself if you prefer that, but I rather dislike doing it after code review has already been done.

cgutman: OK, I'll drop that commit, squash the fixups, and get this merged. Thanks for tackling this one.

cgutman (approved):

PR #2820, opened 2024-07-07
fix(linux/input): don't pass unknown battery values

cgutman:
## Description
The `battery.percentage` value is a real percentage (`0..100`), but it has a special value of `LI_BATTERY_PERCENTAGE_UNKNOWN` (0xFF) to indicate the percentage is unknown. We shouldn't pass this value or the `LI_BATTERY_STATE_UNKNOWN` status into `inputino`.

cc: @ABeltramo 

### Issues Fixed or Closed
Fixes #2715

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/input/inputtino_gamepad.cpp
+++ b/src/platform/linux/input/inputtino_gamepad.cpp
@@ -261,10 +261,12 @@ namespace platf::gamepad {
         case LI_BATTERY_STATE_FULL:
           state = inputtino::PS5Joypad::BATTERY_FULL;
           break;
+        case LI_BATTERY_STATE_UNKNOWN:
+          return;
       }
-      // Battery values in Moonlight are in the range [0, 0xFF (255)]
-      // Inputtino expects them as a percentage [0, 100]
-      std::get<inputtino::PS5Joypad>(*gamepad->joypad).set_battery(state, battery.percentage / 2.55);
+      if (battery.percentage != LI_BATTERY_PERCENTAGE_UNKNOWN) {
+        std::get<inputtino::PS5Joypad>(*gamepad->joypad).set_battery(state, battery.percentage);
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/platform/linux/input/inputtino_gamepad.cpp
```diff
@@ -261,10 +261,12 @@ namespace platf::gamepad {
         case LI_BATTERY_STATE_FULL:
           state = inputtino::PS5Joypad::BATTERY_FULL;
           break;
+        case LI_BATTERY_STATE_UNKNOWN:
+          return;
```
ABeltramo, line 265: Re-looking at this, shouldn't we add `LI_BATTERY_STATE_NOT_PRESENT` and `LI_BATTERY_STATE_NOT_CHARGING` so that we cover all cases?
```c
case LI_BATTERY_STATE_NOT_CHARGING:
 state = inputtino::PS5Joypad::CHARGHING_ERROR;
 break
case LI_BATTERY_STATE_UNKNOWN:
case LI_BATTERY_STATE_NOT_PRESENT:
  return; // We can't set it, let's return
```

cgutman: `LI_BATTERY_STATE_NOT_PRESENT` makes sense, but I'm not sure about mapping `LI_BATTERY_STATE_NOT_CHARGING` to a charging error. We don't really have a specific value in the protocol to indicate that a device is actually broken. "Not charging" could simply indicate the device is holding at 80% or whatever to preserve the life of the battery cells.

ABeltramo: Right, makes sense. To be fair I'm not even sure what the actual meaning for `CHARGING_ERROR` originally is or how (if?) that will be picked up by clients, but I can see the subtle difference that you mention.

cgutman: Okay, I changed the logic here to just ignore status values that the code doesn't understand and those that don't have good mappings to inputino.

If there was a way to just report a percentage without a battery status, maybe we could support `LI_BATTERY_STATE_NOT_CHARGING`.

ReenigneArcher (approved):

PR #2825, opened 2024-07-08
Check API before trying to disable border on WGC capture

ns6089:
## Description

Using the new WGC capture method on Windows 10 currently causes a segmentation fault.
This should resolve it.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/windows/display_wgc.cpp
+++ b/src/platform/windows/display_wgc.cpp
@@ -11,6 +11,7 @@
 
 #include <windows.graphics.capture.interop.h>
 #include <winrt/windows.foundation.h>
+#include <winrt/windows.foundation.metadata.h>
 #include <winrt/windows.graphics.directx.direct3d11.h>
 
 namespace platf {
@@ -19,6 +20,7 @@ namespace platf {
 
 namespace winrt {
   using namespace Windows::Foundation;
+  using namespace Windows::Foundation::Metadata;
   using namespace Windows::Graphics::Capture;
   using namespace Windows::Graphics::DirectX::Direct3D11;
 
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Do you think we should just remove the whole try/catch, especially considering the catch isn't working anyway.

ns6089:
Force-pushed an update, now it looks like this 
```cpp
      return -1;
    }

    /*
    Currently this triggers a segmentation fault on Windows 10
    According to
    https://learn.microsoft.com/en-us/uwp/api/windows.graphics.capture.graphicscapturesession.isborderrequired
    we are expected to request user's consent first, but doing
    winrt::GraphicsCaptureAccess::RequestAccessAsync(winrt::GraphicsCaptureAccessKind::Borderless);
    accomplishes nothing, probably requiring an entry in app's manifest

    try {
      capture_session.IsBorderRequired(false);
    }
    catch (winrt::hresult_error &e) {
      BOOST_LOG(warning) << "Screen capture may not be fully supported on this device for this release of Windows: failed to disable border around capture area: [0x"sv << util::hex(e.code()).to_string_view() << ']';
    }
    */

    try {
      capture_session.StartCapture();
```

cgutman:
Is there a good reason to even allow WGC usage on Windows 10? I believe it was missing useful functionality like `GraphicsCaptureSession.IsCursorCaptureEnabled` and `GraphicsCaptureSession.IsBorderRequired` prior to Windows 11.

ns6089:
Well, some support is better than no support at all. And it performs rather well outside of those two(?) things.
I think we can query the presence of these properties with `winrt::Windows::Foundation::Metadata::ApiInformation::IsPropertyPresent()` and handle them in a way that doesn't hinder Windows 11. I will check it.

ns6089:
Yeah, checking for property presence resolves the segfault. Pushed a proper fix. Now we just need to confirm that the fix works alright on Windows 11.
`GraphicsCaptureSession.IsCursorCaptureEnabled` doesn't seem to be a problem on reasonably updated Windows 10, at least it functions correctly on my system.

ReenigneArcher:
> Now we just need to confirm that the fix works alright on Windows 11.

I tested, and it's working fine.

cgutman (approved):

PR #2873, opened 2024-07-16
Migrate audio pipeline to float from 16-bit integer

ns6089:
## Description

Float is the native format of opus codec and most if not all capture backends.

Implementation progress:
- [x] Windows
- [x] MacOS
- [x] Linux

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/audio.cpp
+++ b/src/platform/linux/audio.cpp
@@ -36,7 +36,7 @@ namespace platf {
   to_string(const char *name, const std::uint8_t *mapping, int channels) {
     std::stringstream ss;
 
-    ss << "rate=48000 sink_name="sv << name << " format=s16le channels="sv << channels << " channel_map="sv;
+    ss << "rate=48000 sink_name="sv << name << " format=float channels="sv << channels << " channel_map="sv;
     std::for_each_n(mapping, channels - 1, [&ss](std::uint8_t pos) {
       ss << pa_channel_position_to_string(position_mapping[pos]) << ',';
     });
@@ -54,12 +54,12 @@ namespace platf {
     util::safe_ptr<pa_simple, pa_simple_free> mic;
 
     capture_e
-    sample(std::vector<std::int16_t> &sample_buf) override {
+    sample(std::vector<float> &sample_buf) override {
       auto sample_size = sample_buf.size();
 
       auto buf = sample_buf.data();
       int status;
-      if (pa_simple_read(mic.get(), buf, sample_size * 2, &status)) {
+      if (pa_simple_read(mic.get(), buf, sample_size * sizeof(float), &status)) {
         BOOST_LOG(error) << "pa_simple_read() failed: "sv << pa_strerror(status);
 
         return capture_e::error;
@@ -73,7 +73,7 @@ namespace platf {
   microphone(const std::uint8_t *mapping, int channels, std::uint32_t sample_rate, std::uint32_t frame_size, std::string source_name) {
     auto mic = std::make_unique<mic_attr_t>();
 
-    pa_sample_spec ss { PA_SAMPLE_S16LE, sample_rate, (std::uint8_t) channels };
+    pa_sample_spec ss { PA_SAMPLE_FLOAT32, sample_rate, (std::uint8_t) channels };
     pa_channel_map pa_map;
 
     pa_map.channels = channels;
@@ -82,7 +82,8 @@ namespace platf {
     });
 
     pa_buffer_attr pa_attr = {};
-    pa_attr.maxlength = frame_size * 8;
+    pa_attr.fragsize = frame_size * channels * sizeof(float);
+    pa_attr.maxlength = pa_attr.fragsize * 2;
 
     int status;
 
--- a/src/platform/windows/audio.cpp
+++ b/src/platform/windows/audio.cpp
@@ -117,50 +117,46 @@ namespace {
     return waveformat;
   }
 
-  using virtual_device_waveformats_t = std::vector<WAVEFORMATEXTENSIBLE>;
+  using virtual_sink_waveformats_t = std::vector<WAVEFORMATEXTENSIBLE>;
 
   template <WORD channel_count>
-  virtual_device_waveformats_t
+  virtual_sink_waveformats_t
   create_virtual_sink_waveformats() {
-    switch (channel_count) {
-      case 2: {
-        auto channel_mask = waveformat_mask_stereo;
-        return {
-          create_waveformat(sample_format_e::f32, channel_count, channel_mask),
-          create_waveformat(sample_format_e::s32, channel_count, channel_mask),
-          create_waveformat(sample_format_e::s24in32, channel_count, channel_mask),
-          create_waveformat(sample_format_e::s24, channel_count, channel_mask),
-          create_waveformat(sample_format_e::s16, channel_count, channel_mask),
-        };
-      }
-
-      case 6: {
-        auto channel_mask1 = waveformat_mask_surround51_with_backspeakers;
-        auto channel_mask2 = waveformat_mask_surround51_with_sidespeakers;
-        return {
-          create_waveformat(sample_format_e::f32, channel_count, channel_mask1),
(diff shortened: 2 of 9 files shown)


cgutman (dismissed):
Successfully tested locally on macOS and Linux

src/platform/linux/audio.cpp
```diff
@@ -82,7 +82,8 @@
     });
 
     pa_buffer_attr pa_attr = {};
-    pa_attr.maxlength = frame_size * 8;
+    pa_attr.fragsize = frame_size * channels * sizeof(float);
```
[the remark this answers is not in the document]
ns6089: > Multiplication result may overflow 'unsigned int' before it is converted to 'unsigned long'. 

I feel like this bot is trolling us. And it probably doesn't have that much of AI inside just yet,,,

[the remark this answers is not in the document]
ns6089: For context
`pa_attr.fragsize` is `uint32_t`
`frame_size` is `uint32_t`
`channels` is `int`
`sizeof(float)` is `size_t`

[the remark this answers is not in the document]
ns6089: At least AI can regurgitate funny jokes, this is just bad static analysis.

Something is better than nothing I suppose, but if you want something actually useful sign up for coverity or pvs-studio, they should have open-source licenses. Those two have been around for a while and have proper engines, not simple clang wrappers.

I'm not going to change code to silence this warning since it's wrong and useless on multiple levels. Anyone else can modify the code after PR is merged of course.

psyke83:
This seems to break audio on my system by default.

Log is [here](https://gist.github.com/psyke83/21e664f5b9849d59e6954e635723b378); relevant lines:
```
[2024:07:19:19:51:47]: Error: Couldn't set Wave Format [0x88890008]
[2024:07:19:19:51:47]: Error: Unable to initialize audio capture. The stream will not have audio.
```

I can make audio work with your PR by disabling the Steam Streaming Speaker and Microphone devices and reconnecting the session; simply setting a real (non-Steam) endpoint to the default doesn't fix the issue, as the Steam virtual devices are always set to default when a new session is started.

ns6089:
Can reproduce, let's see what did I do wrong...

ns6089:
> Can reproduce, let's see what did I do wrong...

Messed with the function picks format for steam virtual speakers device.

@psyke83 Should work now, good catch

ns6089:
2024-07-20 11_33_08-Window

Wasn't me by the way, github did it automatically.
No new changes to Linux or MacOS code, the review is still valid.

ReenigneArcher:
> 2024-07-20 11_33_08-Window
> 
> Wasn't me by the way, github did it automatically. No new changes to Linux or MacOS code, the review is still valid.

Reviews get dismissed when the code changes. That a branch protection rule I have enabled.

psyke83:
> > Can reproduce, let's see what did I do wrong...
> 
> Messed with the function picks format for steam virtual speakers device.
> 
> @psyke83 Should work now, good catch

Confirmed working via Steam endpoints with latest changes on your PR branch - thanks.

src/platform/windows/audio.cpp
```diff
@@ -152,22 +152,22 @@ namespace platf::audio {
...
+    wave_format.SubFormat = KSDATAFORMAT_SUBTYPE_IEEE_FLOAT;
 
     return wave_format;
   }
 
   int
   set_wave_format(audio::wave_format_t &wave_format, const format_t &format) {
     wave_format->nSamplesPerSec = SAMPLE_RATE;
-    wave_format->wBitsPerSample = 16;
+    wave_format->wBitsPerSample = 24;
```
cgutman, line 170: I'm concerned we might need some additional logic for this. I don't know if it's guaranteed that all devices will support 24-bit PCM for all channel configurations. I think it would be best if we tried them in order: 32-bit -> 24-bit -> 16-bit

That would ensure we don't possibly regress any audio devices that are 16-bit only, and also provide the highest quality on devices with 32-bit support.

cgutman (changes requested):

src/platform/windows/audio.cpp
ns6089: It's used only for "virtual" sinks, which I thought only consisted of steam virtual speakers. Am I wrong in this assumption?
```cpp
/**
 * If the requested sink is a virtual sink, meaning no speakers attached to
 * the host, then we can seamlessly set the format to stereo and surround sound.
 *
 * Any virtual sink detected will be prefixed by:
 *    virtual-(format name)
 * If it doesn't contain that prefix, then the format will not be changed
 */
std::optional<std::wstring>
set_format(const std::string &sink) {
  auto sink_info = get_sink_info(sink);

  // If the sink isn't a device name, we'll assume it's a device ID
  auto wstring_device_id = find_device_id_by_name(sink).value_or(from_utf8(sink_info.second.data()));

  if (sink_info.first == format_t::none) {
    // wstring_device_id does not contain virtual-(format name)
    // It's a simple deviceId, just pass it back
    return std::make_optional(std::move(wstring_device_id));
  }

  wave_format_t wave_format;
  auto status = policy->GetMixFormat(wstring_device_id.c_str(), &wave_format);
  if (FAILED(status)) {
    BOOST_LOG(error) << "Couldn't acquire Wave Format [0x"sv << util::hex(status).to_string_view() << ']';

    return std::nullopt;
  }

  set_wave_format(wave_format, formats[(int) sink_info.first - 1]);

  WAVEFORMATEXTENSIBLE p {};
  status = policy->SetDeviceFormat(wstring_device_id.c_str(), wave_format.get(), (WAVEFORMATEX *) &p);

  // Surround 5.1 might contain side-{left, right} instead of speaker in the back
  // Try again with different speaker mask.
  if (status == 0x88890008 && sink_info.first == format_t::surr51) {
    set_wave_format(wave_format, surround_51_side_speakers);
    status = policy->SetDeviceFormat(wstring_device_id.c_str(), wave_format.get(), (WAVEFORMATEX *) &p);
  }
  if (FAILED(status)) {
    BOOST_LOG(error) << "Couldn't set Wave Format [0x"sv << util::hex(status).to_string_view() << ']';

    return std::nullopt;
  }

  return std::make_optional(std::move(wstring_device_id));
}
```

ReenigneArcher: > It's used only for "virtual" sinks, which I thought only consisted of steam virtual speakers. Am I wrong in this assumption?

Other virtual speakers can be used, such as virtual audio cable https://vb-audio.com/Cable/

ns6089: Alright, so sunshine can have default or custom audio sink, and default or custom virtual audio sink.
Will check how it can be handled.

cgutman: I think it could just be as simple as a loop that tries `SetDeviceFormat()` until it succeeds.

ns6089: I thought it would be a simple loop too, but then ended up with a nasty memory corruption. The rest of the code had accumulated quite a bit of technical dept. Either way it should be ready soon.

ns6089: Pushed and tested what I could.

cgutman (dismissed):

src/platform/windows/audio.cpp
```diff
@@ -859,15 +920,40 @@
         prop->GetValue(PKEY_DeviceInterface_FriendlyName, &adapter_friendly_name.prop);
         prop->GetValue(PKEY_Device_DeviceDesc, &device_desc.prop);
 
-        auto adapter_name = no_null((LPWSTR) adapter_friendly_name.prop.pszVal);
-        auto device_name = no_null((LPWSTR) device_friendly_name.prop.pszVal);
-        auto device_description = no_null((LPWSTR) device_desc.prop.pszVal);
+        auto adapter_name = (LPWSTR) adapter_friendly_name.prop.pszVal;
```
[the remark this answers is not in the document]
cgutman: I'll commit a fix for these and merge

PR #2885, opened 2024-07-17
feat: use output/display names for wlgrab

vmsh0:
## Description

TL;DR: This is a proof-of-concept PR for supporting using output names with the Wayland grabber, instead of arbitrarily-generated numeric ID as it is currently the standard in Sunshine.

Long version:

Currently, output IDs are assigned following this logic:

- get_registry() is called on the Wayland display, which causes the server to enumerate all global objects including wl_output objects (https://github.com/LizardByte/Sunshine/blob/eafb47238d92663851d264717f56609ffe92fdb9/src/platform/linux/wlgrab.cpp#L39)
- while receiving the responses, the callback (https://github.com/LizardByte/Sunshine/blob/1dd4b68e1cafbf134dd524cb7b7e1dbdfe686238/src/platform/linux/wayland.cpp#L175) singles out wl_output objects and saves the screen to an `std::vector`, implicitly assigning them numeric IDs in whatever order the Wayland server enumerates the objects

This is nondeterministic in two different ways:
- There is no guarantee that a Wayland server reuses the same objects IDs and enumeration order for two different clients (i.e. might be unstable across Sunshine restarts)
- There is no guarantee that across system reboots the Wayland server will enumerate outputs in the same order

This commit introduces support for indicating a [`wl_output`](https://wayland.app/protocols/wayland#wl_output)/[`zxdg_output_v1`](https://wayland.app/protocols/xdg-output-unstable-v1) (e.g. `eDP-1`, `HDMI-0` for wlroots based compositors) screen name in the `output_name` Sunshine parameter. I think this improves the above situation in the following ways:

- These names are explicitly specified in the interfaces to be stable within a Wayland session. While this is not as good as being stable across reboots (even though at a practical level this is also usually true, as these names are also used in compositor configuration files), at least it allows to reconfigure Sunshine on-the-fly before starting it, by updating its configuration file in a compositor-dependent manner if necessary.
- These same names are, in all Wayland compositors I have tried, the actual user-facing output names that are used in the tools. Thus, this allows users to configure Sunshine using the same output IDs that they use within other Wayland tools (e.g. wlr-randr, wf-recorder, compositor output configuration, etc), improving usability.
- At the same time, this PR retains compatibility with the previously-employed numeric IDs.

**Please note that this PR is a proof of concept**, mainly to kickstart a conversation on the feature. If the feature is deemed desirable, before merging it, some improvements that can be made to this PR are:
- The instructions on the web GUI should be updated to reflect this feature
- The documentation should report this feature
- It should be evaluated whether this is possible/feasible/desirable for other grabbers as well
- It should be evaluated whether to require using a prefix to indicate an output name instead of a numeric ID, as theoretically a numeric ID is also a valid Wayland output name (although I have never seen one which is entirely numeric)

### Screenshot

(not relevant)

### Issues Fixed or Closed

This is somewhat related to #2490.

## Type of Change

## Checklist

## Branch Updates

### DIFF
--- a/src/platform/linux/wlgrab.cpp
+++ b/src/platform/linux/wlgrab.cpp
@@ -50,13 +50,32 @@ namespace wl {
         return -1;
       }
 
+      // load output names, to allow matching by name
+      for (auto &monitor : interface.monitors) {
+        monitor->listen(interface.output_manager);
+      }
+
+      display.roundtrip();
+
       auto monitor = interface.monitors[0].get();
 
       if (!display_name.empty()) {
-        auto streamedMonitor = util::from_view(display_name);
-
-        if (streamedMonitor >= 0 && streamedMonitor < interface.monitors.size()) {
-          monitor = interface.monitors[streamedMonitor].get();
+        // if display_name is numeric, match a display number
+        if (display_name.find_first_not_of("0123456789") == std::string::npos) {
+          auto streamedMonitor = util::from_view(display_name);
+          if (streamedMonitor >= 0 && streamedMonitor < interface.monitors.size()) {
+            monitor = interface.monitors[streamedMonitor].get();
+          }
+        }
+        // otherwise, match a display name
+        else {
+          for (auto &candidate : interface.monitors) {
+            auto candidateMonitor = candidate.get();
+            if (candidateMonitor->name == display_name) {
+              monitor = candidateMonitor;
+              break;
+            }
+          }
         }
       }
 


vmsh0:
Are there any comments or feedback on this PR?

ReenigneArcher:
> Are there any comments or feedback on this PR?

I haven't even looked at it since it was marked draft.

vmsh0:
> > Are there any comments or feedback on this PR?
> 
> I haven't even looked at it since it was marked draft.

Fair enough :)

It's marked a draft because it doesn't include the necessary documentation effort, as before putting that effort in I would like some general feedback on the idea.

Is the project interested in the feature? Is my proposed implementation strategy ok? Etc

ReenigneArcher:
I think this is an improvement overall, but I say this as a non Linux user.

Also, any help on the Linux side is greatly appreciated. Most of the active devs are mostly focused on Windows.

This may affect #2490

vmsh0:
> This may affect #2490

Good catch.

The main improvement of #2490 is to add a "friendly name" to the screen, in addition to an ID, and the relevant UI changes. However, it does not change how ids are computed. My PR, instead, does change how IDs are picked: the actual Wayland IDs can now be used instead of the made-up Sunshine IDs (while retaining the ability to recognize them). This is also the case for the macOS implementation, even though unfortunately the native IDs are not as user-friendly as the ones on Linux/Wayland. While I do not have the chance to perform testing on macOS, I can assume that the IDs are stable across applications and perhaps reboots, 

Thus, this does to some degree interact with #2490, which perhaps should be merged first, and then this PR should be reworked to build upon the respective modified data structures to reimplement this functionality. Similar work could probably be carried out for other platforms, as I think this is in principle very useful and makes configuring Sunshine more intuitive and stable.

To abstract a bit, I think the interaction between these two PRs underlines the following design directions:
1. Should Sunshine, in general, prefer native IDs when available, instead of the arbitrary numeric IDs that are generated now? This is the main design decision. As I have argued, I think it should.
2. Should Sunshine support matching screens by criteria other than an ID -- either made-up or OS-generated? E.g. by resolution, refresh rate, type of physical port, virtual viewport/main screen status, etc? I.e. should the principle of having a more native and intuitive way to select a screen be brought even further?

I would be willing to contribute more work towards both the above directions, particularly on Linux (and perhaps Windows if no one else is interested in picking this up), if deemed worthwhile. However, I think that such structural decisions require a broader discussion and much more feedback. Is this the appropriate place to have such a discussion, or should I post this somewhere else?

ReenigneArcher:
> Should Sunshine, in general, prefer native IDs when available, instead of the arbitrary numeric IDs that are generated now? This is the main design decision. As I have argued, I think it should.

Yes. There's also this for Windows: https://github.com/LizardByte/Sunshine/pull/2894 which uses https://github.com/LizardByte/libdisplaydevice ... currently that library is Windows only, but it could be extended to macOS and Linux as well. Linux probably being the more difficult one due to so many options... not saying this work needs to move to that library (but you can contribute there if you want), just pointing out that we are moving to more stable IDs for Windows as well. e.g. `output_name = \\.\DISPLAY1` becomes `output_name = {daeac860-f4db-5208-b1f5-cf59444fb768}`.

> Should Sunshine support matching screens by criteria other than an ID -- either made-up or OS-generated? E.g. by resolution, refresh rate, type of physical port, virtual viewport/main screen status, etc? I.e. should the principle of having a more native and intuitive way to select a screen be brought even further?

I'm not sure on that. It sounds like it would be complex and probably break easily. I think moving to a more stable ID is sufficient for now.

This has also been one of the longest standing issues with xorg: https://github.com/LizardByte/Sunshine/issues/221

ReenigneArcher:
@vmsh0 I'm really sorry for the delay on this. I'd like to get this feature in, if you're still interested in working on it.

Dregu:
This doesn't seem to work, at least after merging master to get it to build at all. Maybe it worked perfectly before, but now it only works initially when starting sunshine, it picks the correct device with `output_name = HEADLESS-1` ("Selected Headless output 2 for streaming"), but later when starting the stream something has maybe sanitized that string to "0" again and it always picks the primary monitor. I only checked that it's indeed being called with "0" the second time around, but not why.

I like this idea though and would like to see it added.

vmsh0:
Hi,
As far as I can remember, it did work for me :) but it has been half a year since this PR was submitted.

Thank you for the feedback and sorry for my lack of response. I am still interested in working on this because I would like Sunshine to become a part of my permanent desktop stack, and this feature is very important to me. I will try to pick up everything in the next few weeks.

sg-qwt:
I tried this PR with sway. Doesn't seem to work even on first start.
In my conf settings, I put `output_name = HEADLESS-1`
Somehow it's still pick my main monitor instead of the configured one:
```
[2025-02-23 04:40:25.194]: Info: -------- Start of Wayland monitor list --------
[2025-02-23 04:40:25.194]: Info: Monitor 0 is eDP-1: California Institute of Technology 0x1303 (eDP-1)
[2025-02-23 04:40:25.194]: Info: Monitor 1 is DP-7: Dell Inc. DELL U2718QM MYPFK89J15HL (DP-7)
[2025-02-23 04:40:25.194]: Info: Monitor 2 is HEADLESS-1: Headless output 2
[2025-02-23 04:40:25.194]: Info: --------- End of Wayland monitor list ---------
[2025-02-23 04:40:25.195]: Info: Selected monitor [California Institute of Technology 0x1303 (eDP-1)] for streaming
```

ReenigneArcher:
@sg-qwt We don't use that format for `output_name` anymore. https://docs.lizardbyte.dev/projects/sunshine/latest/md_docs_2configuration.html#output_name ... not sure if this PR/branch has been that change in it or not though.

sg-qwt:
> @sg-qwt We don't use that format for `output_name` anymore. https://docs.lizardbyte.dev/projects/sunshine/latest/md_docs_2configuration.html#output_name ... not sure if this PR/branch has been that change in it or not though.


Thanks, put display number to output_name do work as expected(output_name = 2), but the whole point of this PR is to support deterministic output_name  (e.g. eDP-1, HDMI-0 for wlroots based compositors ) , which is what I try to test with this PR.

ReenigneArcher:
This PR was opened before we switched to using a display "id" instead of display name, so in order to consider this PR it will need to be updated to use the new pattern.

PR #2939, opened 2024-07-31
tests(windows): prefer static libraries

ns6089:
Resolves a couple of errors specific to msys.

## Type of Change

## Checklist

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -72,7 +72,7 @@ target_link_options(${PROJECT_NAME} PRIVATE)
 
 if (WIN32)
     # prefer static libraries since we're linking statically
-    # this fixes gtest_main and libcurl linking errors
+    # this fixes gtest_main and libcurl linking errors, when using non MSYS2 version of CMake
     set_target_properties(${PROJECT_NAME} PROPERTIES LINK_SEARCH_START_STATIC 1)
 endif ()
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -70,4 +70,10 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC ${SUNSHINE_DEFINITIONS} ${TEST
 target_compile_options(${PROJECT_NAME} PRIVATE $<$<COMPILE_LANGUAGE:CXX>:${SUNSHINE_COMPILE_OPTIONS}>;$<$<COMPILE_LANGUAGE:CUDA>:${SUNSHINE_COMPILE_OPTIONS_CUDA};-std=c++17>)  # cmake-lint: disable=C0301
 target_link_options(${PROJECT_NAME} PRIVATE)
 
+if (WIN32)
+    # prefer static libraries since we're linking statically
+    # this fixes gtest_main and libcurl linking errors
+    set_target_properties(${PROJECT_NAME} PROPERTIES LINK_SEARCH_START_STATIC 1)
+endif ()
+
 add_test(NAME ${PROJECT_NAME} COMMAND sunshine_test)


tests/CMakeLists.txt
```diff
@@ -70,4 +70,10 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC ${SUNSHINE_DEFINITIONS} ${TEST
 target_compile_options(${PROJECT_NAME} PRIVATE $<$<COMPILE_LANGUAGE:CXX>:${SUNSHINE_COMPILE_OPTIONS}>;$<$<COMPILE_LANGUAGE:CUDA>:${SUNSHINE_COMPILE_OPTIONS_CUDA};-std=c++17>)  # cmake-lint: disable=C0301
 target_link_options(${PROJECT_NAME} PRIVATE)
 
+if (WIN32)
+    # prefer static libraries since we're linking statically
+    # this fixes gtest_main and libcurl linking errors
+    set_target_properties(${PROJECT_NAME} PROPERTIES LINK_SEARCH_START_STATIC 1)
```
ReenigneArcher, line 76: It seems this is more of a vscode condition than a Windows condition?

Could this be an issue in other OSes with vscode?

ns6089: It's msys+cmake condition and on Windows we don't use anything else.
vscode simply calls cmake with various arguments.
Try to use outside cmake with msys in cmake kit and gtest_main will fail to link properly.
And I'm still half-convinced you have dynamically linked libcurl even in clion.

ns6089: About other OSes, I don't think we build sunshine target with `LINK_SEARCH_START_STATIC` anywhere else so in theory we should be fine.

ReenigneArcher: We static link boost by default everywhere now. Got tired of Sunshine breaking when package managers update the boost version: https://github.com/LizardByte/Sunshine/blob/444ca98cbfd4ab5da5b921f6b45cf21968c80b4e/cmake/dependencies/Boost_Sunshine.cmake#L14

Curl "should" be static: https://github.com/LizardByte/Sunshine/blob/444ca98cbfd4ab5da5b921f6b45cf21968c80b4e/cmake/compile_definitions/windows.cmake#L16 ... but that seems to be broken with the updated version of curl.

I will run some tests tonight to see if I can replicate the behavior.

ns6089: ### LINK_SEARCH_START_STATIC[¶](https://cmake.org/cmake/help/latest/prop_tgt/LINK_SEARCH_START_STATIC.html#link-search-start-static)
Assume the linker looks for static libraries by default.
Some linkers support switches such as -Bstatic and -Bdynamic to determine whether to use static or shared libraries for -lXXX options. CMake uses these options to set the link type for libraries whose full paths are not known or (in some cases) are in implicit link directories for the platform. By default the linker search type is assumed to be -Bdynamic at the beginning of the library list. This property switches the assumption to -Bstatic. It is intended for use when linking an executable statically (e.g. with the GNU -static option).

---

May be UCRT toolchain quirk. Honestly, I don't think we should partake in this -Bstatic/-Bdynamic guessing game since we can never be sure of the result.

FrogTheFrog: According to this https://stackoverflow.com/a/68351329 we also need an end variant?

Also it says that:
`It is intended for use when linking an executable statically (e.g. with the GNU -static option).`

So maybe add a compiler guard for to exclude MSVC?

tests/CMakeLists.txt
```diff
@@ -70,4 +70,10 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC ${SUNSHINE_DEFINITIONS} ${TEST
 target_compile_options(${PROJECT_NAME} PRIVATE $<$<COMPILE_LANGUAGE:CXX>:${SUNSHINE_COMPILE_OPTIONS}>;$<$<COMPILE_LANGUAGE:CUDA>:${SUNSHINE_COMPILE_OPTIONS_CUDA};-std=c++17>)  # cmake-lint: disable=C0301
 target_link_options(${PROJECT_NAME} PRIVATE)
 
+if (WIN32)
+    # prefer static libraries since we're linking statically
+    # this fixes gtest_main and libcurl linking errors
```
ReenigneArcher, line 75:
```suggestion
    # this fixes gtest_main and libcurl linking errors, when using non MSYS2 version of CMake
```

tests/CMakeLists.txt
ns6089: Can we worry about it when(if) we start supporting dynamic linking in MSVC? I can't test hypothetical scenarios. But I can state with confidence that right now tests don't work without going through additional hoops and substituting default cmake.

FrogTheFrog: > Can we worry about it when(if) we start supporting dynamic linking in MSVC? I can't test hypothetical scenarios. But I can state with confidence that right now tests don't work without going through additional hoops and substituting default cmake.

Fine with me, but what about `LINK_SEARCH_END_STATIC`?

ns6089: It's not necessary and we don't do it for sunshine.exe. LINK_SEARCH_START_STATIC is enough to remove all appearances of -Bstatic/-Bdynamic from the makefile.

ns6089: Someone can test it of course, but I don't see the point.

FrogTheFrog: LGTM in that case

PR #2963, opened 2024-08-04
feat(logging): include milliseconds in timestamps

ns6089:
## Description

Include milliseconds in logging timestamps and improve formatting while at it.
Should now give better idea how long individual steps take (shader compilation, encoder probing, etc).

### Screenshot

Before
```
[2024:08:04:22:33:15]: Info: Sunshine version: 0.0.0.f8b0be1e
```
After
```
[2024-08-04 22:33:15.623] Info: Sunshine version: 0.0.0.f8b0be1e
```

## Type of Change

## Checklist

### DIFF
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -4,6 +4,7 @@
  */
 // standard includes
 #include <fstream>
+#include <iomanip>
 #include <iostream>
 
 // lib includes
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -8,6 +8,7 @@
 
 // lib includes
 #include <boost/core/null_deleter.hpp>
+#include <boost/format.hpp>
 #include <boost/log/attributes/clock.hpp>
 #include <boost/log/common.hpp>
 #include <boost/log/expressions.hpp>
@@ -67,7 +68,6 @@ namespace logging {
     sink->set_formatter([](const bl::record_view &view, bl::formatting_ostream &os) {
       constexpr const char *message = "Message";
       constexpr const char *severity = "Severity";
-      constexpr int DATE_BUFFER_SIZE = 21 + 2 + 1;  // Full string plus ": \0"
 
       auto log_level = view.attribute_values()[severity].extract<int>().get();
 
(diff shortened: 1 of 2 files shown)


ReenigneArcher:
If we're going to adjust the logging format, can we improve the readability a little more?

I'm not an expert on boost logging, but can we do something like this?

`2024-08-04 22:33:15.623 - INFO :: <module/file that called the message... or the thread> : <the message>`

ns6089:
It's [possible](https://stackoverflow.com/a/33151310), but in my opinion we shouldn't.

It will
- add quite a bit of informational bloat
- break message alignment
- be entirely useless for users 
- be not very useful for us (copy-pasting and searching the codebase for message keywords is near instantaneous and hasn't failed me once for sunshine)

So yeah. Small changes that objectively improve the situation (conventional date separating symbols, removing needless ':') I can do. Large subjective changes, I'd rather not.

ReenigneArcher:
I agree with your points, I was just thinking it would allow us to not have to search and we could go directly the point in the code... I guess it's not incredibly value added, unless the message is from outside of Sunshine's codebase.

I would like the log format to be like this though if you could.

`2024-08-04 22:33:15.623 - INFO :: <the message>`

ns6089:
> I would like the log format to be like this though if you could.
`2024-08-04 22:33:15.623 - INFO :: <the message>`

No strong opinion on this, but I don't want it to be traced to my commit since I have nothing to do with the decision.

I've updated this PR leaving only the timestamp change. The rest can be done outside without any deep c++ knowledge.
https://github.com/ns6089/Sunshine/blob/4dc34b0e670cbad5577cd2bce8d83b2a5e9a833a/src/logging.cpp#L77-L92
https://github.com/ns6089/Sunshine/blob/4dc34b0e670cbad5577cd2bce8d83b2a5e9a833a/src/logging.cpp#L103
https://github.com/ns6089/Sunshine/blob/4dc34b0e670cbad5577cd2bce8d83b2a5e9a833a/src_assets/common/assets/web/index.html#L145

ns6089:
Added an #include statement. It should fix failing builds on some platforms (hopefully).

PR #2973, opened 2024-08-07
ci(brew): add prerelease variant to tap

Hazer:
## Description
The proposal here is to publish a prerelease variant to our homebrew tap, so we can then install it as:

```sh
brew tap LizardByte/homebrew
brew install sunshine-prerelease
```

I'll probably need help here @ReenigneArcher, I'm not sure I'm on the right track.

Regarding the formula, I'm trying to replicate the following sample I made locally `sunshine-prerelease.rb`:
```rb
require "language/node"

class SunshinePrerelease < Formula
  desc "Self-hosted game stream host for Moonlight"
  homepage "https://app.lizardbyte.dev/Sunshine"
  url "https://github.com/LizardByte/Sunshine.git",
    tag: "v2024.805.184417"
  version "2024.805.184417"
  license all_of: ["GPL-3.0-only"]
  head "https://github.com/LizardByte/Sunshine.git", branch: "master"

  # https://docs.brew.sh/Brew-Livecheck#githublatest-strategy-block
  livecheck do
    url :stable
    regex(/^v?(\d+\.\d+\.\d+)$/i)
    strategy :github_latest do |json, regex|
      match = json["tag_name"]&.match(regex)
      next if match.blank?

      match[1]
    end
  end

  option "with-docs-off", "Disable docs"
  option "with-dynamic-boost", "Dynamically link Boost libraries"
  option "without-dynamic-boost", "Statically link Boost libraries" # default option

  depends_on "cmake" => :build
  depends_on "doxygen" => :build
  depends_on "graphviz" => :build
  depends_on "node" => :build
  depends_on "pkg-config" => :build
  depends_on "curl"
  depends_on "miniupnpc"
  depends_on "openssl"
  depends_on "opus"
  depends_on "icu4c" if build.without?("dynamic-boost")

  conflicts_with "sunshine", because: "sunshine-prerelease symlink with the name for compatibility with sunshine"

  on_linux do
    depends_on "avahi"
    depends_on "libcap"
    depends_on "libdrm"
    depends_on "libnotify"
    depends_on "libva"
    depends_on "libvdpau"
    depends_on "libx11"
    depends_on "libxcb"
    depends_on "libxcursor"
    depends_on "libxfixes"
    depends_on "libxi"
    depends_on "libxinerama"
    depends_on "libxrandr"
    depends_on "libxtst"
    depends_on "numactl"
    depends_on "pulseaudio"
    depends_on "systemd"
    depends_on "wayland"
  end

  def install
[...]

PR #2990, opened 2024-08-08
feat(macos): add beta homebrew formula

ReenigneArcher:
## Description

This PR adds a beta homebrew formula. It works by doing the following, AFTER the spec is validated.

1. rename the formula
2. Use sed to replace 2 values
3. Upload it to our homebrew repo

### Issues Fixed or Closed

- Closes #2973 

## Type of Change

## Checklist

### DIFF
--- a/packaging/sunshine.rb
+++ b/packaging/sunshine.rb
@@ -1,6 +1,7 @@
 require "language/node"
 
 class @PROJECT_NAME@ < Formula
+  # conflicts_with "sunshine", because: "sunshine and sunshine-beta cannot be installed at the same time"
   desc "@PROJECT_DESCRIPTION@"
   homepage "@PROJECT_HOMEPAGE_URL@"
   url "@GITHUB_CLONE_URL@",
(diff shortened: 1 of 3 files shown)


packaging/sunshine.rb
```diff
@@ -1,6 +1,7 @@
 require "language/node"
 
 class @PROJECT_NAME@ < Formula
+  # conflicts_with "sunshine-beta", because: "sunshine and sunshine-beta cannot be installed at the same time"
```
Hazer, line 4: I think the right way to handle it is, for Stable sunshine, it should be `sunshine-beta`. For Beta sunshine, it should be `sunshine`. So if you're commenting out for Stable, it should be just `sunshine` here.

Hazer (changes requested):

packaging/sunshine.rb
ReenigneArcher: Commenting out cause it fails the audit. It's not commented out in the beta formula that gets published though.

Anything more advanced than this will require changes to https://github.com/LizardByte/homebrew-release-action

But yea, the conflict needs to be reversed with this logic.

Hazer (approved):

PR #3013, opened 2024-08-13
feat(cmake): use symlink/junction for shaders in build dir

ns6089:
## Description

Use directory junction instead of copy in the build directory during cmake configure on Windows.
This prevents work-in-progress shaders getting overwritten without any confirmation.

## Type of Change

## Checklist

### DIFF
--- a/cmake/packaging/windows.cmake
+++ b/cmake/packaging/windows.cmake
@@ -39,9 +39,15 @@ install(DIRECTORY "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/misc/gamepad/"
 install(DIRECTORY "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/assets/"
         DESTINATION "${SUNSHINE_ASSETS_DIR}"
         COMPONENT assets)
-# copy assets to build directory, for running without install
+
+# copy assets (excluding shaders) to build directory, for running without install
 file(COPY "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/assets/"
-        DESTINATION "${CMAKE_BINARY_DIR}/assets")
+        DESTINATION "${CMAKE_BINARY_DIR}/assets"
+        PATTERN "shaders" EXCLUDE)
+# use junction for shaders directory
+file(TO_NATIVE_PATH "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/assets/shaders" shaders_in_build_src_native)
+file(TO_NATIVE_PATH "${CMAKE_BINARY_DIR}/assets/shaders" shaders_in_build_dest_native)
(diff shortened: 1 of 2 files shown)


cmake/packaging/windows.cmake
```diff
@@ -39,9 +39,15 @@ install(DIRECTORY "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/misc/gamepad/"
...
+
+# copy assets (excluding shaders) to build directory, for running without install
 file(COPY "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/assets/"
-        DESTINATION "${CMAKE_BINARY_DIR}/assets")
+        DESTINATION "${CMAKE_BINARY_DIR}/assets"
+        PATTERN "shaders" EXCLUDE)
+# use junction for shaders directory
+file(TO_NATIVE_PATH "${SUNSHINE_SOURCE_ASSETS_DIR}/windows/assets/shaders" shaders_in_build_src_native)
+file(TO_NATIVE_PATH "${CMAKE_BINARY_DIR}/assets/shaders" shaders_in_build_dest_native)
+execute_process(COMMAND cmd.exe /c mklink /J "${shaders_in_build_dest_native}" "${shaders_in_build_src_native}")
```
ReenigneArcher, line 50: I think we can use cmake to create a symbolic link instead. https://cmake.org/cmake/help/latest/command/file.html#create-link

ReenigneArcher (commented):
Same changes also needed for Linux. https://github.com/LizardByte/Sunshine/blob/299672795c5ef923e5a000d6e11f9421e3a43fb5/cmake/packaging/linux.cmake#L5-L7

cmake/packaging/windows.cmake
ns6089: On Unix-like sure.
On Windows it's complicated. Symlinks require UAC elevation unless usermode symlinks are enabled in group policies or Windows is in Developer mode. Directory junctions always work without elevation.

cgutman: Do we need to go through cmd.exe or can we invoke `mklink` directly?

ns6089: Yes, `mklink` is a command of cmd shell, and there is no `mklink.exe`.

ns6089:
> Same changes also needed for Linux. 

I wouldn't say needed. but I can make them. But I can't say for sure that it won't break some obscure build configuration or packager, unlike windows on linux copying symlinks copy symlinks themselves, not what they point at.

@ReenigneArcher so in light of this information, should I do it or not?

ReenigneArcher:
> I wouldn't say needed

Right, it could just be annoyance when developing on Linux, just like it was for you on Windows. Shouldn't affect anything for packagers as I only added this for allowing unit tests to run without doing a full blown install.

> should I do it or not?

Yes, please.

ns6089:
Added shaders directory symlink to linux cmake.
Untested, but if CI doesn't fail it should work fine.

PR #3029, opened 2024-08-16
fix(win/qsv): skip unsupported 4:4:4 codecs

ns6089:
## Description

Skip particular QSV 4:4:4 codecs bypassing ffmpeg.

## Type of Change

## Checklist

### DIFF
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -1871,6 +1871,13 @@ namespace platf::dxgi {
       if (!boost::algorithm::ends_with(name, "_qsv")) {
         return false;
       }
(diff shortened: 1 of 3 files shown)


cgutman (changes requested):
We also need to cover the H.264 case above on line 2519.

Edit: Actually, it looks like we're missing a `>= 0` check on the return value of `validate_config()` there too. That explains the weird behavior I saw.

ns6089:
Updated the branch.

One day we should migrate all those int-returning function too bool. Some fail on 0, some fail on non-0, and both are implicitly converted.

cgutman (approved):

PR #3077, opened 2024-08-24
fix(nvhttp): wrap TLS socket to ensure graceful closure

cgutman:
## Description
The upstream `SimpleWeb::Server<SimpleWeb::HTTPS>` implementation doesn't ensure the TLS session is torn down gracefully which leads to unexpected TCP RST packets from Sunshine to the client when using TLS v1.3 and setting `close_connection_after_response` (which NvHTTP does). Depending on the client, this spurious TCP RST can lead to the HTTPS request failing even though all the required data did end up on the wire. The `QSslSocket` implementation in Qt seems to be particularly susceptible to the issue, causing computers to constantly bounce between online and offline.

Ideally this would be fixed upstream, but the way that the classes are structured with `SimpleWeb::ServerBase` doesn't provide a straightforward place to insert this code (`ServerBase` has no clue about any TLS stuff). Fixing it on our side using a little wrapper class is fairly simple with most of the diff just being renames of the old class to the new one. We already had a class derived from `SimpleWeb::Server<SimpleWeb::HTTPS>`, so changing it to derive from `SimpleWeb::ServerBase` just required a few more lines from the upstream constructor of `SimpleWeb::Server<SimpleWeb::HTTPS>`.

## Type of Change

## Checklist

### DIFF
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -44,19 +44,38 @@ namespace nvhttp {
 
   crypto::cert_chain_t cert_chain;
 
-  class SunshineHttpsServer: public SimpleWeb::Server<SimpleWeb::HTTPS> {
+  class SunshineHTTPS: public SimpleWeb::HTTPS {
   public:
-    SunshineHttpsServer(const std::string &certification_file, const std::string &private_key_file):
-        SimpleWeb::Server<SimpleWeb::HTTPS>::Server(certification_file, private_key_file) {}
+    SunshineHTTPS(boost::asio::io_service &io_service, boost::asio::ssl::context &ctx):
+        SimpleWeb::HTTPS(io_service, ctx) {}
+
+    virtual ~SunshineHTTPS() {
+      // Gracefully shutdown the TLS connection
+      SimpleWeb::error_code ec;
+      shutdown(ec);
+    }
+  };
+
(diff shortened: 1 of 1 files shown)


src/nvhttp.cpp
```diff
@@ -44,19 +44,38 @@ namespace nvhttp {
...
+      // Gracefully shutdown the TLS connection
+      SimpleWeb::error_code ec;
+      shutdown(ec);
+    }
+  };
+
+  class SunshineHTTPSServer: public SimpleWeb::ServerBase<SunshineHTTPS> {
+  public:
+    SunshineHTTPSServer(const std::string &certification_file, const std::string &private_key_file):
+        ServerBase<SunshineHTTPS>::ServerBase(443),
```
ns6089, line 62: Is it ok to hardcode the port?

cgutman: Yeah, it's just the default port number for the socket type. It's always been there, but it was just previously hidden inside the constructor for `SimpleWeb::ServerBase<SimpleWeb::HTTPS>` before.

Hazer:
@cgutman Any migration steps for current users running this behind some reverse proxy or it's mostly transparent?

cgutman:
It's transparent to any reverse proxies.

PR #3084, opened 2024-08-25
fix(mdns): don't hardcode mDNS instance name

cgutman:
## Description
All platform mDNS implemenations currently hardcode the instance name to `Sunshine`. This causes reliable name collisions during mDNS registration when more than one Sunshine host is running on the same network. The collision is eventually resolved by appending a number to the name, but this is unnecessary work and delay for nothing. We should just register using our hostname as the instance name most other mDNS services do.

## Type of Change

## Checklist

### DIFF
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -206,4 +206,32 @@ namespace net {
 
     return mapped_port;
   }
+
+  /**
+   * @brief Returns a string for use as the instance name for mDNS.
+   * @return Hostname-based instance name or "Sunshine" if hostname is invalid.
+   */
+  std::string
+  mdns_instance_name() {
+    std::string hostname = boost::asio::ip::host_name();
+
+    // Truncate to 63 characters per RFC 6763 section 7.2.
+    if (hostname.size() > 63) {
+      hostname.resize(63);
+    }
+
+    for (auto i = 0; i < hostname.size(); i++) {
+      // Replace any spaces with dashes
+      if (hostname[i] == ' ') {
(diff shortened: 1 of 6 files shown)


src/network.cpp
```diff
@@ -206,4 +206,32 @@ namespace net {
...
     return mapped_port;
   }
+
+  /**
+   * @brief Returns a string for use as the instance name for mDNS.
+   * @return Hostname-based instance name or "Sunshine" if hostname is invalid.
+   */
+  std::string
+  mdns_instance_name() {
+    std::string hostname = boost::asio::ip::host_name();
```
ReenigneArcher, line 216: If this became an argument for the function, the unit testing could be parametrized with various hostnames to check if it will work all the scenarios provided.

src/platform/macos/publish.cpp
```diff
@@ -105,7 +105,8 @@ namespace platf::publish {
       &serviceRef,
       0,  // flags
       0,  // interfaceIndex
-      SERVICE_NAME, SERVICE_TYPE,
+      nullptr,  // name
```
Hazer, line 108: Since you're giving it a nullptr, it won't use `net::mdns_instance_name()` for macOS at all, and the tests aren't useful for macOS right?

cgutman: Correct, the function is unused on macOS.

src/network.cpp
cgutman: Added in the latest commit.

PR #3239, opened 2024-09-29
build(fedora)!: add copr build

ReenigneArcher:
## Description

Add fedora copr build spec. Final build will be here: https://copr.fedorainfracloud.org/coprs/lizardbyte/Sunshine/

Todo:
- [x] trigger copr build from PRs and push events
- [x] dynamically get version number
- [x] cuda for aarch64
- [x] one cmake call
- [x] remove fedora docker builds
- [x] update docs
- [x] unit tests pass
- [x] cuda improvements (in f40+)
- [x] remove og spec file

References:
- https://pavel.raiskup.cz/blog/copr-ci-and-custom-source-method.html
- https://github.com/praiskup/copr-ci-tooling
- https://docs.pagure.org/copr.copr/custom_source_method.html#custom-source-method
- https://docs.pagure.org/copr.copr/user_documentation.html#webhooks

### Screenshot

image

## Type of Change

## Checklist


packaging/linux/fedora/sunshine.spec
```diff
@@ -0,0 +1,237 @@
...
+BuildRequires:  rpm-build
+BuildRequires:  systemd-udev
+BuildRequires:  systemd-rpm-macros
+%{?sysusers_requires_compat}  # TODO: what is this?
+BuildRequires:  wget
+BuildRequires:  which
+
+# Conditional BuildRequires for cuda-gcc based on Fedora version and architecture
+%if 0%{?fedora} >= 40 && "%{_target_cpu}" == "x86_64"
+BuildRequires:  cuda-gcc-c++  # TODO: why not use this for all builds?
```
matte-schwartz, line 51: Wasn't sure if there were any unexpected repercussions for using this package on aarch64 and did not have the ability to check for myself. very likely it can be used for all builds.

packaging/linux/fedora/sunshine.spec
```diff
@@ -0,0 +1,237 @@
+%global build_timestamp %(date +"%Y%m%d")
+%undefine _hardened_build
+
+Name:           sunshine
+Version:        2024.911.215654  # TODO: can we get this dynamically?
```
matte-schwartz, line 5: should be able to import this from a git tag https://docs.pagure.org/copr.copr/user_documentation.html#github-webhooks https://docs.pagure.org/copr.copr/user_documentation/pagure_integration.html#auto-rebuilding 

and use something like
`{{{ git_dir_version }}}`

packaging/linux/fedora/sunshine.spec
```diff
@@ -0,0 +1,237 @@
...
+Summary:        Sunshine is a self-hosted game stream host for Moonlight.
+Release:        1%{?dist}
+License:        GPLv3-only
+URL:            https://github.com/LizardByte/Sunshine
+Source0:        sunshine.sysusers  # TODO: what is this?
```
matte-schwartz, line 10: https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_dynamic_allocation this was my best attempt at following the fedora guidelines for adding a user to a group, i.e. for input.

matte-schwartz (commented):
left a few replies to some of the TODOs

ReenigneArcher:
Thanks for the review!

Ideally, we could just build using our linux build script and not worry about the spec file at all (could hopefully just use the rpm generated by cpack)... but for some reason it's detecting that it's being run on Fedora 41. Currently the script doesn't support chroots (or whatever) and expects to be run on the target system.

matte-schwartz:
Ah I see what you mean here: https://download.copr.fedorainfracloud.org/results/lizardbyte/Sunshine:pr:3239/srpm-builds/08090723/builder-live.log.gz

I wonder if there's a chroot property or something that we're missing then? I'll see if I can find anything useful, otherwise I do wonder if this is a COPR issue of some sort since I don't even see a fedora-41 chroot in your COPR repository.

ReenigneArcher:
Just did a little more checking... I guess it comes from here: 
image

I assumed it would run the script in each of the selected chroots. Maybe we have to cross compile to get this to all work then...

matte-schwartz:
Yes, that's how I understand the ideal COPR workflow to work. For example, I know that Bazzite ends up doing it this way for their packages directly via GitHub webhooks. You give it the .spec file in your git repo and then it rebuilds using that .spec however you set it up to do so. I think in general this is to promote reproducible builds.

packaging/linux/fedora/sunshine.spec
```diff
@@ -0,0 +1,252 @@
...
+path_to_udevadm=$(which udevadm)
+if [ -x "$path_to_udevadm" ] ; then
+  $path_to_udevadm control --reload-rules
+  $path_to_udevadm trigger --property-match=DEVNAME=/dev/uinput
+  $path_to_udevadm trigger --property-match=DEVNAME=/dev/uhid
+fi
+
+%preun
+# Remove udev rule
+rm -f %{_udevrulesdir}/99-uhid.rules  # TODO: why is this removed... doesn't look like a Sunshine file
```
matte-schwartz, line 219: I think this can be dropped... Checked in a VM and no such file gets added during sunshine install.

packaging/linux/fedora/sunshine.spec
```diff
@@ -0,0 +1,252 @@
...
+
+# Remove modules-load configuration
+rm -f /usr/lib/modules-load.d/uhid.conf  # TODO: why is this removed?
```
matte-schwartz, line 222: this gets created during the .spec itself and i had issues with it lingering afterwards after removing sunshine. 

```
# Add modules-load configuration
install -D -m 0644 /dev/stdin %{buildroot}/usr/lib/modules-load.d/uhid.conf <<EOF
uhid
EOF
```
this is done to load the uhid module in initramfs even if it doesn't detect the module as being used during dracut, which must be run every time a new kernel is installed. from what i remember when packaging it, a user might otherwise need to `modprobe uhid` manually.

packaging/linux/fedora/sunshine.spec
```diff
@@ -0,0 +1,252 @@
...
+cd %{_builddir}/Sunshine/build
+%make_install
+
+# Add modules-load configuration
+install -D -m 0644 /dev/stdin %{buildroot}/usr/lib/modules-load.d/uhid.conf <<EOF
+uhid
+EOF
+
+%post
+# TODO: can we just use postinst script directly?
```
matte-schwartz, line 200: unfortunately no, in an `rpm-ostree` environment (which is immutable Fedora) the way that packages get installed into is does not very well with this. I ran into the issue previously and found this way to be the best.

packaging/linux/fedora/sunshine.spec
```diff
@@ -0,0 +1,252 @@
...
+
+# Systemd unit file for user services
+%{_userunitdir}/sunshine.service
+
+# Udev rules
+%{_udevrulesdir}/*-sunshine.rules
+
+# Modules-load configuration
+# TODO: what is this?
+%{_modulesloaddir}/uhid.conf
```
matte-schwartz, line 237: we made this earlier 
```
# Add modules-load configuration
install -D -m 0644 /dev/stdin %{buildroot}/usr/lib/modules-load.d/uhid.conf <<EOF
uhid
EOF
```
this just lists the fact that we end up with a file there in the buildroot while packaging.

packaging/linux/fedora/sunshine.spec
ReenigneArcher: I think this is the last item...

Does this not take care of it though?

```bash
# Trigger udev rule reload for /dev/uinput and /dev/uhid
path_to_udevadm=$(which udevadm)
if [ -x "$path_to_udevadm" ] ; then
  $path_to_udevadm control --reload-rules
  $path_to_udevadm trigger --property-match=DEVNAME=/dev/uinput
  $path_to_udevadm trigger --property-match=DEVNAME=/dev/uhid
fi
```

PR #3294, opened 2024-10-12
fix(audio/windows): don't set virtual speakers higher than 24-bit mode

andygrundman:
## Description
When the Moonlight "Mute host PC speakers" option is selected, Sunshine will change the audio device to Steam Speakers. The changed block of code cycles through some settings in order and stops on the first one found, which happens to be the 32-bit signed int format. What we really want to happen is to have 24-bit as the highest supported format because this mode won't unset an existing spatial audio mode (Atmos for Headphones only supports 16 and 24-bit at <= 48k. 

This patch simply removes the F32 and S32 checks for stereo. F32 isn't supported by Steam Speakers, but S32 is and that's the one we want to avoid.

## Type of Change

## Checklist

### DIFF
--- a/src/platform/windows/audio.cpp
+++ b/src/platform/windows/audio.cpp
@@ -124,9 +124,8 @@ namespace {
   create_virtual_sink_waveformats() {
     if constexpr (channel_count == 2) {
       auto channel_mask = waveformat_mask_stereo;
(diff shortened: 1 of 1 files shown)


cgutman:
@ns6089 thoughts?

ReenigneArcher:
Here's the related thread on discord: https://discord.com/channels/804382334370578482/1293704798574743625

ns6089:
@cgutman Should be fine I think.

PR #3325, opened 2024-10-24
feat(stream)!: remove limit on concurrent sessions and allow quitting apps with active sessions

cgutman:
## Description
This PR reworks session handling to be more flexible in the following cases:
- A user has existing streaming sessions and would like to connect a new client. This would fail with the default value of `channels` prior to this PR and the user would need to go manually disconnect that session.
- A user left a stream idle on another device and would like to quit the game they are streaming. That would fail until all active sessions have disconnected (and apprently in some other situation where session management broke: https://github.com/moonlight-stream/moonlight-qt/issues/1279)

This removes the `channels` configuration options and allows users to connect as many session as they want (up to the arbitrary limit of 128). They can also now quit the running app (which also disconnects all sessions) even if a session is still connected.

## Type of Change

## Checklist

### DIFF
--- a/src/rtsp.cpp
+++ b/src/rtsp.cpp
@@ -639,12 +639,24 @@ namespace rtsp_stream {
     server.session_clear(launch_session_id);
   }
 
+  /**
+   * @brief Get the number of active sessions.
+   * @return Count of active sessions.
+   */
   int
   session_count() {
     // Ensure session_count is up-to-date
     server.clear(false);
 
     return server.session_count();
+  }
+
+  /**
+   * @brief Terminates all running streaming sessions.
+   */
+  void
+  terminate_sessions() {
+    server.clear(true);
   }
 
   int
--- a/src/rtsp.cpp
+++ b/src/rtsp.cpp
@@ -599,7 +599,7 @@ namespace rtsp_stream {
      * @param session The session to remove.
      */
     void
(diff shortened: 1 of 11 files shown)


src/rtsp.cpp
```diff
@@ -644,6 +639,10 @@ namespace rtsp_stream {
     server.session_clear(launch_session_id);
   }
 
+  /**
+   * @brief Get the number of active sessions.
+   * @return Count of active sessions.
+   */
```
ReenigneArcher, line 645: The documentation blocks only need to be in the header assuming the object/function is defined there.

ReenigneArcher (commented):
Can we, or are we, logging the count of active streams? 128 seems excessive and likely to cause huge performance degradation? Might be good to just have that in the logs so the helpers can clearly see it.

cgutman:
> Can we, or are we, logging the count of active streams?

Added logging in the latest commit.

> 128 seems excessive and likely to cause huge performance degradation?

The point of picking 128 wasn't that someone would actually want to stream to 128 clients. We have to pick _some_ number because ENet allocates a fixed size array for holding peers. I picked 128 because it should be comfortably above other hardware and software limitations for almost any conceivable setup while not being excessively large and wasting memory. The intent is for users to hit the limits of their CPU/GPU/network before they would get near 128 clients.

ReenigneArcher (approved):

PR #3332, opened 2024-10-28
feat(vaapi): add option to enable strict enforcement of frame size

cgutman:
## Description
This PR introduces a configuration option for AMD GPUs using the VAAPI encoder to strictly enforce frame bitrate limits (rc_buffer_size) to balance the concerns of video quality (#2864) and frame bitrate excursions (#2788). Intel GPUs appear fine with VBR mode, so this option can be ignored for them.

This PR sets the default to the v0.23.1 (and pre-#2821) behavior which is to _not_ enforce a strict RC buffer size. Even though this isn't ideal for packet loss reasons, the immediate symptom of large frames exceeding the FEC maximum in #2788 was resolved separately by #2803 anyway.

### Screenshot
image

### Issues Fixed or Closed
Fixes #2864

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

Issue #3331, opened 2024-10-27
Unable to decode AV1 stream generated by NVENC on sunshine linux in capable hardware


LuisB79:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

Sunshine on arch produces an AV1 stream that it's not able to be decoded on a snapdragon 8 gen 3 soc, no matter what resolution is set or what framerate is set

### Expected Behavior

Sunshine on arch producing an AV1 stream that's able to decode on snapdragon 8 gen 3 soc.

### Additional Context

I installed as per the documentation instructions, this is not the aur package.
On windows the stream is being produced by nvenc and it's able to be decoded on the same hardware
One noticeable difference is that in windows there's a message saying "NvEnc: created encoder AV1 p1 two-pass rfi", where as in linux there isn't one

Settings on arch 
X11
NVFBC as capture
NVENC as encoder

Client
moonlight-android running on a zfold 6
Client works fine with h265 or h264

related: https://github.com/moonlight-stream/moonlight-android/issues/1434

### Host Operating System

Linux

### Operating System Version

Arch

### Architecture

amd64/x86_64

### Sunshine commit or version

Version v2024.1025.12635

### Package

Linux - LizardByte/pacman-repo

### GPU Type

Nvidia

### GPU Model

4080 mobile

### GPU Driver/Mesa Version

565.57.01

### Capture Method

NvFBC (Linux)

### Config

```Shell
qp = 24
upnp = enabled
resolutions = [
    1920x1080
]
encoder = nvenc
capture = nvfbc
```

### Relevant log output

```shell
[2024-10-26 20:29:26.392]: Info: Sunshine version: v2024.1025.12635
[2024-10-26 20:29:26.392]: Info: Package Publisher: LizardByte
[2024-10-26 20:29:26.392]: Info: Publisher Website: https://app.lizardbyte.dev
[2024-10-26 20:29:26.392]: Info: Get support: https://app.lizardbyte.dev/support
[2024-10-26 20:29:26.417]: Info: System tray created
[2024-10-26 20:29:26.835]: Info: Found [1] outputs
[2024-10-26 20:29:26.835]: Info: Virtual Desktop: 2560x1440
[2024-10-26 20:29:26.835]: Info: XrandR: available
[...]

Issue #2864, opened 2024-07-15
Stream quality regression VAAPI Radeon Linux


neatnoise:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

I regularly update master git version of sunshine. The latest updates (probably this https://github.com/LizardByte/Sunshine/commit/38c13c8fe1f204dcb5383cf34530557e6d829347 ) introduced stream quality regression for VAAPI encoding  (RX 580 GPU). The visual quality is much worse. The changed pixels update to achive good visual quality like with 1 second delay. At 25 Mb/s bitrate 1080p 60fps it shows very noticeable compression effects, it stutters, it looks like 5 Mbit bitrate before updates.

### Expected Behavior

Nice looking stream at 25 Mbit

### Host Operating System

Linux

### Operating System Version

Arch linux latest

### Architecture

64 bit

### Sunshine commit or version

sunshine-git-2024.713.205505.r0.g18e7dfb-1-x86_64.pkg.tar.zst

### Package

Linux - AUR (Third Party)

### GPU Type

AMD

### GPU Model

RX 580

### GPU Driver/Mesa Version

latest mesa-git

### Capture Method

KMX (Linux)

### Config

```Shell
wan_encryption_mode = 2
cert = x
pkey = x
output_name = 0
amd_quality = default
external_ip = x
nv_rc = auto
channels = 3
address_family = both
resolutions = [
    1280x720,
    1920x1080
]
upnp = enabled
nv_preset = default
fps = [60]
qp = -1
encoder = vaapi
lan_encryption_mode = 2
amd_rc = auto
file_state = x
```

### Apps

```json
All, desktop
```

### Relevant log output

```shell
lip 15 14:00:52 tm-stc sunshine[1294]: [2024:07:15:14:00:52]: Info: Trying encoder [vaapi]
lip 15 14:00:52 tm-stc sunshine[1294]: [2024:07:15:14:00:52]: Info: Screencasting with KMS
lip 15 14:00:52 tm-stc sunshine[1294]: [2024:07:15:14:00:52]: Info: /dev/dri/card1 -> amdgpu
lip 15 14:00:52 tm-stc sunshine[1294]: [2024:07:15:14:00:52]: Info: Found monitor for DRM screencasting
[...]

PR #3337, opened 2024-10-29
fix(vaapi): use VBR rate control mode

cgutman:
## Description
This is a possible alternative fix for https://github.com/LizardByte/Sunshine/issues/2864 that could be taken instead of https://github.com/LizardByte/Sunshine/pull/3332.

In the case of single frame VBV size, using CBR with VAAPI seems to have major quality problems with Intel and AMD drivers. VBR behaves much better in my testing (Intel H.264 case) and allows us to still enforce a reasonable frame size limit without causing the video quality to suffer. Interestingly, we also had to do something similar with the QuickSync encoder using the `CBR_WITH_VBR` quirk flag to force VBR mode.

As a nice side effect, it also allows the video bitrate to drop when content is static, rather than adding useless filler data to the bitstream to preserve the exact specified bitrate. This also fixes the decoder compatibility issue reported in #3331.

## Type of Change

## Checklist

### DIFF
--- a/src/platform/linux/vaapi.cpp
+++ b/src/platform/linux/vaapi.cpp
@@ -129,16 +129,6 @@ namespace va {
       return 0;
     }
 
-    void
-    init_codec_options(AVCodecContext *ctx, AVDictionary *options) override {
-      // Don't set the RC buffer size when using H.264 on Intel GPUs. It causes
-      // major encoding quality degradation.
-      auto vendor = vaQueryVendorString(va_display);
-      if (ctx->codec_id != AV_CODEC_ID_H264 || (vendor && !strstr(vendor, "Intel"))) {
-        ctx->rc_buffer_size = ctx->bit_rate * ctx->framerate.den / ctx->framerate.num;
-      }
-    }
-
     int
     set_frame(AVFrame *frame, AVBufferRef *hw_frames_ctx_buf) override {
       this->hwframe.reset(frame);
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -860,6 +860,7 @@ namespace video {
         { "low_power"s, 1 },
         { "async_depth"s, 1 },
         { "idr_interval"s, std::numeric_limits<int>::max() },
+        { "rc_mode"s, "VBR"s },
       },
       {},  // SDR-specific options
       {},  // HDR-specific options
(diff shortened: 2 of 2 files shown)


ns6089:
VBR is required because CBR with reasonably relaxed VBV still results in poor quality on Intel?

cgutman:
What I'm finding in my testing with several AMD and Intel GPUs with a single frame VBV is that VBR appears to behave like CBR except that it allows bitrate undershoots (which is what we want). The quality issues seen on Intel with CBR and a single frame VBV are similar to what we already saw with QSV and solved by forcing VBR there.

I did verify that this change alone does not fix the poor encoding quality on Polaris cards (though my RDNA3 card was totally fine, regardless of the VBV).

So I'm thinking for the full solution:
- VBR rate control
- Remove the Intel H.264 workaround (not needed with VBR)
- Wire up the strict VBV enforcement option in https://github.com/LizardByte/Sunshine/pull/3332 to be specific to non-Intel GPUs

ns6089:
Sporadic 2x-3x overshoots for a frame are expected without pre-analysis pass due to the parallel nature of hardware encoding, but not sure this can be controlled in VAAPI. What I meant by "relaxed" VBV is that we don't have to choose between two extremities - single frame or "off", but may settle on something like 4 or 5 frames. Because otherwise it may default to something like 1 second since it performs better in the usual periodic GOP scenarios. The fact that CBR also enables zero padding is certainly a bummer though, again not sure if it can be controlled in VAAPI.

cgutman:
> What I meant by "relaxed" VBV is that we don't have to choose between two extremities - single frame or "off", but may settle on something like 4 or 5 frames.

Yeah, unfortunately even pretty relaxed VBV buffers like 250 ms still cause Polaris's encoder to produce garbage quality.

What I ended up doing for #3332 is basically choosing between 2 operating modes:
- VBR + single frame VBV
- CBR + default VBV

The former works great for Intel GPUs and AV1 on AMD GPUs (likely _any_ codec for RDNA3 and later, but I wanted to be conservative). It is also what we use for QuickSync (which likely shares RC logic with Intel's VAAPI driver).

The latter is what we will default to for H.264 and HEVC on AMD GPUs, but we'll allow users to override if they want.

The coupling of RC mode and RC buffer size helps to balance the negative effects of each other. Single frame VBV avoids the VBR pitfall of producing massive frames using RC headroom provided by earlier smaller frames. CBR avoids some of the downsides of a larger VBV by ensuring a near constant frame size.

Closing this PR in favor of the full fix in #3332

hebo6:
The bitrate drop feature looks impressive. I hope this PR will be merged.

PR #3339, opened 2024-10-29
fix(openssl): fix build with OPENSSL_NO_DEPRECATED

cgutman:
## Description
Remove dependency on functions that were deprecated in OpenSSL 3.0. No behavior changes at runtime (`SSL_get_peer_certificate` is defined to be an alias of `SSL_get1_peer_certificate()`).

The missing `openssl/rsa.h` include is used for `EVP_PKEY_CTX_set_rsa_keygen_bits()`. `openssl/rsa.h` was pulled in indirectly via `openssl/pem.h` when deprecated APIs are enabled. 

## Type of Change

## Checklist

### DIFF
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -1033,7 +1033,13 @@ namespace nvhttp {
 
     // Verify certificates after establishing connection
     https_server.verify = [add_cert](SSL *ssl) {
-      crypto::x509_t x509 { SSL_get_peer_certificate(ssl) };
+      crypto::x509_t x509 {
+#if OPENSSL_VERSION_MAJOR >= 3
+        SSL_get1_peer_certificate(ssl)
+#else
+        SSL_get_peer_certificate(ssl)
+#endif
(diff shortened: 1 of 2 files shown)


src/nvhttp.cpp
```diff
@@ -1033,7 +1033,13 @@ namespace nvhttp {
 
     // Verify certificates after establishing connection
     https_server.verify = [add_cert](SSL *ssl) {
-      crypto::x509_t x509 { SSL_get_peer_certificate(ssl) };
+      crypto::x509_t x509 {
+#if OPENSSL_VERSION_MAJOR >= 3
```
ReenigneArcher, line 1037: I don't believe any builds still using anything less than 3.0, are you wanting to just keep this check for any who may self compile with an older version?

cgutman: Yeah, it's such a trivial change that it doesn't seem like a problem to keep supporting 1.1 and 3.0+.

PR #3344, opened 2024-10-30
fix(windows): disable shortcut mode for WGC

ReenigneArcher:
## Description

This PR slightly improves WGC experience by doing the following:

- Disables "shortcut" mode when WGC is enabled (either from CLI args or from the config option).
- Add a WGC specific start menu shortcut. This works if running as "Admin", but fails otherwise due to permissions accessing the config directory. I'm not sure how to make the shortcut have "run as administrator" mode, which I know is possible in Windows, but I don't know if it's possible from NSIS.

Additional ideas I have, but not really sure the best way to handle are:

- Auto toggle the service on or off when changing capture mode.

This PR also fixes one small typo from #3343.

## Type of Change

## Checklist

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1259,6 +1259,15 @@ namespace config {
       // the path is incorrect or inaccessible.
       apply_config(std::move(vars));
       config_loaded = true;
+
+      // Disable shortcut launch modes if capture mode is wgc
+#ifdef _WIN32
+      if (config::video.capture == "wgc" && (shortcut_launch || service_admin_launch)) {
+        BOOST_LOG(info) << "Disabling service launch because WGC capture mode is enabled"sv;
+        shortcut_launch = false;
+        service_admin_launch = false;
+      }
+#endif
     }
     catch (const std::filesystem::filesystem_error &err) {
       BOOST_LOG(fatal) << "Failed to apply config: "sv << err.what();
--- a/cmake/packaging/windows.cmake
+++ b/cmake/packaging/windows.cmake
@@ -101,10 +101,13 @@ set(CPACK_NSIS_CREATE_ICONS_EXTRA
         "${CPACK_NSIS_CREATE_ICONS_EXTRA}
         CreateShortCut '\$SMPROGRAMS\\\\$STARTMENU_FOLDER\\\\${CMAKE_PROJECT_NAME}.lnk' \
             '\$INSTDIR\\\\${CMAKE_PROJECT_NAME}.exe' '--shortcut'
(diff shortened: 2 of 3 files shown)


src/config.cpp
```diff
@@ -1259,6 +1259,15 @@ namespace config {
       // the path is incorrect or inaccessible.
       apply_config(std::move(vars));
       config_loaded = true;
+
+      // Disable shortcut launch modes if capture mode is wgc
+#ifdef _WIN32
+      if (config::video.capture == "wgc" && (shortcut_launch || service_admin_launch)) {
+        BOOST_LOG(info) << "Disabling service launch because WGC capture mode is enabled"sv;
+        shortcut_launch = false;
+        service_admin_launch = false;
```
cgutman, line 1268: I think this will mean we launch as non-admin right? I don't think we'll have access to our config files in that case.

ReenigneArcher: I did have to run the new shortcut as admin, but as far as I understand with that variable enabled, it will try to start the service which currently does no work with WGC.

https://github.com/LizardByte/Sunshine/blob/ec0cdcf0fa711df47bda9e54a4177b7b1199e45d/src/config.cpp#L1289-L1295

Perhaps instead of disabling "shortcut_launch", it would be better to add a new `shortcut_launch_wgc` variable?

cgutman: I'm a little hesistant to add another shortcut, especially one that activates a Beta option and requires the user to know to run it as admin. I remember the bad old days when we were telling users not to use our shortcuts because they would run Sunshine in a confusing configuration (non-admin, possible concurrently with the service-based Sunshine instance).

I feel like this is stepping back into that usability mess again. I know that the experience for WGC isn't great today with the current shortcut behavior, but that's a beta feature that probably less than 1% of people use. I don't want to confuse the other 99% to try to make the experience slightly better for those few WGC users.

ReenigneArcher: > I'm a little hesistant to add another shortcut, especially one that activates a Beta option and requires the user to know to run it as admin.

My wording was probably confusing. I mean using the variable `shortcut_launch_wgc`, and then have a similar conditional check to https://github.com/LizardByte/Sunshine/blob/ec0cdcf0fa711df47bda9e54a4177b7b1199e45d/src/config.cpp#L1296-L1328 that would launch it automatically as admin. Looking at that code it seems possible to relaunch as admin, without requiring the user to do "runas" manually. If this could work, no new start menu shortcut would be required and then it would just rely on the user changing their config (I think).

cgutman: We could avoid the service launch if we detect WGC is enabled and just launch as admin, but there is still the concern with coexistence with any running service-based Sunshine process. I'm not really sure how to nicely solve that without creating a scenario where the user could fairly trivially break it by toggling the capture method.

Issue #3340, opened 2024-10-30
macOS 15 connection issues


soerenkampschroer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

On macOS 15 sunshine is not able to connect to the client when started as a service (brew services start sunshine-beta). The server will advertise but upon connecting the logs show that the connection times out.

It may be a permission issue connected to the new Local Network permission. As far as I know it's impossible to add or whitelist programs for the permission and there needs to be a popup where you can confirm or deny access. I did manage to get the last beta build to work as a service by deleting all sunshine related permissions and reinstalling it, but due to the frequent name change of the beta build it needs new permissions after every update. I have not been able to get the latest build running, even though I again removed all permissions and reinstalled the latest build again after doing the update.

However, sunshine works fine if it's started manually by running it in terminal as a command.

This was not an issue on macOS 14, so my guess is it's the new Local Network permission, which has caused issues with other apps as well.

Firewall is turned off.

Related thread on this issue:
https://forums.developer.apple.com/forums/thread/760964

### Expected Behavior

Client being able to connect.

### Host Operating System

macOS

### Operating System Version

15.0.1

### Architecture

amd64/x86_64

### Sunshine commit or version

v2024.1029.45450

### Package

macOS - Homebrew

### GPU Type

AMD

### GPU Model

AMD RX 6800

### GPU Driver/Mesa Version

OS Drivers

### Capture Method

None

### Config

```Shell
encoder = videotoolbox
output_name = 4128836
```

### Apps

```json
I'm only using Desktop.
```

### Relevant log output

```shell
[2024-10-30 01:07:15.619]: Info: // Ignore any errors mentioned above, they are not relevant. //
[2024-10-30 01:07:15.619]: Info: 
[2024-10-30 01:07:15.619]: Info: Found H.264 encoder: h264_videotoolbox [videotoolbox]
[2024-10-30 01:07:15.619]: Info: Found HEVC encoder: hevc_videotoolbox [videotoolbox]
[2024-10-30 01:07:15.619]: Info: Executing Do Cmd: [/Users/soeren/Documents/Projects/sunshine-scripts/start.sh]
[2024-10-30 01:07:22.014]: Info: Executing [Desktop]
[...]

PR #3347, opened 2024-10-31
fix(homebrew): move binary after build and codesign

soerenkampschroer:
## Description
Moving binary after build and adding codesign as per issue #3340 

### Issues Fixed or Closed
Fixes issue #3340 

## Type of Change

## Checklist

### DIFF
--- a/packaging/sunshine.rb
+++ b/packaging/sunshine.rb
@@ -113,13 +113,13 @@ class @PROJECT_NAME@ < Formula
       system "make", "install"
 
       # Move binary instead of symlinking it
-      bin.install "sunshine-#{ENV["BUILD_VERSION"]}" => "sunshine"
+      bin.install "sunshine-#{ENV["BUILD_VERSION"]}" => "sunshine" if OS.mac?
 
       bin.install "tests/test_sunshine"
     end
 
     # Codesign the installed binary
-    system "codesign", "-s", "-", "--force", "--deep", "#{bin}/sunshine"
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Failures are a bit hard to find so I'm pasting them here:

```txt
  Error: 1 problem in 1 formula detected.
    * line 122, col 56: Use `bin/"sunshine"` instead of `"#{bin}/sunshine"`
  Error:  Process [['brew', 'audit', '--os=all', '--arch=all', '--strict', '--online', 'homebrew-release-action/homebrew-test/sunshine']] failed with exit code 1
  Error:  Formula sunshine failed audit
```

And on Linux we probably need to make this conditional? Or use an alternative.

```
==> codesign -s - --force --deep /home/linuxbrew/.linuxbrew/Cellar/sunshine/0.0.3347/bin/sunshine
Failed to execute: codesign
```

soerenkampschroer:
Oh I wasn't aware this would impact building for Linux as well. 
I haven't used Linux in a while, so I don't know if signing is even a thing there. But as it is, it should be conditional only to macOS.

Also I've discovered that the moving part isn't really moving but copying. The build folder will contain two identical binaries, sunshine and sunshine-v2024.XXXX.XXXX. That doesn't bother me personally because the binary is quite small, but maybe there is a better way. The whole copy thing only really matters for macOS too, so that should probably also be conditional.

ReenigneArcher:
Yea, the install will just install that file again. I don't think it's a huge deal as our binary is quite small. There's probably a way to remove the original versioned binary filename though.

PR #3348, opened 2024-10-31
fix(homebrew): codesign binary only on Intel macOS

soerenkampschroer:
## Description
Moving binary after build and adding codesign for macOS as per issue #3340 

### Issues Fixed or Closed
Issue #3340 

## Type of Change

## Checklist

### DIFF
--- a/packaging/sunshine.rb
+++ b/packaging/sunshine.rb
@@ -112,14 +112,11 @@ class @PROJECT_NAME@ < Formula
       system "make"
       system "make", "install"
 
-      # Move binary instead of symlinking it
-      bin.install "sunshine-#{ENV["BUILD_VERSION"]}" => "sunshine" if OS.mac?
-
       bin.install "tests/test_sunshine"
     end
 
-    # Codesign the installed binary
-    system "codesign", "-s", "-", "--force", "--deep", bin/"sunshine" if OS.mac?
+    # codesign the binary on intel macs
+    system "codesign", "-s", "-", "--force", "--deep", bin/"sunshine" if OS.mac? && Hardware::CPU.intel?
 
     bin.install "src_assets/linux/misc/postinst" if OS.linux?
   end
--- a/packaging/sunshine.rb
+++ b/packaging/sunshine.rb
@@ -113,13 +113,13 @@ class @PROJECT_NAME@ < Formula
       system "make", "install"
 
       # Move binary instead of symlinking it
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
FYI, in the future you don't need to close your PR and open a new one to make fixes. You can just push your changes to your PR branch and it will be picked up.

cathyjf:
What is this intended to fix? The issue of having to re-grant permissions for each build of sunshine will _not_ be fixed by this, if that's the intent. That issue occurs because the ad hoc code signature changes each time the underlying binary changes; it's not because the filename is changing. To avoid having to remove and re-grant the permission each build, it's necessary to sign each build with the **same** Apple Developer ID certificate as the previous build, not an ad hoc certificate. Homebrew doesn't provide a way to do this, so the issue of having to remove and re-grant permissions cannot be fixed very easily.

cathyjf:
Also, Homebrew already signs all code it builds, at least on arm64. This pull request is not the correct solution to whatever it is trying to fix.

cathyjf:
The actual solution to this general problem is that Sunshine should be distributed as a proper app bundle, and signed by some trusted certificate (perhaps owned by a trusted maintainer) as part of the GitHub build process. Then the homebrew formula should be replaced by a cask that just installs the trusted binary. This will solve the following problems:

1. Permissions will not need to be removed and re-granted each time Sunshine updates any code.
2. Sunshine will not be a gaping vulnerability on the host machine. Currently, once you grant Sunshine any permissions, you're actually granting those permissions to all programs, because any program can trivially inject code into the Sunshine process. To avoid this, it's necessary for Sunshine and all its dependencies to use the trusted runtime, and be distributed as an app bundle.

cathyjf:
To summarize my comments, everything already works as well as it can be expected to work on arm64 because Homebrew already signs code on arm64. It apparently does not sign code on Intel. As a stopgap for Intel users, I suppose this pull request is fine, but I recommend that it be made specific to Intel to avoid conflicting with Homebrew's built in ad hoc codesigning. You can check for the architecture in the Homebrew formula.

Long term, Homebrew should be distributed as an app bundle, but I think it's valid to have a short term fix for Intel users.

soerenkampschroer:
Thank you so much for clearing that up @cathyjf that makes sense! I am running Intel, so that's probably why I needed to sign it myself. I wasn't aware about brew signing it only for arm64. I'll look into how to make it conditional to Intel and update the pull request accordingly. I'll also remove the line copying the binary.

PR #3360, opened 2024-11-02
fix(macos): prevent indefinite hanging if screen capture is not granted

cathyjf:
## Description

Currently, if Sunshine has not yet been granted the screen capture
permission, the program simply hangs indefinitely on startup when it
attempts to probe for usable encoders. This is not desirable because
it prevents testing and using all of the parts of Sunshine that do not
require the screen capture permission.

With this patch, the encoder probing will simply fail instead of
hanging indefinitely if Sunshine does not yet have the screen capture
permission.

Note that Sunshine already prints out an error message telling the user
that the screen capture permission is needed. The bug is that Sunshine
currently indefinitely hangs shortly after printing that message.

### Screenshot
N/A

### Issues Fixed or Closed
N/A

## Type of Change

## Checklist

### DIFF
--- a/src/platform/macos/misc.h
+++ b/src/platform/macos/misc.h
@@ -8,6 +8,11 @@
 
 #include <CoreGraphics/CoreGraphics.h>
 
+namespace platf {
+  bool
+  is_screen_capture_allowed();
+}
+
 namespace dyn {
   typedef void (*apiproc)();
 
--- a/src/platform/macos/misc.mm
+++ b/src/platform/macos/misc.mm
@@ -42,6 +42,16 @@ namespace platf {
   CGRequestScreenCaptureAccess(void) __attribute__((weak_import));
 #endif
 
+  namespace {
+    auto screen_capture_allowed = std::atomic<bool> { false };
+  }  // namespace
+
+  // Return whether screen capture is allowed for this process.
+  bool
+  is_screen_capture_allowed() {
+    return screen_capture_allowed;
+  }
+
   std::unique_ptr<deinit_t>
   init() {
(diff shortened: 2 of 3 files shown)


src/platform/macos/misc.mm
```diff
@@ -42,6 +42,17 @@
...
+    /// @brief This variable will be set to true if and only if we determine that this
+    ///        process has been granted the screen capture permission.
+    auto screen_capture_allowed = std::atomic<bool>{ false };
+  }  ///< This anonymous namespace prevents access to its contents outside of this file.
+
+  /// Return whether screen capture is allowed for this process.
+  bool is_screen_capture_allowed() {
+    return screen_capture_allowed;
+  }
+
```
ReenigneArcher, line 55:
```suggestion
  namespace {
    auto screen_capture_allowed = std::atomic<bool> { false };
  }  // namespace

  // Return whether screen capture is allowed for this process.
  bool
  is_screen_capture_allowed() {
    return screen_capture_allowed;
  }

```

src/platform/macos/misc.h
```diff
@@ -8,6 +8,10 @@
 
 #include <CoreGraphics/CoreGraphics.h>
 
+namespace platf {
+  bool is_screen_capture_allowed();
```
ReenigneArcher, line 12:
```suggestion
  bool
  is_screen_capture_allowed();
```

ReenigneArcher (changes requested):
Thank you for the PR. The suggestions are mostly to address linting failures.

ReenigneArcher (approved):

ReenigneArcher (approved):

PR #3361, opened 2024-11-02
build(docs): parallelize generation of dot graphs

cathyjf:
Building the Sunshine docs is currently agonizingly slow because it
generates many hundreds of dot graphs in serial, without any
parallelization.

This commits sets `DOT_NUM_THREADS` to 0, which causes `doxygen` to
determine on its own a reasonable level of parallelization, based on
the number of processors available on the system.

The speed difference with this change is incredibly dramatic. I haven't
performed particularly rigorous testing, but on my fairly weak laptop
building the docs after applying this patch takes about 35 seconds.
Without applying this patch, building the docs instead takes a
ridiculous 135 seconds. This is on an M1 machine with only 4 performance
cores and 4 efficiency cores. On a more powerful machine, the difference
will be even more marked.

Unless there's some reason I'm not aware of why the doc files need to
be created in serial, I recommend enabling parallelization so that it
isn't so painful to build the documentation.

See also:
    https://xkcd.com/303

## Type of Change

## Checklist

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -30,6 +30,7 @@ PROJECT_LOGO = ../sunshine.png
 PROJECT_NAME = Sunshine
 
 # project specific settings
+DOT_NUM_THREADS = 0
 DOT_GRAPH_MAX_NODES = 60
 IMAGE_PATH = ../docs/images
 INCLUDE_PATH = ../third-party/build-deps/dist/Linux-x86_64/include/


ReenigneArcher:
I think this fails on Windows, or at least it used to. I'll run the workflow and see if that's still the case though.

ReenigneArcher:
> I think this fails on Windows, or at least it used to. I'll run the workflow and see if that's still the case though.
> 
> Edit: here is the original issue on Doxygen's side: [doxygen/doxygen#9694](https://github.com/doxygen/doxygen/issues/9694)

Seems to be working okay on the first attempt. I'm going to run it a few times because I seem to remember the issue being sort of random.

ReenigneArcher:
Unfortunately, it looks like this needs to be reverted.

image

cathyjf:
How about we make it serial on Windows only?

ReenigneArcher:
Fine with me, might be easier to handle it in https://github.com/LizardByte/doxyconfig

PR #3415, opened 2024-11-24
fix(nvenc): Enable opt-in client refresh by client

TheElixZammuto:
## Description
Context in https://github.com/TheElixZammuto/moonlight-xbox/issues/117

This type of degradation can be fixed on nvidia cards by implementing intra-refresh.
I do not have an nvidia card so I wasn't be able to test, but this change has been validated by the commenters in the issue above using a custom Sunshine build made by me and by other forks of Sunshine made by other people, and it seems to fix the issue.

This change is completely opt-in by the client, so this change should only apply to the Xbox/UWP Client

Related to:

**moonlight-xbox:** https://github.com/TheElixZammuto/moonlight-xbox/pull/155
**moonlight-common-c:** https://github.com/moonlight-stream/moonlight-common-c/pull/97

### Issues Fixed or Closed
https://github.com/TheElixZammuto/moonlight-xbox/issues/117

## Type of Change

## Checklist

### DIFF
--- a/src/nvenc/nvenc_base.cpp
+++ b/src/nvenc/nvenc_base.cpp
@@ -322,10 +322,20 @@ namespace nvenc {
         set_minqp_if_enabled(config.min_qp_hevc);
         fill_h264_hevc_vui(format_config.hevcVUIParameters);
         if (client_config.enableIntraRefresh == 1) {
-          format_config.enableIntraRefresh = 1;
-          format_config.singleSliceIntraRefresh = 1;
-          format_config.intraRefreshPeriod = 300;
-          format_config.intraRefreshCnt = 299;
+          if (get_encoder_cap(NV_ENC_CAPS_SUPPORT_INTRA_REFRESH)) {
+            format_config.enableIntraRefresh = 1;
+            format_config.intraRefreshPeriod = 300;
+            format_config.intraRefreshCnt = 299;
+            if (get_encoder_cap(NV_ENC_CAPS_SINGLE_SLICE_INTRA_REFRESH)) {
+              format_config.singleSliceIntraRefresh = 1;
+            }
+            else {
+              BOOST_LOG(warning) << "NvEnc: Single Slice Intra Refresh not supported";
+            }
+          }
+          else {
+            BOOST_LOG(error) << "NvEnc: Client asked for intra-refresh but the encoder does not support intra-refresh";
+          }
         }
         break;
       }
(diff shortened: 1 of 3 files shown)


ReenigneArcher:
Do we need to wait for the moonlight-common-c PR to be merged for this?

cgutman:
> Do we need to wait for the moonlight-common-c PR to be merged for this?

Not really. The SDP attribute looks fine and that's the only part of the network protocol that's affected by this change.

If we wanted to do this for real, I would suggest allowing the client to specify the parameters of the intra-refresh period, but since this is basically a hack, I'm fine just having a single flag to indicate it.

I'm pretty convinced we (or FFmpeg) probably have a bug here though because I've seen this "artifacting on mostly static images" pop up on other clients too. It's just really tricky to catch in a way that can actually be analyzed. We would need to write the whole video stream to disk, so we could run it through a bitstream analyzer it after the fact. Maybe it's feasible to do on Xbox where it occurs relatively soon, but in other cases it can take hours of streaming for it to happen.

src/nvenc/nvenc_base.cpp
```diff
@@ -321,6 +321,12 @@ namespace nvenc {
         set_ref_frames(format_config.maxNumRefFramesInDPB, format_config.numRefL0, 5);
         set_minqp_if_enabled(config.min_qp_hevc);
         fill_h264_hevc_vui(format_config.hevcVUIParameters);
+        if (client_config.enableIntraRefresh == 1) {
+          format_config.enableIntraRefresh = 1;
+          format_config.singleSliceIntraRefresh = 1;
```
cgutman, line 326: I think `singleSliceIntraRefresh` is a special feature that's not supported by all NVENC encoders. Do we need that?

ns6089: It's possible to check for `NV_ENC_CAPS_SINGLE_SLICE_INTRA_REFRESH` cap and enable single slice conditionally.

TheElixZammuto: Done!

TheElixZammuto:
> I'm pretty convinced we (or FFmpeg) probably have a bug here though because I've seen this "artifacting on mostly static images" pop up on other clients too. It's just really tricky to catch in a way that can actually be analyzed. We would need to write the whole video stream to disk, so we could run it through a bitstream analyzer it after the fact. Maybe it's feasible to do on Xbox where it occurs relatively soon, but in other cases it can take hours of streaming for it to happen.

@cgutman @ns6089 Could it be enough to save the last 300 NAL units onto a flash drive when the user presses an hotkey so that they can upload it to us and can analyse this further?

ns6089:
In my opinion, no.
https://github.com/TheElixZammuto/moonlight-xbox/issues/117#issuecomment-2503358667

PR #3417, opened 2024-11-25
fix(process): proc status lost when streaming

MiroKaku:
## Description
Steps to reproduce:
1. Start any streaming using moonlight.
2. Return to the app selection page.
3. Open the sunshine Web-UI, delete or add any application.
4. Streaming status on moonlight has been lost.

Reason:
```cpp
  void
  refresh(const std::string &file_name) {
    auto proc_opt = proc::parse(file_name);

    if (proc_opt) {
      proc = std::move(*proc_opt); // <---- !!!! Look here, It is overwritten.
    }
  }
```

### Screenshot
None

### Issues Fixed or Closed
None

## Type of Change

## Checklist


ReenigneArcher (commented):
Could you add documentation blocks for the new methods in the header file (not in the cpp file)?

https://docs.lizardbyte.dev/projects/sunshine/en/master/md_third-party_2doxyconfig_2docs_2source__code.html#documentation-blocks

MiroKaku:
> Could you add documentation blocks for the new methods in the header file (not in the cpp file)?
> 
> https://docs.lizardbyte.dev/projects/sunshine/en/master/md_third-party_2doxyconfig_2docs_2source__code.html#documentation-blocks

done.

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
 
     ~proc_t();
 
-    const std::vector<ctx_t> &get_apps() const;
-    std::vector<ctx_t> &get_apps();
+    /**
+     * @brief Gets the list of applications.
```
FrogTheFrog, line 93:
```suggestion
     * @brief Get the list of applications.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @return A constant reference to the vector of applications.
+     */
+    const std::vector<ctx_t>& get_apps() const;
+
+    /**
+     * @brief Gets the list of applications.
```
FrogTheFrog, line 99:
```suggestion
     * @brief Get the list of applications.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @return A reference to the vector of applications.
+     */
+    std::vector<ctx_t>& get_apps();
```
FrogTheFrog, line 102: I suggest to remove this method as it is not used at all.

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+
+    /**
+     * @brief Sets the list of applications.
```
FrogTheFrog, line 105:
```suggestion
     * @brief Set the list of applications.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @param apps The new list of applications.
+     * @note This will overwrite the existing list of applications.
+     * @see refresh(const std::string &file_name)
+     */
+    void set_apps(std::vector<ctx_t> apps);
+
+    /**
+     * @brief Gets the image path of the application with the given app_id.
```
FrogTheFrog, line 113:
```suggestion
     * @brief Get the image path of the application with the given app_id.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @brief Gets the list of applications.
+     * @return A reference to the vector of applications.
+     */
+    std::vector<ctx_t>& get_apps();
+
+    /**
+     * @brief Sets the list of applications.
+     * @param apps The new list of applications.
+     * @note This will overwrite the existing list of applications.
+     * @see refresh(const std::string &file_name)
```
FrogTheFrog, line 108: @ReenigneArcher I thing `@see` is incorrect here as you had some special tag.

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+
+    /**
+     * @brief Gets the image path of the application with the given app_id.
+     * @param app_id The ID of the application.
+     * @return The image path of the application.
+     */
     std::string get_app_image(int app_id);
+
+    /**
+     * @brief Gets the name of the last run application.
```
FrogTheFrog, line 120:
```suggestion
     * @brief Get the name of the last run application.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @brief Gets the environment variables.
+     * @return A constant reference to the environment variables.
+     */
+    const boost::process::v1::environment& get_env() const;
+
+    /**
+     * @brief Gets the environment variables.
+     * @return A reference to the environment variables.
+     */
+    boost::process::v1::environment& get_env();
```
FrogTheFrog, line 135: I suggest to remove this method.

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
     std::string get_app_image(int app_id);
+
+    /**
+     * @brief Gets the name of the last run application.
+     * @return The name of the last run application.
+     */
     std::string get_last_run_app_name();
+
+    /**
+     * @brief Gets the environment variables.
```
FrogTheFrog, line 126:
```suggestion
     * @brief Get the environment variables.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @return A constant reference to the environment variables.
+     */
+    const boost::process::v1::environment& get_env() const;
+
+    /**
+     * @brief Gets the environment variables.
```
FrogTheFrog, line 132:
```suggestion
     * @brief Get the environment variables.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @return A reference to the environment variables.
+     */
+    boost::process::v1::environment& get_env();
+
+    /**
+     * @brief Sets the environment variables.
```
FrogTheFrog, line 138:
```suggestion
     * @brief Set the environment variables.
```

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     * @param env The new environment variables.
+     * @note This will overwrite the existing environment variables.
+     * @see refresh(const std::string &file_name)
```
FrogTheFrog, line 141: Same as the other `@see` comment.

src/process.h
```diff
@@ -89,10 +89,62 @@ namespace proc {
...
+     */
+    void set_env(boost::process::v1::environment env);
+
+    /**
+     * @brief Terminates the currently running process.
```
FrogTheFrog, line 146:
```suggestion
     * @brief Terminate the currently running process.
```

src/process.cpp
```diff
@@ -698,7 +714,11 @@ namespace proc {
     auto proc_opt = proc::parse(file_name);
```
FrogTheFrog, line 714: I strongly suggest to change signature of `std::optional<proc::proc_t> parse(const std::string &file_name)` to `std::optional<std:tuple<std::vector<ctx_t>, boost::process::v1::environment>> parse(const std::string &file_name)` and also remove `proc_t` constructor as it is another disaster waiting to happen (don't forget to initialize `int _app_id{0};`)

src/process.cpp
```diff
@@ -698,7 +714,11 @@ namespace proc {
...
 
     if (proc_opt) {
-      proc = std::move(*proc_opt);
+      // Update the process object with the new environment and apps
+      // And, keep app running status.
+
+      proc.set_env(proc_opt->get_env());
```
FrogTheFrog, line 720: After looking into this I have found another bug related to `_env`. It is used in `terminate` function, this means we have to preserve the `_env` until process terminates :/

The moment we call `execute` we should copy the `_env` to a member variable `_app_env` (name is debatable) and replace every usage of `_env` within `execute` and `terminate` with `_app_env` (except the first one where we copy it of course). Or we could store the new ENV to `_current_env` and the in execute copy it (`_env = _current_env`). I prefer the first renaming option, but whatever is easier is ok.

FrogTheFrog (changes requested):
I've added some comment style suggestion, based on the feedback I've got from RA in my other commits.

Also I've found some other issues with the `env` that I would like you to fix. Sorry, that such a simple PR turned out to explode. If you prefer I can also take over the requested changes, just let me know.

FrogTheFrog:
Also @ReenigneArcher, why do we use `-1` here?
image

Is `0` not good enough?

ReenigneArcher:
Not sure exactly where that code comes from, but 0 is the first app.

FrogTheFrog:
> Not sure exactly where that code comes from, but 0 is the first app.

But then we have a lot of places in the code checking `running() > 0` to determine if any app is running. Which is incorrect...

ReenigneArcher:
That whole system problem needs to be re-thought. As far as I see, `app_id` is not even real, it's more like an app index.

MiroKaku:
> I've added some comment style suggestion, based on the feedback I've got from RA in my other commits.
> 
> Also I've found some other issues with the `env` that I would like you to fix. Sorry, that such a simple PR turned out to explode. If you prefer I can also take over the requested changes, just let me know.

I would prefer if you could take over the requested changes. Thank you! 😀

FrogTheFrog:
@ReenigneArcher I will move some data around without changing logic too much. This should automatically resolve all of the known issues.

ReenigneArcher:
> @MiroKaku Have you perhaps unchecked the checkbox where maintainers or something can push to the branch? If so, can you enable it?
> 
> Otherwise I will have to create a fork :/

It is checked, but I don't think it allows "Collaborators" to push to.

FrogTheFrog:
I will create a fork then.

Yundi339:
It's a bit embarrassing 😂😂😂, our fork repository included this pr a long time ago, to solve the problem described in pr.
Since we merge it, it has been used well. Until recently I tested some special use scenarios.
I found that modifying apps in streams causes env information to be lost.
So I submitted a new PR #4654 , but I realized that there was no merge this PR #3417 .

ReenigneArcher:
@Yundi339 No worries. Unfortunately, both this PR and the replacement PR (https://github.com/LizardByte/Sunshine/pull/3604) went stale.

I would accept a combined PR that includes these fixes as well as what your PR solves, if you're willing.

PR #3455, opened 2024-12-16
feat(web-ui): replace dropdown menus with checkboxes 

FrogTheFrog:
## Description
Replaced the enabled/disabled dropdowns with checkboxes.
Except this one as it has custom dropdown values and it seems that it could possible have more values in the future:
image

Additionally fixed incorrectly labeled default values, some missing/mis-configured translation stuff and styling issues.

##### Before
image
##### After
image

##### Before 1
image
##### After 1
image
##### Before 2
image
##### After 2
image
##### Before 3
image
##### After 3
image
##### Before 4
image
##### After 4
image

##### Before
image
##### After
image

##### Before
image
##### After
image

##### Before
image
##### After
image

##### Before
image
##### After
image

##### Before
image
##### After
image

##### Before
image
##### After
image

##### Before
image
##### After
image

##### Before
image
image
##### After
image
image

## Type of Change

## Checklist


src_assets/common/assets/web/apps.html
```diff
@@ -209,21 +205,21 @@ <h1>{{ $t('apps.applications_title') }}</h1>
         </div>
         <!-- elevation -->
         <div class="mb-3 form-check" v-if="platform === 'windows'">
-          <label for="appElevation" class="form-check-label">{{ $t('_common.run_as') }}</label>
+          <label for="appElevation" class="form-check-label">{{ $t('_common.run_as') }}<div class="mt-0 form-text">{{ $t('_common.disabled_def_cbox') }}</div></label>
```
ReenigneArcher, line 208:
```suggestion
          <label for="appElevation" class="form-check-label">{{ $t('_common.run_as') }}
            <div class="mt-0 form-text">{{ $t('_common.disabled_def_cbox') }}</div>
          </label>
```

Can we try to avoid nested elements on a single line?

src_assets/common/assets/web/configs/tabs/Advanced.vue
```diff
@@ -80,7 +80,7 @@ const config = ref(props.config)
     </div>
 
     <!-- Encoder -->
-    <div class="mb-3">
+    <div>
```
ReenigneArcher, line 83: This will be the only section on this page without `mb-3`?

I think you probably did this because it's the last one, but I fear it will get replicated because new sections are often copied and pasted.

Comment applies on many of the .vue pages.

src_assets/common/assets/web/configs/tabs/AudioVideo.vue
```diff
@@ -54,12 +54,9 @@ const config = ref(props.config)
...
         <!-- Install Steam Audio Drivers -->
-        <div class="mb-3">
-          <label for="install_steam_audio_drivers" class="form-label">{{ $t('config.install_steam_audio_drivers') }}</label>
-          <select id="install_steam_audio_drivers" class="form-select" v-model="config.install_steam_audio_drivers">
-            <option value="disabled">{{ $t('_common.disabled') }}</option>
-            <option value="enabled">{{ $t('_common.enabled_def') }}</option>
-          </select>
+        <div class="mb-3 form-check">
+          <label for="install_steam_audio_drivers" class="form-label">{{ $t('config.install_steam_audio_drivers') }}<div class="mt-0 form-text">{{ $t('_common.enabled_def_cbox') }}</div></label>
+          <input type="checkbox" class="form-check-input" id="install_steam_audio_drivers" v-model="config.install_steam_audio_drivers" true-value="enabled" false-value="disabled" />
```
ReenigneArcher, line 59:
```suggestion
          <input
            type="checkbox"
            class="form-check-input"
            id="install_steam_audio_drivers"
            v-model="config.install_steam_audio_drivers"
            true-value="enabled"
            false-value="disabled"
          />
```

When the line is too long can we do this?

Also, do the true/false values account for all the possible options in the config? Ideally in the cpp we would translate enabled/disabled to boolean before providing it to the web-ui.

src_assets/common/assets/web/configs/tabs/Files.vue
```diff
@@ -47,7 +47,7 @@ const config = ref(props.config)
     </div>
 
     <!-- State File -->
-    <div class="mb-3">
```
ReenigneArcher, line 50: Same comment

ReenigneArcher (changes requested):
I still need to review the input accordion a little closer.

src_assets/common/assets/web/configs/tabs/Advanced.vue
FrogTheFrog: Ok, will undo this.

src_assets/common/assets/web/apps.html
```diff
@@ -116,13 +116,9 @@ <h1>{{ $t('apps.applications_title') }}</h1>
...
-        <div class="mb-3">
-          <label for="excludeGlobalPrep" class="form-label">{{ $t('apps.global_prep_name') }}</label>
-          <select id="excludeGlobalPrep" class="form-select" v-model="editForm['exclude-global-prep-cmd']">
-            <option v-for="val in [false, true]" :value="val">
-              {{ !val ? $t('_common.enabled') : $t('_common.disabled') }}
-            </option>
-          </select>
+        <div class="mb-3 form-check">
+          <label for="excludeGlobalPrep" class="form-label">{{ $t('apps.global_prep_name') }}<div class="mt-0 form-text">{{ $t('_common.enabled_def_cbox') }}</div></label>
+          <input type="checkbox" class="form-check-input" id="excludeGlobalPrep" v-model="editForm['exclude-global-prep-cmd']" true-value="false" false-value="true" />
```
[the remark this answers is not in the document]
FrogTheFrog: It probably should, however this involves more work than just replacing dropdown list with checkbox, which I would like not to do here.

src_assets/common/assets/web/configs/tabs/AudioVideo.vue
FrogTheFrog: > When the line is too long can we do this?

I'll try to come up with something better.

> Also, do the true/false values account for all the possible options in the config? Ideally in the cpp we would translate enabled/disabled to boolean before providing it to the web-ui.

As you have said, only 1 value pair will be accounted for. In case none of the values match, it defaults to `false`. Again, this requires more refactoring due to separate parsing in cpp/vue. I do not want to do this here.

src_assets/common/assets/web/Checkbox.vue
```diff
@@ -0,0 +1,108 @@
...
+
+    // Try mapping strings next (first in the list will be used as fallback)
+    const stringPairs = [
+      ["true", "false"],
+      ["1", "0"],
+      ["enabled", "disabled"],
+      ["enable", "disable"],
+      ["yes", "no"],
+      ["on", "off"]
+    ];
```
ReenigneArcher, line 64: Is there a way we can migrate values over to true/false whenever they make the next save?

ReenigneArcher (commented):
Sorry for the delay, I didn't want to disturb you over the holidays.

src_assets/common/assets/web/Checkbox.vue
FrogTheFrog: Unfortunately, it is not that trivial:

1. Each page handles it's own configuration and has it's own "default" values. Ideally this place (and only) needs to be adapted to migrate the values, which leads me to the next point.
2. Sunshine gives the config as a `map<string, string>` to the UI, whereas the default values from the UI can also be boolean literals (not strings). However, then the literals are saved as strings!? It's a mess...

Based on those 2 points I would strongly suggest against trying Checkbox do any modification (if that was your intention) and properly cleanup the deserialize/serialize code for each page in the UI, but not now plz...

src_assets/common/assets/web/configs/tabs/AudioVideo.vue
```diff
@@ -54,14 +55,12 @@ const config = ref(props.config)
...
-          <label for="install_steam_audio_drivers" class="form-label">{{ $t('config.install_steam_audio_drivers') }}</label>
-          <select id="install_steam_audio_drivers" class="form-select" v-model="config.install_steam_audio_drivers">
-            <option value="disabled">{{ $t('_common.disabled') }}</option>
-            <option value="enabled">{{ $t('_common.enabled_def') }}</option>
-          </select>
-          <div class="form-text">{{ $t('config.install_steam_audio_drivers_desc') }}</div>
-        </div>
+        <Checkbox class="mb-3"
+                  id="install_steam_audio_drivers"
+                  locale-prefix="config"
```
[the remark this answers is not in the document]
FrogTheFrog: It is not always needed. It is used to generated default label and description when they are not provided:
```
const labelField = props.label ?? `${props.localePrefix}.${props.id}`;
const descField = props.desc ?? `${props.localePrefix}.${props.id}_desc`;
```

I noticed that in most of the places the following pattern was used `<some_prefix>.<item_id>...`, so instead of specifying `id, label, desc` I thought it would be simple to just specify `id, prefix`.

I set the default to `missing-prefix` so that when the translation fails it is easily visible that you have missed something.

src_assets/common/assets/web/configs/tabs/Inputs.vue
```diff
@@ -138,56 +125,49 @@ const config = ref(props.config)
...
-        <option value="enabled">{{ $t('_common.enabled_def') }}</option>
-      </select>
-      <div class="form-text">{{ $t('config.key_rightalt_to_key_win_desc') }}</div>
-    </div>
+    <Checkbox v-if="config.keyboard === 'enabled'"
+              class="mb-3"
+              id="key_rightalt_to_key_win"
+              locale-prefix="config"
+              v-model="config.key_rightalt_to_key_win"
+              default="false"
```
ReenigneArcher, line 142:
```suggestion
              default="true"
```

I think this one is reversed?

ReenigneArcher (commented):
I think there may be a slight bug.

This one should be checked, but isn't.
image

The next ones are correctly unchecked by the "Default" text is wrong. Not sure the fix on these ones.
image

image

image

image

image



Also, these options do not appear if DS4 is selected.

image

Perhaps the options shown should apply if set to Automatic OR DS4?

src_assets/common/assets/web/configs/tabs/Inputs.vue
FrogTheFrog: Actually no, according to source code it was always meant to be false by default:
image

FrogTheFrog:
> This one should be checked, but isn't.

This was always meant to be set to `false` by default (check the other reply).

> The next ones are correctly unchecked by the "Default" text is wrong. Not sure the fix on these ones.

Fixed this one

> Also, these options do not appear if DS4 is selected.

They are not supposed to appear since we already "know" that we will be emulating DS4 controller.

Also, the lonely DS4-only checkbox is also only meant to appear when DS4 is selected manually:
image

ReenigneArcher (approved):

PR #3477, opened 2024-12-22
fix(input): Removed batching of move events of PEN_PACKET

aceradryd:
## Description
The changes ensure that move events from pens are no longer batched.
This is necessary when Sunshine is used to work with graphics programs.

### Issues Fixed

[moonlight-stream/moonlight-android#1427](https://github.com/moonlight-stream/moonlight-android/issues/1427)

### Associated pull requests
[moonlight-stream/moonlight-common-c#99](https://github.com/moonlight-stream/moonlight-common-c/pull/99)

## Type of Change

## Checklist

### DIFF
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -1392,9 +1392,9 @@ namespace input {
    */
   batch_result_e
   batch(PSS_PEN_PACKET dest, PSS_PEN_PACKET src) {
-    // Only batch hover or move events
-    if (dest->eventType != LI_TOUCH_EVENT_MOVE &&
-        dest->eventType != LI_TOUCH_EVENT_HOVER) {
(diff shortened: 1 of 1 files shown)


cgutman:
When I was testing the the pen support during initial implementation, I encountered some cases where injecting pen input would take a non-trivial amount of time per event and lead to a massive backlog of input events that caused input latency. Unfortunately, it's not always easy to reproduce (perhaps it depends on the window message processing latency of the focused application). We should probably have _some_ batching to ensure we aren't sending pen events at a completely insane rate that the OS can't keep up with.

Douvahkiin:
@cgutman Is it possible to make it an option? User can choose whether to have batching or not.

PR #3513, opened 2025-01-04
fix(rtp): improve timestamp accuracy

andygrundman:
## Background

I am [experimenting](https://github.com/andygrundman/moonlight-ios/tree/andyg.ios-spatial-audio) with the iOS AVSampleBufferAudioRenderer and AVSampleBufferRenderSynchronizer APIs, which include the ability to use RTP timestamps as a way to pace playback, provide a/v sync that won't drift, use variable-speed playback to cover for short glitches (such as Game Mode's drop in Bluetooth latency), proper spatial audio support for free, and more. It should also be nicely compatible with the AVSampleBufferDisplayLayer API currently used by the iOS video code.  The biggest challenge with these APIs is their higher latency, but I'm hopeful it can get to a reasonable level. Worst case, it could be a good option when you are willing to trade a little latency for solid playback.

## Description

* Video: instead of using now() when the RTP packet is created, use the earlier packet->frame_timestamp that we're already collecting for host latency stats. This timestamp should be more accurate to when we captured the frame. I am not sure if all backends support this timestamp, so there is a fallback to the current method.
* Audio: fix a bug where the RTP timestamps stop advancing when no audio is being sent. Audio now uses the same timer-based source for this field, but remains as milliseconds. Because of this bug, I am pretty sure no client is using these timestamps for anything.
* Both use steady_clock and microseconds / 1000 to maintain precision. Video still uses 90khz time base per the spec, and audio is effectively at packet duration resolution of 5ms/10ms.

Known issue: I had to comment out one assert in Moonlight, because this patch changes one of the FEC timestamps. The FEC RFC says "The timestamp SHALL be set to a time corresponding to the repair packet's transmission time.  Note that the timestamp value has no use in the actual FEC protection process and is usually useful for jitter calculations.", so I don't think this assert is correct anyway. Even though this shouldn't affect Moonlight release builds which don't use asserts or the extra FEC verification code, this probably needs to be resolved before this can be accepted, either by only sending correct timestamp packets to newer Moonlight versions, or fixing it in the FEC code. (Not to get too sidetracked but was the Opus FEC support ever tried? I sometimes wonder if the 40% FEC overhead on audio is too high for benefit.)

RtpAudioQueue.c:300
```LC_ASSERT_VT(existingBlock->fecHeader.baseTimestamp == fecBlockBaseTs);```

## Type of Change

## Checklist

### DIFF
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -366,12 +366,12 @@ namespace stream {
 
     struct {
       crypto::cipher::cbc_t cipher;
-      std::string ping_payload;
 
-      std::uint16_t sequenceNumber;
+      std::string ping_payload;
       // avRiKeyId == util::endian::big(First (sizeof(avRiKeyId)) bytes of launch_session->iv)
       std::uint32_t avRiKeyId;
-      std::uint32_t timestamp;
+      std::uint16_t sequenceNumber;
+      std::chrono::steady_clock::time_point timestamp_epoch;
       udp::endpoint peer;
 
       util::buffer_t<char> shards;
@@ -1274,7 +1274,7 @@ namespace stream {
   videoBroadcastThread(udp::socket &sock) {
     auto shutdown_event = mail::man->event<bool>(mail::broadcast_shutdown);
     auto packets = mail::man->queue<video::packet_t>(mail::video_packets);
-    auto timebase = boost::posix_time::microsec_clock::universal_time();
+    auto timestamp_epoch = std::chrono::steady_clock::now();
 
     // Video traffic is sent on this thread
     platf::adjust_thread_priority(platf::thread_priority_e::high);
@@ -1479,8 +1479,13 @@ namespace stream {
             auto *inspect = (video_packet_raw_t *) shards.data(x);
 
             // RTP video timestamps use a 90 KHz clock
-            auto now = boost::posix_time::microsec_clock::universal_time();
-            auto timestamp = (now - timebase).total_microseconds() / (1000 / 90);
+            auto timestamp = static_cast<std::uint32_t>(
+              std::chrono::duration_cast<std::chrono::microseconds>(
+                packet->frame_timestamp
+                  ? *packet->frame_timestamp - timestamp_epoch
+                  : std::chrono::steady_clock::now() - timestamp_epoch // is this fallback needed?
+              ).count() / (1000.0 / 90)
+            );
 
             inspect->packet.fecInfo =
               (x << 12 |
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by andygrundman** [comment]
Oops, I changed the title in the commit, but forgot to do the PR. I guess it's still too long anyway:

```
fix(rtp): maintain audio timestamp continuity even during periods of silence. Measure video timestamps closer to the time of capture by reusing host latency's frame_timestamp.
```

**Comment by cgutman** [comment]
> Because of this bug, I am pretty sure no client is using these timestamps for anything.

I believe that's accurate. The clients that I know of that try to pace audio do it via monitoring the amount of pending audio (either in moonlight-common-c's queue, in the OS audio stack, or both).

> Known issue: I had to comment out one assert in Moonlight, because this patch changes one of the FEC timestamps. The FEC RFC says "The timestamp SHALL be set to a time corresponding to the repair packet's transmission time. Note that the timestamp value has no use in the actual FEC protection process and is usually useful for jitter calculations.", so I don't think this assert is correct anyway.

AFAIK, Nvidia did their own thing and didn't implement any known RTP FEC standard, so it's asserting what GFE and current Sunshine do.

IIUC, what is happening in your case is that you hit the (common) case where a data packet is received first and we synthesize an FEC header using this code:
```
        // This is a data packet, so we will need to synthesize an FEC header
        fecBlockPayloadType = packet->packetType;
        fecBlockBaseSeqNum = (packet->sequenceNumber / RTPA_DATA_SHARDS) * RTPA_DATA_SHARDS;
        fecBlockBaseTs = packet->timestamp - ((packet->sequenceNumber - fecBlockBaseSeqNum) * AudioPacketDuration);
        fecBlockSsrc = packet->ssrc;
```

Since you've adjusted the logic to no longer guarantee that there is `AudioPacketDuration` milliseconds between the timestamps in each packet, the synthesized FEC header won't always match the real thing that we will get later. If this an accurate description of what you saw going on, then I think removing the assert is fine.

> Even though this shouldn't affect Moonlight release builds which don't use asserts or the extra FEC verification code, this probably needs to be resolved before this can be accepted, either by only sending correct timestamp packets to newer Moonlight versions, or fixing it in the FEC code.

Did you have some idea of how you would "fix it"? AFAICT, it seems basically intrinsic to this model where timestamps are real measured things and not just a glorified counter.

> Not to get too sidetracked but was the Opus FEC support ever tried? I sometimes wonder if the 40% FEC overhead on audio is too high for benefit.

I know Nvidia didn't use it. Given how tiny of a fraction that audio data is of the total traffic, it probably made sense to use the existing Reed-Solomon FEC scheme they were already using for video. They didn't even bother with a variable FEC scheme for audio, unlike video FEC which can change based on client feedback.

File: src/stream.cpp [hunk 1]
```diff
@@ -1629,7 +1634,12 @@ namespace stream {
       auto session = (session_t *) channel_data;
 
       auto sequenceNumber = session->audio.sequenceNumber;
-      auto timestamp = session->audio.timestamp;
+      // Audio timestamps are in milliseconds and should be AudioPacketDuration (5ms or 10ms) apart
+      auto timestamp = static_cast<std::uint32_t>(
+        std::chrono::duration_cast<std::chrono::microseconds>(
+          std::chrono::steady_clock::now() - session->audio.timestamp_epoch
```
cgutman, line 1640: I think it's possible to get capture time from the OS too like we do with video frames. `IAudioCaptureClient::GetBuffer()` takes an optional `pu64QPCPosition` parameter that gives you the timestamp of capture.

andygrundman:
I will be opening a new PR that focuses only on video timestamps.

PR #3529, opened 2025-01-11
feat(display): add display mode remapping option

FrogTheFrog:
## Description
This is part 3 of https://github.com/LizardByte/Sunshine/pull/3441.

This PR adds an option to remap requested resolution and/or FPS to another resolution and/or refresh rate.

### Screenshot
When remapping is possible for both:
image

When remapping is possible for only resolution:
image

When remapping is possible for only FPS:
image

## Type of Change

## Checklist


moi952:
> @moi952 could I ask you to please test this build. Just if the remapping still works? I don't have access to host PC at the moment :/

Yes of course, I'm not at home, but I can test when I get back if it's not too late, or tomorrow morning.
Thanks

moi952:
> @moi952 could I ask you to please test this build. Just if the remapping still works? I don't have access to host PC at the moment :/

I just got back, it works perfectly

FrogTheFrog:
> Shouldn't **Final** fields at least contain some default safe value if they can not be left empty, like a minimum of 60fps?

One of them can be empty.

> Also, entering resolutions by hand is error prone and a user can easily have a typo. I would say, a drop down (combo box) with list of resolutions to choose from is better with a default safe value like 1080p.

I like the combobox idea.

RebelliousX:
> > Shouldn't **Final** fields at least contain some default safe value if they can not be left empty, like a minimum of 60fps?
> 
> One of them can be empty.

Then the explanation text should have bold text or brighter color (light yellow or red) to explain that to bring the user's attention, otherwise it could be easily missed.

moi952:
> I will test this soon, but at first glance from pictures. Shouldn't **Final** fields at least contain some default safe value if they can not be left empty, like a minimum of 60fps? Also, entering resolutions by hand is error prone and a user can easily have a typo. I would say, a drop down (combo box) with list of resolutions to choose from is better with a default safe value like 1080p.

The idea of ​​a drop-down list is interesting at first glance, but the problem is that there are very strange resolutions like those of iPhones and other smartphones.
On the other hand, if it is possible to retrieve the list of resolutions available by the target screen, it can be interesting for the target resolution.

For the received resolution (or base it is too complicated because too many possibilities)

In the future, maybe it would be interesting to add an option to switch to a target resolution if the resolution / framerate requested by the client does not pass on the target screen, rather than using the initial screen and resolution (the behavior when the resolution requested by the client does not exist)

RebelliousX:
> > I will test this soon, but at first glance from pictures. Shouldn't **Final** fields at least contain some default safe value if they can not be left empty, like a minimum of 60fps? Also, entering resolutions by hand is error prone and a user can easily have a typo. I would say, a drop down (combo box) with list of resolutions to choose from is better with a default safe value like 1080p.
> 
> The idea of ​​a drop-down list is interesting at first glance, but the problem is that there are very strange resolutions like those of iPhones and other smartphones. On the other hand, if it is possible to retrieve the list of resolutions available by the target screen, it can be interesting for the target resolution.

We can workaround this by having the last item in the combobox says "Custom Resolution". And if it is chose, then a text box for the resolution appears to enter custom value if the user wants.

ReenigneArcher:
Thank you everyone for testing and giving feedback on this PR, but let's not make these PRs grow too large. It causes burnout for both the contributors and maintainers. The UI can be improved in a later PR, and this is not a small ask since we don't have any patterns in place for resolution inputs currently. The old resolution input also did not have any input validation and strictly relied on the user entering the values correctly.

FrogTheFrog:
As RA has stated, this will be UI for the 1st iteration.

You are welcome to improve it later on as I'm already burned out and no longer plan to work on a feature that I would not be even using in the end.

ReenigneArcher (approved):

PR #3530, opened 2025-01-12
feat(capture/windows): hook APIs to avoid output reparenting that breaks DDA

cgutman:
## Description
This is a new approach for handling multi-GPU systems that could eliminate the need for ddprobe.exe and all the issues that come with locking the process to a single primary GPU. Rather than using registry values, we are hooking the underlying `NtGdiDdDDIGetCachedHybridQueryValue()` function that DXGI uses internally to track the GPU preference of the process and always returning `D3DKMT_GPU_PREFERENCE_STATE_UNSPECIFIED`. In my testing, this appears to completely eliminate the DXGI output reparenting (`dxgi!ReparentOutputs()`) that causes DDA to break.

This PR contains commits that completely remove ddprobe.exe and all options related to it. If we find a case that this new approach doesn't cover, we can drop those commits and preserve the options to use the legacy ddprobe.exe method.

In my testing on a multi-GPU desktop (AMD+Intel) and laptop (Intel+Nvidia) both running Win11 24H2, capture works properly regardless of which GPU the monitors are connected to or which GPU is set as default. The only restriction is that you can't use the monitor swap hotkeys to swap between monitors attached to different GPUs while streaming due to lack of support for switching encoders while streaming (unless you force software encoding).

I'd appreciate if @RebelliousX and @Nonary can test to ensure this doesn't break their streaming scenarios with virtual displays.

## Type of Change

## Checklist


Nonary:
Does that mean we rely entirely on hooks now, since we’re no longer using the standard DXGI outputs? If so, won’t anti-cheat systems potentially block the hook? And if the hook fails, does that mean we’re unable to capture anything at all?

I had figured the only way to realistically fix the multi-GPU problem would be to refactor sunshine to capture in a separate process and then used shared memory, which is way out of my skill level.

src_assets/common/assets/web/configs/tabs/audiovideo/DisplayDeviceOptions.vue
```diff
@@ -108,7 +108,7 @@ const config = ref(props.config)
                 <label for="dd_config_revert_delay" class="form-label">
                   {{ $t('config.dd_config_revert_delay') }}
                 </label>
-                <input type="number" class="form-control" id="dd_config_revert_delay" placeholder="3000" min="0"
+                <input type="text" class="form-control" id="dd_config_revert_delay" placeholder="3000"
```
FrogTheFrog, line 111:
```suggestion
                <input type="number" class="form-control" id="dd_config_revert_delay" placeholder="3000" min="0"
```

Please keep this change that came with the reverted PR

src/platform/windows/display_base.cpp
```diff
@@ -1105,11 +1024,7 @@ namespace platf {
 
     BOOST_LOG(debug) << "Detecting monitors..."sv;
 
-    // We must set the GPU preference before calling any DXGI APIs!
     const auto output_name { display_device::map_output_name(config::video.output_name) };
```
FrogTheFrog, line 1027:
```suggestion
```

This looks like an unused variable.

src/platform/windows/display_base.cpp
```diff
@@ -510,6 +413,18 @@ namespace platf::dxgi {
     return false;
   }
 
+  HRESULT
```
FrogTheFrog, line 416: Should probably copy-paste the explanation from PR explaining why these hooks are needed.

RebelliousX:
@cgutman I would love to test. But Windows build failed, I can't download artifact. 🤔

src/platform/windows/display_base.cpp
```diff
@@ -519,13 +434,22 @@ namespace platf::dxgi {
...
+        if (f) {
+          f(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
+        }
+
+        FreeLibrary(user32);
       }
 
-      FreeLibrary(user32);
+      {
+        MH_Initialize();
```
FrogTheFrog, line 448: We are not calling "uninitialize" which could be reported as a memleak during program exit. Which probably doesn't matter as the OS will cleanup, but maybe a comment could be added?

FrogTheFrog (changes requested):
Code-wise LGTM apart from a few minor requests.

FrogTheFrog:
> @cgutman I would love to test. But Windows build failed, I can't download artifact. 🤔

Please get the build from https://github.com/FrogTheFrog/Sunshine/pull/4, where I hope I have fixed the build issue.

RebelliousX:
> > @cgutman I would love to test. But Windows build failed, I can't download artifact. 🤔
> 
> Please get the build from [FrogTheFrog#4](https://github.com/FrogTheFrog/Sunshine/pull/4), where I hope I have fixed the build issue.

Yours actually worked! That doesn't make any sense.. I am starting to lose faith in computer science, it only took a few "LOL" comments to make the build succeed 🤔

ReenigneArcher:
Doc failures are working as intended. FYI:
https://github.com/LizardByte/doxyconfig/blob/4c9452482bd01cb36764dc914d4537b278ad4218/doxyconfig-Doxyfile#L136-L142

src/platform/windows/display_base.cpp
cgutman: Yep, added one

src/platform/windows/display_base.cpp
cgutman: Added doxygen and a longer form comment to the body of the function.

src/platform/windows/display_base.cpp
```diff
@@ -510,6 +413,22 @@ namespace platf::dxgi {
     return false;
   }
 
+  /**
+   * @brief Hook for NtGdiDdDDIGetCachedHybridQueryValue() from win32u.dll.
+   * @param gpuPreference On output, the preference value.
```
FrogTheFrog, line 418: I'm not sure I understand this description. Is it meant to be like `An output for the preference value.`?

cgutman:
> Does that mean we rely entirely on hooks now, since we’re no longer using the standard DXGI outputs? If so, won’t anti-cheat systems potentially block the hook? And if the hook fails, does that mean we’re unable to capture anything at all?

We've basically just moved from one undocumented GPU preference mechanism (writing UserGpuPreferences registry values manually) to another (hooking the underlying API for preference selection). Both rely on OS implementation details that could break in an OS update, but there's not an official API to request this behavior so we're stuck with it.

We're only hooking APIs in our own process, so anti-cheat won't care about it. GPU preferences on the rest of the system will still behave normally.

src/platform/windows/display_base.cpp
```diff
@@ -510,6 +413,22 @@ namespace platf::dxgi {
...
+   * @return S_OK always.
+   */
+  HRESULT
+  __stdcall NtGdiDdDDIGetCachedHybridQueryValueHook(D3DKMT_GPU_PREFERENCE_QUERY_STATE *gpuPreference) {
+    // By faking a cached GPU preference state of D3DKMT_GPU_PREFERENCE_STATE_UNSPECIFIED, this will
+    // prevent DXGI from performing the normal GPU preference resolution that looks at the registry,
+    // power settings, and the hybrid adapter DDI interface to pick a GPU. Instead, we will not be
+    // bound to any specific GPU. This will prevent DXGI from performing output reparenting (moving
+    // outputs from their true location to the render GPU), which breaks DDA.
+    *gpuPreference = D3DKMT_GPU_PREFERENCE_STATE_UNSPECIFIED;
```
FrogTheFrog, line 428: Should probably handle a `nullptr` somehow.

FrogTheFrog (changes requested):

src/platform/windows/display_base.cpp
cgutman: When I test the real syscall, it returns `STATUS_INVALID_PARAMETER` in that case, so I'll do that.

src/platform/windows/display_base.cpp
cgutman: Will clarify.

FrogTheFrog (approved):
LGTM

cgutman:
I've tested this in Intel+AMD and Intel+Nvidia scenarios on Win10 22H2 and it seems to work fine there too.

PR #3558, opened 2025-01-18
fix(httpcommon): sonarqube warning cleanup

CodyManess:
## Description
Fixes most of the sonarqube warnings on httpcommon.  Was able to fix one of the security issues regarding curl_easy_init, but the other one does not want to go away.  Since it's a false alarm, I added the no sonar comment to prevent it from affecting the security score.

## Type of Change

## Checklist

### DIFF
--- a/src/httpcommon.cpp
+++ b/src/httpcommon.cpp
@@ -58,7 +58,9 @@ namespace http {
         create_creds(config::nvhttp.pkey, config::nvhttp.cert)) {
       return -1;
     }
-    if (user_creds_exist(config::sunshine.credentials_file) && reload_user_creds(config::sunshine.credentials_file)) {
+    if (!user_creds_exist(config::sunshine.credentials_file)) {
+      BOOST_LOG(info) << "Open the Web UI to set your new username and password and getting started";
+    } else if (reload_user_creds(config::sunshine.credentials_file)) {
       return -1;
     }
     return 0;
@@ -106,7 +108,6 @@ namespace http {
       BOOST_LOG(error) << "validating user credentials: "sv << e.what();
     }
 
-    BOOST_LOG(info) << "Open the Web UI to set your new username and password and getting started";
     return false;
   }
 
@@ -175,9 +176,9 @@ namespace http {
     return 0;
   }
 
-  bool download_file(const std::string &url, const std::string &file) {
+  bool download_file(const std::string &url, const std::string &file, long ssl_version) {
     // sonar complains about weak ssl and tls versions; however sonar cannot detect the fix
-    CURL *curl = curl_easy_init(); // NOSONAR
+    CURL *curl = curl_easy_init();  // NOSONAR
     if (!curl) {
       BOOST_LOG(error) << "Couldn't create CURL instance";
       return false;
@@ -196,7 +197,7 @@ namespace http {
       return false;
     }
 
-    curl_easy_setopt(curl, CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_3); // NOSONAR
+    curl_easy_setopt(curl, CURLOPT_SSLVERSION, ssl_version);  // NOSONAR
     curl_easy_setopt(curl, CURLOPT_URL, url.c_str());
     curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, fwrite);
     curl_easy_setopt(curl, CURLOPT_WRITEDATA, fp);
--- a/src/httpcommon.cpp
+++ b/src/httpcommon.cpp
@@ -54,17 +54,12 @@ namespace http {
       config::nvhttp.pkey = (dir / ("pkey-"s + unique_id)).string();
     }
 
-    if (!fs::exists(config::nvhttp.pkey) || !fs::exists(config::nvhttp.cert)) {
-      if (create_creds(config::nvhttp.pkey, config::nvhttp.cert)) {
-        return -1;
-      }
+    if ((!fs::exists(config::nvhttp.pkey) || !fs::exists(config::nvhttp.cert)) &&
+        create_creds(config::nvhttp.pkey, config::nvhttp.cert)) {
+      return -1;
     }
-    if (user_creds_exist(config::sunshine.credentials_file)) {
-      if (reload_user_creds(config::sunshine.credentials_file)) {
-        return -1;
-      }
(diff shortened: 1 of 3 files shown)


ReenigneArcher:
Thank you! I'll take a look at this in a few days. We just published a stable release so I want to have a little cooldown period in case any bug reports start flying in. I also have a few PRs where I'm working on this same file, so I will need to decide what order to merge them in (I don't think there will be any difficult conflicts though).

ReenigneArcher:
@CodyManess I apologize for this, but we adjusted our clang-format rules after the last stable release (since PR activity was pretty lite). This will require fixing some conflicts, luckily this PR is pretty small and there are not too many. Are you comfortable resolving the conflicts or would like some assistance with that?

CodyManess:
@ReenigneArcher No worries, I'd be happy to get it resolved.  Should get it fixed shortly.

ReenigneArcher (changes requested):

src/httpcommon.cpp
```diff
@@ -180,18 +178,14 @@ namespace http {
...
 
   bool download_file(const std::string &url, const std::string &file) {
-    CURL *curl = curl_easy_init();
-    if (curl) {
-      // sonar complains about weak ssl and tls versions
-      // ideally, the setopts should go after the early returns; however sonar cannot detect the fix
-      curl_easy_setopt(curl, CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_2);
-    } else {
+    // sonar complains about weak ssl and tls versions; however sonar cannot detect the fix
+    CURL *curl = curl_easy_init(); // NOSONAR
```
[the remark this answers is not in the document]
CodyManess: There's a few forum posts about [the issue] (https://community.sonarsource.com/t/use-stronger-ssl-and-tls-versions-when-i-use-libcurl-to-connect-to-the-sftp-server/128730/9), but I've tried to refactor different ways with sonarqube still picking it up no matter what.

src/httpcommon.cpp
```diff
@@ -111,6 +106,7 @@ namespace http {
       BOOST_LOG(error) << "validating user credentials: "sv << e.what();
     }
 
+    BOOST_LOG(info) << "Open the Web UI to set your new username and password and getting started";
```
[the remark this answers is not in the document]
CodyManess: Good catch!  I guess I somehow looked over those early returns.  Just added a new change with your recommendation.

src/httpcommon.cpp
```diff
@@ -219,17 +213,15 @@ namespace http {
   }
 
   std::string url_escape(const std::string &url) {
-    CURL *curl = curl_easy_init();
-    char *string = curl_easy_escape(curl, url.c_str(), url.length());
+    char *string = curl_easy_escape(nullptr, url.c_str(), static_cast<int>(url.length()));
     std::string result(string);
     curl_free(string);
-    curl_easy_cleanup(curl);
```
ReenigneArcher, line 226: Could you explain why this was removed?

The other returns have `curl_url_cleanup()` called right before, this return now doesn't have any cleanup?

ReenigneArcher (commented):
This looks pretty good. I just have one question. Also, there are a couple of lint errors to fix up. https://github.com/LizardByte/Sunshine/actions/runs/12893465494/job/35949970099?pr=3558#step:7:19

src/httpcommon.cpp
CodyManess: I was looking for ways to get rid of the security error associated with line 222.  Looking at the [libcurl docs](https://curl.se/libcurl/c/curl_easy_escape.html), I saw that under the History section the curl parameter is ignored with current versions of curl_easy_escape.  So it seems that we don't need to init a curl pointer to use curl_easy_escape and can just pass it a nullptr for the curl parameter, so I got rid of the init and cleanup to remove the extra code as well as the security warning.

ReenigneArcher:
Unfortunately, the unit test for the download file function is using a host that does not support TLS 1.3 (or even 1.2). https://www.ssllabs.com/ssltest/analyze.html?d=httpbin.org

This is where the test lives: https://github.com/LizardByte/Sunshine/blob/8263d8976f0ce8bc78ec31840ce6ed0e750c85f8/tests/unit/test_httpcommon.cpp#L57-L58

Should we specify it as TLS 1.1, or allow this to be passed in as an optional parameter (defaulting to TLS 1.3)?

CodyManess:
@ReenigneArcher Went the default parameter route since it's likely most hosts will be able to use and benefit from TLS1_3.

tests/unit/test_httpcommon.cpp
```diff
@@ -5,11 +5,12 @@
 #include "../tests_common.h"
 
 #include <src/httpcommon.h>
+#include <curl/curl.h>
```
ReenigneArcher, line 8:
```suggestion
// test imports
#include "../tests_common.h"

// lib imports
#include <curl/curl.h>

// local imports
#include <src/httpcommon.h>
```

ReenigneArcher: There's a few other minor linting errors as well.

ReenigneArcher:
I think these ones were missed: https://github.com/LizardByte/Sunshine/actions/runs/12900688192/job/35971645851#step:7:32

(Sorry, I know this is annoying but it ensures we have consistent styling).

ReenigneArcher:
These are actually from `clang-format`, nothing for `clang-tidy` is enforced yet (we should look into that though). If you look in the scripts directory, there is a python `requirements.txt` file, and an `update_clang_format.py` file. Ideally you would setup a venv for the project (CLion can do this for you, not sure about vscode or other IDEs), then just run the script. Note that clang-format from pypi is a little newer than the one in our CI (the action is outdated), so there might be a couple of mismatches if you install the latest clang-format.

src/httpcommon.h
```diff
@@ -4,6 +4,8 @@
  */
 #pragma once
 
+#include <curl/curl.h>
```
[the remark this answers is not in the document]
ReenigneArcher: I took care of this one through the web-ui. Should be able to merge this once all checks pass now.

ReenigneArcher (approved):

CodyManess:
Thanks for all your help and patience with this.  Looking forward to contributing again in the near future!

PR #3594, opened 2025-01-27
fix(video): allow encoder probing when there are no devices at all

FrogTheFrog:
## Description
It is possible that if there are no devices reported, that the API is not functioning properly. Make sure not to prevent encoding in such cases if we can't tell.

## Type of Change

## Checklist

### DIFF
--- a/src/display_device.cpp
+++ b/src/display_device.cpp
@@ -810,19 +810,15 @@ namespace display_device {
     });
   }
 
-  bool is_any_device_active() {
+  EnumeratedDeviceList enumerate_devices() {
     std::lock_guard lock {DD_DATA.mutex};
     if (!DD_DATA.sm_instance) {
-      // Platform is not supported, assume success.
-      return true;
+      // Platform is not supported.
+      return {};
     }
 
     return DD_DATA.sm_instance->execute([](auto &settings_iface) {
-      const auto devices {settings_iface.enumAvailableDevices()};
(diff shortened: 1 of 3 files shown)


src/video.cpp
```diff
@@ -2550,7 +2550,7 @@ namespace video {
   }
 
   int probe_encoders() {
-    if (!display_device::is_any_device_active()) {
+    if (display_device::are_all_devices_inactive()) {
       BOOST_LOG(error) << "No display devices are active at the moment! Cannot probe encoders as this could break Sunshine.";
```
ReenigneArcher, line 2554: The log message no longer makes sense.

src/display_device.cpp
```diff
@@ -819,9 +819,15 @@ namespace display_device {
 
     return DD_DATA.sm_instance->execute([](auto &settings_iface) {
       const auto devices {settings_iface.enumAvailableDevices()};
-      // If at least one device has additional info, it is active.
-      return std::any_of(std::begin(devices), std::end(devices), [](const auto &device) {
-        return static_cast<bool>(device.m_info);
+
+      // If there are no devices, then the API is probably not working...
+      if (devices.empty()) {
+        return false;
```
ReenigneArcher, line 825: Since we are checking if all devices are inactive... shouldn't this return true?

And the first early return should now return false?

Issue #3653, opened 2025-02-10
Systemd user service causes race condition


whacka:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

None

### Describe the Bug

Hi,
I'm currently using Sunshine on Fedora 41, with KDE and an NVIDIA card.
I noticed that my system was randomly crashing before reaching the desktop when signing in.
I'd enabled sunshine as a user service in systemd using systemctl --user enable sunshine.service
When first logging in, it looks like KWin and Sunshine both try to access /dev/dri/card1 at the same time. This causes a race condition and crashes KWin.
Is it possible to add functionality to wait until the system is ready, or avoid locking access to the GPU somehow?

### Host Operating System

Linux

### Operating System Version

Fedora 41 Kinoite

### Architecture

amd64/x86_64

### Sunshine commit or version

v2025.209.200029

### Package

Linux - Fedora Copr

### GPU Type

NVIDIA

### GPU Model

4070 Ti

### GPU Driver/Mesa Version

565.77

### Capture Method

KMS (Linux)

### Config

```Shell

```

### Apps

```json

```

### Relevant log output

```shell
KWin failing to access GPU:

kwin_wayland: No backend specified, automatically choosing drm
kwin_wayland: kwin_core: Failed to open /dev/dri/card1 device (Device or resource busy)
kwin_wayland: kwin_wayland_drm: failed to open drm device at "/dev/dri/card1"
kwin_wayland: kwin_wayland_drm: No suitable DRM devices have been found


Sunshine:

systemd Started sunshine.service - Sunshine self-hosted game stream host for Moonlight..
sunshine Info: Sunshine version: v2025.203.34856
sunshine Info: Package Publisher: LizardByte
sunshine Info: Publisher Website: https://app.lizardbyte.dev
sunshine Info: Get support: https://app.lizardbyte.dev/support
sunshine Error: Failed to create session:
sunshine Info: /dev/dri/card1 -> nvidia-drm
...
sunshine Warning: Failed to create system tray
sunshine Info: Screencasting with KMS
sunshine Error: Couldn't open: /dev/dri/card1: Permission denied
sunshine Error: Couldn't find monitor [0]
sunshine Error: Couldn't open: /dev/dri/card1: Permission denied
sunshine Error: Couldn't find monitor [0]
sunshine Info: Screencasting with KMS
```

[...]

PR #3670, opened 2025-02-17
fix(linux): make service start after desktop target

whacka:
## Description
Starting the Sunshine service at the same time as the desktop causes a race condition, this is currently solved with a delay.
My service file was missing the delay (#3653), but I wanted to find a solution without the delay.
I've added an After= dependency for the desktop, which should prevent the service from starting until the desktop has loaded. This avoids the need for the delay, so I've removed it.
This is working on my system (Fedora 41 with KDE), but I'm not sure whether other desktop environments need testing.

## Type of Change

## Checklist

### DIFF
--- a/packaging/linux/sunshine.service.in
+++ b/packaging/linux/sunshine.service.in
@@ -2,10 +2,9 @@
 Description=@PROJECT_DESCRIPTION@
 StartLimitIntervalSec=500
 StartLimitBurst=5
+After=xdg-desktop-autostart.target
 
 [Service]
-# Avoid starting Sunshine before the desktop is fully initialized.
-ExecStartPre=/bin/sleep 5
 @SUNSHINE_SERVICE_START_COMMAND@
 @SUNSHINE_SERVICE_STOP_COMMAND@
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Thank you for the PR, but I am not sure we can add this.

Please see https://github.com/LizardByte/Sunshine/pull/2691 and the mentioned issue.

Maybe with your changes it won't trigger this issue because it doesn't include `PartOf` and `Wants`? If we could get testing on more distros that would be good.

whacka:
Thanks for letting me know, I didn't see that there was a bug on some systems when using the dependencies. The issue looks specific to GNOME, so I'm happy to close the PR. Hopefully phasing out the hardcoded delay can reconsidered when the issue is fixed upstream.

ReenigneArcher:
@whacka do you think it might work if we didn't use the following though?

```
PartOf=graphical-session.target
Wants=xdg-desktop-autostart.target
```

These are not part of your PR, but were included when we had that previous bug. Perhaps the bug was related to just having a combination of all of them?

PR #3676, opened 2025-02-20
Add support for building on openSUSE Tumbleweed

gw3583:
## Description
Add support for building on openSUSE Tumbleweed

## Type of Change

## Checklist

### DIFF
--- a/cmake/packaging/linux.cmake
+++ b/cmake/packaging/linux.cmake
@@ -55,20 +55,27 @@ set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
             libx11-6, \
             miniupnpc, \
             openssl | libssl3")
-set(CPACK_RPM_PACKAGE_REQUIRES "\
-            ${CPACK_RPM_PLATFORM_PACKAGE_REQUIRES} \
-            libcap >= 2.22, \
-            libcurl >= 7.0, \
-            libdrm >= 2.4.97, \
-            libevdev >= 1.5.6, \
-            libopusenc >= 0.2.1, \
-            libva >= 2.14.0, \
-            libwayland-client >= 1.20.0, \
-            libX11 >= 1.7.3.1, \
-            miniupnpc >= 2.2.4, \
-            numactl-libs >= 2.0.14, \
-            openssl >= 3.0.2, \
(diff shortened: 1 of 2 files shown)


gw3583:
This is a first pass. It successfully builds an RPM on an opensuse tumbleweed fresh install, and I can install it on a fresh opensuse VM and successfully stream from it. Not sure if there's anything obvious I've missed though.

ReenigneArcher (changes requested):
Thank you for the PR!

I've left some comments on the code. In addition here are a couple more.

1. If we add a new CMAKE arg it should be added the `options.cmake` file
2. We don't actually use this `linux_build.sh` script for rpms anymore (I've only left it for self compiling)
3. I think we can support openSuse through our copr repo builds. If you can add whatever conditional logic you need to this spec (https://github.com/LizardByte/Sunshine/blob/master/packaging/linux/fedora/Sunshine.spec), I can enable those chroots in copr. I know this is a bit harder to test locally, so it's fine if we have to go through a few iterations of CI runs in the PR.

ReenigneArcher:
A user just linked this on our discord, figured I'd share here just for reference: https://build.opensuse.org/package/show/games:tools/sunshine

gw3583:
> Thank you for the PR!
> 
> I've left some comments on the code. In addition here are a couple more.
> 
>     1. If we add a new CMAKE arg it should be added the `options.cmake` file
> 
>     2. We don't actually use this `linux_build.sh` script for rpms anymore (I've only left it for self compiling)
> 
>     3. I think we can support openSuse through our copr repo builds. If you can add whatever conditional logic you need to this spec (https://github.com/LizardByte/Sunshine/blob/master/packaging/linux/fedora/Sunshine.spec), I can enable those chroots in copr. I know this is a bit harder to test locally, so it's fine if we have to go through a few iterations of CI runs in the PR.

Thanks for the feedback - will try to take a look at these in the next week or two.

ReenigneArcher:
For anyone who finds this in the future, openSUSE support should be added here: https://github.com/LizardByte/Sunshine/blob/master/packaging/linux/fedora/Sunshine.spec

PRs welcome

PR #3691, opened 2025-02-27
fix(http): lower TLS to v1.2 for downloading covers

ReenigneArcher:
## Description

Some users are unable covers due to requiring TLS v1.3, so the requirement will be lowered for now.

### Issues Fixed or Closed

Fixes ##3583

## Type of Change

## Checklist

### DIFF
--- a/src/httpcommon.h
+++ b/src/httpcommon.h
@@ -23,7 +23,7 @@ namespace http {
   );
 
   int reload_user_creds(const std::string &file);
-  bool download_file(const std::string &url, const std::string &file, long ssl_version = CURL_SSLVERSION_TLSv1_3);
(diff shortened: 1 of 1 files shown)


Danny3:
What if a user's system can handle TLS v1.3 without problems, lt's say it uses the new Debian 13 (Trixie)?
Will then TLS v1.3 be used or still TLS v1.2?
Can't it be made to fallback to TLS v1.2 only when TLS v1.3 is not available, similar to how the web browsers do it?
Thanks!

ReenigneArcher:
The logic doesn't currently exist in Sunshine to detect the maximum TLS version that both IGDB and the user's machine support. I would accept a PR adding that in though.

Also, the TLS version will be bumped to 1.3 in a few months, once Windows 10 is EOL.

PR #3704, opened 2025-03-05
fix(audio): set the bits-per-sample of Steam Streaming Speakers to 16-bit when the default audio device is 16-bit

andygrundman:
## Description
fix(audio): set the bits-per-sample of Steam Streaming Speakers to 16-bit when the default audio device is 16-bit. This may help with audio glitches when enabling the virtual device.

### Issues Fixed or Closed

Resolves https://github.com/orgs/LizardByte/discussions/682

## Type of Change

## Checklist

### DIFF
--- a/src/platform/windows/audio.cpp
+++ b/src/platform/windows/audio.cpp
@@ -117,15 +117,26 @@ namespace {
 
   using virtual_sink_waveformats_t = std::vector<WAVEFORMATEXTENSIBLE>;
 
+  /**
+   * @brief List of supported waveformats for an N-channel virtual audio device
+   * @tparam channel_count Number of virtual audio channels
+   * @returns std::vector<WAVEFORMATEXTENSIBLE>
+   * @note The list of virtual formats returned are sorted in preference order and the first valid
+   *       format will be used. All bits-per-sample options are listed because we try to match
+   *       this to the default audio device. See also: set_format() below.
+   */
   template<WORD channel_count>
   virtual_sink_waveformats_t create_virtual_sink_waveformats() {
     if constexpr (channel_count == 2) {
       auto channel_mask = waveformat_mask_stereo;
-      // only choose 24 or 16-bit formats to avoid clobbering existing Dolby/DTS spatial audio settings
+      // The 32-bit formats are a lower priority for stereo because using one will disable Dolby/DTS
+      // spatial audio mode if the user enabled it on the Steam speaker.
       return {
         create_waveformat(sample_format_e::s24in32, channel_count, channel_mask),
         create_waveformat(sample_format_e::s24, channel_count, channel_mask),
         create_waveformat(sample_format_e::s16, channel_count, channel_mask),
+        create_waveformat(sample_format_e::f32, channel_count, channel_mask),
+        create_waveformat(sample_format_e::s32, channel_count, channel_mask),
       };
     } else if (channel_count == 6) {
       auto channel_mask1 = waveformat_mask_surround51_with_backspeakers;
@@ -298,6 +309,10 @@ namespace platf::audio {
         auto waveformatext_pointer = reinterpret_cast<const WAVEFORMATEXTENSIBLE *>(mixer_waveformat.get());
         capture_waveformat.dwChannelMask = waveformatext_pointer->dwChannelMask;
       }
+
+      BOOST_LOG(info) << "Audio mixer format is "sv << mixer_waveformat->wBitsPerSample << "-bit, "sv
+                      << mixer_waveformat->nSamplesPerSec << " Hz, "sv
+                      << ((mixer_waveformat->nSamplesPerSec != 48000) ? "will be resampled to 48000 by Windows"sv : "no resampling needed"sv);
     }
 
     status = audio_client->Initialize(
@@ -315,7 +330,7 @@ namespace platf::audio {
       return nullptr;
     }
 
-    BOOST_LOG(info) << "Audio capture format is " << logging::bracket(waveformat_to_pretty_string(capture_waveformat));
+    BOOST_LOG(info) << "Audio capture format is "sv << logging::bracket(waveformat_to_pretty_string(capture_waveformat));
 
     return audio_client;
   }
@@ -793,6 +808,22 @@ namespace platf::audio {
         }
       }
 
+      // When switching to a Steam virtual speaker device, try to retain the bit depth of the
+      // default audio device. Switching from a 16-bit device to a 24-bit one has been known to
+      // cause glitches for some users.
+      int wanted_bits_per_sample = 32;
+      auto current_default_dev = default_device(device_enum);
+      if (current_default_dev) {
+        audio::prop_t prop;
+        prop_var_t current_device_format;
+
+        if (SUCCEEDED(current_default_dev->OpenPropertyStore(STGM_READ, &prop)) && SUCCEEDED(prop->GetValue(PKEY_AudioEngine_DeviceFormat, &current_device_format.prop))) {
+          auto *format = (WAVEFORMATEXTENSIBLE *) current_device_format.prop.blob.pBlobData;
+          wanted_bits_per_sample = format->Samples.wValidBitsPerSample;
+          BOOST_LOG(info) << "Virtual audio device will use "sv << wanted_bits_per_sample << "-bit to match default device"sv;
+        }
+      }
+
       auto &device_id = virtual_sink_info->first;
       auto &waveformats = virtual_sink_info->second.get().virtual_sink_waveformats;
       for (const auto &waveformat : waveformats) {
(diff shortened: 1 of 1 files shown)


andygrundman:
I'm going to rewrite this as a method that returns the default device format and actually matches the SSS to the device. It doesn't need to be specific to 16 bits.

andygrundman:
This may have a bad bug :( I somehow caused it to run into some errors. I realized I'm not sure that I'm safely calling `default_device(device_enum)` and will need to read more of the docs. 

Info: Reinitializing audio capture
Error: Couldn't initialize audio client for [Stereo]: [0x8889000A]
Error: Couldn't find supported format for audio
Warning: Couldn't re-initialize audio input

src/platform/windows/audio.cpp
```diff
@@ -793,6 +795,25 @@ namespace platf::audio {
...
+      // When switching to a Steam virtual speaker device, try to retain the bit depth of the
+      // default audio device. Switching from a 16-bit device to a 24-bit one has been known to
+      // cause glitches for some users.
+      int wanted_bits_per_sample = 32;
+      auto current_default_dev = default_device(device_enum);
+      if (current_default_dev) {
+        audio::prop_t prop;
+        prop_var_t current_device_format;
+
+        // clang-format off: avoid a long line and warnings about nested if statements
```
ReenigneArcher, line 807: clang format is warning about nested if statements? I don't think I've seen that before... or do you mean sonar is?

andygrundman: Sorry the if statements weren't clang warnings, just sonar. I'm still working on my C++ style, but I like the way this looks and it does clean up the if statements. (I also removed the one that was only for logging the device name). clang-format doesn't seem to have a line length limit and wanted to put this all on one line.

```
        if ( SUCCEEDED(current_default_dev->OpenPropertyStore(STGM_READ, &prop))
          && SUCCEEDED(prop->GetValue(PKEY_AudioEngine_DeviceFormat, &current_device_format.prop)) )
        {
```

Anyway, do you know if the API calls I'm making are safe? Or maybe the info I need can be found another way?

I forgot to mention one really strange bug I ran into. I use the S/PDIF device for stereo output and normally it's set to 24/48. I was switching to 16/48 for testing this. But at some point Windows must have re-selected it as default but chose the lowest (top-most) format in the list which is 16-bit 32kHz. I only noticed because my DAC displays the active samplerate and I have never seen it display 32 before. This must have happened during `reset_default_device()` but feels more like a Windows bug.

ReenigneArcher: clang format does have a line length limit. Unfortunately it's not very good. When you set it, it not only makes sure that you don't exceed that limit, it also will make lines longer if you broke up it over several and it would fit within the limit. Due to this, I just set it to 0. In some cases I know you can separate it over multiple lines, but it probably depends on specifically what it is.

This is what everything looks like when set to 120. https://github.com/LizardByte/Sunshine/actions/runs/13682786643/job/38259008671?pr=3706

> Anyway, do you know if the API calls I'm making are safe? Or maybe the info I need can be found another way?

I'm not sure, not an expert in this area.

andygrundman:
OK, this is cleaned up and squashed to 1 commit. I discovered that the last patch I made to this file, removing f32 and s32 from stereo would break this new logic if you had a 32-bit default device. The correct fix for that previous issue is just to put these at the bottom of the list so they aren't preferred over 24 and 16.

This now works well for me and I can't reproduce the issue I had. I have also added extra logging that may prove helpful if this doesn't solve the issue for people. Specifically I log the bps and samplerate of the internal mixer format, with a note about auto-resampling if it's not 48k.

src/platform/windows/audio.cpp
```diff
@@ -117,19 +117,24 @@ namespace {
 
   using virtual_sink_waveformats_t = std::vector<WAVEFORMATEXTENSIBLE>;
 
+  // The list of virtual formats are sorted in preference order and the first valid format will be used.
+  // All bits-per-sample options are listed because we try to match this to the default audio device.
```
ReenigneArcher, line 121: This seems like a good opportunity to document the template better. Could you document this template using a doxygen comment block?

```cpp
/*
 * @brief ...
 * @tparam ...
 */
```

- https://docs.lizardbyte.dev/projects/sunshine/latest/md_third-party_2doxyconfig_2docs_2source__code.html#documentation-blocks
- https://stackoverflow.com/a/13359876

src/platform/windows/audio.cpp
```diff
@@ -117,19 +117,24 @@ namespace {
...
         create_waveformat(sample_format_e::s24in32, channel_count, channel_mask),
         create_waveformat(sample_format_e::s24, channel_count, channel_mask),
         create_waveformat(sample_format_e::s16, channel_count, channel_mask),
+        create_waveformat(sample_format_e::f32, channel_count, channel_mask),
+        create_waveformat(sample_format_e::s32, channel_count, channel_mask),
       };
     } else if (channel_count == 6) {
       auto channel_mask1 = waveformat_mask_surround51_with_backspeakers;
-      auto channel_mask2 = waveformat_mask_surround51_with_sidespeakers;
+      auto channel_mask2 = waveformat_mask_surround51_with_sidespeakers;  // XXX will never be used, but is probably the better 5.1 layout
```
ReenigneArcher, line 137: What is `XXX` indicating here? Just calling extra attention to the note?

src/platform/windows/audio.cpp
```diff
@@ -793,6 +802,26 @@ namespace platf::audio {
...
+      // When switching to a Steam virtual speaker device, try to retain the bit depth of the
+      // default audio device. Switching from a 16-bit device to a 24-bit one has been known to
+      // cause glitches for some users.
+      int wanted_bits_per_sample = 32;
+      auto current_default_dev = default_device(device_enum);
+      if (current_default_dev) {
+        audio::prop_t prop;
+        prop_var_t current_device_format;
+
+        // clang-format off: easier to read this split over 2 lines
```
ReenigneArcher, line 814: I know it's not ideal to have this on a long line, but I'd prefer not to have clang-format exceptions for this. I'm open to different clang-format rules though to help make things more readable.

src/platform/windows/audio.cpp
cgutman: I'm not sure it's actually true either. I thought 5.1 with surrounds as back speakers was pretty universal. @andygrundman did you have more context to add here?

andygrundman: Back in the DVD era, 5.1 had the 2 rear speakers on the side walls at 90 degrees and also they were a few feet above ear level. In the modern 5.1 layout for Atmos, Dolby says it's anywhere between 110 and 120 degrees, and everything is at ear level now. One thing that was never an actual thing is 5.1 with 180 degree rear wall speakers. Those speakers are what you add to the side speakers to get a 7.1 system (where sides are back to 90 degrees). So the right 5.1 would be putting them at 120, splitting the difference, but Microsoft is stuck with their legacy options, and an Atmos option that we can't really use.

I think in practice, it doesn't matter where the rear speakers are as long as it's sort of behind you a little bit. My comment preferring 5.1-side is partly because I'm old, but mostly referring to the way the ordering works in Sunshine, those other options won't ever be used and the user can't choose them. Really I'm happy if the channels are all lined up correctly on the client side. 7.1 was different from Windows in SDL for years and nobody noticed until we got it [patched](https://github.com/libsdl-org/SDL/issues/11709) recently.

win-5 1-back
dolby-5 1
dolby-7 1

ReenigneArcher (approved):

massej:
My default device is 24bits and I need to set my output device to 16 bits. It would be nice to have an option in the GUI to force it to 16 bits. Everytime i start a game in moonlight my audio is crackling until I go under the device settings and set it to 16 bits.

andygrundman:
If you change your default device to 16 does it fix the problem? The fix I made was just a guess because I can't reproduce it, but the theory was to avoid any changes. That's why it will stay 24 in your case. Please check the Sunshine log, what do these lines show for you, especially the new line about the mixer format?

```
[2025-06-05 15:14:04.478]: Info: Virtual audio device will use 24-bit to match default device
[2025-06-05 15:14:04.496]: Info: Changed virtual audio sink format to [S24 48000 2.0]
...
[2025-06-05 15:14:04.543]: Info: Audio mixer format is 32-bit, 48000 Hz, no resampling needed
[2025-06-05 15:14:04.553]: Info: Audio capture format is [F32 48000 2.0]
[2025-06-05 15:14:04.554]: Info: Opus initialized: 48 kHz, 2 channels, 512 kbps (total), LOWDELAY
```

massej:
Thank you @andygrundman for the quick reply,

It works, but I have to manually switch the settings every time, which gets a bit irritating.

After gaming, I need to switch the output back to 24-bit/384 kHz to enjoy music properly. If I leave it at 16-bit/48 kHz—my default setting for gaming—I can clearly notice a drop in audio quality. It's not an issue for games, but for music listening, it's far from ideal. (I'm using a Beyerdynamic headset with an audiophile-grade tube preamp.)

andygrundman:
Cool, so you obviously have a nice DAC. I used to think this was some sort of issue with bad motherboard audio, because I also use USB/SPDIF/HDMI audio and never have a problem with Steam speakers. And I always have my default audio set to 24/48. I guess you could test if 24/48 makes any difference to the problem. Oh, and did you check the Sunshine log?

massej:
When the sound cut, there is no log in Sunshine, I guess the interesting log are in Moonlight. I'm not sure where the log are in Moonlight but I can connect using adb shell on the TV.

I tested on my phone to connect to Sunshine and the 24 bits sound work well.

I have the feeling that my TV is not powerful enough to process the 24 bits audio and Moonlight get out of sync.

PR #3727, opened 2025-03-16
fix(packaging/windows): add Sunshine directories to system PATH

ReenigneArcher:
## Description

This PR was inspired after too many users have tried running `sunshine --<command>` from the wrong working directory.

The option via CPACK is too complex (see explanation below).

<details>
<summary>CPACK method</summary>

See https://cmake.org/cmake/help/latest/cpack_gen/nsis.html#variable:CPACK_NSIS_MODIFY_PATH

As far as I can tell there is no way to set the default options.

TODO:

- [ ] This doesn't currently work since we don't use a `bin` dir. https://gitlab.kitware.com/cmake/cmake/-/issues/15635

image

</details>

## Type of Change

## Checklist

### DIFF
--- a/cmake/packaging/windows.cmake
+++ b/cmake/packaging/windows.cmake
@@ -1,17 +1,17 @@
 # windows specific packaging
 
 # see options at: https://cmake.org/cmake/help/latest/cpack_gen/nsis.html
-install(TARGETS sunshine RUNTIME DESTINATION "." COMPONENT application)
+install(TARGETS sunshine RUNTIME DESTINATION "bin" COMPONENT application)
 
 # Hardening: include zlib1.dll (loaded via LoadLibrary() in openssl's libcrypto.a)
-install(FILES "${ZLIB}" DESTINATION "." COMPONENT application)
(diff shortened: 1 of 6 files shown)


ReenigneArcher:
@cgutman My initial idea for this was to use the CPACK option to set the path, but ran into a bunch of issues with it.

1. It doesn't work unless `./bin` exists
2. Moved the exe and tools to the bin directory, but then Sunshine had some issues.
3. Sunshine couldn't find the config directory. I was able to fix this.
4. Sunshine couldn't find assets directory (I think... unknown as Sunshine crashed and nothing useful in logs)... this would be far more complex to fix.
5. Setting the PATH worked sometimes, and other times it didn't work and required a re-install.
6. Cannot set default option to set the PATH or not.
7. The option to add desktop icon is not customizable and didn't ever work.

So, I started thinking it would be easier to just have an install/uninstall script. Any thoughts/opinions/questions on this approach.

These are mine:

1. Should we prompt to set the PATH or not? My thought is no prompt and just set it all the time.
2. Does anything in Sunshine depend on the working directory? I don't believe it does, but I'm not 100% sure.

PR #3741, opened 2025-03-21
build(android): support android platform

mengyanshou:
## Description
To do some preparation for sunshine to support android, submit some macro definitions first

## Type of Change

## Checklist

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -32,7 +32,7 @@
   #include <shellapi.h>
 #endif
 
-#ifndef __APPLE__
+#if !defined(__APPLE__) && !defined(__ANDROID__)
   // For NVENC legacy constants
   #include <ffnvcodec/nvEncodeAPI.h>
 #endif
@@ -1039,9 +1039,12 @@ namespace config {
   }
 
   void apply_config(std::unordered_map<std::string, std::string> &&vars) {
+#ifndef __ANDROID__
+    // TODO(Lin): May android also can support this
     if (!fs::exists(stream.file_apps.c_str())) {
       fs::copy_file(SUNSHINE_ASSETS_DIR "/apps.json", stream.file_apps);
     }
+#endif
 
     for (auto &[name, val] : vars) {
       std::cout << "["sv << name << "] -- ["sv << val << ']' << std::endl;
@@ -1066,7 +1069,7 @@ namespace config {
     bool_f(vars, "nvenc_opengl_vulkan_on_dxgi", video.nv_opengl_vulkan_on_dxgi);
     bool_f(vars, "nvenc_latency_over_power", video.nv_sunshine_high_power_mode);
 
-#ifndef __APPLE__
+#if !defined(__APPLE__) && !defined(__ANDROID__)
     video.nv_legacy.preset = video.nv.quality_preset + 11;
     video.nv_legacy.multipass = video.nv.two_pass == nvenc::nvenc_two_pass::quarter_resolution ? NV_ENC_TWO_PASS_QUARTER_RESOLUTION :
                                 video.nv.two_pass == nvenc::nvenc_two_pass::full_resolution    ? NV_ENC_TWO_PASS_FULL_RESOLUTION :
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -15,7 +15,11 @@
 #include <boost/log/expressions.hpp>
 #include <boost/log/sinks.hpp>
 #include <boost/log/sources/severity_logger.hpp>
-#include <display_device/logging.h>
+#ifdef __ANDROID__
+  #include <android/log.h>
+#else
+  #include <display_device/logging.h>
+#endif
 
 // local includes
 #include "logging.h"
@@ -86,6 +90,32 @@ namespace logging {
 #endif
     };
 
+#ifdef __ANDROID__
+    android_LogPriority android_priority;
+    switch (log_level) {
+      case 0:
+        android_priority = ANDROID_LOG_VERBOSE;
+        break;
+      case 1:
+        android_priority = ANDROID_LOG_DEBUG;
+        break;
+      case 2:
+        android_priority = ANDROID_LOG_INFO;
+        break;
+      case 3:
+        android_priority = ANDROID_LOG_WARN;
+        break;
+      case 4:
+        android_priority = ANDROID_LOG_ERROR;
+        break;
+      case 5:
+        android_priority = ANDROID_LOG_FATAL;
+        break;
+    };
+    // get log adn input to andorid
+    std::string log_message = view.attribute_values()[message].extract<std::string>().get();
+    __android_log_print(android_priority, "Sunshine", "%s%s", log_type.data(), log_message.c_str());
+#endif
     auto now = std::chrono::system_clock::now();
     auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(
       now - std::chrono::time_point_cast<std::chrono::seconds>(now)
@@ -105,7 +135,9 @@ namespace logging {
     }
 
     setup_av_logging(min_log_level);
+#ifndef __ANDROID__
     setup_libdisplaydevice_logging(min_log_level);
+#endif
 
     sink = boost::make_shared<text_sink>();
 
@@ -153,6 +185,7 @@ namespace logging {
     });
   }
 
+#ifndef __ANDROID__
   void setup_libdisplaydevice_logging(int min_log_level) {
     constexpr int min_level {static_cast<int>(display_device::Logger::LogLevel::verbose)};
     constexpr int max_level {static_cast<int>(display_device::Logger::LogLevel::fatal)};
@@ -182,6 +215,7 @@ namespace logging {
       }
     });
   }
+#endif
 
   void log_flush() {
     if (sink) {


src/config.cpp
```diff
@@ -1039,9 +1039,12 @@ namespace config {
   }
 
   void apply_config(std::unordered_map<std::string, std::string> &&vars) {
+#ifndef __ANDROID__
+    // TODO(Lin): May android also can support this
```
ReenigneArcher, line 1043:
```suggestion
    // TODO: Android can possibly support this
```

src/config.cpp
```diff
@@ -32,7 +32,7 @@
   #include <shellapi.h>
 #endif
 
-#ifndef __APPLE__
+#if !defined(__APPLE__) && !defined(__ANDROID__)
```
ReenigneArcher, line 35: Let's put Android first (alphabetically).

This applies in one other place as well.

src/logging.cpp
```diff
@@ -15,7 +15,11 @@
 #include <boost/log/expressions.hpp>
 #include <boost/log/sinks.hpp>
 #include <boost/log/sources/severity_logger.hpp>
-#include <display_device/logging.h>
+#ifdef __ANDROID__
+  #include <android/log.h>
+#else
+  #include <display_device/logging.h>
+#endif
```
ReenigneArcher, line 22:
```suggestion

// conditional includes
#ifdef __ANDROID__
  #include <android/log.h>
#else
  #include <display_device/logging.h>
#endif
```

Can you put this below the `// local includes` ?

ReenigneArcher (changes requested):
Thank you for the PR! I'm looking forward to this.

I expected more changes for in order for initial Android support. Is there more coming? Maybe something in the `./cmake` directory?

src/config.cpp
mengyanshou: OK!

mengyanshou:
Sorry for the delayed response. Unfortunately, I don't have much time either.

My main tech stack is Flutter (UI Framework) and Android Native.

Typically, the compilation of an Android app differs significantly from C/C++ and requires the Java environment and Gradle.

So, from my current perspective, it seems quite difficult to make the Sunshine workflow directly produce an APK.

Additionally, I'm concerned that my lack of proficiency in C/C++ might disrupt the original workflow, so I plan to submit PRs in multiple stages.

However, I can start with some compilation adaptations to ensure that future Sunshine APKs for Android can be imported and compiled smoothly.

As seen in the [patches](https://github.com/nightmare-space/sunshine_android/tree/main/patchs), I can now directly use the [Sunshine Source](https://github.com/nightmare-space/sunshine_android/tree/main/android) along with a [Patch Script](https://github.com/nightmare-space/sunshine_android/blob/main/scripts/apply_patch.sh) to make Android compilation work properly.

My initial idea is to submit these patches to Sunshine so that the Android-compiled Sunshine and the Sunshine source code can remain consistent.

Additionally, I’d like to share some recent comparisons with scrcpy.

I’ve studied a lot of scrcpy’s code. In fact, scrcpy is already very efficient. It’s possible that the client’s use of SDL2 results in rendering that’s not as extremely efficient (just a guess). However, in terms of the server side, scrcpy and the ported Sunshine for Android should be roughly the same.

That said, porting to Android still has significant value because it allows users to use Moonlight to stream their Android devices.

I will continue to push this process forward.

src/logging.cpp
```diff
@@ -86,6 +92,32 @@ namespace logging {
...
+        android_priority = ANDROID_LOG_WARN;
+        break;
+      case 4:
+        android_priority = ANDROID_LOG_ERROR;
+        break;
+      case 5:
+        android_priority = ANDROID_LOG_FATAL;
+        break;
+    };
+    // get log adn input to andorid
```
ReenigneArcher, line 117:
```suggestion
```

This comment seems unnecessary

src/logging.cpp
```diff
@@ -15,7 +15,13 @@
 #include <boost/log/expressions.hpp>
 #include <boost/log/sinks.hpp>
 #include <boost/log/sources/severity_logger.hpp>
-#include <display_device/logging.h>
+
+// conditional includes
+#ifdef __ANDROID__
+  #include <android/log.h>
+#else
+  #include <display_device/logging.h>
```
FrogTheFrog, line 23: Unless there are some compilation issues, I would not touch the `display_device. This function https://github.com/LizardByte/Sunshine/blob/5da5293471272fa643863a0f3cbdb190a9476a4b/src/display_device.cpp#L615 only creates a SettingsManager for Windows at the moment.

This means that every `display_device` function will either do nothing or act as a "pass-trough". Same for logging - nothing should be logged via `setup_libdisplaydevice_logging`, so maybe just keep it as it is?

src/logging.cpp
```diff
@@ -86,6 +92,32 @@ namespace logging {
 #endif
     };
 
+#ifdef __ANDROID__
+    android_LogPriority android_priority;
```
FrogTheFrog, line 96:
```suggestion
    android_LogPriority android_priority{};
```
If this `android_LogPriority` is a built-in type, then it should be initialized to some default value at least. If it's a class/struct you can leave it as it is.

src/logging.cpp
```diff
@@ -86,6 +92,32 @@ namespace logging {
 #endif
     };
 
+#ifdef __ANDROID__
```
FrogTheFrog, line 95: This whole section is not correct way to log to android via boost. You need to create a custom sink (like here https://gist.github.com/cawka/ca21d79605bf51621d6206d2cd76b638) and then probably replace the default sink (https://github.com/LizardByte/Sunshine/blob/5da5293471272fa643863a0f3cbdb190a9476a4b/src/logging.cpp#L116) with your android one.

FrogTheFrog (changes requested):

src/logging.cpp
mengyanshou: Ok, I will try to modify this part, but it will be a little later. Thank you for your review

src/logging.cpp
Laurie-Lin: Additionally, `#include <display_device/logging.h>` fails to compile on Android. 
This is likely due to my unfamiliarity with the build system, which I am still learning. 
I couldn't locate where the `display_device/logging.h` header file comes from, 
but I suspect it might be related to monitor/display headers, which shouldn't be needed on Android.

src/logging.cpp
Laurie-Lin: I rewrote the code because the previous version indeed had issues. I kept the source code for the formatter and added a custom android_sink_backend.

src/logging.cpp
```diff
@@ -97,22 +103,67 @@ namespace logging {
...
   [[nodiscard]] std::unique_ptr<deinit_t> init(int min_log_level, const std::string &log_file) {
     if (sink) {
       // Deinitialize the logging system before reinitializing it. This can probably only ever be hit in tests.
       deinit();
     }
 
+#ifndef __ANDROID__
     setup_av_logging(min_log_level);
     setup_libdisplaydevice_logging(min_log_level);
+#endif
```
ReenigneArcher, line 158: Does it compile if you remove the `ifndef`?

Laurie-Lin: I attempted this, but in reality, Sunshine for Android does not compile FFmpeg. This is because if FFmpeg is used for encoding, it would rely on CPU encoding on Android, which would result in poor performance.

Currently, the approach is to use pure Android APIs to perform hardware encoding on the current screen data.

This means that there is no need to compile FFmpeg for Android. Introducing FFmpeg cross-compilation would only make things more complex.

In this case, setup_av_logging is also unnecessary, as no FFmpeg code is being executed.

Laurie-Lin:
I've been too busy recently and have no idea how to move this PR forward. I used to open it occasionally, but there has been no progress on the parts awaiting review. So I haven't paid much attention to it since then.

ReenigneArcher:
@mengyanshou there's a few comments above that haven't been addressed yet. I'd be happy to discuss in discord if there is some confusion to clear up?

ReenigneArcher:
image

The same comments still apply.

Laurie-Lin:
Sorry, my hand was shaking and I clicked the wrong one. But I think most of them should be the reviews waiting for a reply on my side. If these changes are accepted, I will modify these reviews as resolved to proceed with the subsequent merge.

Then regarding one of the Todos, I currently don't know how to implement it, but I think it's necessary to keep the todo.

截屏2025-07-13 00 51 53

ReenigneArcher:
Don't worry about the TODO warning, those are okay.

mengyanshou (commented):
Look good to me!

Laurie-Lin:
Next, what can I do to help move forward? I have very little time recently and should not be able to continue submitting changes to the code until next month. Then, thank you for your cooperation @ReenigneArcher

src/logging.cpp
ReenigneArcher: It's from here: https://github.com/LizardByte/libdisplaydevice/blob/master/src/common/include/display_device/logging.h

src/logging.cpp
ReenigneArcher: It seems it's possible to use FFmpeg on Android with hw encoding. I didn't do a ton of research on this, but a quick search brought this up. https://www.reddit.com/r/androiddev/comments/clz0ol/comment/evz6vlq/

I think we can probably add an Android build off FFmpeg to our pre-built FFmpeg (https://github.com/LizardByte/build-deps). We're already doing some cross-compiling there, though I have not looked into doing this for android yet. This can definitely come later if you have a different approach for now.

Quackdoc: ffmpeg would be really nice for us at blissOS as we support amd and intel gpus. Even for cpu encoding ffmpeg would be nice instead of relying on platform support.

I don't have a whole bunch of time to help out with this, but media-kit and mpv-android cross compile ffmpeg. You can find the buildscripts for mpv-android here https://github.com/mpv-android/mpv-android/tree/master/buildscripts.

src/config.cpp
```diff
@@ -1038,9 +1038,12 @@ namespace config {
   }
 
   void apply_config(std::unordered_map<std::string, std::string> &&vars) {
+#ifndef __ANDROID__
+    // TODO: Android can possibly support this
```
ReenigneArcher, line 1042: Apologies, I wrote a comment on this for the last review but it didn't appear for some reason.

I think we discussed this on Discord some time back, but can you explain here why it's not supported right now? In my mind this should work, but I might be missing something.

In my mind, the default "Desktop" application could be renamed to "Screen" or something along those lines. Then the user could add the commands for other applications that would launch an app. Maybe I am missing something?

For reference, these are the current three `apps.json` files:

- https://github.com/LizardByte/Sunshine/blob/master/src_assets/linux/assets/apps.json
- https://github.com/LizardByte/Sunshine/blob/master/src_assets/macos/assets/apps.json
- https://github.com/LizardByte/Sunshine/blob/master/src_assets/windows/assets/apps.json

FrogTheFrog:
> If possible @FrogTheFrog, could you check the updated logging section?

LGTM

src/config.cpp
Laurie-Lin: Since I haven't decided on the form of Sunshine Android yet - whether it will be an APK, a binary launched via adb, or both options

The difference is that if it's an APK, the app will need to request screen recording permissions from users, with the readable directory being `/data/data/com.xxx/files`

If it's a binary launched through adb, no permissions would be needed, with the readable directory being `/data/local/tmp`

These two approaches have different readable directories, so I'd like to implement this when I have more time available

Laurie-Lin:
> Seems pretty good to me. In addition to my one question below, I have one more. If possible @FrogTheFrog, could you check the updated logging section?
> 
> What would it take to compile this in our CI (understanding no apk would be provided yet)? Could you provide instructions either here or in https://github.com/LizardByte/Sunshine/blob/master/docs/building.md

Regarding CI, as mentioned above, if it's a binary, we would need to compile the entire Sunshine using NDK, which is more complex and requires additional code. If it's an APK, we would need to introduce gradle for compilation. However, we're still in the very early stages - I've only merged some Android-compatible code into Sunshine, but currently we cannot compile either an APK or binary with these changes. Therefore, CI work will need to wait until the entire Android compatibility code is completed. The initial idea is to make it compatible with Android while not affecting any of Sunshine's original code

CalTrask:
Is this project still being developed? Sunshine hosting on android?

Laurie-Lin:
I'm very sorry that I'm too busy. I have a lot of projects to maintain personally. I don't know how to proceed with the support for Sunshine Android because I'm not very familiar with the original code. Many of the current implementations are at a demo level. If there are many people who need this feature, I can spare half a day to a full day a week to follow up on this function.

Laurie-Lin:
> 这个项目还在开发吗？Sunshine Hosting 在安卓系统上能用吗？

See comment

PR #3767, opened 2025-03-29
fix: handle wide-character output on Windows console to avoid truncation

a492219408:
## Description

This change fixes a bug where the console would stop printing any subsequent output if it encountered Chinese characters. By converting the wide-string output to UTF-8 with `converter.to_bytes()`, the console now continues to properly display messages containing Chinese characters without interruption.

### Screenshot

Snipaste_2025-03-27_23-16-13
Snipaste_2025-03-27_23-16-44

### Issues Fixed or Closed

No specific issue references.

## Type of Change

## Checklist

### DIFF
--- a/tools/audio.cpp
+++ b/tools/audio.cpp
@@ -212,9 +212,9 @@ namespace audio {
     std::wcout
       << L"===== Device ====="sv << std::endl
       << L"Device ID          : "sv << wstring.get() << std::endl
-      << L"Device name        : "sv << no_null((LPWSTR) device_friendly_name.prop.pszVal) << std::endl
-      << L"Adapter name       : "sv << no_null((LPWSTR) adapter_friendly_name.prop.pszVal) << std::endl
-      << L"Device description : "sv << no_null((LPWSTR) device_desc.prop.pszVal) << std::endl
+      << L"Device name        : "sv << converter.to_bytes(no_null((LPWSTR) device_friendly_name.prop.pszVal)).c_str() << std::endl
+      << L"Adapter name       : "sv << converter.to_bytes(no_null((LPWSTR) adapter_friendly_name.prop.pszVal)).c_str() << std::endl
+      << L"Device description : "sv << converter.to_bytes(no_null((LPWSTR) device_desc.prop.pszVal)).c_str() << std::endl
       << L"Device state       : "sv << device_state_string << std::endl
       << L"Current format     : "sv << current_format << std::endl
       << std::endl;


tools/audio.cpp
```diff
@@ -13,6 +13,11 @@
 #include <roapi.h>
 #include <synchapi.h>
 
+#ifdef _WIN32
+#include <fcntl.h>
+#include <io.h>
+#endif
```
ReenigneArcher, line 19: Does it hurt anything if these are included (and not used) on Unix?

a492219408: You're absolutely right to raise the question. Here's a more detailed explanation:

The headers `<io.h>` and the function `_setmode()` are both **specific to Windows**. They do not exist on Unix-like systems (e.g. Linux, macOS), and trying to include or use them on those platforms would normally cause a compilation error.

However, since these lines are wrapped inside an `#ifdef _WIN32` block, they are only **visible to the compiler when building on Windows**. On Unix systems, the preprocessor evaluates this condition as false and **completely removes those lines** before the actual compilation step. This means:
- The code inside the `#ifdef` block is **never seen or compiled** on Unix.
- There is **no runtime or compile-time impact** on non-Windows platforms.
- This is a standard cross-platform practice in C/C++ projects.

Additionally, this particular bug (where `std::wcout` halts output when encountering wide characters like Chinese) only occurs on Windows. So the fix involving `_setmode()` is both **necessary and safe** to restrict to the Windows platform.

Let me know if you'd prefer this handled differently — but the current setup should be fully portable.

FrogTheFrog:
What if we change the code to look some thing like this?
```
void print_device(device_t &device) {
    audio::wstring_t wstring;
    DWORD device_state;

    device->GetState(&device_state);
    device->GetId(&wstring);

    audio::prop_t prop;
    device->OpenPropertyStore(STGM_READ, &prop);

    prop_var_t adapter_friendly_name;
    prop_var_t device_friendly_name;
    prop_var_t device_desc;

    prop->GetValue(PKEY_Device_FriendlyName, &device_friendly_name.prop);
    prop->GetValue(PKEY_DeviceInterface_FriendlyName, &adapter_friendly_name.prop);
    prop->GetValue(PKEY_Device_DeviceDesc, &device_desc.prop);

    if (!(device_state & device_state_filter)) {
      return;
    }

    std::string device_state_string = "Unknown"s;
    switch (device_state) {
      case DEVICE_STATE_ACTIVE:
        device_state_string = "Active"s;
        break;
      case DEVICE_STATE_DISABLED:
        device_state_string = "Disabled"s;
        break;
      case DEVICE_STATE_UNPLUGGED:
        device_state_string = "Unplugged"s;
        break;
      case DEVICE_STATE_NOTPRESENT:
        device_state_string = "Not present"s;
        break;
    }

    std::string current_format = "Unknown"s;
    for (const auto &format : formats) {
      // This will fail for any format that's not the mix format for this device,
      // so we can take the first match as the current format to display.
      auto audio_client = make_audio_client(device, format);
      if (audio_client) {
        current_format = format.name.data();
        break;
      }
    }

    std::cout
      << "===== Device ====="sv << std::endl
      << "Device ID          : "sv << platf::to_utf8(wstring.get()) << std::endl
      << "Device name        : "sv << platf::to_utf8(no_null((LPWSTR) device_friendly_name.prop.pszVal)) << std::endl
      << "Adapter name       : "sv << platf::to_utf8(no_null((LPWSTR) adapter_friendly_name.prop.pszVal)) << std::endl
      << "Device description : "sv << platf::to_utf8(no_null((LPWSTR) device_desc.prop.pszVal)) << std::endl
      << "Device state       : "sv << device_state_string << std::endl
      << "Current format     : "sv << current_format << std::endl
      << std::endl;
  }
```

where the `platf::to_utf8` is from https://github.com/LizardByte/Sunshine/blob/master/src/platform/windows/misc.h#L35? Getting rid of the `static std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t> converter;` altogether.

I have not checked if it compiles, but please do try it out.

ReenigneArcher (changes requested):
Please try the suggestions from Frog.

PR #3777, opened 2025-04-03
fix(config): ensure logging config settings are properly loaded

ReenigneArcher:
## Description

A bug was introduced by https://github.com/LizardByte/Sunshine/pull/3769 where the logging level would not be respected since it was parsed after logging was initialized. This PR changes the behavior to parse the config before logging is initialized (as it previously was); however the modified config values are stored in a global variable so they can be logged during startup.

## Type of Change

## Checklist


nlflint:
Confirmed, this enables debug logging.

PR #3802, opened 2025-04-19
Reducing severity of missing ViGEmBus from fatal to warning

snusstofilen:
ViGEmBus has reached end-of-life and is no longer receiving any patches. We shouldn't require the user to install a discontinued software. In fact, it isn't a fatal error.

## Description

Sunshine\src\platform\windows\input.cpp

Replacing the severity level of the log from fatal to warning regarding missing ViGEmBus installation, as this error isn't fatal, nor should we require installing discontinued software.

### Screenshot
image

## Type of Change

## Checklist

### DIFF
--- a/src/platform/windows/input.cpp
+++ b/src/platform/windows/input.cpp
@@ -203,8 +203,8 @@ namespace platf {
       client_t client {vigem_alloc()};
       VIGEM_ERROR status = vigem_connect(client.get());
       if (!VIGEM_SUCCESS(status)) {
-        // Log a special fatal message for this case to show the error in the web UI
-        BOOST_LOG(fatal) << "ViGEmBus is not installed or running. You must install ViGEmBus for gamepad support!"sv;
+        // Log a special info message for this case to show the error in the web UI
+        BOOST_LOG(info) << "ViGEmBus is not installed or running. You must install ViGEmBus for gamepad support!"sv;
       } else {
         vigem_disconnect(client.get());
       }


ReenigneArcher:
It's required for gamepad support.

snusstofilen:
> It's required for gamepad support.

In which way. I am running Sunshine without it and I am using my gamepad just fine.

ReenigneArcher:
Try without steam

snusstofilen:
> Try without steam

It works just fine even without steam.

snusstofilen:
To use a gamepad you do not require vigembus. There are other ways of setting it up. Most simply, connecting it to the host. Some guests do not even support connecting gamepads to it. Then it would be completely useless, right? I do not see how this is a "fatal" error.

ReenigneArcher:
Of course this is game "streaming" software, so it assumes you are connecting the gamepad to the client. You can obviously get around it by connecting the gamepad directly to the host, but in that case you are not using virtual input...

It's only fatal so it alerts the user in the UI. It's not fatal in the aspect that Sunshine won't work at all.

snusstofilen:
> Of course this is game "streaming" software, so it assumes you are connecting the gamepad to the client. You can obviously get around it by connecting the gamepad directly to the host, but in that case you are not using virtual input...
> 
> It's only fatal so it alerts the user in the UI. It's not fatal in the aspect that Sunshine won't work at all.

If the assumption is that people will connect the gamepad to the client, and that it part of the instructions on how to use Sunshine, then yes, it would be fatal. So if that's how you wish to keep it then I understand. I didn't find that info in the documentation.

PR #3807, opened 2025-04-22
Let users opt-out of auto-opening the browser

cmcaine:
## Description

Let users opt-out of auto-opening the browser when Sunshine is started from a shortcut on windows.

### Screenshot

The commit should add a checkbox on the general page. This is untested.

### Issues Fixed or Closed

Fixes https://github.com/orgs/LizardByte/discussions/358

## Type of Change

## Checklist
  - It looks fine, but I don't have a windows CPP build env handy to test.
  - Doesn't seem relevant?

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -581,7 +581,8 @@ namespace config {
     "ipv4",  // Address family
     platf::appdata().string() + "/sunshine.log",  // log file
     false,  // notify_pre_releases
-    {},  // prep commands
+    {},     // prep commands
+    true,   // open_webui
   };
 
   bool endline(char ch) {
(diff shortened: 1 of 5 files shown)


cmcaine:
I didn't try to understand the relevant code. I think this PR should be tested and accepted or rejected first, before trying to build upon it with any further changes.

For linux, I vaguely remember this being a pain because you xdg-open is not always present and is sometimes broken on wayland, etc. The description in the option could just say that it is windows-only?

ReenigneArcher:
I don't think it makes sense to have this setting apply only to Windows to be honest. It's somewhat complicated and annoying to have config options that are OS dependent so I like to only do there where absolutely necessary.

If this PR won't be extended to handle the cases for the other OSes, then I suppose we could probably just remove the logic that opens the browser when using the shortcut. It's sort of pointless on Windows as users can use the tray icon if they want to open it.

PR #3814, opened 2025-04-25
build(linux): add arch linux support for linux_build.sh script

Clutchnp:
this pull requests adds arch linux support for building via the scripts/linux_build.sh script

## Type of Change

## Checklist


ReenigneArcher (changes requested):
Thank you for the PR!

Can you double check the dependencies are actually needed? Some of what you have listed I am not familiar with and not used to build Sunshine from my experience. This is what we use for our pkgbuild. https://github.com/LizardByte/Sunshine/blob/65b0217a903db803744f62c1fff51cb85b5a4409/packaging/linux/Arch/PKGBUILD#L15-L52

Could you also alphabetize the dependencies?

Clutchnp:
gcc, gcc-fortran , -compat should not be there, i dont know how that went through. as for others, i went by the dependencies of the deb and fedora packages, and some dependencies are present in the debian build but not in the arch pkgbuild (because most of them were to build other deps and were not necessary for arch ) but still I have added some which i think should be present (eg. wayland), I have removed the unneccesary  deps and commited the suggested changes, this was bad on my part, will try that it not happens again

scripts/linux_build.sh
```diff
@@ -153,6 +197,7 @@ function add_ubuntu_deps() {
   )
 }
 
+
```
ReenigneArcher, line 200:
```suggestion
```

looks like another added line snuck in

scripts/linux_build.sh
```diff
@@ -436,7 +485,15 @@ function run_install() {
...
 cat /etc/os-release
-if grep -q "Debian GNU/Linux 12 (bookworm)" /etc/os-release; then
+
+if grep -q "Arch Linux" /etc/os-release; then
+  distro="arch"
+  version=""
+  package_update_command="${sudo_cmd} pacman -Syu --noconfirm"
+  package_install_command="${sudo_cmd} pacman -Sy"
+  nvm_node=0
+
```
ReenigneArcher, line 495:
```suggestion
```

scripts/linux_build.sh
```diff
@@ -90,6 +90,50 @@ shift $((OPTIND -1))
...
+    dependencies+=(
+      "libva"  # VA-API
+    )
+  fi
+
+  if [ "$skip_cuda" == 0 ]; then
+    dependencies+=(
+      "cuda"  # VA-API
+    )
+  fi
```
ReenigneArcher, line 134: This isn't going to work. Check the `install_cuda` function. Currently it downloads cuda from nvidia and installs it using the run file. But as of now there is no way to tell this script to use the system installed cuda.

ReenigneArcher (changes requested):
Thanks for the quick updates! And no worries about missing a few things, I don't expect a first time contributor to be an expert on the code base!

One other thought:

Should there be a command added to create a package like we have for deb and rpm?

Clutchnp:
For CUDA, we can simply check if it's present using command -v nvcc.

As for creating a package, CPack does not support generating Arch Linux–friendly packages. We can create a PKGBUILD wrapper around it to just handle packaging, but I usually use this script  to check builds, and think others do so too, so I feel its not necessary, what do you think?

scripts/linux_build.sh
ReenigneArcher: > For CUDA, we can simply check if it's present using command -v nvcc.
> 
> As for creating a package, CPack does not support generating Arch Linux–friendly packages. We can create a PKGBUILD wrapper around it to just handle packaging, but I usually use this script to check builds, and think others do so too, so I feel its not necessary, what do you think?

That's fine with both aspects. The packaging is probably not that important since we provide a nightly one in our pacman-repo anyway.

Just one more thing to address with cuda.

We will need some logic to determine the line below. Perhaps the `install_cuda` function should return the path of the compiler.

`cmake_args+=("-DCMAKE_CUDA_COMPILER:PATH=${build_dir}/cuda/bin/nvcc")`

ReenigneArcher (approved):
Thanks for all the iterations! I think this is good to go now.

ReenigneArcher:
I'm not sure what's causing it, but all debian builds using this script are now failing.

Clutchnp:
That should not happen but still in case i directed the error output to /dev/null 
Can you check again @ReenigneArcher

ReenigneArcher:
@Clutchnp I re-ran everything but it's the same error. I'm not sure where the error is occurring to be honest.

scripts/linux_build.sh
```diff
@@ -90,6 +90,50 @@ shift $((OPTIND -1))
...
+  'miniupnpc'
+  'ninja'
+  'nodejs'
+  'npm'
+  'numactl'
+  'openssl'
+  'opus'
+  'udev'
+  'wayland'
+)
```
ReenigneArcher, line 122:
```suggestion
  dependencies+=(
    'avahi'
    'base-devel'
    'cmake'
    'curl'
    'git'
    'libayatana-appindicator'
    'libcap'
    'libdrm'
    'libevdev'
    'libmfx'
    'libnotify'
    'libpulse'
    'libva'
    'libx11'
    'libxcb'
    'libxfixes'
    'libxrandr'
    'libxtst'
    'miniupnpc'
    'ninja'
    'nodejs'
    'npm'
    'numactl'
    'openssl'
    'opus'
    'udev'
    'wayland'
  )
```

PR #3821, opened 2025-04-28
Add new form of preparation commands that runs before checking for displays

FiveYellowMice:
## Description
This PR attempts to a new form of preparation commands that runs before checking for displays, to solve feature request https://github.com/orgs/LizardByte/discussions/283 . This will include a minor refactor as suggested in [this comment](https://github.com/orgs/LizardByte/discussions/283#discussioncomment-6757616).

Hopefully this can open up doors for adding "on stream pause/resume" commands in the future.

### Issues Fixed or Closed
https://github.com/orgs/LizardByte/discussions/283

## Type of Change

## Checklist

### DIFF
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -851,6 +851,17 @@ namespace nvhttp {
     host_audio = util::from_view(get_arg(args, "localAudioPlayMode"));
     auto launch_session = make_launch_session(host_audio, args);
 
+    if (appid > 0) {
+      auto err = proc::proc.prepare(appid, launch_session);
+      if (err) {
+        tree.put("root.<xmlattr>.status_code", err);
+        tree.put("root.<xmlattr>.status_message", "Failed to prepare for the specified application");
+        tree.put("root.gamesession", 0);
+
+        return;
+      }
+    }
+
     if (rtsp_stream::session_count() == 0) {
       // The display should be restored in case something fails as there are no other sessions.
       revert_display_configuration = true;
@@ -885,7 +896,7 @@ namespace nvhttp {
     }
 
     if (appid > 0) {
-      auto err = proc::proc.execute(appid, launch_session);
+      auto err = proc::proc.execute();
       if (err) {
         tree.put("root.<xmlattr>.status_code", err);
         tree.put("root.<xmlattr>.status_message", "Failed to start the specified application");
--- a/src/process.cpp
+++ b/src/process.cpp
@@ -130,21 +130,21 @@ namespace proc {
     return cmd_path.parent_path();
   }
 
-  int proc_t::execute(int app_id, std::shared_ptr<rtsp_stream::launch_session_t> launch_session) {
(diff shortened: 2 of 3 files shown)


ReenigneArcher:
Thank you for the PR!

I think we could just move when the prep commands run instead of introducing another type of prep commands?

ReenigneArcher:
@FiveYellowMice any response?

ReenigneArcher:
> If someone have their commands assuming display to be already setup, they will fail.

That's a good point. I was not considering those who are changing resolutions for example. My bad...

Expanding on the roadmap entry here. This is just some brainstorming, nothing written in stone. We probably need a few different stages when a command can run.

- pre display check - this PR (needed to enable displays before streaming)
- post display check - current prep commands
- pre stream start - current main command
- post stream start - nothing available currently
- stream pause / disconnect - nothing available currently
- stream quit / end / termination - current undo commands
- additional user connects to stream - nothing available currently

And the commands and do/undo commands to be refactored a bit. In the apps config it should probably just be a sequence of commands.

e.g.

```json
commands: [
  {
    "detached": True/False,
    "elevated": True/False,
    "ignore_error": True/False,
    "log_file": "e.g. my_command.log",
    "name": "Friendly name",
    "run": "terminal command to run",
    "shell": "shell to use, e.g. auto, bash, cmd, pwsh, etc.",
    "stage": "the point when the command should run",
    "working_directory": "path to start in"
  }
]
```

These would be listed in the order that they should run. In the c++ at each stage loop over the list in order and only run the command if the stage matches the stage we're at.

FiveYellowMice:
Thanks for the additional info. I guess in this case, I can keep the
original plan for this PR, i e. adding a new type of commands, but keep
this future roadmap in mind. Then think about what to do with the roadmap
afterwards.

There does seem to be a backwards incompatibility with the roadmap though.
Currently, the undo prep commands are run in reverse order of definition,
and they will only run if the corresponding do command has exited
successfully. These 2 features allows for graceful exit should one of the
do commands fail. The proposed refactor will discard these 2 features. Has
this been discussed and deemed acceptable?

If not, my opinion is to keep the commands in do/undo pairs, and extend the
pairs to have more type

* Pre-display setup / Post-display revert
* Post-display setup / Pre-display revert
* Pre-stream start / Post-stream stop
* etc.

But that probably makes things more confusing to the user. Some types of
commands (like detached commands) don't fit into a do/undo pair, so they
will not be in a pair, which means more types of data structures, more
complexity.

Nonary:
@FiveYellowMice I totally get what you mean by complexity and confusion with the prep command redesign. But fortunately, AI is rather quite insane right now. 

image

... So yeah, that's my next project to implement into Sunshine a complete rework to prep commands.

Nonary:
Submitted my prep command redesign in draft, this will likely replace this PR once I continue working on it.

https://github.com/LizardByte/Sunshine/pull/4023

Right now, it's at a somewhat usable state but may drastically change over the next few days.

PR #3833, opened 2025-04-30
feat(audio): add an option to disable sound streaming

Clutchnp:
## Description
So I use sunshine to stream my headless monitor to my laptop and use it as a second display, when doing so sunshine automatically changes the default audio sink to one created by sunshine, This is undesirable for my usecase, this pr adds an option disable_audio to the config which allows stopping streaming audio completely

### Issues Fixed or Closed
None

## Type of Change

## Checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -770,25 +770,25 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
-### disable_audio
+### stream_audio
 
 <table>
     <tr>
         <td>Description</td>
         <td colspan="2">
-            This option allows to disable streaming audio completely, useful for streaming of headless displays as second monitors
+            Weather to Stream Audio or Not, Disabling this can be useful for streaming headless displays as second monitors.
         </td>
     </tr>
     <tr>
         <td>Default</td>
         <td colspan="2">@code{}
-            disabled
+            enabled
             @endcode</td>
     </tr>
     <tr>
         <td>Example</td>
         <td colspan="2">@code{}
-            disable_audio = enabled
+            stream_audio = disabled
             @endcode</td>
     </tr>
 </table>
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -776,7 +776,7 @@ editing the `conf` file in a text editor. Use the examples as reference.
     <tr>
         <td>Description</td>
         <td colspan="2">
-            Weather to Stream Audio or Not, Disabling this can be useful for streaming headless displays as second monitors.
+            Whether to stream audio or not. Disabling this can be useful for streaming headless displays as second monitors.
         </td>
     </tr>
     <tr>
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -770,6 +770,29 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### disable_audio
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            This option allows to disable streaming audio completely, useful for streaming of headless displays as second monitors
+        </td>
+    </tr>
+    <tr>
+        <td>Default</td>
+        <td colspan="2">@code{}
+            disabled
+            @endcode</td>
+    </tr>
+    <tr>
+        <td>Example</td>
+        <td colspan="2">@code{}
+            disable_audio = enabled
+            @endcode</td>
+    </tr>
+</table>
+
 ### install_steam_audio_drivers
 
 <table>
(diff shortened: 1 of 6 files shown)


ReenigneArcher (changes requested):
Thanks for the PR!

Is this necessary if you uncheck the option in Moonlight "mute audio on host"?

If this PR is still needed, the web UI will need some updates as well.

Clutchnp:
Thanks, I didn't know such an option on moonlight existed, still this has the issue  that audio is still audible from the the moonlight client. Also we will be sending unnecessary audio data

ReenigneArcher (commented):
It's probably not obvious, but the strings for the UI are located in this file: https://github.com/LizardByte/Sunshine/blob/master/src_assets/common/assets/web/public/assets/locale/en.json

This is the only file that should be updated as the others are done via automation from CrowdIn.

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -227,6 +227,8 @@
     "hevc_mode_desc": "Allows the client to request HEVC Main or HEVC Main10 video streams. HEVC is more CPU-intensive to encode, so enabling this may reduce performance when using software encoding.",
     "high_resolution_scrolling": "High Resolution Scrolling Support",
     "high_resolution_scrolling_desc": "When enabled, Sunshine will pass through high resolution scroll events from Moonlight clients. This can be useful to disable for older applications that scroll too fast with high resolution scroll events.",
+    "disable_audio" : "Disable Audio Streaming",
+    "disable_audio_desc" : "Disable Streaming Audio Completely, useful for streaming headless monitors",
```
ReenigneArcher, line 231:
```suggestion
    "disable_audio_desc" : "Disable Streaming Audio Completely, useful for streaming headless monitors.",
```

docs/configuration.md
```diff
@@ -770,6 +770,29 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### disable_audio
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            This option allows to disable streaming audio completely, useful for streaming of headless displays as second monitors
```
ReenigneArcher, line 779:
```suggestion
            This option allows to disable streaming audio completely, useful for streaming of headless displays as second monitors.
```

ReenigneArcher: Perhaps the wording should more closely match what's in the UI?

ReenigneArcher (commented):
This seems pretty good to me. I think my only possible change would be to reverse the boolean.

e.g. rename it to "Stream Audio" or something along those lines, with the default value of enabled. I think all of our options follow this logic and none of them are like "disable keyboard" for example. What do you think?

Do you have an idea how would could do the same for video? I'm not sure if it's as straight forward as it is for audio, but this has been requested quite a bit.

Clutchnp:
Yup changed the wording to stream_audio instead of disable_audio as you suggested, as for video I don't understand, what you mean, like the whole point of sunshine is streaming video right? if we disable that it dosen't make sense to use sunshine or am I understanding sth wrong?

ReenigneArcher:
> as for video I don't understand what you mean, like the whole point of sunshine is streaming video right?

Yea, that's the primary purpose. There have been requests to stream ONLY input or ONLY audio. It's definitely not a common use case though.

Clutchnp:
Only thing that comes to mind, is creating a dummy encoder that sends blanks or does nothing, what do you think?

ReenigneArcher:
Yea, I guess Moonlight would need the majority of changes for a no video mode.

Thanks for the updates on the PR! I will review in detail tomorrow and if all looks good will get this one merged.

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -227,6 +227,8 @@
     "hevc_mode_desc": "Allows the client to request HEVC Main or HEVC Main10 video streams. HEVC is more CPU-intensive to encode, so enabling this may reduce performance when using software encoding.",
     "high_resolution_scrolling": "High Resolution Scrolling Support",
     "high_resolution_scrolling_desc": "When enabled, Sunshine will pass through high resolution scroll events from Moonlight clients. This can be useful to disable for older applications that scroll too fast with high resolution scroll events.",
+    "stream_audio" : "Stream Audio""
+    "stream_audio_desc" : "Weather to Stream Audio or Not, Disabling this can be useful for streaming headless displays as second monitors.",
```
ReenigneArcher, line 231:
```suggestion
    "stream_audio": "Stream Audio",
    "stream_audio_desc": "Whether to stream audio or not. Disabling this can be useful for streaming headless displays as second monitors.",
```

Could you also put the keys in alphabetical order?

docs/configuration.md
```diff
@@ -770,6 +770,29 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### stream_audio
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            Weather to Stream Audio or Not, Disabling this can be useful for streaming headless displays as second monitors.
```
ReenigneArcher, line 779:
```suggestion
            Whether to atream audio or not. Disabling this can be useful for streaming headless displays as second monitors.
```

ReenigneArcher (changes requested):

ReenigneArcher (dismissed):

ReenigneArcher:
There's a lint failure: https://github.com/LizardByte/Sunshine/actions/runs/14776649715/job/41486304122?pr=3833

Issue #3840, opened 2025-05-02
Error passing argument 4 of notify_notification_add_action


rubin55:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

```
/tmp/sunshine-git/src/Sunshine/third-party/tray/src/tray_linux.c: In function ‘tray_update_internal’:
/tmp/sunshine-git/src/Sunshine/third-party/tray/src/tray_linux.c:103:87: error: passing argument 4 of ‘notify_notification_add_action’ from incompatible pointer type [-Wincompatible-pointer-types]
  103 |         notify_notification_add_action(currentNotification, "default", "Default", tray->notification_cb, NULL, NULL);
      |                                                                                   ~~~~^~~~~~~~~~~~~~~~~
      |                                                                                       |
      |                                                                                       void (*)(void)
In file included from /usr/include/libnotify/notify.h:27,
                 from /tmp/sunshine-git/src/Sunshine/third-party/tray/src/tray_linux.c:19:
/usr/include/libnotify/notification.h:202:85: note: expected ‘NotifyActionCallback’ {aka ‘void (*)(NotifyNotification *, char *, void *)’} but argument is of type ‘void (*)(void)’
  202 |                                                                NotifyActionCallback callback,
      |                                                                ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
/usr/include/libnotify/notification.h:120:19: note: ‘NotifyActionCallback’ declared here
  120 | typedef void    (*NotifyActionCallback) (NotifyNotification *notification,
      |                   ^~~~~~~~~~~~~~~~~~~~
```

### Expected Behavior

Expected sunshine master to compile.

### Additional Context

I build master regularly, this is a recently intruduced issue. I suspect the interface of libnotify changed.

### Host Operating System

Linux

### Operating System Version

Arch Linux, kernel 6.14.4

### Architecture

amd64/x86_64

### Sunshine commit or version

8ef838b

### Package

other (self built)

### GPU Type

NVIDIA

### GPU Model

4090

### GPU Driver/Mesa Version

570.144

### Capture Method

KMS (Linux)

### Config

```Shell
resolutions = [
    1280x720,
    1920x1080,
    1920x1200,
    2880x1920,
    3840x2160,
    7680x4320
]
fps = [30,60]
[...]

PR #3846, opened 2025-05-05
build(linux): fix gcc version and disable docs for build on arch

Clutchnp:
## Description
This commit locks the gcc version to 14 instead of 15.x.x currently default on arch and also disables docs as they are currently broken on arch. This fixes arch builds

### Issues Fixed or Closed
- Fix bug #3840 

## Type of Change

## Checklist

### DIFF
--- a/scripts/linux_build.sh
+++ b/scripts/linux_build.sh
@@ -96,6 +96,8 @@ function add_arch_deps() {
     'base-devel'
     'cmake'
     'curl'
+    "gcc-${gcc_version}"
+    "gcc${gcc_version}-libs"
     'git'
     'libayatana-appindicator'
     'libcap'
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
The docs should automatically be disabled after checking the doxygen version

Clutchnp:
arch is on doxygen version 1.13.2 it will pass through, the scripts only check for lower versions not higher

Clutchnp:
ok i did that and as for max version i went off of this comment https://github.com/LizardByte/Sunshine/issues/3852#issuecomment-2850813592

scripts/linux_build.sh
```diff
@@ -382,8 +386,11 @@ function run_install() {
     "gcc-ranlib"
   )
 
-  # update alternatives for gcc and g++ if a debian based distro
-  if [ "$distro" == "debian" ] || [ "$distro" == "ubuntu" ]; then
+  #set gcc version based on distros 
+  if [ "$distro" == "arch" ]; then
+    export CC=gcc-14
+    export CXX=g++-14
```
ReenigneArcher, line 392: How will this work when using cuda? Last time I checked cuda on arch was using gcc13, but that may have been updated since.

Issue #2439, opened 2024-04-18
Wrong screen orientation on Steam Deck (Linux) using KMS


Lamfi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

Fedora 39 - Bazzite Linux
Steam Deck
Wayland
![image](https://github.com/LizardByte/Sunshine/assets/37219244/346c9641-be22-4f29-a8d6-3e4a05a10941)

Sunshine from ujust setup-sunshine also tried from dnf
Same problem when i try x11 :0

### Host Operating System

Linux

### Operating System Version

Fedora 39 - Bazzite Linux

### Architecture

64 bit

### Sunshine commit or version

Last version

### Package

Linux - pkg.tar.zst

### GPU Type

AMD

### GPU Model

Steam Deck custom

### GPU Driver/Mesa Version

idk

### Config

```Shell
capture = kms
encoder = vaapi
```

### Relevant log output

```shell
SadShrimp, [18.04.2024 9:54]
[2024:04:18:09:56:57]: Info: Sunshine version: 0.23.0.8999264
[2024:04:18:09:56:57]: Info: /dev/dri/card1 -> evdi
[2024:04:18:09:56:57]: Warning: No render device name for: /dev/dri/card1
[2024:04:18:09:56:57]: Info: /dev/dri/card2 -> evdi
[2024:04:18:09:56:57]: Warning: No render device name for: /dev/dri/card2
[2024:04:18:09:56:57]: Info: /dev/dri/card3 -> evdi
[2024:04:18:09:56:57]: Warning: No render device name for: /dev/dri/card3
[2024:04:18:09:56:57]: Info: /dev/dri/card4 -> evdi
[2024:04:18:09:56:57]: Warning: No render device name for: /dev/dri/card4
[2024:04:18:09:56:57]: Info: /dev/dri/card5 -> amdgpu
[2024:04:18:09:56:57]: Error: Environment variable WAYLAND_DISPLAY has not been defined
[2024:04:18:09:56:57]: Info: Trying encoder [vaapi]
[2024:04:18:09:56:57]: Info: Screencasting with KMS
[2024:04:18:09:56:57]: Info: /dev/dri/card1 -> evdi
[2024:04:18:09:56:57]: Warning: No render device name for: /dev/dri/card1
[...]

PR #711, opened 2023-01-05
Steam Deck Gamingmode fix

safijari:
## Description
Added a function to check panel orientation in kmsgrab and for the case when the orientation is 270 implemented swapping width and height. This change allows the Steam deck gamingmode to stream correctly.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [X] I want maintainers to keep my branch updated

PR #3893, opened 2025-05-20
fix(linux): steamdeck pointer orientation

alanjds:
## Description
I am experiencing an issue when using Moonlight Android that also affects its fork Artemis: When serving Sunshine from a SteamDeck, the pointer orientation does not match the screen orientation. I mean, when moving the finger horizontally, the mouse pointer on the streamed screen moves vertically.

My current suspicion is that the rotation sourced from DRM_MODE_ROTATE_270 changes the _image_ orientation but does not change the _input_ orientation.

So, instead of advocating for applying the same solution to input handling in Sunshine, I am experimenting with removing the _image_ rotation on Sunshine and handling both the _image and input_ rotation on client side. 

That would be cleaner IMHO because Sunshine would be responsible to provide accurate image and input handling, and the decision on how to _present_ these would be kept in full on the client side. Would still be useful if Sunshine provides the rotation info it discovered, _as metadata_ to the client.

### Why a Draft PR?
This PR is opened as Draft because I could compile a working binary from current `master` Sunshine on Ubuntu 24.04 but I failed to make it build any package with `cpak`, and the bare binary does not run on Steamdeck because some library missing.

So I am pushing a barebones PR just to make CI build the needed packages on my behalf, instead of preparing Steamdeck to build the stuff directly.

### Issues Fixed or Closed
- Relates to #3203
- Relates to #2439 
- Practically reverts #711 

## Type of Change

## Checklist

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -561,6 +561,7 @@ namespace config {
     true,  // always send scancodes
     true,  // high resolution scrolling
     true,  // native pen/touch support
+    0,  // input rotation
   };
 
   sunshine_t sunshine {
@@ -1222,6 +1223,7 @@ namespace config {
 
     bool_f(vars, "high_resolution_scrolling", input.high_resolution_scrolling);
     bool_f(vars, "native_pen_touch", input.native_pen_touch);
+    int_f(vars, "output_rotation", input.output_rotation);
 
     bool_f(vars, "notify_pre_releases", sunshine.notify_pre_releases);
 
--- a/src/config.h
+++ b/src/config.h
@@ -202,6 +202,7 @@ namespace config {
 
     bool high_resolution_scrolling;
     bool native_pen_touch;
+    int output_rotation;
   };
 
   namespace flag {
--- a/src/platform/linux/input/inputtino.cpp
+++ b/src/platform/linux/input/inputtino.cpp
@@ -34,9 +34,26 @@ namespace platf {
     delete input;
   }
 
+  std::pair<int, int> rotate_delta(int delta_x, int delta_y, int rotation) {
+    switch (rotation) {
+      case 0:
+        return {delta_x, delta_y};
+      case 90:
+        return {delta_y, -delta_x};
+      case 180:
+        return {-delta_x, -delta_y};
+      case 270:
+        return {-delta_y, delta_x};
+      default:
+        return {delta_x, delta_y};
+    }
+  }
+
   void move_mouse(input_t &input, int deltaX, int deltaY) {
     auto raw = (input_raw_t *) input.get();
-    platf::mouse::move(raw, deltaX, deltaY);
+
+    auto rotated = rotate_delta(deltaX, deltaY, config::input.output_rotation);
+    platf::mouse::move(raw, rotated.first, rotated.second);
   }
 
   void abs_mouse(input_t &input, const touch_port_t &touch_port, float x, float y) {
--- a/src/platform/linux/kmsgrab.cpp
+++ b/src/platform/linux/kmsgrab.cpp
@@ -695,13 +695,13 @@ namespace platf {
 
               switch (card.get_panel_orientation(plane->plane_id)) {
                 case DRM_MODE_ROTATE_270:
-                  BOOST_LOG(debug) << "Detected panel orientation at 90, swapping width and height.";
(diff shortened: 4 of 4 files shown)


ReenigneArcher:
Thank you for the PR

> This PR is opened as Draft because I could compile a working binary from current master Sunshine on Ubuntu 24.04 but I failed to make it build any package with cpak, and the bare binary does not run on Steamdeck because some library missing.

> So I am pushing a barebones PR just to make CI build the needed packages on my behalf, instead of preparing Steamdeck to build the stuff directly.

I ran the workflow, artifacts should be available shortly. `cpak` won't create a package for ArchLinux, but there will be one generated by our CI.

alanjds:
> I ran the workflow, artifacts should be available shortly. `cpak` won't create a package for ArchLinux, but there will be one generated by our CI.

Hum... Thanks for triggering the CI. However I could not find the built artifacts. Where are they stored?

alanjds:
Btw, I am running Bazzite on that SteamDeck, so may AppImage be the best option, if I understood correctly the installation docs.

alanjds:
I just tested the built AppImage in a standard Lenovo laptop with the screen correct and then rotated. Moonlight Android input handling worked correctly in both cases.

On Moonlight Android I now see black borders but correct input handling. On Artemis I can use some option to get rid of black borders if using the cellphone in Portrait mode when connecting. Then just turn the phone and kept using :)

Will test the same AppImage in SteamDeck before speaking more, but seems that this custom handling is indeed messing with the input handling of the clients.

ReenigneArcher:
> > I ran the workflow, artifacts should be available shortly. `cpak` won't create a package for ArchLinux, but there will be one generated by our CI.
> 
> Hum... Thanks for triggering the CI. ~However I could not find the built artifacts. Where are they stored?~
> 
> Just found the link inside the CI action: https://github.com/LizardByte/Sunshine/actions/runs/15139957404/artifacts/3161359794. Sorry for nudging.

Deb and Arch packages are in ci-docker. https://github.com/LizardByte/Sunshine/actions/runs/15139957392?pr=3893

alanjds:
Done tests on SteamDeck running Bazzite. Both `master` and this custom patched version behave exactly the same. That is surprising.

On both servers (master and patched) the Moonlight Android seems to always respect the _hardware_ orientation for image and the _desktop orientation_ for input.

I am inclined to think that:
- The orientation patch now on `master` should be removed
- Input should be transformed from desktop to hardware _somewhere_.

This can be changed on Sunshine or on Moonlight. But as Sunshine is the one sending _hardware orientation_ and interpreting input as _desktop orientation_, I think so far that Sunshine is the one that should "fix" this by making the orientation unique.

As changing the _hardware output_ orientation seems harder (or impossible) as demonstrated by the lack of effect of this PR change, seems that _desktop input_ orientation is the place possible to be patched.

_For now_ I think that this is easier do this on Sunshine side, because it already have the desktop orientation information available (DRM_MODE_ROTATION_* detection).

What do you think?

ReenigneArcher:
I agree. The transformations should be done on the Sunshine side.

alanjds:
@ReenigneArcher could you please release the CI again?

alanjds:
More findings ahead:

> Done tests on SteamDeck running Bazzite. Both `master` and this custom patched version behave exactly the same. That is surprising.

They behave the same _on Desktop Mode_, including Steam Big Picture. When on Gamescope Mode, the patches for DRM_ROTATION are needed.

> On both servers (master and patched) the Moonlight Android seems to always respect the _hardware orientation_ for image and the _desktop orientation_ for input.

The patches I introduced for `move_mouse()` do fix the pointer behavior on Desktop Mode, but breaks the behavior on Gamescope Mode.

So, I think that is better to have the input rotation transform as _an option per App_, allowing a client to choose when to apply them depending on the use. That also frees Sunshine from doing bad assumptions and wrongly applying transformations.

> I am inclined to think that:
> 
>     * The orientation patch now on `master` should be removed
It should not. The orientation patch :point_up:  is needed on Gamescope Mode, and do not harm the Desktop Mode.

>     * Input should be transformed from desktop to hardware _somewhere_.
It should. But not always. Better if is an option. Best if the option is per-app.

alanjds:
I still want to put patches for `touch_update` and `pen_update` events, but the current code do fix the issue if `output_rotation` is set on ~/.config/sunshine.sunshine.conf

I will also rethink about the config name, up to when the PR gets ready for review

PR #2006, opened 2024-01-08
package/linux: Provide launchable tags in appstream metainfo file

joebonrichie:
## Description
By default the .desktop file gets installed to sunshine.desktop, as such provide the launchable tag for it to help `appstream-builder` "find" the .desktop file due to the id (Sunshine.desktop) not matching.

Rename the id to just @PRODUCT_NAME@ as well since the .desktop suffix is no longer required.

Provide URL as well.

Spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-launchable

Fixes appstream-builder generation for this package in Solus.

### Screenshot
N/A

### Issues Fixed or Closed
N/A

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I want maintainers to keep my branch updated

PR #3901, opened 2025-05-24
package/linux: Fixup metainfo to keep appstream composers happy

joebonrichie:
## Description

1. By default the .desktop file gets installed to sunshine.desktop, as such provide the launchable tag for it to help `appstream-builder` "find" the .desktop file due to the id (Sunshine.desktop) not matching.

Rename the id to just @PRODUCT_NAME@ as well since the .desktop suffix is no longer required.

Provide URL as well.

Spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-launchable

Fixes `appstream-builder` generation.

2. Set the metadata_license to FSFAP. To be valid, it has to be one of the licenses vetted by the spec, otherwise `appstreamcli-compose` will error.

Spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-metadata_license

Fixes generation with `appstreamcli-compose` and thus distros which use `appstream-generator`.

### Screenshot
N/A

### Issues Fixed or Closed
Follow up to #2006.

## Type of Change


packaging/linux/sunshine.appdata.xml
```diff
@@ -11,6 +11,8 @@
             @PROJECT_LONG_DESCRIPTION@
         </p>
     </description>
+    <launchable type="desktop-id">sunshine.desktop</launchable>
+    <launchable type="url">https://localhost:47990</launchable>
```
ReenigneArcher, line 15: The port is customizable, so I'm not sure it should be listed here?

packaging/linux/sunshine.appdata.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
-    <id>@PROJECT_NAME@.desktop</id>
-    <metadata_license>@PROJECT_LICENSE@</metadata_license>
+    <id>@PROJECT_NAME@</id>
+    <metadata_license>FSFAP</metadata_license>
```
ReenigneArcher, line 4: We're using `CC0-1.0` for flatpak, probably better to commonize it.

ReenigneArcher (commented):
Thank you for the PR! I have a few comments below.

Do you think there is anything else we can commonize between the flatpak version and the standard one?

ReenigneArcher:
@joebonrichie Apologies for the delay. There were several CI build failures that I had to address before merging this. Anyway, this will be merged, and then I will work on a follow up PR to validate these in CI (I may request a review from you on that if it's okay).

Issue #3469, opened 2024-12-20
Windows service start type gets overwritten on Sunshine updates


ReenigneArcher:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

Windows service start type gets overwritten on Sunshine updates

### Expected Behavior

Query for the existing service, get the start method, and apply that to the new service when re-installing.

### Host Operating System

Windows

### Operating System Version

*

### Architecture

amd64/x86_64

### Sunshine commit or version

-

### Package

Windows - installer (recommended)

### GPU Type

AMD

### GPU Model

-

### GPU Driver/Mesa Version

-

### Capture Method

None

### Config

```Shell

```

### Apps

```json

```

### Relevant log output

```shell
-
```

PR #3902, opened 2025-05-24
fix(installer/windows): remember service start type

ReenigneArcher:
## Description

This PR will make the installer "remember" the previous type of service start type on Windows. If the user selects to "auto start" during installation it will overwrite the previous service start type.

### Issues Fixed or Closed

- Fixes https://github.com/LizardByte/Sunshine/issues/3469
- Closes https://github.com/LizardByte/Sunshine/pull/3867

## Type of Change

## Checklist

### DIFF
--- a/src_assets/windows/misc/service/install-service.bat
+++ b/src_assets/windows/misc/service/install-service.bat
@@ -1,10 +1,13 @@
 @echo off
+setlocal enabledelayedexpansion
 
 rem Get sunshine root directory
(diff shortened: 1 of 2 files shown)


Nonary:
Not sure why you removed the part where it doesn't always set it to auto start.

Users are not going to uncheck that option every time it's upgraded.

We want to make it so users can easily upgrade sunshine by inheriting the same settings during an upgrade. If we make it so they have to go into the installer options and uncheck it, nobody is going to benefit from these changes at all

Personally I think after these modifications the entire PR is pointless.

PR #3999, opened 2025-06-22
feat: rework login authentication and add API token support

Nonary:
## Description  
Adds **token-based authentication** to Sunshine’s API and admin UI.  
Adds token based sessions as an alternative to basic authentication.
Admins can now create, view, and revoke login-free API tokens that grant **only the routes they choose**.  
The existing Basic-Auth flow is untouched, this adds on top of the existing code; therefore there is no breaking changes.

### Why this is secure  

| Measure | What it means |
|---------|---------------|
| **Hash-only storage** | Raw token is shown once, then discarded; only its SHA-256 hash is stored, so a leaked state file reveals nothing usable. |
| **Least-privilege scopes** | Each token carries an allow-list of API routes + verbs; any request outside that list is rejected. |
| **One-click revocation** | Tokens can be invalidated instantly from the new **API Tokens** page or via the API. |
| **Non-breaking** | Token logic is additive; Basic-Auth requests are processed exactly as before. |

### Benefits  

* **Safe automation & integrations** – third-party scripts and apps can call the Sunshine API without embedding the admin’s full credentials.  
* **Minimal blast-radius** – each token can be limited to just the endpoints it needs (e.g., `/api/apps` GET only), reducing risk, where as using basic auth would give full access to everything in Sunshine.  
* ** Password manager support ** - Browsers can now safely autofill the login.

### Screenshot  
image

### Issues Fixed or Closed  
_None – this is a new capability._

## Type of Change  
- [ ] Bug fix  
- [x] New feature  
- [ ] Breaking change  
- [ ] Dependency update  
- [x] Documentation update  
- [x] Repository update  

## Checklist  
- [x] Code follows project style guidelines  
- [x] Self-review completed  
- [x] Security-critical sections are commented  
- [x] Docs and inline docstrings updated


Nonary:
There appears to be a weird bug in current master branch where there is like a ghost session showing in Moonlight. Thought this may have caused it but it does not.

Nonary:
> There appears to be a weird bug in current master branch where there is like a ghost session showing in Moonlight. Thought this may have caused it but it does not.

Nevermind, only happens when you add a blank app to Sunshine... still a weird bug though

Nonary:
Okay the unit tests have now been added for both frontend and backend for all of the code changes.

Nonary:
Flipping back to draft as I do extensive testing against the login screen and session handling.

Right now, the session handling and redirect is a prototype and has not been security reviewed by me yet.

Nonary:
Okay, it should be secure enough now. Next steps is to polish the code and remove any dead code, pointless comments and bugfixes if applicable.

Nonary:
> I like the idea of adding tests to the web-ui part, but I think that should be done as a separate PR and before this one. It might be better to do after https://github.com/LizardByte/Sunshine/pull/3855 (or part of) as well.
> 
> 
> 
> Also, using common strategies to the following where possible (to ease the maintenance burden).
> 
> 
> 
> - https://github.com/LizardByte/shared-web
> 
> - https://github.com/LizardByte/gamepad-helper
> 
> - https://github.com/LizardByte/contribkit (doesn't actually have tests, but the framework is there)
> 
> 
> 
> We'll also need to upload coverage to codecov for the web-ui. We already have upload coverage steps, but we'll need to include the reports from this part. There are examples in all 3 of the above.

Fine with me, it was mostly there to help me confirm everything is working without having me to constantly recompile sunshine and test by hand.

It is also kind of hacked in, I had to configure jest instead of using vite testing framework because our Vue setup is kinda chaotic. I am way more experienced in jest so went with that.

But probably would focus more on trying to figure out getting the built in testing framework for vite working instead on the separate PR.

Nonary:
I removed the frontend unit tests, did another sweep through in cleanup some dead code and comments. I've run through a lot of manual testing and automated testing, so I believe it is ready.

Nonary:
I was going to re-submit this one with it being stripped down to just the login portion to reduce PR size, since it technically was not asked for specifically in the roadmap. Putting back in draft for now, right now my top priority is WGC capture so not sure when I will come back to this one.

Nonary:
After reviewing this again nothing needs to be split but there is one more feature I need to add. The frontend should redirect to login screen when an auth error is encountered, so will add that in over the weekend.

Nonary:
Added better session timeouts, cleaned up the dead code. Login redirects are now working properly, added a new config option for users to define the session ttl for login sessions. Improved UX of the API token page, it's looks much better compared to the original version.

I think that covers everything this needed

PR #4063, opened 2025-07-10
Add missing libdrm dependency to Wayland backend

chewi:
## Description

wayland.cpp makes use of libdrm, so it should be pulled in as a dependency. The build fails otherwise.

## Type of Change

## Checklist

### DIFF
--- a/cmake/compile_definitions/linux.cmake
+++ b/cmake/compile_definitions/linux.cmake
@@ -123,6 +123,7 @@ endif()
 # wayland
 if(${SUNSHINE_ENABLE_WAYLAND})
     find_package(Wayland REQUIRED)
+    find_package(LIBDRM REQUIRED)
 else()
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Maybe we should just fail if DRM is OFF and WAYLAND is ON? https://github.com/LizardByte/Sunshine/blob/feac80656d211569c3c4eef839f284d386c90093/cmake/compile_definitions/linux.cmake#L88-L103

chewi:
I thought that, but it's about as much code either way. I figured I might as well give Gentoo users what they're asking for. Also, the DRM backend additionally requires libcap.

ReenigneArcher:
@chewi, thank you for the PR! I ended going an alternate route in https://github.com/LizardByte/Sunshine/pull/4070 which wouldn't result in duplicated includes or platform libraries.

PR #4071, opened 2025-07-13
fix(windows): Add dynamic cursor fix (mousekeys) for KVM switches

tstokes22:
## Description

This PR implements a dynamic fix that listens for the DBT_DEVNODES_CHANGED message.

It hooks into the WM_DEVICECHANGE window message in main.cpp. When triggered, it calls a new function, platf::check_and_force_cursor_visibility(), which safely checks if a mouse is present.

If no mouse is detected, it toggles the Windows "Mouse Keys" accessibility feature. This forces a system-wide cursor redraw, making the cursor visible on the stream again without affecting user settings.

This same function is now also used at stream startup to unify the cursor handling logic.

## Type of Change

## Checklist


src/main.cpp
```diff
@@ -74,6 +84,14 @@ LRESULT CALLBACK SessionMonitorWindowProc(HWND hwnd, UINT uMsg, WPARAM wParam, L
         lifetime::exit_sunshine(0, false);
         return 0;
       }
+    case WM_DEVICECHANGE:
+      {
+        // Check if devices change
+        if (wParam == DBT_DEVNODES_CHANGED) {
+          platf::check_and_force_cursor_visibility();
```
cgutman: You should only enable this behavior if we're actually streaming.

cgutman (changes requested):

src/main.cpp
tstokes22: Thank you, updated the PR to include a check is_streaming.

Issue #4199, opened 2025-08-27
SIGTRAP crash in `stream::session::join` due to deadlock with system tray/notification thread on Wayland


Horcag:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

When a streaming session is terminated (client disconnects), Sunshine attempts to update its system tray icon and send a notification. On a Wayland session (Hyprland) with a notification daemon (`dunst` or `swaync`), this action causes the notification thread to hang indefinitely. This, in turn, blocks the main session teardown logic. After a timeout, a watchdog in the session management code triggers a deliberate `debug_trap`, terminating the entire Sunshine process with a `SIGTRAP` signal.
This crash often leads to a GPU hang, causing the entire graphical session to freeze. The session may recover after several minutes (likely due to a GPU reset by the kernel driver), but sometimes it requires a hard reboot (`REISUB`).

**Reproduction Steps:**
1.  Run Sunshine (built from source) on Arch Linux with Hyprland (Wayland) and a running notification daemon (e.g., `dunst`).
2.  Connect a client and start a desktop stream.
3.  Disconnect the client.
4.  Observe that the Sunshine process crashes after a short hang.

### Expected Behavior

Sunshine should handle an unresponsive notification daemon gracefully. If updating the system tray or sending a notification times out, it should log an error for that specific subsystem and continue the session teardown process without crashing the entire application. A non-critical component like notifications should not be able to cause a fatal application error that destabilizes the host's graphical environment.

### Additional Context

This is not a random segfault but a deliberate application termination. The provided coredump with debug symbols clearly shows the chain of events:
1.  The crashing thread (3886 in the attached dump) is terminated by `lifetime::debug_trap()`.
2.  This trap is called from a lambda function within `stream::session::join`, which acts as a watchdog timer.
3.  The reason for the timeout is that `stream::session::join` is blocked waiting for the system tray update to complete (see thread 4027).
4.  The system tray thread (4027) is, in turn, stuck inside a call to `tray_update`, which then calls into `libnotify.so.4` (see thread 3887 calling `notify_notification_close`). This call to the notification daemon never returns, causing a deadlock.

[...]

PR #3818, opened 2025-04-27
fix(tray): run tray in main event loop enabling support for macOS

ReenigneArcher:
## Description

This PR aims to allow the system tray to function in macOS which requires the tray to run in the main thread (from my understanding).

TODO:
- [ ] Sunshine does not exit from the sigterm (ctrl+c for example), although it does exit when selecting exit in the system tray menu
- [ ] Double check cmake files
- [ ] Test tray
  - [ ] Windows
  - [ ] Linux
  - [ ] macOS

### Issues Fixed or Closed

- Closes https://github.com/LizardByte/roadmap/issues/28

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

PR #4208, opened 2025-08-29
feat(tray): Add runtime config option to enable/disable system tray

Kishi85:
## Description

Add a runtime config option to enable/disable system tray (and desktop notifications associated with it) so Users can disable the tray icon if they wish to without having to recompile Sunshine. This is especially helpful if there are issues related to the system try (like #4199).

### Issues Fixed or Closed

- Resolves https://github.com/orgs/LizardByte/discussions/334

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -112,7 +112,7 @@ void mainThreadLoop(const std::shared_ptr<safe::event_t<bool>> &shutdown_event) 
   while (true) {
     if (shutdown_event->peek()) {
       BOOST_LOG(info) << "Shutdown event detected, breaking main loop"sv;
-      if (tray_is_enabled) {
+      if (tray_is_enabled && config::sunshine.system_tray) {
         system_tray::end_tray();
       }
       break;
@@ -418,7 +418,7 @@ int main(int argc, char *argv[]) {
   }
 
   // Stop the threaded tray if it was started
-  if (tray_is_enabled) {
+  if (tray_is_enabled && config::sunshine.system_tray) {
     system_tray::end_tray_threaded();
   }
 #endif
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -384,7 +384,7 @@ int main(int argc, char *argv[]) {
   }
 #endif
 
-  if (tray_is_enabled) {
+  if (tray_is_enabled && config::sunshine.system_tray) {
     BOOST_LOG(info) << "Starting system tray"sv;
 #ifdef _WIN32
     // TODO: Windows has a weird bug where when running as a service and on the first Windows boot,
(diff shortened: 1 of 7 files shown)


Kishi85:
> Thank you for another PR!
> 
> This looks pretty good to me. There's a few places that need additional updates when new config options are added.
> 
> * docs: docs/configuration.md
> * config ui: src_assets/common/assets/web/config.html
> 
> The option should be in the same order in all places.

I've made the changes as requested. Thanks for clarifying what I've missed!

This made me realize that I've also missed the config ui point for the other option I've added so I'll put up another PR with a fix for that in a minute. Also it might be a good idea to add some kind of unit test for config options (maybe just grepping for the option in all relevant files would be enough for a start) so this gets caught early.

ReenigneArcher:
Lint workflow might be failing now because the branch is out of date. I'm not 100% sure on that.

Kishi85:
I've rebased it to current master. Let's see if that fixes it.

src_assets/common/assets/web/configs/tabs/General.vue
```diff
@@ -133,6 +133,14 @@ function removeCmd(index) {
               v-model="config.notify_pre_releases"
               default="false"
     ></Checkbox>
+
+    <!-- Enable system tray -->
+    <Checkbox class="mb-3"
```
ReenigneArcher, line 138: This should not be shown for macOS.

ReenigneArcher (commented):
I thought of one more thing, since the tray doesn't work on macOS yet.

I'm trying to change that in https://github.com/LizardByte/Sunshine/pull/3818 but there is one bug that needs to be fixed before that can be merged.

ReenigneArcher:
Apologies, but I ended up wrapping up https://github.com/LizardByte/Sunshine/pull/3818 last night.

This will require fixing a few conflicts and the tray now works on macOS as well.

Kishi85:
> Apologies, but I ended up wrapping up #3818 last night.
> 
> This will require fixing a few conflicts and the tray now works on macOS as well.

I've rebased to the new master and squashed my changes in the process to simplify adapting this to #3818. 

I moved the config value check from system_tray.cpp to main.cpp right where the system tray is initially started. I'm not sure if this is the best way to do it though. I've also removed the commit with the MacOS restriction.

src/main.cpp
```diff
@@ -384,7 +384,7 @@ int main(int argc, char *argv[]) {
   }
 #endif
 
-  if (tray_is_enabled) {
+  if (tray_is_enabled && config::sunshine.system_tray) {
```
ReenigneArcher, line 387: There's another on line 421 that probably needs this as well.

Kishi85: IMHO it's not necessary because the way ```end_tray_threaded``` is implemented it checks if the thread is running first thing when called and stops further processing if it is not running. Same thing for ```end_tray``` on Line 115 which checks if tray was initialized before cleaning up.

I'll add both to be consistent but it's basically double checking that way and would work fine without those if-Statements on Line 115+421 from logical standpoint.

PR #4209, opened 2025-08-29
feat(macOS): Capture audio on macOS using Tap API

### LINKED ISSUE #27: Capture audio on macOS using Tap API (Sunshine)
### Is there an existing issue for this item?

### Repositories

LizardByte/Sunshine

### Languages/Skills/Technologies

C++, macOS

### Description

Capture audio using Tap API for macOS.

### Estimated Effort

effort:Large

### Priority

priority:Low

### Target Milestone

3-6 months

### Dependencies

_No response_

ThomVanL:
## Description

This PR adds **system-wide audio tap support for macOS**. The implementation introduces:

* System-wide audio tap functionality to capture audio from all system sources
  * Uses an audio converter to handle varying client audio requirements
* Unit tests covering the new system tap methods, along with additional coverage for the existing microphone path
* Updated UI with a macOS-specific toggle (see screenshots)
* Updated configuration options to support the new audio tap feature
* Added Doxygen documentation for new APIs

**Additional changes**

* Adjusted `cmake` files for compatibility with Homebrew-based setup.
    *  Tweaked dependency detection so that `openssl` and `opus` are found automatically. This replaces the need to run manual `ln` commands, but I’m not sure if this is the best long-term approach. _Feedback welcome_.
    * Updated `cmake/compile_definitions/unix.cmake` to ensure `SUNSHINE_ASSETS_DIR` resolves correctly.
* Updated `src_assets/macos/assets/Info.plist` to prepare for required macOS permission prompts.
* Added a `macos_system_wide_audio_tap` config option to the `audio_t` struct.

**Testing**

* Verified functionality with multiple (Moonlight) clients requiring mono, stereo, 5.1, and 7.1 audio configurations.
* Confirmed audio conversion works as expected across varying client setups.
* Stress-tested with multiple concurrent clients for several hours without memory leaks or race conditions observed.
* Host ran an arm64 build during testing

**Notes**

* My background is primarily in .NET, with some experience in C, C++, and Rust. Objective-C is new to me, but I carefully reviewed and tested the bits on memory management and synchronization.
* I leaned on GitHub Copilot and AI assistance heavily for the Objective-C parts, especially syntax and boilerplate; but I reviewed, debugged and tested everything myself a bunch of times.
* Some of the unit tests were bordering on integration tests, so I tried to keep them focused and not blur the lines too much.
* I developed this feature to make Sunshine more accessible on macOS, especially for less technical users. I have to admit that I fumbled quite a bit getting BlackHole to work. 🙂 But I also wanted to make sure to not break the existing `setupMicrophone` functionality as it is also a viable option!
* I am not entirely convinced on the naming or location of the `macos_system_wide_audio_tap` setting.
    * Open to suggestions here!
* Built this on arm64 macOS 15.6.1 (24G90)

If the AI involvement is a blocker for merging, no worries. I can totally understand! This was also a learning project for me and I had a lot of fun building it.

### Screenshot

**Web UI – Audio/Video Configuration**
New option for enabling system-wide audio recording on macOS. Disables the audio sink option when checked.
a

**macOS Permission Prompt**
System permission request when Sunshine first tries to access system audio:
b

**System Settings – Screen & System Audio Recording**
macOS privacy settings showing Sunshine/Terminal access to "System Audio Recording Only":
c

#### Roadmap Issues

- Closes https://github.com/LizardByte/roadmap/issues/27

## Type of Change

## Checklist

### AI Usage

- [X] **Heavy**: AI generated most or all of the code changes


ReenigneArcher:
Thank you for the PR!

> If the AI involvement is a blocker for merging, no worries. I can totally understand! This was also a learning project for me and I had a lot of fun building it.

This isn't a blocker, we just want to understand how much it was used when developing the PR.

I've went ahead and approved the workflow run so the basic linting and whatnot can be taken care of before full reviews take place.

Looks like there's an issue with the tests. I don't see an error, just kind of looks like it crashes out. https://github.com/LizardByte/Sunshine/actions/runs/17332988293/job/49216925169?pr=4209#step:8:3666

Linting:

- https://github.com/LizardByte/Sunshine/actions/runs/17332988295/job/49216903453?pr=4209#step:11:19
- https://github.com/LizardByte/Sunshine/actions/runs/17332988295/job/49216903453?pr=4209#step:14:87

We'll also need another way to compile the tests without the .mm file on non macOS platforms: https://github.com/LizardByte/Sunshine/actions/runs/17332988293/job/49216934161?pr=4209#step:11:1886

ThomVanL:
Hey @ReenigneArcher, thanks for kicking things off and for clarifying the AI usage policy! 

I’ll dig into those CI issues next. 

Got one question regarding the cross-platform test compilation issue: would you be okay with me updating `tests/CMakeLists.txt` so that those `.mm` files are only included when building on macOS? That seems like the cleanest way to keep non-macOS builds happy while still testing the new audio tap functionality where it applies.

I’ll push updates once I have fixes in place. And thanks again for the feedback! 🙂

ReenigneArcher:
> Got one question regarding the cross-platform test compilation issue: would you be okay with me updating tests/CMakeLists.txt so that those .mm files are only included when building on macOS?

Sounds fine to me and I also believe that's the cleanest and easiest approach.

src/platform/windows/display_base.cpp
ReenigneArcher, line 1: Thanks, I'm not sure if the ci will pass. A new version of clang-format just dropped.

We have two actions, one using v20 and one using the latest. I have a fix here to make them use the same version: https://github.com/LizardByte/.github/pull/510

ReenigneArcher: Yea, looks like the other one failed with this change. You can revert it, it should pass now with the PR I mentioned above.

ThomVanL: Sure thing, should be done now!

ReenigneArcher:
https://github.com/LizardByte/Sunshine/actions/runs/17337137215/job/49225275832?pr=4209#step:8:635

I don't think this error has anything to do with your changes, must be something changed in homebrew recently. We use a custom python based action from https://github.com/LizardByte/actions/tree/master/actions/release_homebrew to validate the homebrew builds. I don't see anything obvious in the release notes for what would cause this: https://github.com/Homebrew/brew/releases

andygrundman:
Thanks for the PR. I have built this on my M1 Pro MBP and have been trying to get it to work, but haven't had much success so far. I am not able to get any sound to be captured and/or sent. So far I am only testing with the simplest use case of audio playing out of my MBP speakers. I edited the code to make the tap and aggregate non-private, so I could try to view the tap/aggregate with Apple's [sample app](https://developer.apple.com/documentation/coreaudio/capturing-system-audio-with-core-audio-taps?language=objc) I can see the tap, but not the aggregate. I'm not sure what's wrong. Can you detail your testing process?

I do seem to be getting OK log entries and since I'm running from iTerm, all my permissions seem to be in order (iTerm has access to many things).

```
[2025-08-29 23:56:39.783]: Info: Detected display: Built-in Retina Display (id: 1) connected: true
[2025-08-29 23:56:39.783]: Info: Configuring selected display (1) to stream
[2025-08-29 23:56:39.841]: Info: Creating encoder [hevc_videotoolbox]
[2025-08-29 23:56:39.841]: Info: Color coding: SDR (Rec. 601)
[2025-08-29 23:56:39.841]: Info: Color depth: 10-bit
[2025-08-29 23:56:39.841]: Info: Color range: MPEG
[2025-08-29 23:56:39.841]: Info: Streaming bitrate is 55987000
[2025-08-29 23:56:39.850]: Info: [hevc_videotoolbox @ 0x143e60e70] This device does not support the max_ref_frames option. Value ignored.
[2025-08-29 23:56:40.712]: Info: Using macOS system audio tap for capture.
[2025-08-29 23:56:40.712]: Info: Sample rate: 48000, Frame size: 240, Channels: 2
[2025-08-29 23:56:40.738]: Info: Aggregate device created with ID: 203
[2025-08-29 23:56:40.738]: Info: Aggregate device created and configured successfully
[2025-08-29 23:56:40.739]: Info: No conversion needed - formats match (device: 48000Hz/2ch)
[2025-08-29 23:56:40.739]: Info: Device properties and converter configuration completed
[2025-08-29 23:56:40.793]: Info: System tap IO proc created and started successfully
[2025-08-29 23:56:40.793]: Info: Audio buffer initialized successfully with size: 8192 bytes
[2025-08-29 23:56:40.793]: Info: System tap setup completed successfully!
[2025-08-29 23:56:40.793]: Info: macOS system audio tap capturing.
[2025-08-29 23:56:40.794]: Info: Opus initialized: 48 kHz, 2 channels, 512 kbps (total), LOWDELAY
```

I had to make the following changes to get it to build:
```
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunguarded-availability-new"
-    AVCaptureDeviceDiscoverySession *discoverySession = [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:@[AVCaptureDeviceTypeBuiltInMicrophone, AVCaptureDeviceTypeExternalUnknown]
+    AVCaptureDeviceDiscoverySession *discoverySession = [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:@[AVCaptureDeviceTypeMicrophone, AVCaptureDeviceTypeExternal]
                                                                                                                mediaType:AVMediaTypeAudio
                                                                                                                 position:AVCaptureDevicePositionUnspecified];
     NSArray *devices = discoverySession.devices;
     BOOST_LOG(debug) << "Found "sv << [devices count] << " devices using discovery session"sv;
     return devices;
-#pragma clang diagnostic pop
```

plus a fix in our input.cpp that breaks the latest Xcode 16.4, I'm surprised if you didn't run into this one. I am running Xcode 16.4 clang-1700.0.13.5).
```
--- a/src/platform/macos/input.cpp
+++ b/src/platform/macos/input.cpp
@@ -534,7 +534,7 @@ const KeyCodeMap kKeyCodesMap[] = {
     if (!output_name.empty()) {
       uint32_t max_display = 32;
       uint32_t display_count;
-      CGDirectDisplayID displays[max_display];
+      CGDirectDisplayID displays[32];
```

ReenigneArcher:
Posting this here for reference in case it's needed for permissions of unit tests. We used to need something like this for macports, but it was never necessary for homebrew.

https://github.com/LizardByte/Sunshine/blob/b662b8e7c2aa23a4673edb1e601f7337458dcc7e/.github/workflows/CI.yml#L841-L898

ThomVanL:
Hey @andygrundman, thanks for taking the time to test the PR. Sorry to hear it’s not working correctly.

Here’s my setup on an M4 MBP and what I did.

```shell
clang --version
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

I made the changes in VS Code and followed the steps in `building.md`:

```shell
mkdir build
cmake -B build -G Ninja -S . 
ninja -C build
```

For testing I did the following steps:

- In the VS Code terminal: `./build/sunshine`
- Accept the allow connections prompt
- Connect with multiple Moonlight clients with the following sample rate/frame size/channel combos:
    - iPad @ 1080p → 48000Hz/240/2ch audio → works
    - Android TV @ 1080p → 48000Hz/240/__8ch__ audio → works
    - iOS @ 360p → 48000Hz/__480__/2ch audio → works

So at one point I'm running three streams simultaneously and the audio plays through the devices. 

My `sunshine.conf` looks like:

```ini
audio_sink = Steam Streaming Speakers
macos_system_wide_audio_tap = true
origin_web_ui_allowed = pc
stream_audio = true
wan_encryption_mode = 2
```

With `min_log_level = debug`, here’s the output when connecting from the 360p device (trimmed for brevity).

```log
[2025-08-30 10:42:27.456]: Info: CLIENT CONNECTED
[2025-08-30 10:42:27.463]: Debug: Start capturing Video
[2025-08-30 10:42:27.463]: Info: Detecting displays
[2025-08-30 10:42:27.463]: Info: Detected display: Built-in Retina Display (id: 1) connected: true
[2025-08-30 10:42:27.463]: Info: Configuring selected display (1) to stream
[2025-08-30 10:42:27.541]: Info: Creating encoder [hevc_videotoolbox]
[2025-08-30 10:42:27.541]: Info: Color coding: SDR (Rec. 601)
[2025-08-30 10:42:27.541]: Info: Color depth: 8-bit
[2025-08-30 10:42:27.541]: Info: Color range: MPEG
[2025-08-30 10:42:27.541]: Info: Streaming bitrate is 1268000
[2025-08-30 10:42:27.550]: Info: [hevc_videotoolbox @ 0x13d01de00] This device does not support the max_ref_frames option. Value ignored.
[2025-08-30 10:42:27.945]: Debug: Start capturing Audio
[2025-08-30 10:42:27.946]: Warning: audio_control_t::set_sink() unimplemented: Steam Streaming Speakers
[2025-08-30 10:42:27.946]: Info: Using macOS system audio tap for capture.
[2025-08-30 10:42:27.946]: Info: Sample rate: 48000, Frame size: 480, Channels: 2
[2025-08-30 10:42:27.946]: Debug: setupSystemTap called with sampleRate:48000 frameSize:480 channels:2
[2025-08-30 10:42:27.946]: Debug: macOS version check passed (running 15.6.1)
[2025-08-30 10:42:27.946]: Debug: System tap initialization completed
[2025-08-30 10:42:27.946]: Debug: Creating tap description for 2 channels
[2025-08-30 10:42:27.946]: Debug: Creating process tap with name: SunshineAVAudio-Tap-0x6000007d00c0
[2025-08-30 10:42:27.949]: Debug: AudioHardwareCreateProcessTap returned status: 0
[2025-08-30 10:42:27.949]: Debug: Process tap created successfully with ID: 140
[2025-08-30 10:42:27.949]: Debug: Creating aggregate device with tap UID: 675EC59C-D7CC-4A25-A093-F2B4B4227895
[2025-08-30 10:42:27.957]: Debug: AudioHardwareCreateAggregateDevice returned status: 0
[2025-08-30 10:42:27.957]: Info: Aggregate device created with ID: 141
[2025-08-30 10:42:27.959]: Debug: Set aggregate device sample rate to 48000Hz
[2025-08-30 10:42:27.959]: Debug: Set aggregate device buffer size to 480 frames
[2025-08-30 10:42:27.959]: Info: Aggregate device created and configured successfully
[2025-08-30 10:42:27.960]: Debug: Device reports 2 input channels
[2025-08-30 10:42:27.960]: Debug: Device properties - Sample Rate: 48000Hz, Channels: 2
[2025-08-30 10:42:27.960]: Debug: needsConversion: NO (device: 48000Hz/2ch -> client: 48000Hz/2ch)
[2025-08-30 10:42:27.960]: Info: No conversion needed - formats match (device: 48000Hz/2ch)
[2025-08-30 10:42:27.960]: Info: Device properties and converter configuration completed
[2025-08-30 10:42:27.960]: Debug: Creating IOProc for aggregate device ID: 141
[2025-08-30 10:42:27.977]: Debug: AudioDeviceCreateIOProcID returned status: 0
[2025-08-30 10:42:27.978]: Debug: Starting IOProc for aggregate device
[2025-08-30 10:42:27.995]: Debug: AudioDeviceStart returned status: 0
[2025-08-30 10:42:27.995]: Info: System tap IO proc created and started successfully
[2025-08-30 10:42:27.995]: Debug: Initializing audio buffer for 2 channels
[2025-08-30 10:42:27.995]: Info: Audio buffer initialized successfully with size: 8192 bytes
[2025-08-30 10:42:27.995]: Info: System tap setup completed successfully!
[2025-08-30 10:42:27.995]: Info: macOS system audio tap capturing.
[2025-08-30 10:42:27.995]: Info: Opus initialized: 48 kHz, 2 channels, 96 kbps (total), LOWDELAY
```

I’m familiar with the sample app! It might be the case that the aggregate device settings are probably still marked as private. You’ll need to flip those to `NO` in two places, once for the tap description and once for the aggregate device.

https://github.com/LizardByte/Sunshine/blob/64047055b86c378fb79835101a78e55b2e33f6e2/src/platform/macos/av_audio.mm#L604

https://github.com/LizardByte/Sunshine/blob/64047055b86c378fb79835101a78e55b2e33f6e2/src/platform/macos/av_audio.mm#L649

Then the tap will show up in Apple's sample app's UI.

Screenshot 2025-08-30 at 10 56 15

And the aggregate device shows up as well.

Screenshot 2025-08-30 at 10 56 22

But even with those values flipped to `YES`, I can still hear audio on the 360p device!

andygrundman:
Thanks for the detailed info. I forgot my log only had Info level, when using Debug my log does look exactly like yours. I feel like I must just have a permission issue, maybe using iTerm as the permission "owner" isn't correct.

Do you see both of these items? What process names are they using? I only get a System Audio item when recording a test file in AudioTapSample.
permissions

Here's basically what my Screen & System Audio Recording settings look like:
privacy

If this is the issue, I wonder how Sunshine can detect that it doesn't actually have permission, hmm.

ThomVanL:
You're right, because I ran into a similar issue with VS Code. Granting it "Screen & System Audio Recording" wasn’t enough; I had to explicitly allow "System Audio Recording Only." From what I found online, the VS Code app bundle itself might be causing the problem. I also tried running `tccutil reset All` to clear permissions, but the behavior stayed the same. I had to explicitly add permissions, but only for VS Code.

Screenshot 2025-08-30 at 13 26 59

I did not even notice the little privacy notice at the top until just now, thanks for that. Here's what it's like on my end.

Screenshot 2025-08-30 at 13 28 52

When I launched `./build/sunshine` directly from the macOS Terminal (not iTerm), I did get the prompt mentioned in my initial message on this PR.

Edit: just to be clear, my dev loop consists of launching sunshine builds through the VS Code integrated terminal.

andygrundman:
Great, adding `usr/local/bin/sushine` manually worked. (I run `sudo make install` probably because I used to have problems with the assets directory or something.)

I worry that this is a nightmare for the average user. I think we'll probably have to ship a proper dmg package so that only the actual app gets the permission and not whatever terminal the user ran brew install from. But I'm getting ahead of myself... I will put in some other review notes I have, after I do some more testing.

ThomVanL:
I completely agree that the user experience shouldn’t be tied to a specific terminal. At some point, you’ll likely want to create an app bundle. I've also added an additional property list key in the `Info.plist` file to help facilitate this a little:

https://github.com/LizardByte/Sunshine/blob/64047055b86c378fb79835101a78e55b2e33f6e2/src_assets/macos/assets/Info.plist#L11-L12

I will check for a way to properly check those required permissions because right now this is missing.

By the way, I ran into the same issue with loading the assets. I initially worked around it by creating a symlink:

```shell
ln -s /Users/<username>/Sunshine/build/assets/web /usr/local/assets/web
```

However, I didn’t like having that lingering around so I updated the `unix.cmake` file to allow passing in the absolute path via `-DSUNSHINE_ASSETS_DIR`. But we could revert this if needed.

https://github.com/LizardByte/Sunshine/blob/64047055b86c378fb79835101a78e55b2e33f6e2/cmake/compile_definitions/unix.cmake#L1-L11

ReenigneArcher:
> By the way, I ran into the same issue with loading the assets. I initially worked around it by creating a symlink:

a `make install` is required after compiling

andygrundman:
I spent a bit of time looking about making a proper app bundle with cmake. It's certainly doable, and mostly just a matter of refactoring the maze of all the include files. App bundle would give us easier permissions, asset storage, signing, much easier for users.

The more I run the Mac version the more I realize the video side of things needs a lot of work too: host latency stat is missing, min/max framerate support, HDR doesn't work, frametime is uneven, it's using BGRA textures so probably using much more memory/cpu/gpu than it should be. It's one of the only apps that runs my MBP fan...  Lots of fun stuff to work on I guess.

What's your use case with Mac Sunshine, if you don't mind me asking? We should assume controller support will also be a lot of work. Apple's "Screen Sharing" app already covers the remote desktop use cases, and no one plays games on their Mac (yet).

ReenigneArcher:
> I spent a bit of time looking about making a proper app bundle with cmake. It's certainly doable, and mostly just a matter of refactoring the maze of all the include files.

I assume static linking would also be needed? We had a pkg or dmg a while back but I removed it because the size was huge with dynamic linking and having to include every single dependency and sub dependency. We may not have been creating the package properly though. I think we were relying on macports to create it for us back then.

Edit: There's a roadmap entry for this here (https://github.com/LizardByte/roadmap/issues/18). I closed it as I was planning to do this via homebrew with a homebrew bottle, but they were not willing to accept all of our required dependencies. If it'll be a dmg instead that's fine, I can re-open it. dmg seems like the better option for macOS anyway.

andygrundman:
I'll have to read up on how cmake handles it but I don't think the size will be that bad. We could probably remove all the software encoders to save space, for example.

The most annoying part will probably be that someone will need to use their Apple dev account to sign & notarize official builds.

docs/configuration.md
```diff
@@ -825,6 +825,31 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### macos_system_wide_audio_tap
```
andygrundman: Is there a reason this needs to be a setting? The main point of confusion with users will be the fact that you don't link your capture to an audio device, the way you do on Windows. If we just defaulted to this, is there a downside? 

Just FYI, when I was first [playing around](https://github.com/LizardByte/Sunshine/compare/master...andygrundman:Sunshine:andyg.macos-audio) with the Tap API I went about it using the
```
CATapDescription *tapDesc = [[CATapDescription alloc] initExcludingProcesses:toExclude
                                                                andDeviceUID:audioSinkUID
                                                                  withStream:0];
```
version which used the default audio device. I never tried the system audio tap which I suppose works better. For example system tap captures audio even when the default audio device is muted, I think device capture would not capture audio in this case, but I haven't tested it.

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,822 @@
...
+ * @param inDevice The audio device identifier
+ * @param inNow Current audio time stamp
+ * @param inInputData Input audio buffer list from the device
+ * @param inInputTime Time stamp for input data
+ * @param outOutputData Output audio buffer list (not used in our implementation)
+ * @param inOutputTime Time stamp for output data
+ * @param inClientData Client data containing AVAudioIOProcData structure
+ * @return OSStatus indicating success (noErr) or error code
+ */
+static OSStatus systemAudioIOProcWrapper(AudioObjectID inDevice, const AudioTimeStamp *inNow, const AudioBufferList *inInputData, const AudioTimeStamp *inInputTime, AudioBufferList *outOutputData, const AudioTimeStamp *inOutputTime, void *inClientData) {
```
andygrundman: Do you think this ObjC call will have any performance overhead? One of the main things the CoreAudio docs drill into you is that these callbacks that run at the highest priority on the audio thread should stay in pure C++ and avoid a lot of things that you wouldn't normally worry about: No malloc/free, no locks, no system calls, and I think no Objective-C method calls, because the ObjC runtime isn't realtime safe. 

That said, maybe it's not something we need to worry about.

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,822 @@
...
+  if ([[NSProcessInfo processInfo] isOperatingSystemAtLeastVersion:((NSOperatingSystemVersion) {10, 15, 0})]) {
+    BOOST_LOG(debug) << "Using modern AVCaptureDeviceDiscoverySession API"sv;
+    // This will generate a warning about AVCaptureDeviceDiscoverySession being
+    // unavailable before macOS 10.15, but we have a guard to prevent it from
+    // being called on those earlier systems.
+    // Unfortunately the supported way to silence this warning, using @available,
+    // produces linker errors for __isPlatformVersionAtLeast, so we have to use
+    // a different method.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunguarded-availability-new"
```
andygrundman, line 187: I had to patch this section to get it to compile, due to AVCaptureDeviceTypeBuiltInMicrophone and AVCaptureDeviceTypeExternalUnknown being deprecated.  It probably wouldn't be the end of the world if we simply required at least macOS 15 and dropped any of this legacy stuff completely. 

```
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunguarded-availability-new"
-    AVCaptureDeviceDiscoverySession *discoverySession = [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:@[AVCaptureDeviceTypeBuiltInMicrophone, AVCaptureDeviceTypeExternalUnknown]
+    AVCaptureDeviceDiscoverySession *discoverySession = [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:@[AVCaptureDeviceTypeMicrophone, AVCaptureDeviceTypeExternal]
                                                                                                                mediaType:AVMediaTypeAudio
                                                                                                                 position:AVCaptureDevicePositionUnspecified];
     NSArray *devices = discoverySession.devices;
     BOOST_LOG(debug) << "Found "sv << [devices count] << " devices using discovery session"sv;
     return devices;
-#pragma clang diagnostic pop
```

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,822 @@
...
+    (NSString *) AVFormatIDKey: [NSNumber numberWithUnsignedInt:kAudioFormatLinearPCM],
+    (NSString *) AVSampleRateKey: [NSNumber numberWithUnsignedInt:sampleRate],
+    (NSString *) AVNumberOfChannelsKey: [NSNumber numberWithUnsignedInt:channels],
+    (NSString *) AVLinearPCMBitDepthKey: [NSNumber numberWithUnsignedInt:32],
+    (NSString *) AVLinearPCMIsFloatKey: @YES,
+    (NSString *) AVLinearPCMIsNonInterleaved: @NO
+  }];
+  BOOST_LOG(debug) << "Configured audio output with settings: "sv << sampleRate << "Hz, "sv << (int) channels << " channels, 32-bit float"sv;
+
+  dispatch_queue_attr_t qos = dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_CONCURRENT, QOS_CLASS_USER_INITIATED, DISPATCH_QUEUE_PRIORITY_HIGH);
```
andygrundman: It might not ever matter in practice, but the docs for AVCaptureAudioDataOutput say "A serial dispatch queue (DISPATCH_QUEUE_SERIAL) must be used to guarantee that audio samples will be delivered in order.".

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,822 @@
...
+      UInt32 deviceChannels = self->ioProcData ? self->ioProcData->aggregateDeviceChannels : 2;
+      UInt32 inputFrames = inputBuffer.mDataByteSize / (deviceChannels * sizeof(float));
+
+      // Use AudioConverter if we need any conversion, otherwise pass through
+      if (self->ioProcData && self->ioProcData->audioConverter) {
+        // Let AudioConverter determine optimal output size - it knows best!
+        // We'll provide a generous buffer and let it tell us what it actually used
+        UInt32 maxOutputFrames = inputFrames * 4;  // Very generous for any upsampling scenario
+        UInt32 outputBytes = maxOutputFrames * clientChannels * sizeof(float);
+        float *outputBuffer = (float *) malloc(outputBytes);
```
andygrundman: (Here's a malloc in an IOProc.) But I'm not clear on how AudioConverter works with system taps. Since it's not tied to a device running at a particular sample rate, do you always create the tap at 48k? When do you have to do explicit resampling? It would be great if the OS just handled this automatically.

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,822 @@
...
+}
+
+- (CATapDescription *)createSystemTapDescriptionForChannels:(UInt8)channels {
+  using namespace std::literals;
+
+  BOOST_LOG(debug) << "Creating tap description for "sv << (int) channels << " channels"sv;
+  CATapDescription *tapDescription;
+  NSArray *excludeProcesses = @[];
+
+  if (channels == 1) {
```
andygrundman: I don't think we need to support mono, since the way initStereo handles mono is to duplicate it to left/right, which is what you want.

Speaking of this, I am assuming that stereo is the only supported configuration here. So we should probably be setting the tapDescription's stereo mixdown setting to true. If you've tested multichannel, how does it work? Here is the main use case I often use:

Moonlight 5.1 -> Sunshine playback of a 5.1 FLAC file, 5.1 mkv movies, and obviously games. What audio device has to be used to support 5.1 playback like this, without a virtual multichannel driver like Blackhole?

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,822 @@
...
+    tapDescription = [[CATapDescription alloc] initStereoGlobalTapButExcludeProcesses:excludeProcesses];
+  }
+
+  // Set unique name and UUID for this instance
+  NSString *uniqueName = [NSString stringWithFormat:@"SunshineAVAudio-Tap-%p", (void *) self];
+  NSUUID *uniqueUUID = [[NSUUID alloc] init];
+
+  tapDescription.name = uniqueName;
+  tapDescription.UUID = uniqueUUID;
+  [tapDescription setPrivate:YES];
```
andygrundman: We probably want to use `tapDescription.muteBehavior` to support Moonlight's "Mute host" option. If Mute Host is set, use CATapMuted. The WhenTapped option seems dangerous if it would suddenly unmute, so I wouldn't use that.

 @constant	CATapUnmuted
 	Audio is captured by the tap and also sent to the audio hardware
 @constant	CATapMuted
 	Audio is captured by the tap but no audio is sent from the process to the audio hardware
 @constant	CATapMutedWhenTapped
 	Audio is captured by the tap and also sent to the audio hardware until the tap is read by another audio client. 
 	For the duration of the read activity on the tap no audio is sent to the audio hardware.

docs/configuration.md
```diff
@@ -825,6 +825,31 @@ editing the `conf` file in a text editor. Use the examples as reference.
...
+        <td colspan="2">disabled</td>
+    </tr>
+    <tr>
+        <td>Example</td>
+        <td colspan="2">@code{}
+            macos_system_wide_audio_tap = disabled
+            @endcode</td>
+    </tr>
+</table>
+
```
andygrundman: At the top of this page there is the text "The name of the audio sink used for Audio Loopback. Sunshine can only access microphones on macOS due to system limitations. To stream system audio using Soundflower or BlackHole." We should remove that text.

tests/unit/platform/test_macos_av_audio.mm
```diff
@@ -0,0 +1,648 @@
...
+}
+
+/**
+ * @brief Test that findMicrophone handles nil input gracefully.
+ * Verifies the method returns nil when passed a nil microphone name.
+ */
+TEST_F(AVAudioTest, FindMicrophoneWithNilNameReturnsNil) {
+  #pragma clang diagnostic push
+  #pragma clang diagnostic ignored "-Wnonnull"
+  AVCaptureDevice *device = [AVAudio findMicrophone:nil];
```
andygrundman: Many of these tests that check for nil input are crashing, because the first line in findMicrophone is `BOOST_LOG(debug) << "Searching for microphone: "sv << [name UTF8String];` and you can't call UTF8String on nil. So there are probably a lot of additional checks for nil that need to go into those methods, or maybe the nil tests are a bit overkill.

andygrundman (commented):
Overall I think this is a well done patch, and is a good example of how to do an AI-assisted PR.

I originally was worried about the frequent use of manual release, e.g. 
```
[audioInput release];
[audioOutput release];
```

but then I learned that ObjC built by cmake doesn't use the newer [ARC memory management model](https://en.wikipedia.org/wiki/Automatic_Reference_Counting) that I was more familiar with from developing in Xcode. ARC dates from 2012 in OSX Lion if you can believe it, but since devs had to opt into using ARC, nobody ever did for this code. So, I guess that's a point for AI here. I would have probably not thought about this and been quite confused at all the memory leaks I was creating. Or more likely it would just result in a lot of compiler errors. (In -fobjc-arc mode, these kinds of release calls are compiler errors.)

docs/configuration.md
ThomVanL: I kept it as a setting to avoid breaking existing setups (users relying on BlackHole or similar). This way an update wouldn’t disrupt their workflow. That said, I'm happy to defer to you and the other maintainers on whether it makes sense to just default to the system tap.

Also, I’m not entirely sure `initStereoGlobalTapButExcludeProcesses` is the best long-term choice, since it enforces stereo and might complicate 5.1/7.1 setups? I don’t have a great way to test multichannel properly, so I’d appreciate your input here.

src/platform/macos/av_audio.mm
ThomVanL: Good point... My bad! Malloc definitely shouldn’t be in this callback. I went with the ObjC approach initially, but I agree that staying pure C/C++ here would be safer. I’ll adjust that.

src/platform/macos/av_audio.mm
ThomVanL: I’m fine with either direction, whether we keep backward compatibility or just require macOS 15+. I’ll follow whatever you and the team prefer.

src/platform/macos/av_audio.mm
ThomVanL: My bad, that malloc/free definitely shouldn’t be there in a real-time callback. I remember noticing the `__attribute__((nonblocking))` in the [CoreAudio docs](https://developer.apple.com/documentation/coreaudio/audiodeviceioproc?language=objc&utm_source=chatgpt.com) but forgot to account for it here. I’ll fix that.

docs/configuration.md
ReenigneArcher: Without looking at the code super closely, I believe it should just use the Tap API if audio/virtual sink are unset. If those are set then use whatever they are set to. With this approach their setup will not change as they had to manually set blackhole or whatever.

src/platform/macos/av_audio.mm
ReenigneArcher: I don't have a strong preference about supporting 14. We just recently dropped support for 13 a little bit earlier than Apple because of missing c++23 features.

If we decide to drop support for 14, we can set that in the homebrew formula here: https://github.com/LizardByte/Sunshine/blob/705d76372996372795b021501ad65a9835ebaec3/packaging/sunshine.rb#L40-L42 by adding `depends_on macos: :sequoia`

Also, we'll need to update the minimum version in the readme.

docs/configuration.md
ReenigneArcher: Also here:

- https://github.com/LizardByte/Sunshine/blob/705d76372996372795b021501ad65a9835ebaec3/packaging/sunshine.rb#L153-L154
- https://github.com/LizardByte/Sunshine/blob/705d76372996372795b021501ad65a9835ebaec3/docs/getting_started.md?plain=1#L376-L378
- https://github.com/LizardByte/Sunshine/blob/705d76372996372795b021501ad65a9835ebaec3/src_assets/common/assets/web/configs/tabs/AudioVideo.vue#L37-L40

Maybe we should keep a little blurb somewhere about how to use these two if they don't want to use the Tap API?

src/platform/macos/av_audio.mm
ThomVanL: I can remove the mono handling as you suggested.

In regards to multichannel, I should clarify this a little. I tested on my Android TV Moonlight client by toggling 5.1/7.1 and did get audio, but I’m starting to think it was probably just L+R playback (hard to tell for sure. 

In a previous iteration of the code, enabling 5.1 or 7.1 caused crackling on the client, while stereo worked fine. With the changes in this PR, all settings seem to play without issues. On the host side, I tested by launching a game or playing audio through YouTube, but I’m not sure how to verify that all channels are actually coming through correctly. I imagine properly testing multichannel would require additional hardware on the Android TV client.

Since the tap is set up with `initStereoGlobalTapButExcludeProcesses`, the host always mixes down to stereo. I’m not sure what would happen if we switched to `initExcludingProcesses` instead and tried it that way... As you suggested, we’d probably need a virtual multichannel driver to tap into that device’s full stream.

src/platform/macos/av_audio.mm
ThomVanL: Sure thing, I can hook up `tapDescription.muteBehavior` to support the “Mute host” option, as you suggested.

docs/configuration.md
ThomVanL: I'll remove that.

tests/unit/platform/test_macos_av_audio.mm
ThomVanL: Understood, I can fix that.

ThomVanL:
> Overall I think this is a well done patch, and is a good example of how to do an AI-assisted PR.
> 
> I originally was worried about the frequent use of manual release, e.g.
> 
> ```
> [audioInput release];
> [audioOutput release];
> ```
> 
> but then I learned that ObjC built by cmake doesn't use the newer [ARC memory management model](https://en.wikipedia.org/wiki/Automatic_Reference_Counting) that I was more familiar with from developing in Xcode. ARC dates from 2012 in OSX Lion if you can believe it, but since devs had to opt into using ARC, nobody ever did for this code. So, I guess that's a point for AI here. I would have probably not thought about this and been quite confused at all the memory leaks I was creating. Or more likely it would just result in a lot of compiler errors. (In -fobjc-arc mode, these kinds of release calls are compiler errors.)

Yeah, I wondered about ARC too and did a quick search for `-fobjc-arc` and `@autoreleasepool` usage. Everything pointed to ARC not being enabled, so I followed the existing pattern of manual release calls. 

Good to hear that lines up with your findings as well.

src/platform/common.h
```diff
@@ -554,7 +554,7 @@ namespace platf {
   public:
     virtual int set_sink(const std::string &sink) = 0;
 
-    virtual std::unique_ptr<mic_t> microphone(const std::uint8_t *mapping, int channels, std::uint32_t sample_rate, std::uint32_t frame_size, bool host_audio_enabled) = 0;
+    virtual std::unique_ptr<mic_t> microphone(const std::uint8_t *mapping, int channels, std::uint32_t sample_rate, std::uint32_t frame_size, [[maybe_unused]] bool host_audio_enabled) = 0;
```
ThomVanL: Not a perfect approach, but it should work for now.

ThomVanL:
Hi @ReenigneArcher and @andygrundman, happy new year! 🎉

Also, my apologies for the delay in regards to this PR. I've removed some of the unit tests that turned out to be integration tests in disguise (particularly the TCC ones). After poking around for best practices on running integration tests in a CI/CD pipeline with TCC, I mostly found workarounds that felt a little hacky, so I decided to strip those integration tests out and stick with pure unit tests instead. Not sure if that sits right with you both, but feel free to let me know if it needs to be changed.

andygrundman:
Thanks Thom, I will take a look at this very soon.

andygrundman:
This is looking pretty good, thanks! What I did was create a branch from yours, cherry-pick in some helper things (my build script, input fixes, my previous compile fix (how does this build for you in Tahoe?), and moved a bunch of noisy logging from debug to verbose). With all that out of the way, the main commit in my branch has a bunch of improvements for you to consider. Here's the items in that commit:

https://github.com/andygrundman/Sunshine/commits/andyg.feat-macos-ca-taps-review/

    * Refactor microphone.mm sample() to memcpy the buffer directly instead of copying twice by using std::vector, and optimize the case where avail < neededBytes.
    * Move initializeAudioBuffer before anything can use the buffer.
    * Change size of ring buffer to 30ms (6 * 240 * channels * sizeof(float)).
    * Avoid confusion in converter setup between float and int samplerate.
    * Add helper for logging CoreAudio error codes in FourCC text mode. Remove some redundant logging, and always log the exact API call that is failing.
    * Add optional way to create the aggregate in public mode so it's visible in Audio MIDI Setup/HALLab. Set SUNSHINE_PUBLIC_AUDIO_TAP=1.

I'm not married to these changes, so feel free to counter any of them. I'm not sure what the right buffer size should be, but I felt like 4096 * channels seemed small and just as arbitrary. The public tap thing is neat to look at, especially check out the Apple utility HALLab which exposes all the internals of audio devices. I worry that public taps may not get cleaned up in some situations such as a crash, although maybe private taps get automatically cleaned up. Using an env var for devs is a bit ugly but there's no other way to view the aggregate device.

Re: resampling: do we need to have so much or any resampling code? Doesn't CoreAudio handle a lot of this behind the scenes? I couldn't get the system tap to ever need resampling (I think because it's not using a specific device), and when using BlackHole 16ch sink directly, the resampling code doesn't seem to be called anyway. Even if I run BlackHole at 192k or something, it seems to work fine. I did a trace in Activity Monitor and you can find the bit where the OS is resampling from 192k -> 48k if you search for `acv2::Resampler2Wrapper::ProduceOutput`. I just need some help on how to test your resampling code, since I couldn't get it to use the code path with procData->audioConverter.

This all works quite well for stereo. For surround, I guess we need to think about it some more. I just stumbled across a project called SoundPusher that creates a virtual 5.1 device and does on-demand AC3 encoding. I want to study it some more. https://codeberg.org/q-p/SoundPusher I found it via https://www.maven.de/2025/04/coreaudio-taps-for-dummies/ by the way.

Edit: I think I just realized the conversion code is mostly there for multichannel right? I honestly have not even tested surround because I think that still requires a virtual audio device like what SoundPusher includes, or BlackHole. I'm not sure it works with Tap. This PR should probably just focus on making stereo taps work in the most user-friendly way possible.

andygrundman:
Another edit: I had a few moments of panic while testing this, you know the feeling: the audio is skipping and glitching out real bad, and your first thought is there's some bug in the realtime buffer code. Nope, it's all just AWDL. This code basically becomes unusable on wifi devices that don't have N1 chips. The iPad I have been using as a client still had 6ghz 6E mode enabled which is a death sentence for Moonlight. My N1-equipped iPhone Air streams it perfectly. Thanks Apple.

cmake/compile_definitions/macos.cmake
```diff
@@ -24,13 +24,16 @@ list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
         ${CORE_MEDIA_LIBRARY}
         ${CORE_VIDEO_LIBRARY}
         ${FOUNDATION_LIBRARY}
+        ${AUDIO_TOOLBOX_LIBRARY}
+        ${AUDIO_UNIT_LIBRARY}
+        ${CORE_AUDIO_LIBRARY}
         ${VIDEO_TOOLBOX_LIBRARY})
```
ReenigneArcher, line 38: Could this list be alphabetized?

cmake/dependencies/common.cmake
```diff
@@ -18,6 +18,9 @@ add_subdirectory("${CMAKE_SOURCE_DIR}/third-party/libdisplaydevice")
 # common dependencies
 include("${CMAKE_MODULE_PATH}/dependencies/nlohmann_json.cmake")
 find_package(OpenSSL REQUIRED)
+if(OPENSSL_FOUND)
+    include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
+endif()
```
ReenigneArcher: This shouldn't be necessary?

https://github.com/LizardByte/Sunshine/blob/66cdaf9d14c890e165b10a1f94be8452afd67b76/cmake/compile_definitions/common.cmake#L158

cmake/dependencies/macos.cmake
```diff
@@ -11,3 +11,11 @@ FIND_LIBRARY(VIDEO_TOOLBOX_LIBRARY VideoToolbox)
...
     FIND_LIBRARY(COCOA Cocoa REQUIRED)
 endif()
+
+# Audio frameworks required for audio capture/processing
+FIND_LIBRARY(AUDIO_TOOLBOX_LIBRARY AudioToolbox)
+FIND_LIBRARY(AUDIO_UNIT_LIBRARY AudioUnit)
+FIND_LIBRARY(CORE_AUDIO_LIBRARY CoreAudio)
+
+include_directories(/opt/homebrew/opt/opus/include)
+link_directories(/opt/homebrew/opt/opus/lib)
```
ReenigneArcher: These paths depend on the architecture of homebrew, and can't be hardcoded.

docs/getting_started.md
```diff
@@ -455,9 +455,10 @@ systemctl --user enable sunshine
 ### macOS
 The first time you start Sunshine, you will be asked to grant access to screen recording and your microphone.
 
-Sunshine can only access microphones on macOS due to system limitations. To stream system audio use
-[Soundflower](https://github.com/mattingalls/Soundflower) or
-[BlackHole](https://github.com/ExistentialAudio/BlackHole).
+Sunshine supports native system audio capture on macOS 14.0 (Sonoma) and newer via Apple’s Audio Tap API.
+To use it, simply leave the **Audio Sink** setting blank.
+
+If you are running macOS 13 (Ventura) or earlier—or if you prefer to manage your own loopback device—you can still use [Soundflower](https://github.com/mattingalls/Soundflower) or [BlackHole](https://github.com/ExistentialAudio/BlackHole) and enter its device name in the **Audio Sink** field.
```
ReenigneArcher:
```suggestion
If you prefer to manage your own loopback device—you can still use [Soundflower](https://github.com/mattingalls/Soundflower) or [BlackHole](https://github.com/ExistentialAudio/BlackHole) and enter its device name in the **Audio Sink** field.
```

macos-13 isn't supported anymore

packaging/sunshine.rb
```diff
@@ -194,8 +194,11 @@ def post_install
...
       opoo <<~EOS
-        Sunshine can only access microphones on macOS due to system limitations.
-        To stream system audio use "Soundflower" or "BlackHole".
+        Sunshine now supports system audio capture natively on macOS 14.0 (Sonoma) and later,
+        using the built-in Core Audio Tap API.
+
+        On macOS 13 or earlier, or if you prefer a virtual loopback device,
+        you can still use "Soundflower" or "BlackHole" for system audio capture.
 
         Gamepads are not currently supported on macOS.
```
ReenigneArcher:
```suggestion
        Gamepads are not currently supported on macOS.
```

Since we don't need to worry about macos-13 now, and audio will work out of the box... we can remove the comment about audio altogether.

src_assets/macos/assets/Info.plist
```diff
@@ -8,5 +8,8 @@
   <string>Sunshine</string>
   <key>NSMicrophoneUsageDescription</key>
   <string>This app requires access to your microphone to stream audio.</string>
```
ReenigneArcher: Maybe this comment should be updated?

.gitignore
```diff
@@ -56,3 +56,6 @@ package-lock.json
 # Python
 *.pyc
 venv/
+
+# Caches
+.cache/
```
ReenigneArcher: What, specifically, is generating this directory?

tests/unit/platform/test_macos_av_audio.mm
```diff
@@ -0,0 +1,388 @@
+/**
+ * @file tests/unit/platform/test_macos_av_audio.mm
```
ReenigneArcher:
```suggestion
 * @file tests/unit/platform/macos/test_av_audio.mm
```

Can you rename the file to be in a macos specific folder. We also have a folder for windows there now as well.

tests/unit/platform/test_macos_av_audio.mm
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/**
+ * @brief Test basic AVAudio object lifecycle.
+ * Verifies that AVAudio objects can be created and destroyed without issues.
+ */
+TEST_F(AVAudioTest, ObjectLifecycle) {
+  AVAudio *avAudio = [[AVAudio alloc] init];
+  EXPECT_NE(avAudio, nil);  // Should create successfully
+  [avAudio release];  // Should not crash
```
ReenigneArcher: If this can crash, I think it should be wrapped in a try/catch. If we hit the catch, then it should trigger GTEST to fail.

Same for anywhere else it could crash. This is because when a test crashes it crashes the whole test run.

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -126,7 +126,7 @@
     "apply_note": "Click 'Apply' to restart Sunshine and apply changes. This will terminate any running sessions.",
     "audio_sink": "Audio Sink",
     "audio_sink_desc_linux": "The name of the audio sink used for Audio Loopback. If you do not specify this variable, pulseaudio will select the default monitor device. You can find the name of the audio sink using either command:",
-    "audio_sink_desc_macos": "The name of the audio sink used for Audio Loopback. Sunshine can only access microphones on macOS due to system limitations. To stream system audio using Soundflower or BlackHole.",
+    "audio_sink_desc_macos": "The name of the audio sink used for Audio Loopback. Leave this blank to use the built-in system audio capture (requires macOS 14.0 or later, using the Core Audio Tap API). Alternatively, specify a virtual device such as Soundflower or BlackHole if you prefer or are running an older version of macOS.",
```
ReenigneArcher:
```suggestion
    "audio_sink_desc_macos": "The name of the audio sink used for Audio Loopback. Leave this blank to use the built-in system audio capture. Alternatively, specify a virtual device such as Soundflower or BlackHole if you prefer or are running an older version of macOS.",
```

ReenigneArcher (changes requested):

.gitignore
ThomVanL: This was due to one of the VS Code extensions I have installed, mainly [clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd). I removed this entry and now keep it in a global .gitignore file in my home directory!

tests/unit/platform/test_macos_av_audio.mm
ThomVanL: Done!

tests/unit/platform/test_macos_av_audio.mm
ThomVanL: I've added some try/catch blocks and combined it with gtest's [`FAIL()`](https://google.github.io/googletest/reference/assertions.html#FAIL)  but I have a feeling we can fine-tune the messages a bit more.

src_assets/common/assets/web/public/assets/locale/en.json
ThomVanL: Done!

src_assets/macos/assets/Info.plist
ThomVanL: I've updated it with something similar to what we use in `en.json`.

packaging/sunshine.rb
ThomVanL: Done, I've removed that bit of info.

docs/getting_started.md
ThomVanL: Done, removed it!

cmake/dependencies/macos.cmake
ThomVanL: I've removed it and managed to sort out my dev environment's quirks. :)

cmake/dependencies/common.cmake
ThomVanL: Done, I've removed this.

cmake/compile_definitions/macos.cmake
ThomVanL: Done, alphabetized!

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,846 @@
...
+  BOOST_LOG(debug) << "Retrieving microphone names"sv;
+  NSMutableArray *result = [[NSMutableArray alloc] init];
+
+  for (AVCaptureDevice *device in [AVAudio microphones]) {
+    [result addObject:[device localizedName]];
+  }
+
+  BOOST_LOG(info) << "Found "sv << [result count] << " microphones"sv;
+  return result;
+}
```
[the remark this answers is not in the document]
ThomVanL: Swapped the result to `NSMutableArray *result = [NSMutableArray array]`.

src/platform/macos/av_audio.mm
```diff
@@ -0,0 +1,846 @@
...
+        BOOST_LOG(warning) << "Failed to get stream configuration, using default 2 channels: "sv << streamConfigStatus;
+      }
+      free(streamConfig);
+    }
+  } else {
+    BOOST_LOG(warning) << "Failed to get stream configuration size, using default 2 channels: "sv << streamConfigSizeStatus;
+  }
+
+  BOOST_LOG(debug) << "Device properties - Sample Rate: "sv << aggregateDeviceSampleRate << "Hz, Channels: "sv << aggregateDeviceChannels;
+
```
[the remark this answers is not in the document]
ThomVanL: Made some changes to this code as part of [commit 3831795](https://github.com/LizardByte/Sunshine/pull/4209/commits/3831795c350284770e8d4b13c149bd863aa5b39d).

ReenigneArcher:
@ThomVanL apologies, but we just merged a PR that caused a few conflicts in this one. Could you resolve the conflicts and rebase? We're looking to get your PR merged in relatively soon.

docs/getting_started.md
```diff
@@ -446,9 +446,10 @@ systemctl --user --now enable sunshine
 ### macOS
 The first time you start Sunshine, you will be asked to grant access to screen recording and your microphone.
 
-Sunshine can only access microphones on macOS due to system limitations. To stream system audio use
-[Soundflower](https://github.com/mattingalls/Soundflower) or
-[BlackHole](https://github.com/ExistentialAudio/BlackHole).
+Sunshine supports native system audio capture on macOS 14.0 (Sonoma) and newer via Apple’s Audio Tap API.
+To use it, simply leave the **Audio Sink** setting blank.
+
+If you prefer to manage your own loopback device, you can still use [Soundflower](https://github.com/mattingalls/Soundflower) or [BlackHole](https://github.com/ExistentialAudio/BlackHole) and enter its device name in the **Audio Sink** field.
```
ReenigneArcher:
```suggestion
If you prefer to manage your own loopback device, you can still use
[Soundflower](https://github.com/mattingalls/Soundflower) or
[BlackHole](https://github.com/ExistentialAudio/BlackHole)
and enter its device name in the [audio_sink](configuration.md#audio_sink) field.
```

Let's keep line lengths under 120 characters where possible. We can also link directly to the config option.

src/platform/macos/coreaudio_helpers.h
```diff
@@ -0,0 +1,68 @@
...
+  // kAudioHardwareUnknownPropertyError      = 'who?',
+  // kAudioHardwareBadPropertySizeError      = '!siz',
+  // kAudioHardwareIllegalOperationError     = 'nope',
+  // kAudioHardwareBadObjectError            = '!obj',
+  // kAudioHardwareBadDeviceError            = '!dev',
+  // kAudioHardwareBadStreamError            = '!str',
+  // kAudioHardwareUnsupportedOperationError = 'unop',
+  // kAudioHardwareNotReadyError             = 'nrdy',
+  // kAudioDeviceUnsupportedFormatError      = '!dat',
+  // kAudioDevicePermissionsError            = '!hog'
```
ReenigneArcher:
```suggestion
  // Some CoreAudio error examples:
  // kAudioHardwareNoError = 0,
  // kAudioHardwareNotRunningError = 'stop',
  // kAudioHardwareUnspecifiedError = 'what',
  // kAudioHardwareUnknownPropertyError = 'who?',
  // kAudioHardwareBadPropertySizeError = '!siz',
  // kAudioHardwareIllegalOperationError = 'nope',
  // kAudioHardwareBadObjectError = '!obj',
  // kAudioHardwareBadDeviceError = '!dev',
  // kAudioHardwareBadStreamError = '!str',
  // kAudioHardwareUnsupportedOperationError = 'unop',
  // kAudioHardwareNotReadyError = 'nrdy',
  // kAudioDeviceUnsupportedFormatError = '!dat',
  // kAudioDevicePermissionsError = '!hog'
```

Could you remove the alignmnet? Or alternatively remove this altogether and list a URL where the information can be found?

cmake/compile_definitions/unix.cmake
```diff
@@ -5,6 +5,7 @@ list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
         ${CURL_LIBRARIES})
 
 # add install prefix to assets path if not already there
-if(NOT SUNSHINE_ASSETS_DIR MATCHES "^${CMAKE_INSTALL_PREFIX}")
+# Skip prefix addition for absolute paths or development builds
```
ReenigneArcher: I think this may not be necessary after a similar change was made here: https://github.com/LizardByte/Sunshine/pull/4759/changes#diff-4b6cfad26a99f30b4ac47ede71939639d83e78c0617ac79cfebd22b0a28203cfR7

ReenigneArcher (commented):
Also, what was decided on dropping support macOS-14 or not? https://github.com/LizardByte/Sunshine/pull/4209#discussion_r2312632748

cmake/compile_definitions/unix.cmake
ThomVanL: Done!

src/platform/macos/coreaudio_helpers.h
ThomVanL: I've removed the alignment. I considered adding an Apple documentation URL, but the pages were sparse and not very useful, so I left them out!

ThomVanL:
> @ThomVanL apologies, but we just merged a PR that caused a few conflicts in this one. Could you resolve the conflicts and rebase? We're looking to get your PR merged in relatively soon.

This should be OK now, I think. Note that I updated the `Info.plist.in` template with the values that were previously stored in the old `Info.plist`. I also removed the unused audio constant `kBufferLength` since it no longer appeared to be referenced anywhere.

> Also, what was deciding on dropping support macOS-14 or not? [#4209 (comment)](https://github.com/LizardByte/Sunshine/pull/4209#discussion_r2312632748)

I don’t think a final decision was made on that. I also wasn’t able to test these changes on macOS 14, so I’ll leave that call up to you.

ThomVanL:
> Another edit: I had a few moments of panic while testing this, you know the feeling: the audio is skipping and glitching out real bad, and your first thought is there's some bug in the realtime buffer code. Nope, it's all just AWDL. This code basically becomes unusable on wifi devices that don't have N1 chips. The iPad I have been using as a client still had 6ghz 6E mode enabled which is a death sentence for Moonlight. My N1-equipped iPhone Air streams it perfectly. Thanks Apple.

I had a similar moment of panic while testing. 😅 On my iPhone 15 Pro the audio skips every few seconds, but it works fine on my iPad (2022, 10th gen)! Streaming from my MacBook to our Android TV also works without any issues.

andygrundman (commented):
I'm good with these changes, it's working well in my testing. Good work!

ReenigneArcher:
@ThomVanL this seems good to me as well, though I cannot test as I do not own any Apple computer.

A few things to resolve before this gets merged:

1. Could you fix the lint issues https://github.com/LizardByte/Sunshine/actions/runs/22737285736/job/66112327529?pr=4209 and rebase the PR?
2. It seems the macOS build is failing due to a deployment target mismatch. https://github.com/LizardByte/Sunshine/actions/runs/22737285707/job/66112352267?pr=4209#step:6:3890
  - perhaps we need to set the deployment target to 14.2 in ci-macos.yml, like we do here? https://github.com/LizardByte/build-deps/blob/cb5f0de130d62600d24201e59b6d740d23dcb0d8/.github/workflows/ci.yml#L18-L21 ... if we do update the target to 14.2, we should also update the readme here: https://github.com/LizardByte/Sunshine/blob/c4c10a16b889d9a9fd977ff66059d099ebaf5c77/README.md?plain=1#L245

The Arch and Flatpak failures are not due to your PR and should resolve with a rebase,

README.md
```diff
@@ -242,7 +242,7 @@ LizardByte has the full documentation hosted on [Read the Docs](https://docs.liz
         <td>Linux/Ubuntu: 22.04+ (jammy)</td>
     </tr>
     <tr>
-        <td>macOS: 14+</td>
+        <td>macOS: 14.2+</td>
```
ReenigneArcher, line 245: Also need to add the env variable to the ci-macos.yml workflow, like here: https://github.com/LizardByte/build-deps/blob/cb5f0de130d62600d24201e59b6d740d23dcb0d8/.github/workflows/ci.yml#L18-L21

ReenigneArcher:
It appears that tests are crashing on macOS 15 and 26, at these tests

- 15 - `Configurations/AudioTest.TestEncode/HIGH_STEREO`
- 26 - `Configurations/AudioTest.TestEncode/SURROUND51`

Also, still one lint issue: https://github.com/LizardByte/Sunshine/actions/runs/22879151110/job/66380516025?pr=4209#step:10:57

ThomVanL:
> It appears that tests are crashing on macOS 15 and 26, at these tests
> 
> * 15 - `Configurations/AudioTest.TestEncode/HIGH_STEREO`
> * 26 - `Configurations/AudioTest.TestEncode/SURROUND51`
> 

It looks like we might still need your TCC fix after all… but I will run a few tests first!

ReenigneArcher:
> It looks like we might still need your TCC fix after all… but I will run a few tests first!

It's somewhat odd because the macOS-26 completes some of the audio tests before crashing out.

ThomVanL:
Hey @ReenigneArcher I've had some time to really dive into this a bit more and it's, like you mentioned, worth noting that the TCC prompting is _not happening_ during the Homebrew validation on the master branch. I thought that was a little strange, as well.

You can see that I've pushed a small (preliminary) change that I'd love to see run through the workflow. Your feedback is very much welcome.

I had a closer look at the CI workflow logs (the Homebrew macOS validation) and I noticed other parts of the code also interact with TCC but appear to skip cleanly:

On [macOS 14](https://github.com/LizardByte/Sunshine/actions/runs/23081980223/job/67052734246#step:8:5431):

```
  [----------] 2 tests from EncoderVariants/EncoderTest
  [ RUN      ] EncoderVariants/EncoderTest.ValidateEncoder/videotoolbox
  IOServiceMatchingfailed for: AppleM2ScalerCSCDriver
  /private/tmp/sunshine-20260314-59779-wfzgqc/tests/unit/test_video.cpp:18: Skipped
  Encoder not available
  [  SKIPPED ] EncoderVariants/EncoderTest.ValidateEncoder/videotoolbox (1055 ms)
```

[macOS 15](https://github.com/LizardByte/Sunshine/actions/runs/23081980223/job/67052734252#step:8:5392)
```
  [----------] 2 tests from EncoderVariants/EncoderTest
  [ RUN      ] EncoderVariants/EncoderTest.ValidateEncoder/videotoolbox
  IOServiceMatchingfailed for: AppleM2ScalerParavirtDriver
  /private/tmp/sunshine-20260314-30146-nfg1wh/tests/unit/test_video.cpp:18: Skipped
  Encoder not available
  
  [  SKIPPED ] EncoderVariants/EncoderTest.ValidateEncoder/videotoolbox (2287 ms)
  [ RUN      ] EncoderVariants/EncoderTest.ValidateEncoder/software
  [       OK ] EncoderVariants/EncoderTest.ValidateEncoder/software (86 ms)
  [----------] 2 tests from EncoderVariants/EncoderTest (2374 ms total)
```

[macOS 26](https://github.com/LizardByte/Sunshine/actions/runs/23081980223/job/67052734248)

```
  [----------] 2 tests from EncoderVariants/EncoderTest
  [ RUN      ] EncoderVariants/EncoderTest.ValidateEncoder/videotoolbox
  IOServiceMatchingfailed for: AppleM2ScalerParavirtDriver
  /private/tmp/sunshine-20260314-71238-cegf4y/tests/unit/test_video.cpp:18: Skipped
  Encoder not available
  
  [  SKIPPED ] EncoderVariants/EncoderTest.ValidateEncoder/videotoolbox (1954 ms)
  [ RUN      ] EncoderVariants/EncoderTest.ValidateEncoder/software
  [       OK ] EncoderVariants/EncoderTest.ValidateEncoder/software (88 ms)
  [----------] 2 tests from EncoderVariants/EncoderTest (2043 ms total)
```

So, I went and implemented something in `test_audio.cpp` similar to what happens in `test_video.cpp`, where encoder availability is checked before running.

But about those audio tests... I think the reason these pass on the _master_ branch is a bit subtle. When no audio sink is set in `AudioTest`, `config::audio.sink` is empty, so the local `audio_sink` variable inside `microphone()` stays at its default of `""`, which gets passed to `findMicrophone("")`. That function iterates through all available `AVCaptureDevice` instances and compares each device's `localizedName` to `""`. Since no device matches, it returns `nil`. That in turn propagates back up as `nullptr` from `microphone()` and the null check in `audio.cpp` returns immediately without starting capture. The test's timer fires after 100ms, stops the queue, and the capture thread exits cleanly. Since no packets ever arrive, the `FAIL()` on empty packet data is never hit, and the test "passes". In other words, that test can pass even when audio capture never initialized at all — it isn't actually verifying that audio is working.

With the CoreAudio tap implementation, however, not setting a sink in `AudioTest` now causes execution to reach the CoreAudio tap creation flow. I don't think I can insert a guard in `microphone()` to fail gracefully, as creating the tap itself triggers the TCC prompt and blocks.

As an alternative, we could inject permissions directly into the TCC database, but that feels a little brittle and error-prone.

So the question is: should we add a `GTEST_SKIP()` guard (which would skip the test on machines without the permission rather than actually verifying capture works) or should we go ahead with TCC injection in CI? I'm leaning toward the skip at the moment, but wanted to get input before going any further.

(I'll sort out rebasing my commits, too.)

ReenigneArcher (dismissed):

ReenigneArcher:
I also noticed the audio test wasn't really the greatest test. We could probably improve it in a later PR. For now, I think I'm good with this PR as is. Would you want to submit a follow up PR to help improve the tests?

My preference would be to actually verify audio works, but I understand sometimes this isn't the easiest to do in a CI environment where we don't have control of the hardware. For the encoder tests, I just make them skip if they don't work, unless software encoding fails in which case it is a failure. Software encoding should always work, whether in a CI environment or not.

ReenigneArcher:
Looks like the audio tests are still failing on macos 15 and 26, so maybe it needs a better fix before this is merged.

ThomVanL:
> I also noticed the audio test wasn't really the greatest test. We could probably improve it in a later PR. For now, I think I'm good with this PR as is. Would you want to submit a follow up PR to help improve the tests?

Sure thing, I don't mind taking a look.

> Looks like the audio tests are still failing on macos 15 and 26, so maybe it needs a better fix before this is merged.

That's unfortunate... Though I don't mind digging into it some more!

ReenigneArcher:
@ThomVanL so you think it's a permissions issue?

- macOS-14:  are all skipped now
- macOS-15: `HIGH_STEREO` crashes
- macOS-26: `HIGH_STEREO` is skipped, and the next one crashes

Would adding a virtual audio device allow it to pass? Looks like blackhole is available in homebrew with 3 variants. We could make these dependencies in the homebrew package?

1. https://formulae.brew.sh/cask/blackhole-2ch#default
2. https://formulae.brew.sh/cask/blackhole-16ch#default
3. https://formulae.brew.sh/cask/blackhole-64ch#default

ThomVanL:
@ReenigneArcher Yes, I think it's got to do with the kTCCServiceAudioCapture permission. I have been also considering adding a sink, so it's definitely something we could try.

I have been conducting quite a bit of testing on a separate repository, with just the code path that creates the tap and aggregate device, gtest and a workflow; least with enough code to prompt the TCC. But so far I've been unable to add those permissions to the TCC database. 

It also looks like the permission is not [configured](https://github.com/actions/runner-images/blob/main/images/macos/scripts/build/configure-tccdb-macos.sh) in GitHub's macOS runners, based on what I can tell. 

I will do some more testing this evening.

ReenigneArcher:
@ThomVanL the script you linked is where I originally discovered how to adjust the permissions in the runner as well. You can add anything you want in the same way they do because SIP is disabled in the runners. The hard part, when I originally did it, was figuring out what to give the permission to... but that was with macports. I think with homebrew it should be easier as it's supported on the runners out of the box.

ThomVanL:
@ReenigneArcher I did some more testing on this across a few different approaches, but I wasn’t able to get the TCC settings to apply correctly. I tried configuring it in multiple ways, but no success so far.

As a fallback, I replicated the current behavior by forcing the sink to a non-existent value:

(I have not yet committed this, by the way)

```cpp
struct AudioTest: PlatformTestSuite, testing::WithParamInterface<std::tuple<std::basic_string_view<char>, config_t>> {
  void SetUp() override {
    m_config = std::get<1>(GetParam());
    m_mail = std::make_shared<safe::mail_raw_t>();

#ifdef __APPLE__
    m_saved_sink = config::audio.sink;
    if (config::audio.sink.empty()) {
      config::audio.sink = "__nonexistent_audio_sink__";
      m_overrode_sink = true;
    }
#endif
  }

  void TearDown() override {
#ifdef __APPLE__
    if (m_overrode_sink) {
      config::audio.sink = m_saved_sink;
    }
#endif
  }

  config_t m_config;
  safe::mail_t m_mail;
#ifdef __APPLE__
  std::string m_saved_sink;
  bool m_overrode_sink {};
#endif
};
``` 

That said, I’m not particularly a big fan of this approach. I find it to be more along the lines of a quick fix rather than a proper solution.. Though as far as I can tell it mirrors the current behavior in master. 

I’ll keep digging into the TCC side tomorrow evening to see if there’s a way to get the permissions set correctly in the runner.

ThomVanL:
@ReenigneArcher Even better, I've been able to add the correct TCC permissions, and in my testing I've been able to get OK results from all four `Configurations/AudioTest` tests. I've added the script as a conditional step in the `ci-homebrew.yaml` and rolled back the changes to `test_audio.cpp`. 

I ended up just expanding upon that script from the "actions/runner-images"  I linked earlier. It didn't occur to me to add `kTCCServiceAudioCapture` to those recurring clients, my bad.

ReenigneArcher:
@ThomVanL nice work! Looks like everything is green!

PR #4211, opened 2025-08-30
fix(input/ui): Add missing option for ds5_inputtino_randomize_mac in config.html

Kishi85:
## Description

https://github.com/LizardByte/Sunshine/pull/4208#pullrequestreview-3170332165 made me realize that the change to config.html is missing for ```ds5_inputtino_randomize_mac``` as well. So this is adding it since I've also missed it during #4158 

This will hopefully be the last PR for this option. Sorry for the inconvenience.

## Type of Change

- [ ] **feat**: New feature (non-breaking change which adds functionality)
- [X] **fix**: Bug fix (non-breaking change which fixes an issue)
- [ ] **docs**: Documentation only changes
- [ ] **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semicolons, etc.)
- [ ] **refactor**: Code change that neither fixes a bug nor adds a feature
- [ ] **perf**: Code change that improves performance
- [ ] **test**: Adding missing tests or correcting existing tests
- [ ] **build**: Changes that affect the build system or external dependencies
- [ ] **ci**: Changes to CI configuration files and scripts
- [ ] **chore**: Other changes that don't modify src or test files
- [ ] **revert**: Reverts a previous commit
- [ ] **BREAKING CHANGE**: Introduces a breaking change (can be combined with any type above)

## Checklist

- [X] Code follows the style guidelines of this project
- [X] Code has been self-reviewed
- [X] Code has been commented, particularly in hard-to-understand areas
- [ ] Code docstring/documentation-blocks for new or existing methods/components have been added or updated
- [ ] Unit tests have been added or updated for any new or modified functionality

### AI Usage

- [X] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [ ] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

Issue #4127, opened 2025-08-04
Stream Audio option display error in config UI


jhj0411jhj:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

I'm too lazy to test

### Describe the Bug

​​Latest Sunshine Release:​​ v2025.628.4510
After a clean installation, navigate to the web UI → Configuration → Audio/Video. The "Stream Audio" option shows as unchecked by default, despite its functional default being true. No relevant entry exists in `sunshine.conf` for this setting. Notably, audio streams normally in this seemingly contradictory state.

### Expected Behavior

The "Stream Audio" option checkbox shows as checked by default

### Host Operating System

Windows

### Operating System Version

win10

### Architecture

amd64/x86_64

### Sunshine commit or version

v2025.628.4510

### Package

Windows - installer (recommended)

### GPU Type

NVIDIA

### GPU Model

None

### GPU Driver/Mesa Version

None

### Capture Method

None

### Config

```Shell
empty
```

### Apps

```json

```

### Relevant log output

```shell
None
```

Kishi85:
Since I've been working on a new config option I've had a look at this as well and I don't see anything missing apart from the option setting in config.html. So maybe thats the culprit? If it's not then i'm not sure if this option is working as intended with it's current implementation.

ReenigneArcher:
Could you test out https://github.com/LizardByte/Sunshine/pull/4215

I've implemented a test in this PR to check for issues with config options and it detected `stream_audio` as having an issue initially.

"build" artifacts can be downloaded from here: https://github.com/LizardByte/Sunshine/actions/runs/17350064845?pr=4215#artifacts

jhj0411jhj:
> Could you test out [#4215](https://github.com/LizardByte/Sunshine/pull/4215)
> 
> I've implemented a test in this PR to check for issues with config options and it detected `stream_audio` as having an issue initially.
> 
> "build" artifacts can be downloaded from here: https://github.com/LizardByte/Sunshine/actions/runs/17350064845?pr=4215#artifacts

@ReenigneArcher I've tried build-Windows-AMD64 portable, and everything works fine. Thank you.

PR #4158, opened 2025-08-14
feat(input/linux): allow ds5 gamepads to have a fixed device mac based on controller index

Kishi85:
## Description
This PR adds an option to change how virtual DS5 controllers are created that allows them to retain controller settings (when stored based on the virtual devices MAC-Address, like Steam does) by disabling the random MAC generation and manually specifiying a private MAC-based on the connected gamepads global index. The behaviour then should be equal to the way XBox- and Switchcontrollers and DS4 on Windows are handled currently.

### Issues Fixed or Closed

- Resolves #3404

## Type of Change

- [X] **feat**: New feature (non-breaking change which adds functionality)
- [ ] **fix**: Bug fix (non-breaking change which fixes an issue)
- [ ] **docs**: Documentation only changes
- [ ] **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semicolons, etc.)
- [ ] **refactor**: Code change that neither fixes a bug nor adds a feature
- [ ] **perf**: Code change that improves performance
- [ ] **test**: Adding missing tests or correcting existing tests
- [ ] **build**: Changes that affect the build system or external dependencies
- [ ] **ci**: Changes to CI configuration files and scripts
- [ ] **chore**: Other changes that don't modify src or test files
- [ ] **revert**: Reverts a previous commit
- [ ] **BREAKING CHANGE**: Introduces a breaking change (can be combined with any type above)

## Checklist

- [X] Code follows the style guidelines of this project
- [X] Code has been self-reviewed
- [X] Code has been commented, particularly in hard-to-understand areas
- [X] Code docstring/documentation-blocks for new or existing methods/components have been added or updated
- [ ] Unit tests have been added or updated for any new or modified functionality

### AI Usage

- [X] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [ ] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

PR #4215, opened 2025-08-30
test(config): add test for consistent config

ReenigneArcher:
## Description

This PR was inspired after (during a review) I missed a few spots where config options need to be defined when adding new options.

See:
- https://github.com/LizardByte/Sunshine/pull/4158
- https://github.com/LizardByte/Sunshine/pull/4207
- https://github.com/LizardByte/Sunshine/pull/4211
- https://github.com/LizardByte/Sunshine/pull/4208#pullrequestreview-3170332165

I did use copilot to help generate the test, and due to frustrations with agent mode not being able to understand how to work with the ucrt64 environment, I have also added a copilot-instructions.md file.

### Issues Fixed or Closed

- Fixes https://github.com/LizardByte/Sunshine/issues/4127
- Closes https://github.com/LizardByte/Sunshine/pull/4211

## Type of Change

- [x] **test**: Adding missing tests or correcting existing tests

## Checklist

### AI Usage


Kishi85:
From what I'm seeing this is not checking the C++ defaults yet (as fixed in #4207). 

This might actually be a bit tricky to do with how they are currently initialized as those do not reference back to the config value names (and even the comments are inconsistent) so maybe a check if there are enough values would be a start? To properly parse them it might be necessary to define C++ defaults in another way where the variable name is used explicitly so it can be checked.

Kishi85:
I'm also unsure if this catches differing config value names for UI and in C++ (e.g. ```stream_audio``` in UI currently maps to ```audio.stream``` and not ```audio.stream_audio```) but IMHO those cases should be changed/fix to be uniform throughout the code as it makes it more readable as a whole.

Kishi85:
> This might actually be a bit tricky to do with how they are currently initialized as those do not reference back to the config value names (and even the comments are inconsistent) so maybe a check if there are enough values would be a start? To properly parse them it might be necessary to define C++ defaults in another way where the variable name is used explicitly so it can be checked.

It actually might be enough to make the comments uniform (like ```// <variable_name>[:<optional comment>]```) and add a secondary C++ parser that checks for that pattern (```// <variable_name>``` ignoring everything else after including the colon), create a list like the first parser for C++ does and add that second resultlist to the final comparison.

ReenigneArcher:
> From what I'm seeing this is not checking the C++ defaults yet (as fixed in https://github.com/LizardByte/Sunshine/pull/4207).

Correct. That would be pretty tricky to parse out. Probably defining the config as a map/dictionary type of object would be ideal. The defaults could then be defined as a key, as well as other properties which could help automatically generate the config ui. It'd probably be quite an undertaking though, definitely outside the scope of this PR.

> I'm also unsure if this catches differing config value names for UI and in C++ (e.g. stream_audio in UI currently maps to audio.stream and not audio.stream_audio)

I believe the ui names always map to what goes in the config file and docs

e.g. in config.cpp

```cpp
bool_f(vars, "stream_audio", audio.stream);
```

This PR parses out the second argument, to make sure it exists in the config.html, configuration.md, and localization file.

There could certainly be improvements around the whole config structure of the project IMO.

- https://github.com/LizardByte/roadmap/issues/37
- https://github.com/LizardByte/roadmap/issues/25
- https://github.com/LizardByte/roadmap/issues/87 (if going the sql route for 25)

I'd say this PR is definitely not perfect, but an improvement overall, and caught several inconsistencies so far. I'll merge this as is, but will accept PRs to further improve it.

PR #4244, opened 2025-09-06
feat(config): Added ability to listen source rather than sink

Wohlstand:
## Description
Adding the `audio_source` option that allows to capture explicit `source` device (like `line-in` or `microphone`) rather than capturing an application's sound by making a virtual audio output device to trick the application. In my case I wanted to route the sound source from hardware, and with existing setup it's impossible to do without serious complicated tricks on the audio server by making loopbacks that connects sources and sinks.

By default, `audio_source` is empty, and behaviour is stay as before. But once `audio_source` is defined, then Sunshine will explicitly capture audio from the audio source rather than from the sink's monitor.

### Screenshot
I don't have any UI changes, only backend.

### Issues Fixed or Closed
Discussion:  https://github.com/orgs/LizardByte/discussions/830

#### Roadmap Issues
Nothing

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -510,6 +510,7 @@ namespace config {
 
   audio_t audio {
     {},  // audio_sink
+    {},  // audio_source
     {},  // virtual_sink
     true,  // stream audio
     true,  // install_steam_drivers
@@ -1165,6 +1166,7 @@ namespace config {
     list_prep_cmd_f(vars, "global_prep_cmd", config::sunshine.prep_cmds);
 
     string_f(vars, "audio_sink", audio.sink);
+    string_f(vars, "audio_source", audio.source);
     string_f(vars, "virtual_sink", audio.virtual_sink);
     bool_f(vars, "stream_audio", audio.stream);
     bool_f(vars, "install_steam_audio_drivers", audio.install_steam_drivers);
--- a/src/config.h
+++ b/src/config.h
@@ -146,6 +146,7 @@ namespace config {
 
   struct audio_t {
     std::string sink;
+    std::string source;
     std::string virtual_sink;
     bool stream;
     bool install_steam_drivers;
--- a/src/platform/linux/audio.cpp
+++ b/src/platform/linux/audio.cpp
@@ -456,7 +456,8 @@ namespace platf {
           sink_name = get_default_sink_name();
         }
 
-        return ::platf::microphone(mapping, channels, sample_rate, frame_size, get_monitor_name(sink_name));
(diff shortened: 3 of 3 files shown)


Wohlstand:
Okay, now I got another idea how to make this better: maybe implement this as an option of the "Application" entry to capture the hardware source rather than make a fake output device for the work? Otherwise the thing I did will work just globally for all the "applications" if the thing supposed to be used for many people on the same server. In my case I just share stuff from my local PC to my female-friend and we play the same thing.

Wohlstand:
I guess I will need to refresh it and make sure it's compatible with the latest state.

ReenigneArcher:
> I guess I will need to refresh it and make sure it's compatible with the latest state.

I don't think anything has changed in regards to audio, so shouldn't be any conflicts.

Wohlstand:
The another case:
- I didn't documented anything yet.
- Option I made that works on Linux only, I am not sure does it really necessary on Windows and other platforms.
- The option I made, works just globally, but after reviewing how Sunshine is supposed to work, I got a think to try implement it as the "Application-based".

And so, just how I used it actually:
- I combined Sunshine with the modified NXBT thing (NXBT is a daemon that allows to control Nintendo Switch from PC by simulating Pro Controller via raw Bluetooth manipulation).
- I ran Sunshine with "Desktop" only "application" and removed everything.
- I passed to my female-friend who connects Sunshine the screen of Nintendo Switch via full-screen OBS's projector widget and the capture of Line-IN.
- The thing I made allows to don't make me an inconvenience that it steals the default input and this confuses our call done via messenger running on same computer.
- We sometimes might play the game running on same PC, but I making a custom loopback where I manually tell the game to send output into custom sink and I split the sink between loopback and the Sunshine's capturer (based on my line-in thing).

And, generally thing I made I would like to make cleaner and being more stable. I additionally made (but not yet sent) an option that tells Sunshine to don't change audio defaults at all to don't ruin our call conversation.

Wohlstand:
Hey, bot, thanks for a reminder! I need to finish this thing and design it properly, then I can send this as a final contribution.

Wohlstand:
The only problem that updated thing is not buildable locally on my end, and I have some other things to do, so I keep the state at the old version so it's usable until I will figure for the build. (Actually, dropping the support of old libraries too early is a pain, especially for the stable-focused distros, even it's not so large problem to build them from the source code manually and install to the isolated directory).

Wohlstand:
I will need to refresh my branch and ensire it's buildable.

PR #4293, opened 2025-09-24
build(macos): Remove XCode dependency

eduardomozart:
## Description

On PR https://github.com/LizardByte/Sunshine/pull/4133 has been enforced XCode to build Sunshine, but it's does not seems required. It's a 12GB download, and I was able to compile Sunshine on macOS Tahoe without it.

## Type of Change

## Checklist

- [ ] Code follows the style guidelines of this project
- [ ] Code has been self-reviewed

### AI Usage


ReenigneArcher:
Thank you for the PR!

How would you propose we limit `std::format`, which is what the code you're removing does. Relying strictly on the clang version didn't work for me in that PR, though I don't know if anything changed since then.

eduardomozart:
I didn't face any issues while compiling. My C++ knowledge is little to none, but I do believe that downloading a 12GB file and having to keep it updated (since a new XCode version is released with every macOS version, at least), into a non-develop machine (macOS users don't have any option, there's no pre-built package for us and we have to build sunshine if we want to use it), as I was always able to compile any C++ program using command-line tools.

Here are my build log files from Homebrew:
[sunshine.zip](https://github.com/user-attachments/files/22504836/sunshine.zip)

Here's my ``clang``/``gcc`` version:

```
mac@MacBook-Pro-de-Eduardo Downloads % clang --version
Apple clang version 17.0.0 (clang-1700.3.19.1)
Target: arm64-apple-darwin25.0.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

CLang 13 and higher should offer support to ``std::format`` just fine.

If ``std::format`` is an issue, Gemini AI suggested the following:

> ``fmtlib`` is a header-only library, meaning you can include it directly in your project without needing a separate compilation step. It is also highly portable and works with older C++ standards (C++11 and later), making it a great alternative when a modern C++20 compiler isn't available or desired.
> 
> 1. Add ``fmtlib`` as a dependency to the Homebrew formula. This would involve a line like ``depends_on "fmt"`` or a similar method to ensure the library is available during the build.
> 2. In the Sunshine source code, change all ``#include <format>`` to ``#include <fmt/format.h>`` and replace all calls to ``std::format`` with ``fmt::format``.

ReenigneArcher:
> CLang 13 and higher should offer support to std::format just fine.

I think there is more too it than that, as with clang 14 on macOS-13 it still didn't work.

I'm fine with removing the xcode block though if it's truly not needed.

@andygrundman do you have any thoughts on this?

eduardomozart:
On [StackOverflow](https://stackoverflow.com/questions/71777564/cant-use-stdformat-in-c20) they suggest try to add ``-std=c++20`` when calling ``clang`` since Clang doesn't use C++20 by default, so you must specify it explicitly.

It could also be possible to use GNU GCC from Homebrew instead of Apple one if it doesn't include ``std::format`` on older macOS versions.
It's available on ``brew install gcc`` and can be used with ``gcc-<version>``/``g++-<version>``.

Another alternative would be requiring XCode only for macOS 13 and not for newer versions.

andygrundman:
My instinct is that Xcode/clang is probably required for the Mac-specific Objective-C code and it's likely going to give the best performance on Mac hardware, since it's the only compiler 99% of people use. It's also free, I don't see the problem in requiring it? BTW it's a 2GB download, weird complaint to make.

Edit: The real solution here is that Sunshine needs to be an .app bundle, I started working on this but cmake is kind of a nightmare in the project...

eduardomozart:
It's a 12GB download on macOS Tahoe (ARM):

image

Not only is the size an issue, but the bandwidth required to download it is also an issue (it can take an eternity to download on 2.4GHz Wi-Fi, and it can take up to 10% of disk space in a low-capacity SSD, both of which are still a reality in Brazil). Additionally, it's a bad practice to have development tools installed on endpoint (client-side) machines, as this could allow an attacker to compile code on the fly and run arbitrary code on the machine. I understand the nature of Sunshine and the lack of developers to help maintain it and develop new features, so I appreciate the effort you put into making it possible. Thank you.

PR #4304, opened 2025-09-26
build(linux): fix build packages with systemd services and udev

### LINKED ISSUE #4243: System D unit not being created on clean install (Unit sunshine.service not found)

### Describe the Bug

Brand new install on Debian Trixie using the latest pre release build  v2025.628.4510 but after installing this I am unable to start sunshine as I get the following error:

Failed to start sunshine.service: Unit sunshine.service not found.

This will also prevent it starting on boot. 

I have tried other pre release build v2025.828.181854 as the latest stable build does not yet support Trixie. It has the same issue. 

I resolved the issue in the discord server thread [here](https://discordapp.com/channels/804382334370578482/1411303388162822154) by manually creating the service:

```sudo nano /etc/systemd/system/sunshine.service

[Unit]
Description=Sunshine Service
After=network.target
[Service]
ExecStart=/usr/bin/sunshine
Restart=on-failure
[Install]
WantedBy=multi-user.target

sudo systemctl daemon-reload
sudo systemctl enable sunshine.service
sudo systemctl start sunshine.service
```

I did not think on this further until another discord user reported having the same issue on Arch Linux. They also had to manually create the service file. 

### Expected Behavior

sunshine.service should be created 

### Additional Context

User blue.c on discord has also reported that a clean install on Arch and derivatives like CachyOS using the LizardByte repo has the same issue for them. 

### Host Operating System

Linux

### Operating System Version

Debian 13

### Architecture

amd64/x86_64

2025.829.135256

### Package

Linux - deb

### GPU Type

AMD

### GPU Model

RX 6600XT

### GPU Driver/Mesa Version

25.0.7-2

### Capture Method

KMS (Linux)

### Config

```Shell

```

### Apps

```json

```

### Relevant log output

```shell
No logs created. Will try to recreate in a fresh Debian 13 VM with DPKG flags to log install.
```

PVermeer:
## Description

**Issues**
I've found the following issues while testing the build system:
- Builds do not fail anymore when systemd or udev is not found. Adding `if(UDEV_FOUND) ` and `if(SYSTEMD_FOUND)` to:
https://github.com/LizardByte/Sunshine/blob/c4e5a69c3903d7f3abce0f7388fcb4a6acbc2f08/cmake/packaging/linux.cmake#L22-L34
will make the build silently fails if something changes to the build environment and thus will not install the associating files. I believe this change was made to allow builds on other init systems.

- The newly added test that uses udevadm would silently not run when udev is missing when it should be there.

- Some of the docker images that build the packages are failing on install or at runtime.

**Goals**
- Docker and Fedora COPR build run from the Sunshine CI should fail on missing systemd/udev.
- Docker images should not fail to install or run the package.

**Changes**
- Added a flag in Cmake that sets SUNSHINE_CI. This can be used to check if the build is running in the Sunshine ci. Other possibilities could be to set something like SUNSHINE_NO_SYSTEMD or pass the init system. This is now used to fail the build if systemd/udev/udevadm is not found. I've als implemented it into the build script (--ci) and added it to the COPR build.

- Added the missing packages for the Debian build. It now need *-dev packages to allow pkg-config to query it.

- This project uses recent c++ syntax that requires an up-to-date compiler. Ubuntu-22 does not provide a recent enough compiler. This is added with a ppa: https://github.com/LizardByte/Sunshine/blob/c4e5a69c3903d7f3abce0f7388fcb4a6acbc2f08/scripts/linux_build.sh#L262-L266 This install also requires a newer `libstdc++` library. If the user tries to run sunshine without updating their system with this ppa, it will give a runtime link error. So I tested a bit and the solution I could come up with is to static link this lib when the compiler is < 14 (56cd3737a66389f208cbbfca6992e9a695cd7ca7).

- Some of the debian-based builds also had a runtime error for missing the package `libicu`. This lib is versioning itself within its name... which is :(. This a unicode lib, I believe, so I would expect it to be there on almost all systems. To be sure I've added the debian-base distro versions  (libicu76 | libicu74 | libicu70) to the deb package dependencies (e1286f67a685084c64ffc107b9ba369592aa382d).

This fixed all the build / run issues with docker and the ci.

**Notes**
As a bonus I've added an extra info step to the docker builds. This prints all the relevant information about the package to debug faster.

Let me know what you think. Also I'm not sure about the --ci flag naming (maybe be more specific with naming it SYSTEMD something).

### Issues Fixed or Closed

Fixes #4243

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,8 +21,6 @@ crowdin.yml
 build/
 cmake-*/
 venv/
-.vscode/
-.github/
 
 # ignore artifacts
 artifacts/
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@
 build/
 cmake-*/
 docs/doxyconfig*
+.flatpak-builder/
 
 # npm
 node_modules/
--- a/packaging/linux/fedora/Sunshine.spec
+++ b/packaging/linux/fedora/Sunshine.spec
@@ -121,6 +121,7 @@ cmake_args=(
   "-DSUNSHINE_ENABLE_WAYLAND=ON"
   "-DSUNSHINE_ENABLE_X11=ON"
   "-DSUNSHINE_ENABLE_DRM=ON"
+  "-DSUNSHINE_CI=ON"
   "-DSUNSHINE_PUBLISHER_NAME=LizardByte"
   "-DSUNSHINE_PUBLISHER_WEBSITE=https://app.lizardbyte.dev"
   "-DSUNSHINE_PUBLISHER_ISSUE_URL=https://app.lizardbyte.dev/support"


PVermeer:
@ReenigneArcher Started this PR to discuss furter.

Docker works differently with systemd (it doesn't work unless...) but it should work with the variable query that Sunshine does. I will look into it.

PVermeer:
@ReenigneArcher I've fixed the debian issue but came across some more issues. See updated op.

PVermeer:
I've created PR #4321 to fix the ubuntu-22 build. I do agree that most of it should be static linked.

When looking into `libicu` on debian I've also noted that this comes from Boost. `boost::locale` Uses this and when I looked in the build structure of this module I could not found a way to also static link this besides creating or editing the build config of Boost. I don;t think they support this. We can omit it... but this is not recommended by Boost.

Besides that all other distros have this included in their base except for Debian (docker). The desktop versions of Debian all have this package installed so these should have no problems. I think the current solution is still the best (add it to package dependencies) so we can ensure it's there for at least the Debian based builds.

Concerning the `ci` argument I was not happy with this too. We can leave it out or have an other solution. I would just remove the `if(SYSTEMD_FOUND)` statements or replace the logic with a build flag `NO_SYSTEMD` because I do think it should fail in the CI and not get pushed to users on stable.

ReenigneArcher:
For libicu, is there a way to detect which version to use?

We can also put it into the else section of this block, since I believe it only applies when we static link boost.

```cmake

if(NOT BOOST_USE_STATIC)
    set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
                ${CPACK_DEBIAN_PACKAGE_DEPENDS}, \
                libboost-filesystem${Boost_VERSION}, \
                libboost-locale${Boost_VERSION}, \
                libboost-log${Boost_VERSION}, \
                libboost-program-options${Boost_VERSION}")
    set(CPACK_RPM_PACKAGE_REQUIRES "\
                ${CPACK_RPM_PACKAGE_REQUIRES}, \
                boost-filesystem >= ${Boost_VERSION}, \
                boost-locale >= ${Boost_VERSION}, \
                boost-log >= ${Boost_VERSION}, \
                boost-program-options >= ${Boost_VERSION}")
endif()
```

PVermeer:
I think I found the issue with libicu.
https://github.com/LizardByte/Sunshine/blob/f52891d6fc7063efd96849134ca60837878085b0/cmake/packaging/linux.cmake#L96-L97
Is this still nessecary? I only found an issue with the `file` package not installed. Turning this back on and adding `file` to the debian package list fixes all dependency issues. Just like in the COPR build it would automatically add all shared dependencies, including `libicu`.

Anyway if create PR #4323 with this build setup.

**Alternative:**
I found out that cmake provides a module for ICU (like `FindUdev``) [FindICU](https://cmake.org/cmake/help/latest/module/FindICU.html) . This could be used like this:
```cmake
    find_package(ICU COMPONENTS uc data i18n REQUIRED)
    # ICU does not provide a minor / major version variable.
    string(REGEX MATCH "^[0-9]+" ICU_VERSION_MAJOR ${ICU_VERSION})
    # Debian-based distros have libicu versioned in the name.
    set(CPACK_DEBIAN_PACKAGE_DEPENDS "\
                ${CPACK_DEBIAN_PACKAGE_DEPENDS}, \
                libicu${ICU_VERSION_MAJOR}")
    # Fedora based distros provide a general package
    set(CPACK_RPM_PACKAGE_REQUIRES "\
                ${CPACK_RPM_PACKAGE_REQUIRES}, \
                libicu")
```

PVermeer:
I've created the last PR #4325 for Fedora builds. I personally could never build sunshine properly on my own Fedora (docker) but with some changes implemented to the build script taken from the rpm spec I managed to make it work in docker.

Closing this because I think we got just about everything done.

PR #4325, opened 2025-10-09
build(linux): explicitely set CC and CXX compilers

PVermeer:
## Description
This is hopefully the last PR. This fixes the fedora package build via cmake. When I build sunshine locally I could never get cuda to work so I always build it without it since I don't need it.

However I looked into the rpm spec file and transferred those changes to the `linux_build.sh` script and it now all seems to work.

**Issues & solutions**
- `wget` Complained about the relative link in the -O argument when downloading cuda so I made it resolve to an an absolute path before it gets used (2beecac11587c791e457e300560c69a612c59ec4).
- `cmake` Complained it could not find compilers so these are now explicitly set. This was already done in the `copr` build (0279b20f506034357428b5df977cc31fc18436f8). 
- `cuda` Could not build so I copied some config over from the rpm spec (16918a8d67267ead2565668c10c7971b6fe30c1f):
   - added the `--override` flag conditionally
   - added `CMAKE_CUDA_HOST_COMPILER` flag to cmake

Every docker (including my own Fedora) is now building and packaging correctly. I don't know anything about building cuda so please look carefully.

I also want to note that it should also be possible to use the `linux_build.sh` script in the `copr` build so you don't have to maintain two separate builds. Things like cuda, the build and install could be run from the script.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/scripts/linux_build.sh
+++ b/scripts/linux_build.sh
@@ -749,6 +749,15 @@ if [ "$architecture" != "x86_64" ] && [ "$architecture" != "aarch64" ]; then
   exit 1
 fi
 
+# export variables for github actions ci
+if [ -f "$GITHUB_ENV" ]; then
+  {
+    echo "CC=gcc-${gcc_version}"
+    echo "CXX=g++-${gcc_version}"
+    echo "GCC_VERSION=${gcc_version}"
+  } >> "$GITHUB_ENV"
+fi
+
 # get directory of this script
 script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 build_dir=$(readlink -f "$script_dir/../build")
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -186,7 +186,7 @@ jobs:
         working-directory: build
         run: |
           ${{ steps.python.outputs.python-path }} -m pip install gcovr
-          ${{ steps.python.outputs.python-path }} -m gcovr . -r ../src \
+          ${{ steps.python.outputs.python-path }} -m gcovr --gcov-executable "gcov-${GCC_VERSION}" . -r ../src \
             --exclude-noncode-lines \
             --exclude-throw-branches \
             --exclude-unreachable-branches \


scripts/linux_build.sh
```diff
@@ -455,6 +460,9 @@ function run_step_deps() {
   if [ "$distro" == "arch" ]; then
     export CC=gcc-14
     export CXX=g++-14
+  elif [ "$distro" == "fedora" ]; then
```
ReenigneArcher: I think we could probably use this logic for all distro types and then remove the update alternatives section. `gcc_version` variable is already set for arch too, so I don't know why that's hard coded here.

PVermeer:
> As a note: we don't use this script for Fedora or Arch internally, so it's likely those 2 will break from time to time. We used it for Fedora before we changed to the copr builds.

I know, I'm just saying you could ;). The logic in the script and the rpm spec is now basically the same.

scripts/linux_build.sh
PVermeer: I think you're right

PVermeer:
I've removed all the compiler version logic with alternative compilers and just set the compiler with version before cmake runs. It's much cleaner indeed.

Tested all build again, no issues found.

ReenigneArcher:
Just one thing to fix/tweak.

This step (https://github.com/PVermeer/Sunshine/blob/699e8347da229243c190cea1fb8950a5b0f35ecd/.github/workflows/ci-linux.yml#L180) is failing because it's finding a different gcov version than what version of gcc the code was compiled with. I guess because the export variables are not getting picked up by this step.

I think to solve it we can echo the variables to the GITHUB_ENV file, like so. Then they will picked up in later steps automatically.

```sh
{
  echo "CC=gcc-${gcc_version}"
  echo "CXX=g++-${gcc_version}"
} >> "${GITHUB_ENV}"
```

This assumes gcovr/gcov will use the CC/CXX variables.

PVermeer:
Setting the variables doesn't work.
We need to do this: https://gcovr.com/en/latest/guide/compiling.html#choosing-the-right-gcov-executable.
I've done some testing and this works:
```sh
python3 -m gcovr --gcov-executable gcov-14 . -r ../src \
    --exclude-noncode-lines \
    --exclude-throw-branches \
    --exclude-unreachable-branches \
    --verbose \
    --xml-pretty \
    -o coverage.xml
```

I've implemented your suggestion with the info above (20cec9805b72f0cad19459be65e3a03fe186d4e7).

Issue #4338, opened 2025-10-15
No Video received from host, seems to be memory leak


void-512:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

No video received from host, asked me to check firewall rules. Already check the rules for UDP 47998 and UDP 48000

### Expected Behavior

Should have normal connection

### Additional Context

Laptop is 13900H+4060
The connection is successful if I enable the dGPU and choose "Only Nvidia GPU" in the Nvidia control panel. 
But if I disable the dGPU and use the iGPU only, or if I enable the dGPU but the monitor is connected to the iGPU, I can only get the error above. 
And if I restart the sunshine and try to make a connection from my phone, the sunshine will use up all the memory.
Have to paste the photo here because screen shot doesn't work.

![Image](https://github.com/user-attachments/assets/4c1ac0d9-1000-4d27-9030-80977dbadced)
![Image](https://github.com/user-attachments/assets/896eb2a1-d0b7-4770-9877-ea0303bf260a)

### Host Operating System

Windows

### Operating System Version

Win11 Pro 24H2

### Architecture

amd64/x86_64

### Sunshine commit or version

v2025.1014.193231

### Package

Windows - installer (recommended)

### GPU Type

Intel

### GPU Model

13900H iGPU

### GPU Driver/Mesa Version

32.0.101.7029

### Capture Method

Desktop Duplication API (Windows)

### Config

```Shell
address_family = both
av1_mode = 3
capture = ddx
fec_percentage = 40
gamepad = x360
hevc_mode = 3
locale = zh
notify_pre_releases = enabled
nvenc_preset = 5
nvenc_twopass = full_res
nvenc_vbv_increase = 400
qp = 20
qsv_coder = cabac
qsv_preset = slow
sw_preset = slow
sw_tune = film
upnp = enabled
```

### Apps

```json

```

### Relevant log output

```shell
[2025-10-15 18:06:52.768]: Info: Sunshine version: 2025.1014.193231 commit: 179c01a2eaf80f8562454b1b5fdccedfff603876
[2025-10-15 18:06:52.768]: Info: Package Publisher: LizardByte
[2025-10-15 18:06:52.768]: Info: Publisher Website: https://app.lizardbyte.dev
[...]

PR #4340, opened 2025-10-16
fix(windows): Memory leak when EnumAdapters1 uses dxgi::adapter_t address as parameter

xd-byte:
This PR fixes a memory leak issue in the Windows DXGI adapter enumeration code. Previously, dxgi::adapter_t instances were passed directly to EnumAdapters1(), which expects a raw IDXGIAdapter1**. This could cause the internal pointer of dxgi::adapter_t (a uniq_ptr wrapper) to be overwritten, leading to potential memory leaks or invalid pointer access.

### DIFF
--- a/src/platform/windows/display_base.cpp
+++ b/src/platform/windows/display_base.cpp
@@ -1059,8 +1059,9 @@ namespace platf {
       return {};
     }
 
-    dxgi::adapter_t adapter;
-    for (int x = 0; factory->EnumAdapters1(x, &adapter) != DXGI_ERROR_NOT_FOUND; ++x) {
+    dxgi::adapter_t::pointer adapter_p;
+    for (int x = 0; factory->EnumAdapters1(x, &adapter_p) != DXGI_ERROR_NOT_FOUND; ++x) {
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Thank you for the PR!

Could you fix the lint errors?

Also, do you think this will fix the memory leak reported in https://github.com/LizardByte/Sunshine/issues/4338?

xd-byte:
> Thank you for the PR!
> 
> Could you fix the lint errors?
> 
> Also, do you think this will fix the memory leak reported in #4338?

I have fixed the lint errors. Also, I don't think this fixes #4338, IDXGIAdapter1 won't take up that much memory.

ReenigneArcher:
@xd-byte the commit is coauthored by another git account, is that intentional?

xd-byte:
> @xd-byte the commit is coauthored by another git account, is that intentional?

Thanks for noticing! I’ve updated the commits so the author now matches my GitHub account and force-pushed the branch.

Rick-laboratory:
@xd-byte the user on #4338 did mention he had troubles to choose the igpu, i think you fixed this.

PR #4626, opened 2026-01-25
fix(linux): install vendor preset for sunshine-kms service

psyke83:
## Description

* Install vendor preset so that sunshine-kms will default to disabled
* Remove kms service from flatpak and AppImage (as it can't work)
* Small documentation cleanups

## Type of Change

- [x] **docs**: Documentation only changes

## Checklist

### AI Usage


ReenigneArcher (commented):
Does this section of the docs need to be updated as well?

https://github.com/LizardByte/Sunshine/blob/3e2f4c284630ad1b832271843b4250c2136307fc/docs/getting_started.md?plain=1#L282-L285

Users already had a hard time getting this to work, if it even worked at all.

psyke83:
Let me give the flatpak version a proper test and get back to you. My first instinct is that we should just remove it, as flatpak should never gain privileged capabilities; if users want KMS capture, they should be using another package type.

psyke83:
In my opinion, the Flatpak instructions for KMS should be removed from the docs and packaging/linux/dev.lizardbyte.app.Sunshine.metainfo.xml. It's telling users basically to run as root, which seems risky even when you acknowledge it's sandboxed.

Edit: also, to clarify, you cannot install a service vendor preset to any folder in the user's /home that systemd will recognize. That's another reason why the KMS service is not a good idea to install in this case, as both services will default to enabled and thus may conflict in unpredictable ways.

I also investigated the AppImage "install" behaviour: Ubuntu 25.10's default AppArmor policy blocks setuid from being performed in a fusermount3 fs (which is used by AppImage), so there's no point in trying to install a kms service & preset for that case as well, since we can't even set the cap_sys_admin capability needed for the KMS service.

Will I update the PR with these conclusions in mind? I can keep the lines in the AppArmor scripts, but they won't work unless/until a workaround is devised for the AppArmor policy, which may never happen.

ReenigneArcher:
Thanks @psyke83 !

If you could update the PR with your findings that would be appreciated. Removing the parts from the AppImage script is probably a good idea. I think it worked on an older version of fuse, or I didn't test it well enough when I first implemented the custom AppImage script. Instead I think we can update the docs to indicate that KMS will not work with Flatpak and AppImage.

PR #4627, opened 2026-01-25
feat(api): add application image endpoint

cilps:
Introduces a new API endpoint that returns a application's image.

## Description

Creates a new API route and function. Adds an error_message argument to the not_found function in confighttp.cpp. Updated docs to include new endpoint.

### Issues Fixed or Closed

Resolves https://discord.com/channels/804382334370578482/1464609795012169988

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -191,6 +191,7 @@ namespace confighttp {
    * @brief Send a 404 Not Found response.
    * @param response The HTTP response object.
    * @param request The HTTP request object.
+   * @param error_message The error message to include in the response.
    */
   void not_found(resp_https_t response, [[maybe_unused]] req_https_t request, const std::string &error_message = "Not Found") {
     constexpr SimpleWeb::StatusCode code = SimpleWeb::StatusCode::client_error_not_found;
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -192,12 +192,12 @@ namespace confighttp {
    * @param response The HTTP response object.
    * @param request The HTTP request object.
    */
-  void not_found(resp_https_t response, [[maybe_unused]] req_https_t request) {
+  void not_found(resp_https_t response, [[maybe_unused]] req_https_t request, const std::string &error_message = "Not Found") {
     constexpr SimpleWeb::StatusCode code = SimpleWeb::StatusCode::client_error_not_found;
 
     nlohmann::json tree;
     tree["status_code"] = code;
-    tree["error"] = "Not Found";
+    tree["error"] = error_message;
 
     SimpleWeb::CaseInsensitiveMultimap headers;
     headers.emplace("Content-Type", "application/json");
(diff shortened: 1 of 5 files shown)


src/confighttp.cpp
```diff
@@ -921,6 +921,73 @@ namespace confighttp {
...
+      if (index < 0 || index >= static_cast<int>(apps_node.size())) {
+        std::string error;
+        if (const int max_index = static_cast<int>(apps_node.size()) - 1; max_index < 0) {
+          error = "No applications found";
+        } else {
+          error = std::format("'index' {} out of range, max index is {}", index, max_index);
+        }
+        bad_request(response, request, error);
+        return;
+      }
```
ReenigneArcher: This is basically duplicated from the `deleteApp` function. We should add a reusable function to avoid the repetition.

src/confighttp.cpp
```diff
@@ -921,6 +921,73 @@ namespace confighttp {
...
+
+      auto &app = apps_node[index];
+      if (!app.contains("image-path") || app["image-path"].is_null()) {
+        not_found(response, request, "This application has no image");
```
ReenigneArcher:
```suggestion
        not_found(response, request, "'image-path' not set or does not have a 'png' file extension");
```

src/confighttp.cpp
```diff
@@ -921,6 +921,73 @@ namespace confighttp {
...
+          error = "No applications found";
+        } else {
+          error = std::format("'index' {} out of range, max index is {}", index, max_index);
+        }
+        bad_request(response, request, error);
+        return;
+      }
+
+      auto &app = apps_node[index];
+      if (!app.contains("image-path") || app["image-path"].is_null()) {
```
ReenigneArcher: We should add a condition to check if the lowercase value ends with `.png`, as we only support png.

src/confighttp.cpp
```diff
@@ -921,6 +921,73 @@ namespace confighttp {
...
+        return;
+      }
+
+      std::string path = app["image-path"];
+
+      if (!std::filesystem::exists(path)) {
+        BOOST_LOG(debug) << "File not found: " << path;
+        not_found(response, request, "Image not found, check if the path exists");
+        return;
+      }
```
ReenigneArcher: I am not a security expert, but I feel like we should not indicate if the file was found. With the png check above this is less of an issue, but without it, it would be possible to generate a file map of the operating system if an attacker were to gain access to the API.

src/confighttp.cpp
```diff
@@ -921,6 +921,73 @@ namespace confighttp {
...
+      std::ifstream in(path, std::ios::binary);
+      if (!in) {
+        BOOST_LOG(debug) << "Unable to read file: " << path;
+        not_found(response, request, "Unable to read image");
+        return;
+      }
```
ReenigneArcher: Can we do a check to ensure the file is in fact a `png` image, other than the file extension check above?

https://stackoverflow.com/a/3770670/11214013

ReenigneArcher (changes requested):

src/confighttp.cpp
cilps: Should be fixed. I named the function `check_app_index`, in snake case because other helper functions also used snake case.

src/confighttp.cpp
cilps: Should be fixed.

ReenigneArcher:
@Cilps turns out I actually need this for some changes I'm making in https://github.com/LizardByte/Sunshine/pull/4631

Are you able to resolve the few remaining things or would you like help completing it?

I also noticed one other thing. The default images will not be found by the "path" as these exist in the covers directory. So first we need to check the covers directory and then fallback to checking the entire path.

cilps:
I can resolve the changes today, I only have to test the png validation (and make the /covers path check).

cilps:
I added the PNG validation function, but that function is entirely AI generated, please double check that function. In my testing it works though.

src/confighttp.cpp
cilps: Fixed.

ReenigneArcher:
I made a few tweaks to this PR.

1. It re-uses cover image logic from `process.cpp/h`, so it will return the same as what is provided to Moonlight. This will be the default image if no image is specified. It also means we don't have to specifically check all the directories like the cover directory, assets directory, etc. as this logic is already defined.
2. Moves the png validation logic to the process.cpp/h, and also uses it when Moonlight requests the cover now.
3. Adds some unit tests.

ReenigneArcher (approved):

cilps:
Why did you add the check_content_type function in the getCover function? It checks for the Content-Type application/json header, which is weird because you don't have to send any JSON in the request body to use this endpoint.

ReenigneArcher:
I'm fixing it in https://github.com/LizardByte/Sunshine/pull/4631

PR #4641, opened 2026-01-28
fix(linux/xdgportal): basic frame pacing support

psyke83:
## Description

* Fetch pts and seq pipewire metadata to use as duplicate detection and
  for accurate frame_timestamp calculation (to 100ms limit).
* Duplicate frames are interpreted as timeouts, even though snapshot()
  usually returns duplicate frames immediately.
* Don't send fake frames on timeouts. Rationale: timeouts are recoverable.
* Change capture flow to better accommodate pipewire behaviour:
  - Old behaviour: sleep until next_frame deadline, grab frame, send.
  - New behaviour: grab, send. On timeout, sleep 1ms and retry grab
    until deadline is exceeded, then resync deadline. After 100x1ms retries,
    enter regular deadline waiting mode. On success, sleep until deadline.
  - This flow greatly reduces missed frames due to pipewire sending frames
    at unpredictable intervals, especially when vsync is disabled.
  - This flow (particularly sleeping after send) may also improve latency.
  - New flow seems to have reasonable CPU usage. ~5% CPU in top for 60fps,
    3% for 30fps in-game or static desktop on 60fps stream.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/graphics.h
+++ b/src/platform/linux/graphics.h
@@ -295,6 +295,8 @@ namespace egl {
 
     // Increment sequence when new rgb_t needs to be created
     std::uint64_t sequence;
+    // pts metadata from pipewire
+    std::optional<std::uint64_t> pts;
   };
 
   class sws_t {
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -735,9 +735,23 @@ namespace portal {
       if (stream_data.current_buffer) {
         struct spa_buffer *buf;
         buf = stream_data.current_buffer->buffer;
+        struct spa_meta_header *h;
+        h = static_cast<struct spa_meta_header *>(
+          spa_buffer_find_meta_data(buf, SPA_META_Header, sizeof(*h))
+        );
+
         if (buf->datas[0].chunk->size != 0) {
+          const auto img_descriptor = static_cast<egl::img_descriptor_t *>(img);
+          // Set fallback frame_timestamp in case metadata is unavailable
+          img_descriptor->frame_timestamp = std::chrono::steady_clock::now();
+          if (h) {
+            img_descriptor->pts = h->pts;
+            img_descriptor->sequence = h->seq;
+          } else {
+            img_descriptor->sequence = std::numeric_limits<uint64_t>::max();
+          }
+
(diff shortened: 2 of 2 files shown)


psyke83:
I think this is ready for review.

Slightly OT; mutter has frame pacing issues with portal capture when vsync is disabled and host framerate = stream framerate. It happens without this PR, and I can't really figure out to solve it completely other than by enabling vsync, or letting the host render at a higher FPS than the stream rate. Unrelated to this PR, mutter also has an issue with format conversion, with this being spammed constantly:
```
Jan 28 12:31:49 ryzenl gnome-shell[75647]: Could not destroy EGLImage attached to GBM BO: One or more argument values are invalid.
Jan 28 12:31:49 ryzenl gnome-shell[75647]: Error blitting to screencast framebuffer: cogl_framebuffer_blit premult mismatch.
```
mutter sets up supported formats like so:
```
static const struct {
  CoglPixelFormat cogl_format;
  enum spa_video_format spa_video_format;
} supported_formats[] = {
  { COGL_PIXEL_FORMAT_BGRX_8888, SPA_VIDEO_FORMAT_BGRx },
  { COGL_PIXEL_FORMAT_BGRA_8888_PRE, SPA_VIDEO_FORMAT_BGRA },
};
```

The portalgrab negotiates SPA_VIDEO_FORMAT_BGRA, but I guess mutter doesn't like it because it wants it premultiplied and thus has to do format conversion. If we only advertise `{DRM_FORMAT_XRGB8888, SPA_VIDEO_FORMAT_BGRx}` (reordering it to first place doesn't work), it partially helps, but the BO warning remains only in fullscreen games, and we may need BGRA to be supported for HDR, etc. Maybe if the logspam is addressed, performance will also improve related to the vsync off case. Also, mutter may not allow a framerate higher than the host refresh rate. Apart from that, when vsync is engaged, performance is almost stutter free and input latency is better than KMS.

Kwin is working very well with vsync on or off. Input latency is better than KMS and the stream is very smooth. High fps is also working. Something to keep in mind is that the pipewire pts-derived frame_timestamp seems on average 1 frame higher in Kwin than mutter's reporting. If it turns out one or both of these derived timestamps are inaccurate, we can go back to the calculation based on Sunshine's clock pretty easily.

PR #3905, opened 2025-05-25
build(windows): add arm64 support

rbqvq:
## Description
This PR adds an arm64 build for Windows.

## Issues Fixed or Closed
- Resolves https://github.com/orgs/LizardByte/discussions/898
- Resolves https://github.com/orgs/LizardByte/discussions/688

## Type of Change
* [ ]  Bug fix (non-breaking change which fixes an issue)
* [x]  New feature (non-breaking change which adds functionality)
* [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [x]  Dependency update (updates to dependencies)
* [x]  Documentation update (changes to documentation)
* [x]  Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
* [x]  My code follows the style guidelines of this project
* [x]  I have performed a self-review of my own code
* [x]  I have commented my code, particularly in hard-to-understand areas
* [x]  I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

PR #4687, opened 2026-02-08
build(web-ui): fix rollup failing

rbqvq:
## Description
Use actions/setup-node instead of MSYS2.

Revert commit `fix(build): using @rollup/wasm-node` partial changes in #3905

## Type of Change

- [x] **revert**: Reverts a previous commit

## Checklist

### AI Usage

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -71,7 +71,6 @@ jobs:
             "mingw-w64-${TOOLCHAIN}-graphviz"
             "mingw-w64-${TOOLCHAIN}-miniupnpc"
             "mingw-w64-${TOOLCHAIN}-nlohmann-json"
-            "mingw-w64-${TOOLCHAIN}-nodejs"
             "mingw-w64-${TOOLCHAIN}-onevpl"
             "mingw-w64-${TOOLCHAIN}-openssl"
             "mingw-w64-${TOOLCHAIN}-opus"
@@ -82,6 +81,7 @@ jobs:
             dependencies+=(
               "mingw-w64-${TOOLCHAIN}-MinHook"
               "mingw-w64-${TOOLCHAIN}-nsis"
+              "mingw-w64-${TOOLCHAIN}-nodejs"
             )
           fi
 
@@ -146,6 +146,28 @@ jobs:
         with:
           dotnet-version: '10.x'
 
+      - name: Setup NodeJS
+        # Clang compiled NodeJS has issues when running rollup webpack
+        if: matrix.msystem != 'ucrt64'
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238  # v6.2.0
+        with:
+          node-version: 'lts/*'
+
+      - name: NodeJS Path
+        if: matrix.msystem != 'ucrt64'
+        shell: pwsh
+        run: |
+          # get NodeJS PATH
+          $NODEJS_BINARY_PATH = (Get-Command node).Source
+          $NODEJS_PATH = Split-Path -Path "$NODEJS_BINARY_PATH" -Parent
+
+          # setup environment variables
+          echo "NODEJS_PATH=$NODEJS_PATH" >> $env:GITHUB_ENV
+
+          # step output
+          echo "nodejs-path=$NODEJS_PATH"
+          echo "nodejs-path=$NODEJS_PATH" >> $env:GITHUB_OUTPUT
+
       - name: Setup python
         id: setup-python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
@@ -166,10 +188,19 @@ jobs:
       - name: Build Windows
         shell: msys2 {0}
         env:
+          # MSYSTEM is a built-in environment variable of MSYS2.
+          # Do not use this environment variable name.
+          MATRIX_MSYSTEM: ${{ matrix.msystem }}
           BRANCH: ${{ github.head_ref || github.ref_name }}
           BUILD_VERSION: ${{ inputs.release_version }}
           COMMIT: ${{ inputs.release_commit }}
         run: |
+          # setup NodeJS PATH
+          if [[ "${MATRIX_MSYSTEM}" != "ucrt64" ]]; then
+            NODEJS_PATH=$(cygpath "$NODEJS_PATH")
+            export PATH="$PATH:$NODEJS_PATH"
+          fi
+
           mkdir -p build
           cmake \
             -B build \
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-windows.yml
```diff
@@ -141,6 +141,11 @@ jobs:
           # Clean up
           Remove-Item -Path doxygen-setup.exe
 
+      - name: Setup nodejs
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238  # v6.2.0
```
ReenigneArcher:
```suggestion
      - name: Setup nodejs
        if: matrix.msystem != 'ucrt64'
        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238  # v6.2.0
```

.github/workflows/ci-windows.yml
```diff
@@ -71,7 +72,6 @@ jobs:
             "mingw-w64-${TOOLCHAIN}-graphviz"
             "mingw-w64-${TOOLCHAIN}-miniupnpc"
             "mingw-w64-${TOOLCHAIN}-nlohmann-json"
-            "mingw-w64-${TOOLCHAIN}-nodejs"
```
ReenigneArcher, line 74: Can you put this into the ucrt64 conventional block below? Same for docs.

I'd prefer not to complicate the build for the primary use case, when arm is going to a pretty low percentage of overall usage.

.github/workflows/ci-windows.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         uses: msys2/setup-msys2@4f806de0a5a7294ffabaff804b38a9b435a73bda  # v2.30.0
         with:
           msystem: ${{ matrix.msystem }}
+          path-type: inherit
```
ReenigneArcher: I'm not a huge fan of this because there is so much junk installed on the github runners by default.

Could you just do an `export` in the `Build Windows` step to update the path?

ReenigneArcher (commented):
Thank you for the fast resolution on this!

.github/workflows/ci-windows.yml
rbqvq: I don't know where Actions installed NodeJS. :(

rbqvq: Screenshot_2026-02-09 004216

I think this is the problem we encountered with CLANGARM64.

rollup doesn't provide clang's win32 native binding.

Therefore, my personal suggestion is to install the official Node.JS package.

This can avoid most native binding issues (although GNU, which UCRT64 uses, is mostly provided, I think MSVC is more mainstream).

---

If you feel that using UCRT64's built-in NodeJS is better, that seems fine too (at least for now).

However, I personally think that using actions/setup-node for both AMD64 and ARM64 is better.

ReenigneArcher: > I don't know where Actions installed NodeJS. :(

I think printing the contents of `GITHUB_PATH` env variable, before and after the action could help find it?

I had opened a request in the action to output the path a while ago, but nothing has been implemented yet: https://github.com/actions/setup-node/issues/1122

rbqvq: Looks seems a bit difficult because the paths appear to change with the Node.js version:

/c/npm/prefix
/c/hostedtoolcache/windows/node/24.12.0/arm64

rbqvq: Done.

rbqvq:
It looks like fine.

I've manually mounted the NodeJS PATH.

I think we don't need a PATH for npm prefix.

---

CI PASSED at https://github.com/rbqvq/Sunshine/actions/runs/21802858672/job/62901261780

---

I'm going to sleep now.
I think everything's fine.
Please feel free to make any further changes.

ReenigneArcher (approved):

Yundi339:
Hello, we have encountered the same problem as you, but we are also very lucky that this problem has not been triggered until today.
I don't understand the reason for this rollover, it seems to be introduced because of the version update?
I created PR to slove this problem by COPILOT AI. 

https://github.com/qiin2333/foundation-sunshine/pull/503

The following is the translation.

When Rollup's module pipeline processes an entry file, it invokes the `load` hooks sequentially according to the plugin order. The first plugin to return a non-`null` value dictates the result.

**Older versions of Rollup (worked previously)**:

- The `load` hook of Vite's internal `vite:build-html` plugin was ordered **before** your plugin.
- It intercepted the HTML entry first, extracting the `<script>` tags and converting them into JS module code.
- Your plugin's `load` hook was **never even called** (preempted by Vite's internal plugin).

**Newer versions of Rollup (currently throwing errors)**:

- The execution order or scheduling logic for plugin `load` hooks has changed.
- Your plugin's `load` hook is now invoked **before** Vite's internal HTML plugin.
- It returned raw HTML → Rollup tried to parse it as JS → `parseAsync` threw an error.

So it's not that "the new version started using it," but rather that **the new version changed the invocation order of the `load` hooks**, causing previously bypassed code to suddenly execute. It should never have returned HTML to Rollup in the first place; it was just pure luck that it wasn't being called before.

------

Theoretically, you could add `enforce: 'post'` to your `load` hook to order it after Vite's internal plugins. This way, `vite:build-html` would grab the HTML first, and your `load` wouldn't be invoked.

**But this is not a good fix**:

- Relying on plugin execution order is fragile; it might break again in the next Rollup/Vite update.
- Rollup will **always** parse whatever the `load` hook returns as a JS module, so returning HTML is fundamentally wrong.
- `transformIndexHtml` is the official API in Vite for processing HTML, and it fully covers this functionality.

Deleting the `load` hook is the correct solution—it doesn't rely on order, carries no risk of being accidentally invoked, and has zero negative impact on functionality.

ReenigneArcher:
This PR only addressed an issue when compiling for debian on arm64. Due to adding `"@rollup/wasm-node": "4.57.1"` in https://github.com/LizardByte/Sunshine/pull/3905

It wasn't detected before merge, because we don't compile arm64 docker builds in CI/CD for PRs.

Yundi339:
Alright Ok.
That's very strange. My question is a low probability trigger. And I just need to click to rebuild, and it works smoothly.

As described by the last comment:
> So it's not that "the new version started using it," but rather that **the new version changed the invocation order of the `load` hooks**, causing previously bypassed code to suddenly execute. It should never have returned HTML to Rollup in the first place; it was just pure luck that it wasn't being called before.

If you encounter similar problems in future upgrades, you could consider to deal with it like what I did. I hope you will don't encounter it, it's too metaphysical.

rbqvq:
Thank you for sharing, the problem here is stable trigger, last year's version did not have this problem, the problem has been solved.

PR #4645, opened 2026-01-30
refactor(confighttp): HTML page handlers into generic getPage function

ReenigneArcher:
## Description

Initial motivation for the PR was to consolidate multiple individual HTML page handler functions into a single `getPage` function. This works by providing the page as a parameter. Then HTTP handler functions were updated to take response and request by const reference, which was a clang-tidy suggestion. New helper functions for page handlers were added to simplify the server route setup.

Introduced check_request_body_empty to validate that certain API endpoints receive no request body, replacing previous content-type checks where appropriate. I am not 100% sure if this is necessary. Docs were updated to only include the `Content-Type` headers when necessary.

Also, added some initial tests for `confighttp.cpp/h`.

Lastly, I had to add macros to ignore some error/warnings in tests/tests_common.h due to failures in the ArchLinux build. I am not fully understanding what caused these errors to start occurring.

## Type of Change

- [ ] **feat**: New feature (non-breaking change which adds functionality)
- [ ] **fix**: Bug fix (non-breaking change which fixes an issue)
- [x] **docs**: Documentation only changes
- [ ] **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semicolons, etc.)
- [x] **refactor**: Code change that neither fixes a bug nor adds a feature
- [ ] **perf**: Code change that improves performance
- [x] **test**: Adding missing tests or correcting existing tests
- [ ] **build**: Changes that affect the build system or external dependencies
- [ ] **ci**: Changes to CI configuration files and scripts
- [ ] **chore**: Other changes that don't modify src or test files
- [ ] **revert**: Reverts a previous commit
- [ ] **BREAKING CHANGE**: Introduces a breaking change (can be combined with any type above)

## Checklist

- [x] Code follows the style guidelines of this project
- [x] Code has been self-reviewed
- [x] Code has been commented, particularly in hard-to-understand areas
- [x] Code docstring/documentation-blocks for new or existing methods/components have been added or updated
- [x] Unit tests have been added or updated for any new or modified functionality

### AI Usage

- [ ] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [x] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

PR #4688, opened 2026-02-08
feat(confighttp): make the api/logs more elegant

Yundi339:
## Description

I think the log interface returns the full amount every time, which is an unreasonable mechanism.
Of course, I did consider using SSE, but it's not a good fit here. It would just introduce unnecessary complexity and over-engineering. So I tried to do these things.

Logs API improvements:

- Incremental fetch: client sends X-Log-Offset in request header, server returns only new bytes and X-Log-Size in response. Avoids re-sending the full file every 5s.
- HTTP 304 when offset already equals file size (no body).
- Backend cache: re-read log file only when size or mtime changes; use std::atomic<std::shared_ptr<const std::string>> for thread-safe access without locks.
- Pause polling when tab is hidden (Page Visibility API); resume and refresh when visible.
- Handle non-2xx and 304 so failed responses don’t overwrite existing log content.
- Sunshine uses `try_incremental_log_read` to avoid getting the full logs every time.

============
I know that modifying code related to file operations is an act that is prone to dangerous problems.
Therefore, I suggest that this PR should not be in a hurry to integrate, I will use it locally for a long time to see the stability.

BTW, I added log files to  `sunshinesvc.cpp` on PC, so I could know when Sunshine.exe had a safe and unsafe restart.

### Screenshot

image

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -1111,7 +1111,7 @@ namespace confighttp {
     if (!in || !in.seekg(static_cast<std::streamoff>(prev_size))) {
       return nullptr;
     }
-    const auto tail_len = static_cast<std::size_t>(current_size - prev_size);
+    const auto tail_len = current_size - prev_size;
     std::string tail(tail_len, '\0');
     if (!in.read(tail.data(), static_cast<std::streamsize>(tail_len))) {
       return nullptr;
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -11,6 +11,7 @@
 #include <format>
 #include <fstream>
 #include <set>
+#include <atomic>
 
 // lib includes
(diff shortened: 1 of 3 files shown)


Yundi339:
To prevent indefinite memory growth, a setting should be added in the future to limit the buffer size (e.g., 1MB, 10MB, 50MB). This would not affect the local log files. For example, if set to 10MB, the system should only read the trailing 10MB of the file.

ReenigneArcher:
Thanks for the PRs! For this one specifically, I think I want to get https://github.com/LizardByte/Sunshine/pull/4645 finished and merged first.

ReenigneArcher:
@Yundi339 the previously mentioned PR was merged

PR #4711, opened 2026-02-12
feat(macos/tray): allow tray icon paths to be set at runtime from .app bundle

andygrundman:
## Description
This PR is to prep for building a macOS .app bundle, where the icon paths need to be set at runtime from the .app's Resources directory. A future PR will change the TRAY_ICON * paths to be relative on Mac.

#### Roadmap Issues

Part of https://github.com/LizardByte/roadmap/issues/18

## Type of Change

## Checklist

### AI Usage

GetResourcePath() and the static cache were suggested by GPT.

### DIFF
--- a/src/system_tray.cpp
+++ b/src/system_tray.cpp
@@ -23,7 +23,9 @@
     #define TRAY_ICON_PLAYING WEB_DIR "images/sunshine-playing-16.png"
     #define TRAY_ICON_PAUSING WEB_DIR "images/sunshine-pausing-16.png"
     #define TRAY_ICON_LOCKED WEB_DIR "images/sunshine-locked-16.png"
+    #include <CoreFoundation/CoreFoundation.h>
     #include <dispatch/dispatch.h>
+    #include <unordered_map>
   #endif
 
   // standard includes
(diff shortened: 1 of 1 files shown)


src/system_tray.cpp
```diff
@@ -127,6 +129,58 @@ namespace system_tray {
...
+    static std::unordered_map<std::string, std::string> g_cache;
+    auto search = g_cache.find(relativePath);
+    if (search != g_cache.end()) {
+      return search->second.c_str();
+    }
+
+    // If we're running from an .app bundle, get the internal Resources dir
+    CFBundleRef bundle = CFBundleGetMainBundle();
+    if (!bundle) {
+      return nullptr;
```
ReenigneArcher: should we return the existing value, otherwise the tray would break if it's not app bundle?

andygrundman: The resourcesPath seems to just be the current working dir of the binary (/usr/local/bin) when run as a normal binary today.

But you're right, if this code returns nullptr it will crash later on. I'll return the relativeString instead.

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** +[NSString stringWithUTF8String:]: NULL cString'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000019a2fb8fc __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x0000000199dd2418 objc_exception_throw + 88
	2   Foundation                          0x000000019c5348a4 +[NSString stringWithBytes:length:encoding:] + 0
	3   sunshine-0.0.0-dirty                0x0000000100fe7b20 tray_init + 128
```

PR #4712, opened 2026-02-13
fix(linux): update systemd service units

psyke83:
## Description

* xdg-desktop-autostart.target is not used by vanilla GNOME sessions, so use graphical-session.target as the service enablement dependency.
* Specify graphical-session.target, xdg-desktop-autostart.target and xdg-desktop-portal.service (unprivileged) to run before Sunshine, but leave as weak ordering hints by only specifying via the After unit fields.

### Issues Fixed or Closed

- Closes #4561

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/packaging/linux/sunshine-kms.service.in
+++ b/packaging/linux/sunshine-kms.service.in
@@ -3,6 +3,7 @@ Description=@PROJECT_DESCRIPTION@
 StartLimitIntervalSec=500
 StartLimitBurst=5
 Conflicts=sunshine.service
+After=graphical-session.target xdg-desktop-autostart.target
 
 [Service]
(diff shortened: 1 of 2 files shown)


psyke83:
A change that could be made in the case of the unprivileged service is to add:
```
[Unit]
Wants = xdg-desktop-portal.service
```

This would set up a weak dependency, so if the service exists on the system but was not scheduled to start, it would be started. XDG Desktop Portal needs this, but the unprivileged user service can also be used for x11 and wlroots capture, so I don't think we should assume it to be a requirement. Users may have deliberately disabled this service and may not appreciate Sunshine starting it, so I don't recommend setting this.

PR #4724, opened 2026-02-14
feat(capture/linux): add KWin direct screencast capture method

Ramalama2:
## Description

Add a new `kwin` capture method that bypasses `xdg-desktop-portal` entirely by using KWin's `zkde_screencast_unstable_v1` Wayland protocol directly.

**Before (Portal path):**
```
Sunshine → D-Bus → xdg-desktop-portal → D-Bus → xdg-desktop-portal-kde → Wayland → KWin → PipeWire → Sunshine
```

**After (Direct path):**
```
Sunshine → Wayland → KWin → PipeWire → Sunshine
```

### Why?

The portal path requires two additional D-Bus services (`xdg-desktop-portal` and `xdg-desktop-portal-kde`) that act as pure pass-through layers with no added functionality for Sunshine's use case. In containerized environments (LXC/Docker), the portal frontend also has a known bug (`EACCES` on `/proc/<pid>/root` in user namespaces) that requires patching.

By speaking the KWin screencast Wayland protocol directly, Sunshine:
- Eliminates two D-Bus intermediaries and their associated services
- Removes the need for portal auto-approve patches/environment variables
- Removes the need for `dbus-update-activation-environment` hacks
- Works reliably in unprivileged LXC containers without portal patches
- Reuses the existing PipeWire stream infrastructure from `portalgrab.cpp` (`pipewire_t` class)

### How it works

New file `kwingrab.cpp` implements a standalone Wayland client that:
1. Connects to the Wayland display and binds `zkde_screencast_unstable_v1` from the registry
2. Enumerates `wl_output` objects and selects the target output
3. Calls `stream_output()` to request a screencast stream from KWin
4. Receives the PipeWire node ID via the `created` callback
5. Connects to PipeWire using `pw_context_connect()` (local socket, no FD passing needed)
6. From there, the existing PipeWire DMA-BUF frame pipeline handles everything identically to the portal path

Enabled via `capture = kwin` in the Sunshine config and the CMake flag `-DSUNSHINE_ENABLE_KWIN=ON`.

### Real-world testing

Tested extensively over WAN (VDSL) with Moonlight at 1080p/4K, 60/120fps. The KWin direct path delivers noticeably smoother streaming compared to NvFBC + Xorg, particularly over WAN where consistent frame cadence matters most. KWin's vsync-aligned, compositor-driven frame delivery produces a more even packet flow than NvFBC's software-timer-based polling.

### Screenshot

N/A (no UI changes)

## Type of Change
- [ ] **perf**: Code change that improves performance                                     - [ ] **test**: Adding missing tests or correcting existing tests

## Checklist

### AI Usage

### DIFF
--- a/cmake/compile_definitions/linux.cmake
+++ b/cmake/compile_definitions/linux.cmake
@@ -76,8 +76,8 @@ if(CUDA_FOUND)
     add_compile_definitions(SUNSHINE_BUILD_CUDA)
 endif()
 
-# libdrm is required for both DRM (KMS) and Wayland
-if(${SUNSHINE_ENABLE_DRM} OR ${SUNSHINE_ENABLE_WAYLAND})
+# libdrm is required for DRM (KMS), Wayland, and KWin ScreenCast
+if(${SUNSHINE_ENABLE_DRM} OR ${SUNSHINE_ENABLE_WAYLAND} OR ${SUNSHINE_ENABLE_KWIN})
     find_package(LIBDRM REQUIRED)
 else()
     set(LIBDRM_FOUND OFF)
@@ -184,10 +184,33 @@ if(PIPEWIRE_FOUND)
             "${CMAKE_SOURCE_DIR}/src/platform/linux/portalgrab.cpp")
 endif()
 
+# KWin ScreenCast (direct Wayland protocol, bypasses portal)
+# Requires WAYLAND for wl::display_t and PipeWire for frame transport
+if(${SUNSHINE_ENABLE_KWIN})
+    if(NOT PIPEWIRE_FOUND)
+        pkg_check_modules(PIPEWIRE libpipewire-0.3 REQUIRED)
+    endif()
+endif()
+set(KWIN_FOUND OFF)
+if(WAYLAND_FOUND AND PIPEWIRE_FOUND AND ${SUNSHINE_ENABLE_KWIN})
+    set(KWIN_FOUND ON)
+    add_compile_definitions(SUNSHINE_BUILD_KWIN)
+
+    GEN_WAYLAND("${CMAKE_SOURCE_DIR}/third-party/kde-protocols" "" zkde-screencast-unstable-v1)
+
+    include_directories(SYSTEM ${PIPEWIRE_INCLUDE_DIRS})
+    list(APPEND PLATFORM_LIBRARIES ${PIPEWIRE_LIBRARIES})
+    list(APPEND PLATFORM_TARGET_FILES
+            "${CMAKE_SOURCE_DIR}/src/platform/linux/kwingrab.cpp")
+elseif(${SUNSHINE_ENABLE_KWIN} AND NOT WAYLAND_FOUND)
+    message(WARNING "SUNSHINE_ENABLE_KWIN requires SUNSHINE_ENABLE_WAYLAND — KWin capture disabled")
+endif()
+
 if(NOT ${CUDA_FOUND}
         AND NOT ${WAYLAND_FOUND}
         AND NOT ${X11_FOUND}
         AND NOT ${PIPEWIRE_FOUND}
+        AND NOT ${KWIN_FOUND}
         AND NOT (${LIBDRM_FOUND} AND ${LIBCAP_FOUND})
         AND NOT ${LIBVA_FOUND})
     message(FATAL_ERROR "Couldn't find either cuda, libva, pipewire, wayland, x11, or (libdrm and libcap)")
--- a/src/platform/linux/misc.cpp
+++ b/src/platform/linux/misc.cpp
@@ -892,6 +892,9 @@ namespace platf {
 #ifdef SUNSHINE_BUILD_PORTAL
       PORTAL,  ///< XDG PORTAL
 #endif
+#ifdef SUNSHINE_BUILD_KWIN
+      KWIN,  ///< KWin ScreenCast
+#endif
       MAX_FLAGS  ///< The maximum number of flags
     };
   }  // namespace source
@@ -943,6 +946,15 @@ namespace platf {
   }
 #endif
 
+#ifdef SUNSHINE_BUILD_KWIN
+  std::vector<std::string> kwin_display_names();
+  std::shared_ptr<display_t> kwin_display(mem_type_e hwdevice_type, const std::string &display_name, const video::config_t &config);
+
+  bool verify_kwin() {
+    return window_system == window_system_e::WAYLAND && !kwin_display_names().empty();
+  }
+#endif
+
   std::vector<std::string> display_names(mem_type_e hwdevice_type) {
 #ifdef SUNSHINE_BUILD_CUDA
     // display using NvFBC only supports mem_type_e::cuda
@@ -968,6 +980,11 @@ namespace platf {
 #ifdef SUNSHINE_BUILD_PORTAL
     if (sources[source::PORTAL]) {
       return portal_display_names();
+    }
+#endif
+#ifdef SUNSHINE_BUILD_KWIN
+    if (sources[source::KWIN]) {
+      return kwin_display_names();
     }
 #endif
     return {};
@@ -1011,6 +1028,12 @@ namespace platf {
     if (sources[source::PORTAL]) {
       BOOST_LOG(info) << "Screencasting with XDG portal"sv;
       return portal_display(hwdevice_type, display_name, config);
+    }
+#endif
+#ifdef SUNSHINE_BUILD_KWIN
+    if (sources[source::KWIN]) {
+      BOOST_LOG(info) << "Screencasting with KWin ScreenCast"sv;
+      return kwin_display(hwdevice_type, display_name, config);
     }
 #endif
 
@@ -1066,6 +1089,11 @@ namespace platf {
 #ifdef SUNSHINE_BUILD_PORTAL
     if ((config::video.capture.empty() || config::video.capture == "portal") && verify_portal()) {
       sources[source::PORTAL] = true;
+    }
+#endif
+#ifdef SUNSHINE_BUILD_KWIN
+    if (((config::video.capture.empty() && sources.none()) || config::video.capture == "kwin") && verify_kwin()) {
+      sources[source::KWIN] = true;
     }
 #endif
 
--- a/third-party/kde-protocols/zkde-screencast-unstable-v1.xml
+++ b/third-party/kde-protocols/zkde-screencast-unstable-v1.xml
@@ -1,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="zkde_screencast_unstable_v1">
+  <copyright><![CDATA[
+    SPDX-FileCopyrightText: 2020-2021 Aleix Pol Gonzalez <aleixpol@kde.org>
+
+    SPDX-License-Identifier: LGPL-2.1-or-later
+  ]]></copyright>
+  <interface name="zkde_screencast_unstable_v1" version="5">
+    <description summary="Protocol for managing PipeWire feeds of the different displays and windows">
+      Warning! The protocol described in this file is a desktop environment
+      implementation detail. Regular clients must not use this protocol.
+      Backward incompatible changes may be added without bumping the major
+      version of the extension.
+    </description>
+
+    <enum name="pointer">
+      <description summary="Stream consumer attachment attributes" />
+      <entry name="hidden" value="1" summary="No cursor"/>
+      <entry name="embedded" value="2" summary="Render the cursor on the stream"/>
+      <entry name="metadata" value="4" summary="Send metadata about where the cursor is through PipeWire"/>
+    </enum>
+
+    <request name="stream_output">
+      <description summary="requests a feed from a given source"/>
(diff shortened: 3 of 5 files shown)


Ramalama2:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=LizardByte_Sunshine&pullRequest=4724) **Quality Gate failed**
> 
> Failed conditions  [54 New issues](https://sonarcloud.io/project/issues?id=LizardByte_Sunshine&pullRequest=4724&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [15.2% Duplication on New Code](https://sonarcloud.io/component_measures?id=LizardByte_Sunshine&pullRequest=4724&metric=new_duplicated_lines_density&view=list) (required ≤ 2%)  [6 Duplicated Blocks on New Code](https://sonarcloud.io/dashboard?id=LizardByte_Sunshine&pullRequest=4724) (required ≤ 0)  [54 New Code Smells](https://sonarcloud.io/dashboard?id=LizardByte_Sunshine&pullRequest=4724) (required ≤ 0)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=LizardByte_Sunshine&pullRequest=4724)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Regarding SonarCloud findings:

The duplication is expected — `kwingrab.cpp` reuses the same PipeWire stream handling patterns as `portalgrab.cpp` because the PipeWire side is nearly identical. The key difference is how the PipeWire node ID is obtained: directly via KWin's Wayland protocol instead of through the Portal D-Bus chain. A follow-up refactor to extract the shared PipeWire code into a common module would be welcome if the maintainers prefer that.

Two legitimate issues to fix:
- **L729**: `new uint8_t[]` without matching `delete` — will replace with `std::make_unique`
- **L768**: `malloc` without `free` — will replace with `std::vector<uint64_t>`

The remaining 52 findings are all Wayland C API constraints — the callback signatures (`void *data`, unused parameters, empty handlers like `on_output_done`) are dictated by `wl_output_listener` and the Wayland protocol. These cannot be changed without breaking the API contract.

cmake/compile_definitions/linux.cmake
```diff
@@ -184,10 +184,33 @@ if(PIPEWIRE_FOUND)
             "${CMAKE_SOURCE_DIR}/src/platform/linux/portalgrab.cpp")
 endif()
 
+# KWin ScreenCast (direct Wayland protocol, bypasses portal)
+# Requires WAYLAND for wl::display_t and PipeWire for frame transport
+if(${SUNSHINE_ENABLE_KWIN})
+    if(NOT PIPEWIRE_FOUND)
+        pkg_check_modules(PIPEWIRE libpipewire-0.3 REQUIRED)
+    endif()
+endif()
```
ReenigneArcher, line 193: I think it would be better to a section above XDG portal that checks if portal or kwin are set, and does the pkg_check_modules for the shared dependencies once.

src/platform/linux/misc.cpp
```diff
@@ -891,6 +891,9 @@ namespace platf {
 #endif
 #ifdef SUNSHINE_BUILD_PORTAL
       PORTAL,  ///< XDG PORTAL
+#endif
+#ifdef SUNSHINE_BUILD_KWIN
+      KWIN,  ///< KWin ScreenCast
```
ReenigneArcher, line 896: tests also need to be udpated

third-party/kde-protocols/zkde-screencast-unstable-v1.xml
ReenigneArcher, line 1: Should add https://github.com/KDE/plasma-wayland-protocols as a submodule instead of copy/pasting external source code.

ReenigneArcher (changes requested):
I'm not sure how I feel about another "unstable" protocol being added to Sunshine. The current one has limped along for it's entire existence. @psyke83 do you have any thoughts on adding this?

Regarding the sonar issues, I understand you can't change the return types of signatures, but a good number of the reported issues can be fixed. Certainly all of these could be fixed fairly easily.

- https://sonarcloud.io/project/issues?rules=cpp%3AS1172&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4724&id=LizardByte_Sunshine can by declaring them `[[maybe_unused]]`. Not sure what IDE you're using, but CLion can help you do this automatically.
- https://sonarcloud.io/project/issues?rules=cpp%3AS1186&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4724&id=LizardByte_Sunshine
- https://sonarcloud.io/project/issues?rules=cpp%3AS1121&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4724&id=LizardByte_Sunshine
- https://sonarcloud.io/project/issues?rules=cpp%3AS1905&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4724&id=LizardByte_Sunshine

Others might take more effort to fix, but it's certainly possible.

Anyway, my review comments are below. In addition to those comments, config, web-ui, and docs would also need to be udpated to allow users to force KWin.

psyke83:
@ReenigneArcher I've not yet looked at the code thoroughly, but the actual runtime performance seems to be identical to portalgrab (same downsides are visible on my hardware, namely periodic stuttering at 60fps vsync on, and 60 fps vsync off often degrades to perceptual 30fps). The capture method doesn't really offer any major advantage in that aspect, as the bottleneck (on my hardware at least) seems to be KWin's pipewire handling. This capture method also required me to set `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1` in my Fedora 43 environment even though the specific error check in the code that mentions this didn't trigger correctly.

@Ramalama2 The downside of XDG Portal grab is understandable, but can't you overcome this issue by using a rootless container so that the sunshine process is not using the root namespace?

Ramalama2:
> I'm not sure how I feel about another "unstable" protocol being added to Sunshine.

Same here. Since I built this backend for my own needs, I figured I should at least put up a PR. But to be honest, neither portalgrab nor kwingrab is a great solution.

NvFBC and DRM/KMS are both amazing backends. NvFBC works in unprivileged containers, but it will die with Xorg at some point — and in real-world experience over WAN, it wasn't that great either. That said, the approach itself is excellent. DRM/KMS is equally impressive and future-proof, but it will never work in containers, at least not in unprivileged ones.

Portalgrab and kwingrab are essentially the same thing in the end. In theory, portalgrab offers a bit more compatibility with other environments going forward, but it will never work with headless containers. Kwingrab was born out of that need, but I'm not entirely happy with the solution myself.

I just don't have a better idea.

That said — you're the ones who decide. If you see this as something that adds maintenance burden with no real benefit, we don't need to continue and can simply close this PR. If, on the other hand, you see value in it as the only capture method that works in an unprivileged container (or potentially Docker) with Wayland + KDE, then I'm happy to do the refactoring you've outlined.

Cheers

Ramalama2:
> @ReenigneArcher I've not yet looked at the code thoroughly, but the actual runtime performance seems to be identical to portalgrab (same downsides are visible on my hardware, namely periodic stuttering at 60fps vsync on, and 60 fps vsync off often degrades to perceptual 30fps). The capture method doesn't really offer any major advantage in that aspect, as the bottleneck (on my hardware at least) seems to be KWin's pipewire handling. This capture method also required me to set `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1` in my Fedora 43 environment even though the specific error check in the code that mentions this didn't trigger correctly.
> 
> @Ramalama2 The downside of XDG Portal grab is understandable, but can't you overcome this issue by using a rootless container so that the sunshine process is not using the root namespace?

@psyke83 Thanks for testing. The identical performance to portalgrab is expected — both methods end up at the same KWin PipeWire stream, so the capture path from KWin onward is the same. The advantage of kwingrab isn't capture performance, it's eliminating the two D-Bus intermediaries (xdg-desktop-portal and xdg-desktop-portal-kde) and the issues they bring in containerized environments.

Regarding `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1`: yes, that's expected. KWin restricts access to `zkde_screencast_unstable_v1` via a desktop file whitelist (`org.kde.portal.desktop` and similar). Without the env var, only whitelisted clients can bind the protocol. This is the same restriction that xdg-desktop-portal-kde bypasses by being a whitelisted KDE component. I should document this requirement more clearly.

Regarding the rootless container suggestion: unfortunately, no. The EACCES error is not about whether the process runs as root or not — it's caused by user namespace isolation in the kernel. When xdg-desktop-portal tries to open `/proc/<pid>/root` of the calling process (to check if it's a Flatpak), the kernel denies access because the caller lives in a different user namespace. This happens regardless of the UID inside the container. Both privileged and unprivileged LXC containers have separate user namespaces, so the EACCES occurs in either case. The only fix is patching xdg-desktop-portal to treat EACCES as "not a Flatpak" and fall through to host app detection — which is what I had to do for portalgrab to work in the first place.

Ramalama2:
I'm planning to implement HDR into KWin's `zkde_screencast_unstable_v1` protocol. It's actually a straightforward task because KWin itself already supports HDR fully — the infrastructure is all there (10-bit GL formats, BT.2020 colorimetry, color pipeline). From my initial source code analysis, the changes on the KWin side are minimal:

- `screencaststream.cpp` — add 10-bit formats (ARGB2101010 etc.) to the supported format array
- `outputscreencastsource.cpp` — `drmFormat()` should query the actual output format instead of hardcoding ARGB8888
- `outputscreencastsource.cpp` — `render()` should use `GL_RGB10_A2` instead of always `GL_RGBA8`

However, I'm not sure if I'll get a PR approved on the KWin side when I submit it.

I'll probably do it for myself anyway — just two additional patches, one for KWin and one for kwingrab in Sunshine. But this raises a question about how to handle the Sunshine side:

**Option A:** I add the HDR code to kwingrab in this PR already (10-bit format map, colorspace parsing from PipeWire, passing it to NVENC). That way it's one less Sunshine PR later, and the kwingrab code would be HDR-ready from day one.

**Option B:** I keep this PR as-is (SDR only) and submit a separate Sunshine PR for HDR support later, once I've gotten the KWin changes through.

Option B is probably safer since there's a risk the KWin PR doesn't get accepted. But honestly, on the KWin side it shouldn't be controversial — it's purely additive code with no downsides to existing behavior (SDR fallback stays intact, format negotiation is backwards-compatible).

What do you think? If you see kwingrab as worth merging, I could include the HDR groundwork right away. If kwingrab itself is still uncertain, I'll keep the scope minimal and do HDR separately.

ReenigneArcher:
@Ramalama2 I would say to keep this PR as is, until the changes are merged on the KWin side. Then update this PR once it's available. Otherwise a secondary PR would be fine as well.

Also, I'm good with the idea of adding this capture method. Definitely happy to have more people contributing on the Linux side of things!

Ramalama2:
> @Ramalama2 I would say to keep this PR as is, until the changes are merged on the KWin side. Then update this PR once it's available. Otherwise a secondary PR would be fine as well.
> 
> Also, I'm good with the idea of adding this capture method. Definitely happy to have more people contributing on the Linux side of things!

Great! I'll refactor this PR.
Just need a bit of time, as I'm currently deep into testing and bug-fixing on my unprivileged LXC container setup. Especially with KDE Plasma, since getting a minimal, reliable installation in a headless container without all the bloat is quite a journey.

I'll keep the HDR changes in a separate PR then, as suggested. Will update here once the refactor is ready.

ReenigneArcher:
Marking this PR as draft. Please mark as ready for review once you make the requested changes. Thanks!

Ramalama2:
> Marking this PR as draft. Please mark as ready for review once you make the requested changes. Thanks!

Yes sorry for the delay.
My work hit me hard lately. Im building a complex AD and have 1000 other things todo for the work in the meantime.
So i sadly didn't continued here.

And additionally i fought hard and debugged a lot of Unpriviliged LXC Container Problems related to this Project, it felt like i debugged half of the Linux Source Code (Kernel/HID/Wayland/udev and so on).
And additionally the Update to Plasma/KWin 6.6 needed debugging either since it didn't worked in an Unpriviliged LXC Container. Needed in the end another ENV Variable (DRM Master issues).

However, my LXC Container works absolutely Amazing after all the Work, its insane how good it is.
I'll Write maybe a How-To.
(Have a dual-setup now with Xorg+Gamescope+NvFBC for Steam Gaming and Plasma 6.6+Kwin+Kwin Backend)
Both are working insanely amazing over WAN, 20ms latency, but 0 stalls, and butter smooth 4k@120fps.

However, im building currently a Python Proxy, to switch through Moonlight between the NvFBC/Gamescope and KWin/Plasma pipeline.
Since they both can't run at the same time (DRM Master issue).

Whatever, im sorry for the delay.
I will surely refactor the Code here once im done with all the testing and my Work for my Company calms down a bit.
Probably it will take another 2-3 Weeks until i return to this here.

Thanks for the pation @ReenigneArcher 
Cheers

ReenigneArcher:
No worries as at all! And certainly no rush on this.

PR #4740, opened 2026-02-18
feat(linux/xdgportal): implement reactive capture with duplicate detection

psyke83:
## Description

Replaces polling-based frame capture with event-driven approach using condition variables for immediate notification when new frames arrive.

Changes:
- on_process: Notify snapshot() via condition variable on new buffer
- fill_img: Extract and pass through PipeWire seq/pts metadata
- snapshot: Block on condition variable with timeout; detect and skip duplicate frames using seq/pts metadata to avoid redundant processing
- capture: Simplify catch-up logic to maintain timeline consistency by advancing in delay intervals without resetting origin

This eliminates tight polling loops and reduces unnecessary frame processing when compositor reuses buffers.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/graphics.h
+++ b/src/platform/linux/graphics.h
@@ -295,6 +295,10 @@ namespace egl {
 
     // Increment sequence when new rgb_t needs to be created
     std::uint64_t sequence;
+
+    // PipeWire metadata
+    std::optional<uint64_t> pts;
+    std::optional<uint64_t> seq;
   };
 
   class sws_t {
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -1077,8 +1077,18 @@ namespace portal {
   class portal_t: public platf::display_t {
   public:
     int init(platf::mem_type_e hwdevice_type, const std::string &display_name, const ::video::config_t &config) {
+      // calculate frame interval we should capture at
       framerate = config.framerate;
-      delay = std::chrono::nanoseconds {1s} / framerate;
+      if (config.framerateX100 > 0) {
+        AVRational fps_strict = ::video::framerateX100_to_rational(config.framerateX100);
+        delay = std::chrono::nanoseconds(
+          (static_cast<int64_t>(fps_strict.den) * 1'000'000'000LL) / fps_strict.num
+        );
+        BOOST_LOG(info) << "Requested frame rate [" << fps_strict.num << "/" << fps_strict.den << ", approx. " << av_q2d(fps_strict) << " fps]";
+      } else {
+        delay = std::chrono::nanoseconds {1s} / framerate;
+        BOOST_LOG(info) << "Requested frame rate [" << framerate << "fps]";
+      }
       mem_type = hwdevice_type;
 
       if (get_dmabuf_modifiers() < 0) {
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -162,7 +162,37 @@ namespace portal {
 
   class dbus_t {
   public:
-    ~dbus_t() {
+    ~dbus_t() noexcept {
+      try {
+        if (conn && !session_handle.empty()) {
+          g_autoptr(GError) err = nullptr;
+          // This is a blocking C call; it won't throw, but we wrap for safety
+          g_dbus_connection_call_sync(
+            conn,
+            "org.freedesktop.portal.Desktop",
+            session_handle.c_str(),
+            "org.freedesktop.portal.Session",
+            "Close",
+            nullptr,
+            nullptr,
+            G_DBUS_CALL_FLAGS_NONE,
+            -1,
+            nullptr,
+            &err
+          );
+
+          if (err) {
+            BOOST_LOG(warning) << "Failed to explicitly close portal session: "sv << err->message;
+          } else {
+            BOOST_LOG(debug) << "Explicitly closed portal session: "sv << session_handle;
+          }
+        }
+      } catch (const std::exception &e) {
+        BOOST_LOG(error) << "Standard exception caught in ~dbus_t: "sv << e.what();
+      } catch (...) {
+        BOOST_LOG(error) << "Unknown exception caught in ~dbus_t"sv;
+      }
+
       if (screencast_proxy) {
         g_object_unref(screencast_proxy);
       }
@@ -264,6 +294,7 @@ namespace portal {
     GDBusConnection *conn;
     GDBusProxy *screencast_proxy;
     GDBusProxy *remote_desktop_proxy;
+    std::string session_handle;
 
     int create_portal_session(GMainLoop *loop, gchar **session_path_out, const gchar *session_token, bool use_screencast) {
       GDBusProxy *proxy = use_screencast ? screencast_proxy : remote_desktop_proxy;
@@ -326,6 +357,8 @@ namespace portal {
       }
 
       BOOST_LOG(debug) << session_type << " CreateSession: got session handle: "sv << *session_path_out;
+      // Save it for the destructor to use during cleanup
+      this->session_handle = *session_path_out;
       return 0;
     }
 
@@ -632,16 +665,24 @@ namespace portal {
      *
      * Call this when the session becomes invalid (e.g., on error).
      */
-    void invalidate() {
-      std::scoped_lock lock(mutex_);
-      if (valid_) {
-        BOOST_LOG(debug) << "Invalidating cached portal session"sv;
-        if (pipewire_fd_ >= 0) {
-          close(pipewire_fd_);
-          pipewire_fd_ = -1;
+    void invalidate() noexcept {
+      try {
+        std::scoped_lock lock(mutex_);
+        if (valid_) {
+          BOOST_LOG(debug) << "Invalidating cached portal session"sv;
+          if (pipewire_fd_ >= 0) {
+            close(pipewire_fd_);
+            pipewire_fd_ = -1;
+          }
+
+          dbus_.reset();
+
+          valid_ = false;
         }
-        dbus_.reset();
-        valid_ = false;
+      } catch (const std::exception &e) {
+        BOOST_LOG(error) << "Exception during session invalidation: "sv << e.what();
+      } catch (...) {
+        BOOST_LOG(error) << "Unknown error during session invalidation"sv;
       }
     }
 
@@ -681,27 +722,7 @@ namespace portal {
     }
 
     ~pipewire_t() {
-      pw_thread_loop_lock(loop);
-
-      if (stream_data.stream) {
-        pw_stream_destroy(stream_data.stream);
-        stream_data.stream = nullptr;
-      }
-      if (core) {
-        pw_core_disconnect(core);
-        core = nullptr;
-      }
-      if (context) {
-        pw_context_destroy(context);
-        context = nullptr;
-      }
-
-      pw_thread_loop_unlock(loop);
-
-      pw_thread_loop_stop(loop);
-      if (fd >= 0) {
-        close(fd);
-      }
+      cleanup_stream();
       pw_thread_loop_destroy(loop);
     }
 
@@ -742,11 +763,25 @@ namespace portal {
           stream_data.current_buffer = nullptr;
         }
 
-        pw_stream_disconnect(stream_data.stream);
-        pw_stream_destroy(stream_data.stream);
-        stream_data.stream = nullptr;
+        if (stream_data.stream) {
+          pw_stream_destroy(stream_data.stream);
+          stream_data.stream = nullptr;
+        }
+        if (core) {
+          pw_core_disconnect(core);
+          core = nullptr;
+        }
+        if (context) {
+          pw_context_destroy(context);
+          context = nullptr;
+        }
 
         pw_thread_loop_unlock(loop);
+
+        pw_thread_loop_stop(loop);
+        if (fd >= 0) {
+          close(fd);
+        }
       }
       session_cache_t::instance().invalidate();
     }
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -137,6 +137,8 @@ namespace portal {
     struct pw_buffer *current_buffer;
     uint64_t drm_format;
     std::shared_ptr<shared_state_t> shared;
+    std::mutex frame_mutex;
+    std::condition_variable frame_cv;
   };
 
   struct dmabuf_format_info_t {
@@ -690,6 +692,14 @@ namespace portal {
       pw_thread_loop_destroy(loop);
     }
 
+    std::mutex &frame_mutex() {
+      return stream_data.frame_mutex;
+    }
+
+    std::condition_variable &frame_cv() {
+      return stream_data.frame_cv;
+    }
+
     void init(int stream_fd, int stream_node, std::shared_ptr<shared_state_t> shared_state) {
       fd = stream_fd;
       node = stream_node;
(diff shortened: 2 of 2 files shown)


psyke83:
TLDR; test with `minimum_fps_target = 1`. I'd appreciate some testing feedback before merging - thanks.

It's vital to ~disable (set to 1 second timeout) the "Minimum FPS target" setting in order for XDG Portal capture pacing to work as expected. With the default setting of 0 (half framerate), this setting can randomly cause 50% or more extra frames (60fps -> 90fps) to be inserted, which causes network jitter that completely breaks pacing. I believe this may be happening because the arrival of pipewire buffers is not perfectly aligned with the presentation on screen, which confuses the encoder thread more than kmsgrab, which likely grabs frames at virtually identical intervals.

Results with RX 6600 on Fedora 43: with the above config set appropriately, pacing is perfect on KDE and GNOME. With vsync disabled and the game capped to 60fps, sometimes there may be some slight jitter that clears up after a few seconds, but I notice the same issue with kmsgrab. Enabling vsync seems to prevent the random jitter, likely due to the compositor pacing the frames closer to the vblank interval. I need to test more to be absolutely sure it's bulletproof, but my current impression is that portal grab pacing is on par with or possibly better than kmsgrab.

ReenigneArcher:
@psyke83 This feature also doesn't work quite right on Windows, and is the source of a lot of confusion over the years.

@andygrundman tagging you as FYI, because you've expressed interest in improving this feature in the past.

andygrundman:
It would be fantastic if Sunshine behaved the same as it does on Windows when it comes to capturing frames. See `src/platform/windows/display_base.cpp` (search for the word pacing). The Sunshine architecture is not very good about defining how capture works, because I think the Windows logic should probably be refactored into a higher base class that can be used by all platforms.

When I say "like Windows", what I mean specifically is:

* Captures frames using a frame pacing timer based on the stream framerate and optionally a client-requested fractional framerate for fine-tuning. This is used on Xbox which runs at NTSC refresh rates. Frames captured must have an accurate frame_timestamp indicating when the frame was captured.
* The capture process will naturally capture duplicate frames, and these should not be streamed, until  the framerate drops below min_fps_target. To keep the framerate above the minimum, duplicate frames are sent as needed. Duplicate frames should have their frame_timestamp removed.
* This timer needs to support the framerate requested in x-nv-video[0].clientRefreshRateX100, you can use the Xbox client for testing this. There are 2 special case NTSC framerates that use defined intervals. Multiples of 29.97, and 23.976, and for every other value it's a simple 1/fps.
* Capture should have no relation to the host's refresh rate or vsync timing.

I have tried to argue for the min_fps_target default to be a value like 20, which would allow streaming of 24fps, 30fps, content without duplicate frames. The 0 value is very unintuitive and wasteful. The old value was more like 10, and there will be a tradeoff here with bursty data when the stream suddenly ramps back up to max, moving the mouse is where some people complain about this burst.

ns6089:
> Capture should have no relation to the host's refresh rate or vsync timing

I won't be dying on this hill, but please don't do this. Capture should provide the lowest possible latency within given constraints. Yes, this requires frame pacing on host, but not extremely tight frame pacing. Client still has to implement its own and proper frame pacing anyway to deal with frame transfer time jitter (not ping jitter, caused by variation in encoded frame sizes), and encoding time jitter (caused by gpu scheduling outliers). And demanding tight host-side frame pacing will only produce worse latency in the outliers on the client.

A generic frame pacer looks like this. For each frame, there's three timepoints:
1. Instant after — frame arriving/timestamped after this point are instantly processed, frames arriving/timestamped before are subjected to (2)
2. Delay until — frame arriving/timestamped before (1) are delayed until this point. Delayed frames can get preempted by instant frames and be subsequently dropped.
3. Repeat at — if there's no new frames at this point, the old frame is repeated

Then these points are dynamically adjusted depending on your particular requirements and recent frame history. Take it as you will.

andygrundman:
The host frame pacer is only there to ensure it doesn't capture too many frames, since sending a 60fps stream to a 59.94fps client means the client has to drop a frame a few times a minute. That results in stutter that no amount of client frame pacing can fix. That's the only hill I'm dying on. :) 

Hey, didn't you write that frame pacing code? I think it's great, it just needs to work on all platforms, that's the main issue right?

psyke83:
As an aside, I'd also recommend testing the master branch with `minimum_fps_target = 1`. 

I'm not fully convinced that my retry logic is fully correct; it currently rejects frames that have identical seq and pts, but there are also cases where the pts is identical but seq differs (on kwin). There's still some infrequent jitter appearing on my hardware when testing this branch and also master too.

andygrundman:
I built your patch on my Cachy system with a GTX 1070 in it. It appears to work very well, and I think the min_fps_target works as expected. I noticed your timer uses a timeout of 1000ms whereas the Windows code uses 200ms. I guess this is why the minimum is more fuzzy in Windows.

I added a few lines for Xbox support and I'm able to play Hades 2 at 1440p120 smooth as butter. If I had a faster GPU in this box you could fool me that it's Windows. Here are my changes, they are very simple. Obviously every other backend will need to adopt our changes for variable framerate, I'm interested in what everyone thinks about that. What's the tier list for the 5 capture methods, are some likely to be retired? I'm surprised to see NvFBC alive and kicking.

cc @ns6089 @ReenigneArcher @cgutman 

```
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -1019,8 +1019,18 @@ namespace portal {
   class portal_t: public platf::display_t {
   public:
     int init(platf::mem_type_e hwdevice_type, const std::string &display_name, const ::video::config_t &config) {
+      // calculate frame interval we should capture at
       framerate = config.framerate;
-      delay = std::chrono::nanoseconds {1s} / framerate;
+      if (config.framerateX100 > 0) {
+        AVRational fps_strict = ::video::framerateX100_to_rational(config.framerateX100);
+        delay = std::chrono::nanoseconds(
+          (static_cast<int64_t>(fps_strict.den) * 1'000'000'000LL) / fps_strict.num
+        );
+        BOOST_LOG(info) << "Requested frame rate [" << fps_strict.num << "/" << fps_strict.den << ", approx. " << av_q2d(fps_strict) << " fps]";
+      } else {
+        delay = std::chrono::nanoseconds {1s} / framerate;
+        BOOST_LOG(info) << "Requested frame rate [" << framerate << "fps]";
+      }
       mem_type = hwdevice_type;

       if (get_dmabuf_modifiers() < 0) {
```

(note the nice flat host frametime graph, and 119.88 on the nose. That's what I like to see.)
hades2_linunx_pipewire_xbox

ReenigneArcher:
> I have tried to argue for the min_fps_target default to be a value like 20, which would allow streaming of 24fps, 30fps, content without duplicate frames. The 0 value is very unintuitive and wasteful. The old value was more like 10, and there will be a tradeoff here with bursty data when the stream suddenly ramps back up to max, moving the mouse is where some people complain about this burst.

I'm okay with changing the default. The old default was 1 which is equivalent to 10, but there was a bug in the code where it would never use the client requested framerate (config.framerate).

old code:

```cpp
    // set minimum frame time, avoiding violation of client-requested target framerate
    auto minimum_frame_time = std::chrono::milliseconds(1000 / std::min(config.framerate, (config::video.min_fps_factor * 10)));
    BOOST_LOG(debug) << "Minimum frame time set to "sv << minimum_frame_time.count() << "ms, based on min fps factor of "sv << config::video.min_fps_factor << "."sv;
```

The client could be set to 30, 60, 90, 120., While the config::video.min_fps_factor could have been 1, 2, or 3... resulting in 10, 20, or 30.

So all of the client side options were greater or equal to the setting in Sunshine.

psyke83:
The Min FPS threshold is definitely a problem for me, but it's a pain in the arse to reproduce easily.

The most reliable way I've found is this config:
* Fedora 43 KDE, Plasma 6.6 (copr beta repo), host GPU is RX 6600. I've tested a little bit on GNOME, but it seems that mutter is less susceptible to this issue.
* Stream is set to 60fps, 1080p. 
* Host has a fullscreen game (e.g. Resident Evil 3) with vsync off but framerate capped in-game to 60fps. The game is not fully loading the GPU (average usage is ~75%).
* Min FPS target is left to the default.

The behaviour with master (in summary):
* Stream is often smooth, but at random intervals, the stream itself will slow down to 30fps, even though Moonlight's statistics don't reflect a deviation from 60fps. For brevity let's just call this the "30fps effect", which is simply the snapshot function grabbing the same buffer and therefore pushing duplicates to the encoder.
* Changing the FPS limit, toggling vsync or swapping windowed -> fullscreen mode can temporarily alleviate the slowdown, but it always comes back.
* Bursting / runaway bursting (where the stream violates the framerate cap) doesn't seem to occur.

The behaviour is as follows with this PR:
* Stream is usually smooth at the beginning, but after a few minutes, you will notice some jitter appear. Moonlight's Incoming and Decoding rate will briefly violate the limit (~61-63) every few seconds which is accompanied by network jitter increasing to 3-5%. These small bursts alone impacts overall smoothness quite negatively.
* After a (seemingly random) period of time, the jitter will worsen, which culminates in runaway bursting where the stream remains locked at 90fps (incoming and decoding) with constant network jitter of 50% or so.
* The runaway bursting can be temporarily alleviated by changing the in-game FPS target to 30, 120, enabling vsync, or triggering a fullscreen -> windows -> fullscreen transition via alt+enter, but the pattern will always return.
* With extra debugging I saw an erratic time variance between push_captured_image_cb calls; I believe when the runaway bursting occurred, it would proceed at an odd 8 -> 22 -> 8ms pattern instead of the steady ~16.6ms cadence that would be expected for a 60fps stream.
* To calm the runaway bursting with this PR, reducing the snapshot timeout from 1-2ms helps, but it reintroduces the 30fps effect. Increasing even to 6-8ms alleviates the 30fps effect, but doesn't prevent eventual runaway bursting.

Given the above behaviour, I believe that the min fps target is not working robustly when the capture method pushes frames at variable rates (which is what happens when snapshot starts blocking to filter duplicates). 

"Disabling" the min fps target completely alleviates both types of bursting with this PR, but it has the expected downside of making 30fps content on a 60fps stream less smooth, as the stream will push 30fps when in-game at 30fps, but mouse movement will trigger new frames past that rate, which causes the stream rate to bounce between 30-60.

psyke83:
> I built your patch on my Cachy system with a GTX 1070 in it. It appears to work very well, and I think the min_fps_target works as expected. I noticed your timer uses a timeout of 1000ms whereas the Windows code uses 200ms. I guess this is why the minimum is more fuzzy in Windows.
> 
> I added a few lines for Xbox support and I'm able to play Hades 2 at 1440p120 smooth as butter. If I had a faster GPU in this box you could fool me that it's Windows. Here are my changes, they are very simple. Obviously every other backend will need to adopt our changes for variable framerate, I'm interested in what everyone thinks about that. What's the tier list for the 5 capture methods, are some likely to be retired? I'm surprised to see NvFBC alive and kicking.
> 
> cc @ns6089 @ReenigneArcher @cgutman
> 
> ```
> --- a/src/platform/linux/portalgrab.cpp
> +++ b/src/platform/linux/portalgrab.cpp
> @@ -1019,8 +1019,18 @@ namespace portal {
>    class portal_t: public platf::display_t {
>    public:
>      int init(platf::mem_type_e hwdevice_type, const std::string &display_name, const ::video::config_t &config) {
> +      // calculate frame interval we should capture at
>        framerate = config.framerate;
> -      delay = std::chrono::nanoseconds {1s} / framerate;
> +      if (config.framerateX100 > 0) {
> +        AVRational fps_strict = ::video::framerateX100_to_rational(config.framerateX100);
> +        delay = std::chrono::nanoseconds(
> +          (static_cast<int64_t>(fps_strict.den) * 1'000'000'000LL) / fps_strict.num
> +        );
> +        BOOST_LOG(info) << "Requested frame rate [" << fps_strict.num << "/" << fps_strict.den << ", approx. " << av_q2d(fps_strict) << " fps]";
> +      } else {
> +        delay = std::chrono::nanoseconds {1s} / framerate;
> +        BOOST_LOG(info) << "Requested frame rate [" << framerate << "fps]";
> +      }
>        mem_type = hwdevice_type;
> 
>        if (get_dmabuf_modifiers() < 0) {
> ```
> 
> (note the nice flat host frametime graph, and 119.88 on the nose. That's what I like to see.) hades2_linunx_pipewire_xbox

I tested this; I'm certainly not against it, but I don't think it will impact the specific issue I'm having on my configuration. When connecting to the stream it will use the updated calculation but then falls back to the original:

```
Feb 19 16:37:22 ryzenl sunshine[11629]: [2026-02-19 16:37:22.624]: Info: Requested frame rate [60/1, approx. 60 fps]
Feb 19 16:37:22 ryzenl sunshine[11629]: [2026-02-19 16:37:22.745]: Info: Requested frame rate [60fps]
```

andygrundman:
> I tested this; I'm certainly not against it, but I don't think it will impact the specific issue I'm having on my configuration. When connecting to the stream it will use the updated calculation but then falls back to the original:
> 
> ```
> Feb 19 16:37:22 ryzenl sunshine[11629]: [2026-02-19 16:37:22.624]: Info: Requested frame rate [60/1, approx. 60 fps]
> Feb 19 16:37:22 ryzenl sunshine[11629]: [2026-02-19 16:37:22.745]: Info: Requested frame rate [60fps]
> ```

I think when you see this in the log, it's from the "// Testing for available encoders" part of the log, which might be setting the X100 field in the config. (I kinda hate the testing section, it makes skimming the log a lot harder than it should be.)

psyke83:
I added a commit to resolve performance when using the software encoder, as it's somewhat related to this change. The stream reports 60fps for software encoding on master branch, but it's actually closer to 45fps when accounting for duplicated buffers.  With reactive capture you can now see the true capture rate, so I've added the change that fixes performance, and also included some fixes for crashing on mode change that only happen on the MemPtr path.

Slightly OT: I tested the changes on Kwin and GNOME to confirm they work as expected, but it also came to my attention that GNOME freezes the stream when entering a fullscreen game with software encoding (with this PR and master branch). It's caused by a direct scanout conflict that can be worked around by installing the "Disable Unredirect" extension. When I have time I'll see if this can be worked around in Sunshine itself.

psyke83:
> > I tested this; I'm certainly not against it, but I don't think it will impact the specific issue I'm having on my configuration. When connecting to the stream it will use the updated calculation but then falls back to the original:
> > ```
> > Feb 19 16:37:22 ryzenl sunshine[11629]: [2026-02-19 16:37:22.624]: Info: Requested frame rate [60/1, approx. 60 fps]
> > Feb 19 16:37:22 ryzenl sunshine[11629]: [2026-02-19 16:37:22.745]: Info: Requested frame rate [60fps]
> > ```
> 
> I think when you see this in the log, it's from the "// Testing for available encoders" part of the log, which might be setting the X100 field in the config. (I kinda hate the testing section, it makes skimming the log a lot harder than it should be.)

I added your change and assigned you as the author - hopefully you don't mind. We can back these changes out and split into another PR if necessary.

psyke83:
I refreshed the original patch to clear the SonarQube warnings. I'm fairly happy with the state of the PR right now.

I do have an additional change that leverages the pts metadata to let us calculate more accurate frame_timestamps, but I'd prefer to change that in another PR, as the pts data from mutter vs kwin is different to each other and thus needs to be handled carefully.

andygrundman:
> I refreshed the original patch to clear the SonarQube warnings. I'm fairly happy with the state of the PR right now.
> 
> I do have an additional change that leverages the pts metadata to let us calculate more accurate frame_timestamps, but I'd prefer to change that in another PR, as the pts data from mutter vs kwin is different to each other and thus needs to be handled carefully.

Looks good to me. If you think it's possible to get more accurate frame_timestamp values, that would be useful. The pts is used by at least the Xbox for frame pacing in one of its two rendering modes, so the closer to the capture time we can get the better. The resolution of RTP's pts field is 1/90000 (11.1 microseconds) so it doesn't need to be better than that.

What are your thoughts on the other capture backends? Wayland, NvFBC, X11, KMS? Linux has too much choice in this area, I hate to say it.

psyke83:
> > I refreshed the original patch to clear the SonarQube warnings. I'm fairly happy with the state of the PR right now.
> > I do have an additional change that leverages the pts metadata to let us calculate more accurate frame_timestamps, but I'd prefer to change that in another PR, as the pts data from mutter vs kwin is different to each other and thus needs to be handled carefully.
> 
> Looks good to me. If you think it's possible to get more accurate frame_timestamp values, that would be useful. The pts is used by at least the Xbox for frame pacing in one of its two rendering modes, so the closer to the capture time we can get the better. The resolution of RTP's pts field is 1/90000 (11.1 microseconds) so it doesn't need to be better than that.
> 
> What are your thoughts on the other capture backends? Wayland, NvFBC, X11, KMS? Linux has too much choice in this area, I hate to say it.

I haven't been using Sunshine on Linux for too long, but my ability to make comparisons are limited as I only have AMD GPUs:
* X11 capture was quite slow and prone to stuttering on my hardware. 
* KMS via X11 or Wayland works fairly reliably, but the input latency is worse than DXGI on the same hardware.
* PortalGrab seems to have good potential. With this PR it is mostly smooth, and it feels like input latency is superior to KMS, but I suspect it still needs a lot of work. I know for sure it doesn't seem to work with absolute mouse mode, and I suspect that multi-GPU setups may have trouble (which I unfortunately can't test).

PR #4768, opened 2026-02-24
feat(linux/xdgportal): implement event-driven capture

psyke83:
## Description

```
feat(linux/xdgportal): implement event-driven capture

Summary of changes:
* Synchronized capture:
  - Move frame_cv wait to beginning of loop as a more efficient way to
    wait for a new buffer whilst ensuring that unintentional buffer
    reuse doesn't occur.
* Improved duplicate detection:
  - Transition to duplicate detection logic using a combination
    of SPA_META_Header (PTS) and SPA_META_VideoDamage metadata.
  - Drop frames when PTS delta is zero and no damage is reported;
    this helps to smoothen out stuttering unique to KWin.
```

Additional fixes:
```
fix(linux/xdgportal): fix pipewire teardown logic, GNOME XDG stop icon

* Destroying the core and context during a reinit can cause a deadlock
which can be reproduced when PW_STREAM_FLAG_RT_PROCESS is set. Fix
by ensuring that full teardown is only done in pipewire destructor.
* Add a delay before sending the interrupt signal, as the GNOME
XDG portal needs time to detect changes, otherwise Sunshine will
crash when the users manually stops the stream on mutter.
```

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/graphics.h
+++ b/src/platform/linux/graphics.h
@@ -301,6 +301,8 @@ namespace egl {
     // PipeWire metadata
     std::optional<uint64_t> pts;
     std::optional<uint64_t> seq;
+    std::optional<bool> pw_damage;
+    std::optional<uint32_t> pw_flags;
   };
 
   class sws_t {
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -723,6 +723,24 @@ namespace portal {
 
     ~pipewire_t() {
       cleanup_stream();
+
+      pw_thread_loop_lock(loop);
+
+      if (core) {
+        pw_core_disconnect(core);
+        core = nullptr;
+      }
+      if (context) {
+        pw_context_destroy(context);
+        context = nullptr;
+      }
+
+      pw_thread_loop_unlock(loop);
+
+      pw_thread_loop_stop(loop);
+      if (fd >= 0) {
+        close(fd);
+      }
       pw_thread_loop_destroy(loop);
     }
 
@@ -775,21 +793,8 @@ namespace portal {
           pw_stream_destroy(stream_data.stream);
           stream_data.stream = nullptr;
         }
-        if (core) {
-          pw_core_disconnect(core);
-          core = nullptr;
-        }
-        if (context) {
-          pw_context_destroy(context);
-          context = nullptr;
-        }
 
         pw_thread_loop_unlock(loop);
-
-        pw_thread_loop_stop(loop);
-        if (fd >= 0) {
-          close(fd);
-        }
       }
       session_cache_t::instance().invalidate();
     }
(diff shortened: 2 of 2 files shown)


psyke83:
I'm marking this ready for review, but I'd appreciate if some other folks could test and verify it works OK on their configuration before considering this for merging.

I've tested on GNOME and KDE 6.6.0 via Fedora 43. The updated duplication logic solves a few issues:
* intermittent stuttering when host rate = stream rate (i.e. 60 -> 60), especially seen in KWin, seems to be completely gone.
* When min_fps_target is set to 1, the stream will still go idle, but ramping from ~1fps to 60fps is instantaneous. This means that there's no obvious stuttering in KWin's animations when transitioning between idle to active due to the damage tracking properly tracking when new frames actually have changes.
* Streaming 30fps host content on a 60fps stream no longer jumps erratically between 30-60; the damage+pts delta tracking ensures that the framerate is accurate, so it remains at a solid ~30fps (or whatever rate the host is rendering). The only exception is when the host cursor is visible, which shows that damage tracking is working as expected, as the cursor updates can run faster than the content.

neatnoise:
Works ok for me. I haven't noticed the difference to the previous portalgrab state though. Tested on the up to date CachyOS with KDE, AMD GPU

ReenigneArcher:
> I'm marking this ready for review, but I'd appreciate if some other folks could test and verify it works OK on their configuration before considering this for merging.

Do you want to ask for more volunteers on Discord? There is a `tester-linux` role we could tag.

psyke83:
> Do you want to ask for more volunteers on Discord? There is a `tester-linux` role we could tag.

Perhaps. I've refreshed the PR to make the check less strict as I've identified that stuttering with KWin may actually be a bug on the compositor side related to XWayland applications. It's either [this](https://bugs.kde.org/show_bug.cgi?id=516432) bug, or a variation of the same issue that only applies to screencasting.

If testing this PR (or portalgrab on master) with Steam games, you're going to see stuttering in games due to Proton using XWayland by default. An effective workaround is to use a Proton fork such as Proton GE or Proton-cachyos, and enable the Wayland driver via this launch argument for your games:

```
PROTON_ENABLE_WAYLAND=1 %command%
```

This will [currently](https://github.com/ValveSoftware/steam-for-linux/issues/8020) break Steam's overlay, but it does resolve all stuttering when streaming via KWin (at least on my system).

This workaround doesn't seem to be necessary on GNOME/mutter, but it may be worth trying to yield improved latency and/or performance.

psyke83:
@ReenigneArcher 

The commit in this PR that fixes pipewire teardown logic is now mandatory for PipeWire 1.6 compatibility. I've been testing and feel satisfied that this whole PR is ready to be being merged as-is (if nobody objects), but I can split out the specific fix and open a separate PR if you prefer. 

I think we should get this fixed ASAP since it's causing issues with the Vulkan PR; the user reporting green screen with vulkan encoding is actually hitting the teardown logic bug that's now exacerbated by pipewire 1.6.

ReenigneArcher:
@psyke83 I'm good with having that fix in this PR. Could you rebase the PR onto master? GitHub UI is broken (https://github.com/orgs/community/discussions/188567)

PR #4771, opened 2026-02-24
feat(api/ui): add client enable/disable access control

neatnoise:
## Description

Add ability to enable/disable paired clients via the web UI and API. Disabled clients are rejected at the TLS verification level, preventing them from connecting.

Changes:
- Add `enabled` boolean field to `named_cert_t` struct, persisted in client credentials file
- Add `POST /api/clients/update` endpoint accepting `{ "uuid": "<uuid>", "enabled": true/false }`
- Add `set_client_enabled()` and `is_client_enabled()` functions in nvhttp
- Reject disabled clients during TLS certificate verification
- Add Enable/Disable toggle buttons with status badges in the troubleshooting page client list

This provides basic access control for paired clients without requiring a full unpair/re-pair cycle. Existing clients default to `enabled = true` for backward compatibility. When the client is disabled the visual behaviour of the Moonlight client is that the PC looks like turned off. 

### Screenshot

Screenshot 2026-02-24 at 15 35 14

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -1759,6 +1759,7 @@ namespace confighttp {
     server.resource["^/api/clients/list$"]["GET"] = getClients;
     server.resource["^/api/clients/unpair$"]["POST"] = unpair;
     server.resource["^/api/clients/unpair-all$"]["POST"] = unpairAll;
+    server.resource["^/api/clients/update$"]["POST"] = updateClient;
     server.resource["^/api/config$"]["GET"] = getConfig;
     server.resource["^/api/config$"]["POST"] = saveConfig;
     server.resource["^/api/configLocale$"]["GET"] = getLocale;
@@ -1772,15 +1773,6 @@ namespace confighttp {
     server.resource["^/api/restart$"]["POST"] = restart;
     server.resource["^/api/vigembus/status$"]["GET"] = getViGEmBusStatus;
     server.resource["^/api/vigembus/install$"]["POST"] = installViGEmBus;
-    server.resource["^/api/password$"]["POST"] = savePassword;
-    server.resource["^/api/apps/([0-9]+)$"]["DELETE"] = deleteApp;
-    server.resource["^/api/clients/unpair-all$"]["POST"] = unpairAll;
-    server.resource["^/api/clients/list$"]["GET"] = getClients;
-    server.resource["^/api/clients/update$"]["POST"] = updateClient;
-    server.resource["^/api/clients/unpair$"]["POST"] = unpair;
-    server.resource["^/api/apps/close$"]["POST"] = closeApp;
-    server.resource["^/api/covers/upload$"]["POST"] = uploadCover;
-    server.resource["^/api/covers/([0-9]+)$"]["GET"] = getCover;
 
     // static/dynamic resources
     server.resource["^/images/sunshine.ico$"]["GET"] = getFaviconImage;
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -879,6 +879,11 @@ namespace confighttp {
       std::string uuid = input_tree.value("uuid", "");
       bool enabled = input_tree.value("enabled", true);
       output_tree["status"] = nvhttp::set_client_enabled(uuid, enabled);
+
+      if (!enabled && output_tree["status"] && !nvhttp::has_enabled_clients()) {
+        proc::proc.terminate();
+      }
+
       send_response(response, output_tree);
     } catch (nlohmann::json::exception &e) {
       BOOST_LOG(warning) << "Update Client: "sv << e.what();
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -847,7 +847,6 @@ namespace confighttp {
    * @brief Enable or disable a client.
    * @param response The HTTP response object.
    * @param request The HTTP request object.
-   *
    * The body for the POST request should be JSON serialized in the following format:
    * @code{.json}
    * {
@@ -855,9 +854,18 @@ namespace confighttp {
    *   "enabled": true
    * }
    * @endcode
+   *
+   * @api_examples{/api/clients/update| POST| {"uuid":"<uuid>","enabled":true}}
    */
   void updateClient(resp_https_t response, req_https_t request) {
-    if (!check_content_type(response, request, "application/json") || !authenticate(response, request)) {
+    if (!check_content_type(response, request, "application/json")) {
+      return;
+    }
+    if (!authenticate(response, request)) {
+      return;
+    }
+    std::string client_id = get_client_id(request);
+    if (!validate_csrf_token(response, request, client_id)) {
       return;
     }
 
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -872,7 +872,7 @@ namespace confighttp {
       bool enabled = input_tree.value("enabled", true);
       output_tree["status"] = nvhttp::set_client_enabled(uuid, enabled);
       send_response(response, output_tree);
-    } catch (std::exception &e) {
+    } catch (nlohmann::json::exception &e) {
       BOOST_LOG(warning) << "Update Client: "sv << e.what();
       bad_request(response, request, e.what());
     }
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -41,6 +41,7 @@
 #include "nvhttp.h"
 #include "platform/common.h"
 #include "process.h"
+#include "rtsp.h"
 #include "utility.h"
 #include "uuid.h"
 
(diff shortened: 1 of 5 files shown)


ReenigneArcher:
@netnoise Could you rebase this?

src_assets/common/assets/web/troubleshooting.html
```diff
@@ -135,8 +135,16 @@ <h2 id="unpair" class="mb-0">{{ $t('troubleshooting.unpair_title') }}</h2>
...
-          <button class="btn btn-danger ms-auto" @click="unpairSingle(client.uuid)">
+          <div class="flex-grow-1">
+            {{ client.name !== "" ? client.name : $t('troubleshooting.unpair_single_unknown') }}
+            <span class="badge ms-2" :class="client.enabled ? 'bg-success' : 'bg-secondary'">
+              {{ client.enabled ? 'Enabled' : 'Disabled' }}
+            </span>
+          </div>
+          <button class="btn btn-sm ms-2" :class="client.enabled ? 'btn-warning' : 'btn-success'" @click="toggleClient(client.uuid, !client.enabled)">
+            {{ client.enabled ? 'Disable' : 'Enable' }}
+          </button>
```
ReenigneArcher: Instead of a button with text, how about a toggle?

Option 1: https://getbootstrap.com/docs/5.0/forms/checks-radios/#switches and keep badges
Option 2: https://www.npmjs.com/package/bootstrap5-toggle and remove badges

src/confighttp.cpp
```diff
@@ -833,6 +833,41 @@ namespace confighttp {
...
+   * {
+   *   "uuid": "<uuid>",
+   *   "enabled": true
+   * }
+   * @endcode
+   */
+  void updateClient(resp_https_t response, req_https_t request) {
+    if (!check_content_type(response, request, "application/json") || !authenticate(response, request)) {
+      return;
+    }
```
ReenigneArcher:
```suggestion
    if (!check_content_type(response, request, "application/json")) {
      return;
    }
    if (!authenticate(response, request)) {
      return;
    }

    std::string client_id = get_client_id(request);
    if (!validate_csrf_token(response, request, client_id)) {
      return;
    }
```

There is a new pattern since your PR was created.

src/confighttp.cpp
```diff
@@ -833,6 +833,41 @@ namespace confighttp {
     send_response(response, output_tree);
   }
 
+  /**
+   * @brief Enable or disable a client.
+   * @param response The HTTP response object.
+   * @param request The HTTP request object.
+   *
```
ReenigneArcher:
```suggestion
```

src/confighttp.cpp
```diff
@@ -833,6 +833,41 @@ namespace confighttp {
...
+   * The body for the POST request should be JSON serialized in the following format:
+   * @code{.json}
+   * {
+   *   "uuid": "<uuid>",
+   *   "enabled": true
+   * }
+   * @endcode
+   */
```
ReenigneArcher, line 860: Add `@api_examples`, which will auto generate code in our api docs.

src_assets/common/assets/web/troubleshooting.html
```diff
@@ -135,8 +135,16 @@ <h2 id="unpair" class="mb-0">{{ $t('troubleshooting.unpair_title') }}</h2>
...
-          <div class="flex-grow-1">{{ client.name !== "" ? client.name : $t('troubleshooting.unpair_single_unknown') }}</div>
-          <button class="btn btn-danger ms-auto" @click="unpairSingle(client.uuid)">
+          <div class="flex-grow-1">
+            {{ client.name !== "" ? client.name : $t('troubleshooting.unpair_single_unknown') }}
+            <span class="badge ms-2" :class="client.enabled ? 'bg-success' : 'bg-secondary'">
+              {{ client.enabled ? 'Enabled' : 'Disabled' }}
+            </span>
+          </div>
+          <button class="btn btn-sm ms-2" :class="client.enabled ? 'btn-warning' : 'btn-success'" @click="toggleClient(client.uuid, !client.enabled)">
+            {{ client.enabled ? 'Disable' : 'Enable' }}
```
ReenigneArcher: Strings need to be in `en.json` so they can be localized.

ReenigneArcher (changes requested):
Thanks for the PR!

In addition to the review below, could you also update the api.md doc?

src_assets/common/assets/web/troubleshooting.html
neatnoise: Replaced with a toggle

src/confighttp.cpp
neatnoise: Changed pattern

src/confighttp.cpp
neatnoise: Removed

src/confighttp.cpp
neatnoise: Added

src_assets/common/assets/web/troubleshooting.html
neatnoise: It's a toggle now, no need

neatnoise:
@ReenigneArcher Added changes to api.md, other suggestions, rebased the branch

src/confighttp.cpp
```diff
@@ -1729,6 +1772,15 @@ namespace confighttp {
...
     server.resource["^/api/vigembus/install$"]["POST"] = installViGEmBus;
+    server.resource["^/api/password$"]["POST"] = savePassword;
+    server.resource["^/api/apps/([0-9]+)$"]["DELETE"] = deleteApp;
+    server.resource["^/api/clients/unpair-all$"]["POST"] = unpairAll;
+    server.resource["^/api/clients/list$"]["GET"] = getClients;
+    server.resource["^/api/clients/update$"]["POST"] = updateClient;
+    server.resource["^/api/clients/unpair$"]["POST"] = unpair;
+    server.resource["^/api/apps/close$"]["POST"] = closeApp;
+    server.resource["^/api/covers/upload$"]["POST"] = uploadCover;
+    server.resource["^/api/covers/([0-9]+)$"]["GET"] = getCover;
```
ReenigneArcher: Something's wrong here. Looks like these got duplicated in the rebase?

ReenigneArcher (commented):
Thanks for the updates! Seems pretty good now, just one issue I noticed below. If you could fix that as well as the sonar warning (sorry to be a stickler about that, just trying to slowly get the code base to a nice state).

src/confighttp.cpp
neatnoise: fixed

ReenigneArcher (dismissed):

ReenigneArcher:
There are two small issues.

1. If you disable a client while it is paused, the tray icon remains in that state. We should only clear the icon state if it's the only connected client. May want to reference https://github.com/LizardByte/Sunshine/pull/4890
2. If you disable a client while it is streaming, it does not disconnect that client.

neatnoise:
> There are two small issues.
> 
> 1. If you disable a client while it is paused, the tray icon remains in that state. We should only clear the icon state if it's the only connected client. May want to reference [fix: clear tray icon when unpairing the last paired client #4890](https://github.com/LizardByte/Sunshine/pull/4890)
> 2. If you disable a client while it is streaming, it does not disconnect that client.

1. I changed similarly as it is in unpairing.
2. I added disconnect for all clients when disabled the client. There is no straightforward way to disconnect a particular client currently.

ReenigneArcher:
> > There are two small issues.
> > 
> > 1. If you disable a client while it is paused, the tray icon remains in that state. We should only clear the icon state if it's the only connected client. May want to reference [fix: clear tray icon when unpairing the last paired client #4890](https://github.com/LizardByte/Sunshine/pull/4890)
> > 2. If you disable a client while it is streaming, it does not disconnect that client.
> 
>     1. I changed similarly as it is in unpairing.
> 
>     2. I added disconnect for all clients when disabled the client. There is no straightforward way to disconnect a particular client currently.

I don't think the fix addresses the issues. They still behave exactly the same.

You are checking `has_enabled_clients`, but you need to check for connected clients instead.

To clarify, I performed those tests with 2 clients paired. I only disabled 1 client while streaming and while streaming was paused.

neatnoise:
@ReenigneArcher can you check now?

ReenigneArcher:
> @ReenigneArcher can you check now?

That works!

I noticed one other edge case. I don't know if there's a simple way to fix it.

A disabled client will show an unpaired state in Moonlight. If you try to pair the client again, it appears to succeed on Sunshine's side, adds a new entry and everything, but there are errors on the Moonlight side.

Then in Sunshine's side, the original client is still shown as disabled, and the newly added one is shown as enabled.

I guess the real problem is that Moonlight does not have a way to detect if a client is enabled/disabled.

neatnoise:
> > @ReenigneArcher can you check now?
> 
> That works!
> 
> I noticed one other edge case. I don't know if there's a simple way to fix it.
> 
> A disabled client will show an unpaired state in Moonlight. If you try to pair the client again, it appears to succeed on Sunshine's side, adds a new entry and everything, but there are errors on the Moonlight side.
> 
> Then in Sunshine's side, the original client is still shown as disabled, and the newly added one is shown as enabled.
> 
> I guess the real problem is that Moonlight does not have a way to detect if a client is enabled/disabled.

This would require to change the protocol unfortunately. The user should be aware when enabling/disabling though.

PR #4779, opened 2026-02-26
fix(linux): use FQDN naming for all Linux packaging types

psyke83:
Normalize install files to use the FQDN naming scheme for all package types (native/AppImage/Flatpak) to allow XDG compliance.

* Use FQDN naming convention for icons.
* Use "app-FQDN" naming convention for unprivileged systemd service; fixes XDG Desktop Portal app_id detection.
* Restore privileged "sunshine-kms" service unit back to "sunshine" and restore as the default in documentation and via preset unit. The unprivileged service only needs XDG compliance.

## Type of Change

## Checklist

### AI Usage


psyke83:
These changes are enough to fix XDG detection when streaming via Portal grab, but if we want more complete compliance, the real binary should probably be named "PROJECT_FQDN" as well. Should we change it or leave as-is? Detection works fine on Fedora 43 with the native package and these current changes.

ReenigneArcher:
> These changes are enough to fix XDG detection when streaming via Portal grab, but if we want more complete compliance, the real binary should probably be named "PROJECT_FQDN" as well. Should we change it or leave as-is? Detection works fine on Fedora 43 with the native package and these current changes.

I think we should change it. Maybe for this one we can create an alias in the post install scripts?

packaging/linux/AppImage/AppRun
```diff
@@ -57,7 +57,7 @@ function install() {
   mkdir -p ~/.config/systemd/user
   cp -r "$SUNSHINE_SHARE_HERE/systemd/user/" ~/.config/systemd/
   # patch service executable path
-  sed -i -e "s#$SUNSHINE_PATH#$(readlink -f $ARGV0)#g" ~/.config/systemd/user/sunshine.service
+  sed -i -e "s#$SUNSHINE_PATH#$(readlink -f $ARGV0)#g" ~/.config/systemd/user/app-dev.lizardbyte.app.Sunshine.service
```
ReenigneArcher, line 60: I don't think KMS capture will work in AppImage by default, because we can't do the setcap. Will there be problems by using the "privileged" service?

packaging/linux/flatpak/scripts/additional-install.sh
```diff
@@ -2,9 +2,9 @@
 
 # User Service
 mkdir -p ~/.config/systemd/user
-cp "/app/share/sunshine/systemd/user/sunshine.service" "$HOME/.config/systemd/user/sunshine.service"
+cp "/app/share/sunshine/systemd/user/app-dev.lizardbyte.app.Sunshine.service" "$HOME/.config/systemd/user/app-dev.lizardbyte.app.Sunshine.service"
```
ReenigneArcher, line 5: Same question for Flatpak.

psyke83:
> > These changes are enough to fix XDG detection when streaming via Portal grab, but if we want more complete compliance, the real binary should probably be named "PROJECT_FQDN" as well. Should we change it or leave as-is? Detection works fine on Fedora 43 with the native package and these current changes.
> 
> I think we should change it. Maybe for this one we can create an alias in the post install scripts?

I'm not sure what kind of alias you mean? XDG detects the app_id using multiple matching strategies, including the name of the executable via /proc/$(pidof sunshine)/exe. This would point to the real file and not a symlink. But it's already matching correctly without this change.

packaging/linux/AppImage/AppRun
psyke83: That should be the unprivileged service, though? I left the setcap invocation on the following line but it can technically be removed as well. Or does kms work with appimages? If yes, we need to install both service units.

packaging/linux/flatpak/scripts/additional-install.sh
psyke83: I can understand the confusion here because GitHub marked the service renames incorrectly. I'm sure that I performed "git mv" of the correct files, but I'll verify when I'm back at my PC. Regardless of what was moved, the end content is currently correct though.

packaging/linux/AppImage/AppRun
ReenigneArcher: > Or does kms work with appimages? If yes, we need to install both service units.

It doesn't work, unless the user extracts the AppImage. It's been a pain point of the AppImage since it was creation.

I think some users may work around it by just running the AppImage with sudo.

ReenigneArcher:
> I'm not sure what kind of alias you mean?

Disregard, for I was mistakenly thinking this PR was changing the binary name.

packaging/linux/flatpak/scripts/additional-install.sh
psyke83: I've researched git's behaviour and it seems that I did nothing wrong in terms of renaming the files. The similarity index of both files changed because I also had to make changes to the Conflicts, and the operation ended up making it seem that the wrong files were renamed. The only way to "fix" that would be to do a vanilla rename in one commit, then commit changes to the renamed files in a separate commit -- but once you squash all the commits to master, it would end up doing the same thing. I'll resolve this thread.

psyke83:
> > I'm not sure what kind of alias you mean?
> 
> Disregard, for I was mistakenly thinking this PR was changing the binary name.

So we'll leave the binary name as-is? I think it's ok. Are you comfortable with the hardcoding of the FQDN in scripts, though? It would cause issues for forks that change to their own FQDN. Should I modify the cmake scripts to replace the strings in these files in the same way that the service files are processed?

ReenigneArcher:
> Are you comfortable with the hardcoding of the FQDN in scripts

Hardcoding is fine.

The configure files was not really meant to help forks anyway. It was more to have less places when an adjustment was made.

ReenigneArcher:
Sorry for the delay on this PR. There was a bunch of build failures that needed to be addressed, but finally caught up now.

> But if we want more complete compliance, the real binary should probably be named "PROJECT_FQDN" as well.

Back to this original comment, I guess there was some confusion with the back and forth. I think renaming the binary is a good idea if it makes the project more compliant with standard practices.

What I meant by creating an alias is something like `alias sunshine='dev.lizardbyte.app.Sunshine'` in the post install scripts. Maybe this is not necessary though?

This could always be done later if we don't want to do it now though.

NOTE: If you do push changes to this PR, I had pushed a fix for the new sonar issues, so please keep that.

ReenigneArcher:
I will merge and we can decide about the binary name later.

PR #4780, opened 2026-02-26
feat(linux/xdgportal): use accurate frame timestamps from PipeWire metadata

psyke83:
## Description

feat(linux/xdgportal): use accurate frame timestamps from PipeWire metadata

On new frame:
* Sample Sunshine's steady_clock and PipeWire clock.
* Measure frame age using difference between PW clock "now" and pts.
* Sanity check age (limit of 100ms or 5 frames in past, 2ms in future).
  This is needed to filter large time differences in the last received pts
  when the desktop goes idle, clock is reset, etc.
* If age check passes, use hardware timestamp, otherwise use steady clock.

(Note: this PR temporarily includes the event driven capture changes of #4768 as the are a prerequisite).

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -752,6 +752,10 @@ namespace portal {
       return stream_data.frame_cv;
     }
 
+    uint64_t stream_get_nsec() {
+      return pw_stream_get_nsec(stream_data.stream);
+    }
+
     bool is_frame_ready() const {
       return stream_data.frame_ready;
     }
@@ -861,7 +865,6 @@ namespace portal {
 
         if (buf->datas[0].chunk->size != 0) {
           const auto img_descriptor = static_cast<egl::img_descriptor_t *>(img);
-          img_descriptor->frame_timestamp = std::chrono::steady_clock::now();
 
           // PipeWire header metadata
           struct spa_meta_header *h = static_cast<struct spa_meta_header *>(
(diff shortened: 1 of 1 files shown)


psyke83:
Pinging @andygrundman in case you want to try it out. 

This PR temporarily includes the same commits in #4768 as a prerequisite.

When in-game, the real timestamps are being processed consecutively without any noticeable breaks, and generally they look a bit more inflated than the current (less accurate) steady_clock::now() sampling. Unless I made a mistake, however, these larger values should be more reflective of the actual compositor overhead rather than simply sampling the time when the buffer is received on Sunshine's side. When the desktop is idle, it falls back to the inaccurate sampling due to the "pw now - pts" delta being too large, but reverts back to the hardware timestamps when fresh data is sampled that meets the age check thresholds.

Unfortunately, this current logic doesn't seem to work well on mutter, as it reports very low min/avg/max host latency values (sub 0.3ms, for ex), so it either needs work, or perhaps pw timestamps are unworkable due to compositor variance in pts calculations.

src/platform/linux/portalgrab.cpp
```diff
@@ -1342,6 +1377,92 @@ namespace portal {
...
+
+      return false;
+    }
+
+    void update_frame_metadata(egl::img_descriptor_t *img_egl, int retries) {
+      if (img_egl->seq.has_value() && img_egl->pts.has_value()) {
+        last_seq = img_egl->seq.value();
+        last_pts = img_egl->pts.value();
+
+        // Snapshot both clock domains simultaneously
```
andygrundman, line 1414: What is all this time code doing? If the frame is a duplicate, frame_timestamp should be left null. It gets filled in by code around stream.cpp:1471.

src/platform/linux/portalgrab.cpp
```diff
@@ -1342,6 +1377,92 @@ namespace portal {
...
+        // Anchor the timestamp
+        if (age_ns >= future_tolerance_ns && age_ns < max_age_ns) {
+          // FRESH FRAME: Apply hardware-measured age to the steady_clock anchor
+          img_egl->frame_timestamp = steady_now - std::chrono::nanoseconds(age_ns);
+        } else {
+          // STALE/FUTURE: Fallback to current time
+          img_egl->frame_timestamp = steady_now;
+
+          if (age_ns < future_tolerance_ns) {
+            BOOST_LOG(debug) << "Frame age was future dated: " << (age_ns / 1000) << "us";
```
andygrundman, line 1438: I know this code will be removed but just FYI, use verbose log level for anything that logs very frequently or in a loop, etc.

src/platform/linux/portalgrab.cpp
psyke83: update_frame_metadata() is only called on frames that pass a duplicate and snapshot timeout check. Previously, portalgrab's snapshot timeout was a no-op, but reactive capture (and the updates in this PR) make it actually block for new data, so it was only timing out for extreme cases where no data was being received.

The timeout is 1000ms, so only frames that are extremely late will hit the timeout.

src/platform/linux/portalgrab.cpp
psyke83: Good advice, but I suspect any logging level might be too chatty to be included in a final PR state. Right now, the change is not appropriate as there is just way too much compositor variance to be confident that this is working as expected, so either it needs some changes to the logic, or might just be better to drop it entirely. More eyes on the code might help figure out if the former is possible.

In the case of KWin (6.6.1), stale PTS data explodes the host latency when returning from idle - that necessitates the positive (past) age check. In the case of mutter, I don't think that frames returning from idle have stale pts timestamps, but it has the opposite issue where timestamps end up consistently future dated sometimes by as high as -2ms for an active stream running at 60fps.

andygrundman: OK, if we can't trust pipewire it's probably fine to just use now().

neatnoise:
Reported by the client host latency count increased for me from ~2-3ms to like ~40-50ms. The stream feels great, stable and very low latency though. Is it expected?

psyke83:
I think there's too much variance in pts metadata provided by compositors to make this usable right now, so we'll just stick to to the current timestamps sampled on buffer arrival instead.

PR #4839, opened 2026-03-11
fix(linux/xdgportal): avoid duplicate frame insertion

psyke83:
## Description

Portal capture is event driven/blocking, so duplicate frames can compete
with pacing due to irregular buffer arrival.

Resolve by ensuring that event-driven capture methods default to 0.5fps/
2000ms instead of half the host framerate. This will not impact other
capture methods, but if future event-driven capture methods are added, they
can also utilise the is_event_driven() override.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/common.h
+++ b/src/platform/common.h
@@ -534,6 +534,10 @@ namespace platf {
       return true;
     }
 
+    virtual bool is_event_driven() {
+      return false;
+    }
+
     virtual ~display_t() = default;
 
     // Offsets for when streaming a specific monitor. By default, they are 0.
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -1365,6 +1365,11 @@ namespace portal {
       return 0;
     }
 
+    // This capture method is event driven; don't insert duplicate frames
+    bool is_event_driven() override {
+      return true;
+    }
+
   private:
     bool is_buffer_redundant(const egl::img_descriptor_t *img) {
       // Check for corrupted frame
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -1972,8 +1972,9 @@ namespace video {
       }
     });
 
-    // set max frame time based on client-requested target framerate.
-    double minimum_fps_target = (config::video.minimum_fps_target > 0.0) ? config::video.minimum_fps_target : config.framerate;
+    // set max frame time based on client-requested target framerate (or 0.5fps/2000ms for event-driven capture)
+    double def_fps_target = (disp->is_event_driven() ? 1 : config.framerate);
+    double minimum_fps_target = (config::video.minimum_fps_target > 0.0) ? config::video.minimum_fps_target : def_fps_target;
     std::chrono::duration<double, std::milli> max_frametime {1000.0 / minimum_fps_target};
     BOOST_LOG(info) << "Minimum FPS target set to ~"sv << (minimum_fps_target / 2) << "fps ("sv << max_frametime.count() * 2 << "ms)"sv;
 


psyke83:
Without this PR, I need to set `min_fps_target = 1` to avoid random small FPS bursts that causes intermittent stuttering, which in certain circumstances can degrade to runaway bursting (constant 90fps on a 60fps stream while in a fullscreen game that can only be temporarily fixed by toggling vsync, alt-tabbing or reconnecting the stream). With the PR, the min_fps_target defaults to 1 for event-driven capture methods (currently only portalgrab)  so no user intervention will be necessary, but custom config will be respected.

Pinging @andygrundman - let me know if you object to this change (as I believe you last touched the fake frame logic). 
Also pinging @shelterx due to your comment in the vulkan PR. If you see stuttering with VAAPI + portalgrab, then you can try this PR and see if it helps (this PR will also work with Vulkan encoding when the other PR is merged).

shelterx:
I tried various things KMS/XDG/Vaapi, disabling VRR in KDE, setting monitor to 60Hz in KDE, enabling vsync in game but there's always this slight hitching every 2 second or so.

Fwiw, Apollo (the sunshine fork) does not have this issue.

psyke83:
Just a note: the original version of this PR was disabling fake frame insertion completely, but I didn't realize that images->pop() could timeout indefinitely when a reinit was called (which happens on disconnect or mode change), so I changed it so that portalgrab defaults to the lowest setting when no user config is present.

> I tried various things KMS/XDG/Vaapi, disabling VRR in KDE, setting monitor to 60Hz in KDE, enabling vsync in game but there's always this slight hitching every 2 second or so.
> 
> Fwiw, Apollo (the sunshine fork) does not have this issue.

I assumed from your comment it was an issue exclusively with portal capture. If you're seeing the issue across different capture methods and encoders, this PR is not relevant to you.

ReenigneArcher:
apollo also does not have xdg portal capture

shelterx:
@psyke83 yeah well no worries and sorry for being vague. For the moment it doesnt hitch anything at all. Pipewire got updated along with some other system components. Wish I knew what "fixed" it, I'll keep you updated if it comes back.

andygrundman:
Let's not use the term "fake" for these frames, they are duplicate frames. I think we both want the same thing, which is to change the default, but I haven't tested such a low value to see how clients react to it. Do you have problems if you use 20, or even 10? I wonder if your 90fps issue might be something else. I have never seen Sunshine send more than the requested fps on Windows.

psyke83:
> Let's not use the term "fake" for these frames, they are duplicate frames. I think we both want the same thing, which is to change the default, but I haven't tested such a low value to see how clients react to it. Do you have problems if you use 20, or even 10? I wonder if your 90fps issue might be something else. I have never seen Sunshine send more than the requested fps on Windows.

I'll change the naming to duplicate frames on the next refresh. The 90fps (and more frequent random bursts of ~63fps) is definitely related to this setting, as I no longer have any issues when it's set to 1. I spent a lot of time trying to troubleshoot the issue via changes to portalgrab's code (in vain) before I noticed the generic duplicate insertion logic, and changing this setting immediately fixed the issue. The smoking gun that it's responsible is that the issues only started after snapshot was changed to a blocking call, as the original portalgrab was happily grabbing duplicate buffers and sending frames (some of which were recaptured duplicates) to the encoder at predictable intervals just like kmsgrab, etc.

This PR is not aimed at changing the default value for any other capture method besides portalgrab; I never really noticed any issue with kmsgrab or Windows related to this, so I can't judge as to whether changing the overall default to something else is desirable.

I do believe that for portalgrab, the setting should be as low as possible (effectively disabled), as the timing of snapshot is dictated by on_process callbacks whenever pipewire sends buffers. This behaviour varies by compositor, and can't be compared to a traditional snapshot function that's capture driven with a predictable capture timing for each frame (like kmsgrab or dxgi), as the buffers can arrive in bursts or at irregular intervals which causes extra duplicate frames to be randomly inserted and violate the cap.

If I understand correctly, the original purpose of duplicate frame insertion was to ameliorate input latency or visual lag caused by encoder framerate ramping in Windows, but with this PR (and a RX 6600, testing against libx264, VAAPI and Vulkan), the desktop is always fluid when ramping from idle and I see no input lag when typing. The PR only changes the default and user config is still respected, if needed.

PR #3061, opened 2024-08-22
feat(win/nvenc): add split frame encoding option

ns6089:
## Description

Support [split frame encoding](https://developer.nvidia.com/blog/video-encoding-at-8k60-with-split-frame-encoding-and-nvidia-ada-lovelace-architecture/)
- Depends on #3040

### Screenshot

2024-08-22 14_53_32-Window

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

PR #4892, opened 2026-03-23
feat(nvenc): support for split frame encoding on GPUs with 2+ nvenc blocks

andygrundman:
## Description
This PR adds new option `split_frame_encoding` with options `disabled`, `enabled`, and `driver_decides`. The majority of the changes are from #3061 .

When `NV_ENC_CAPS_NUM_ENCODER_ENGINES` > 1, these options set `init_params.splitEncodeMode` to `NV_ENC_SPLIT_DISABLE_MODE`, `NV_ENC_SPLIT_AUTO_FORCED_MODE`, or `NV_ENC_SPLIT_AUTO_MODE` respectively.

The encoder options string in the log will show `sfe` or `sfe-auto` when nvenc count > 1 and split_frame_encoding is not set to disabled.

This also bumps build-deps in order to pull in ffmpeg 8.1 and nvenc SDK 13.0. (ffmpeg is not used when encoding with nvenc but is needed here due to the fact that ffmpeg owns `third-party/nv-codec-headers/include/ffnvcodec/nvEncodeAPI.h` by including it in the prebuilt ffmpeg binary that comes from build-deps.)

cc @ns6089 @Nonary

## Type of Change

## Checklist

### AI Usage


ns6089:
Currently sunshine supports every nvidia gpu with nvenc ever made, and dropping runtime support for nvenc-11.0 (or more specifically nvenc-11.1) will cut off the ones for which the drivers stopped being updated years ago. Personally I'm indifferent about how sunshine will want to handle this, but you will get bug reports.

andygrundman:
> Currently sunshine supports every nvidia gpu with nvenc ever made, and dropping runtime support for nvenc-11.0 (or more specifically nvenc-11.1) will cut off the ones for which the drivers stopped being updated years ago. Personally I'm indifferent about how sunshine will want to handle this, but you will get bug reports.

You are probably right, but what cards are we talking about here? I have a GTX 1070 which has been EOL'ed recently and runs driver version 580 with SDK 13.0 (this is in Linux). NVENC goes back to the 600 series, so the issue is cards from 600-900? Wikipedia says NVENC 1st gen was used on all those cards with Maxwell 1000 series getting 2nd gen with HEVC. Do we support cards that old?

I don't want to break anything, and I admit I did take the easy route here. But I'm also not worried about supporting people who choose to run older drivers (in this case they would have to be MUCH older drivers since 13.0 shipped in Jan 2025). It's sometimes smart to avoid the current Nvidia driver, but choosing to be a year out of date, probably on Windows 10, I find it hard to understand the mindset. Maybe these users are ok running an older Sunshine too.

ns6089:
> NVENC goes back to the 600 series, so the issue is cards from 600-900? Wikipedia says NVENC 1st gen was used on all those cards with Maxwell 1000 series getting 2nd gen with HEVC. Do we support cards that old?

Yes, current sunshine does support them, on Windows at least. And you're going to have the same dilemma for every future version of nvenc sdk, unless dynamic sdk version selection is implemented. But to repeat myself, both paths are somewhat viable (even though one is superior) and I don't have a strong opinion regarding which one sunshine ends up going for.

andygrundman:
> Yes, current sunshine does support them, on Windows at least. And you're going to have the same dilemma for every future version of nvenc sdk, unless dynamic sdk version selection is implemented. But to repeat myself, both paths are somewhat viable (even though one is superior) and I don't have a strong opinion regarding which one sunshine ends up going for.

I am coming around to this concern a bit more after thinking about it, because a bug report that someone is stuck with software encoding without a good workaround really sucks.

Do you think the approach you used in #3061 is still applicable? I notice you removed the `min_struct_version()`, I assume this is the older way that back-compat was handled? How well did that work? What led you to need to replace it for 12.1?

I worry that supporting so many versions makes it difficult to test properly. Do you have any tips on testing older versions? Thanks for all your work on this issue.

docs/configuration.md
```diff
@@ -2347,6 +2347,46 @@ editing the `conf` file in a text editor. Use the examples as reference.
...
+### nvenc_split_encode
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            Split the encoding of each video frame over multiple NVENC hardware units.
+            Significantly reduces encoding latency with a marginal compression efficiency penalty.
+            This option is ignored if your GPU has a singular NVENC unit.
+            @note{This option only applies when using NVENC [encoder](#encoderhttpslocalhost47990configencoder) with HEVC or AV1.}
```
ReenigneArcher:
```suggestion
            @note{This option only applies when using NVENC [encoder](#encoder) with HEVC or AV1.}
```

This should work.

andygrundman:
> @ns6089 Would supporting all these versions also mean that FFmpeg also needs to be build with all these versions simultaneously?

Can either of you can give me a quick history lesson around switching from ffmpeg's *_nvenc encoders to the native nvenc encoder on Windows? I was curious if you could use the ffmpeg hevc_nvenc encoder and it does still work. If I had to guess, was the reason the support for RFI? That doesn't seem possible through ffmpeg's API.

To your question about ffmpeg, I think that on Windows, ffmpeg never calls nvenc functions so it shouldn't matter if it was compiled with a different version. On Linux though, we do use ffmpeg nvenc and it would be a problem. This is probably why the SFE PR is Windows-only?

ReenigneArcher:
> > @ns6089 Would supporting all these versions also mean that FFmpeg also needs to be build with all these versions simultaneously?
> 
> Can either of you can give me a quick history lesson around switching from ffmpeg's *_nvenc encoders to the native nvenc encoder on Windows? I was curious if you could use the ffmpeg hevc_nvenc encoder and it does still work. If I had to guess, was the reason the support for RFI? That doesn't seem possible through ffmpeg's API.
> 
> To your question about ffmpeg, I think that on Windows, ffmpeg never calls nvenc functions so it shouldn't matter if it was compiled with a different version. On Linux though, we do use ffmpeg nvenc and it would be a problem. This is probably why the SFE PR is Windows-only?

I am not too sure, but there must be a reason why this was added.

```cpp
#if NVENCAPI_VERSION != MAKE_NVENC_VER(12U, 0U)
  #error Check and update NVENC code for backwards compatibility!
#endif
```

This makes Sunshine fail to build with the new build-deps. Though not with your PR where you've adjusted `12U` to `13U`.

andygrundman:
> > > @ns6089 Would supporting all these versions also mean that FFmpeg also needs to be build with all these versions simultaneously?
> > 
> > 
> > Can either of you can give me a quick history lesson around switching from ffmpeg's *_nvenc encoders to the native nvenc encoder on Windows? I was curious if you could use the ffmpeg hevc_nvenc encoder and it does still work. If I had to guess, was the reason the support for RFI? That doesn't seem possible through ffmpeg's API.
> > To your question about ffmpeg, I think that on Windows, ffmpeg never calls nvenc functions so it shouldn't matter if it was compiled with a different version. On Linux though, we do use ffmpeg nvenc and it would be a problem. This is probably why the SFE PR is Windows-only?
> 
> I am not too sure, but there must be a reason why this was added.
> 
> ```c++
> #if NVENCAPI_VERSION != MAKE_NVENC_VER(12U, 0U)
>   #error Check and update NVENC code for backwards compatibility!
> #endif
> ```
> 
> This makes Sunshine fail to build with the new build-deps. Though not with your PR where you've adjusted `12U` to `13U`.

Right, if you try to build Sunshine with the new build-deps but without this PR, it would fail to build due to various changes in 13.0 that are fixed in the first commit of this PR. There are a few breaking changes such as inputBitDepth and outputBitDepth replacing pixelBitDepthMinus8.

ns6089:
> Can either of you can give me a quick history lesson around switching from ffmpeg's *_nvenc encoders to the native nvenc encoder on Windows? I was curious if you could use the ffmpeg hevc_nvenc encoder and it does still work. If I had to guess, was the reason the support for RFI? That doesn't seem possible through ffmpeg's API.

RFI, easy cuda interop for 10-bit 4:4:4, some other options not exposed in ffmpeg, future-proof backward compatibility and the ability to support new nvenc features quickly. ffmpeg today exposes significantly more options compared to when I was writing sunshine's nvenc implementation, but I would still consider the lack of RFI and manual LTR a deal breaker for realtime streaming software.

patryk-matis:
Does it also support 3 nvenc blocks (rtx 5090) or is it limited to 2 for now?

ReenigneArcher:
> Does it also support 3 nvenc blocks (rtx 5090) or is it limited to 2 for now?

PR title says "2+"

ReenigneArcher:
Going to move forward with this as is. If anyone wants to work on the dynamic SDK path a PR would be welcomed.

PR #4918, opened 2026-03-28
build(copr): update rpm-spec

PVermeer:
## Description
Hey @ReenigneArcher, I've been looking into COPR for stable builds and would like suggest these changes. These changes should allow for any Fedora build in COPR.  I've tested all changes in a test copr: https://copr.fedorainfracloud.org/coprs/pvermeer/sunshine-test/monitor/

Please take a look at it. It would be great to have stable builds again on COPR.

### Updated COPR spec-file so all builds from stable to latest can be build on all targets.

On Fedora and derivatives there hasn't been a stable build available for some time. The latest beta releases were causing some issues on Bazzite so I started investigating the reason and came to the conclusion that the cuda compilation is depended on the c compiler version. After inspecting this spec file I found that there is lot going on with switching compilers for targets. Fedora however does not support this very well and only provides a previous compiler and that's it.

This PR includes a different strategy for compiling Sunshine on different targets. It's using the host compiler for building and linking Sunshine except for the cuda files. The cuda files are compiled with recent and compatible compilers provided by [conda](https://anaconda.org/anaconda/repo). Conda is available on Fedora as an official package and I would recommend using this but openSuse Leap does not provide this package. Therefore conda is provided by [micromamba](https://github.com/mamba-org/micromamba-releases/) which is a lean implementation of the conda package manager.

The cuda compilers `nvcc` and compatible c compilers are now installed via [cuda-nvcc](https://anaconda.org/channels/conda-forge/packages/cuda-nvcc/overview) for all build targets. They are passed to cmake for compilation of the cuda files. This also a much leaner package then the package from nvidia.

So basically instead changing the host compilers to build Sunshine with cuda, only the cuda files are compiled with a compatible compiler and then linked with Sunshine binary. This setup increases compatibility with all COPR targets.

#### Build dependencies
- **gcc***: use the system compiler for max compat and conda nvcc for cuda. Only Leap is so old there is a need to use a modern compiler.
- **libX***: X11 is already installed on target and is/will be removed from Fedora. cMake detects it when it's still available on the target so let cmake figure it out if it's still supported.
- **nodeJs/npm**: use the target provided versions. As long as Sunshine uses the LTS for development it should be fine. 
- **python***: no idea why these are here.

#### Runtime dependencies
These are all removed.
Copr (and cmake) can perfectly fine figure out the runtime dependencies from linking. No need to specify them unless your using a dependency directly (e.g. calling git as a command) and thus is not linked. If something is missing I can add it.

### Testing
Not all tests on all targets will run successfully. I'm not sure if this can be fixed. My recommendation is to run them in a controlled environment. What I can tell these are mostly unit-tests and imo don't need te be run when building packages in COPR. I have them enabled so logs are produced in case of issues.

What I do advise to test is the generated output files and this already mostly done with validators.
 
#### Coverage
This will not work because of mixing compilers:
```sh
Version mismatch - expected 15.2 (release) (B52*) got 15.2 (experimental) (B52 )
```
I'm not sure if this can be fixed.

#### Test suites
Success on current Fedora.
Fail on older targets.
Also not sure if this can be fixed.

### -Werror
Disabled because stable builds will trigger some warnings to errors. This can be changed to a conditional to only disable on stable builds. Or fix/allow the errors on future stable builds.

### Cuda patches
Removed cuda patching, I assume these where cause build errors?

### Post script
Simplify modprobe and udev commands. No need to check for rpm-ostree and udevadm, just silently fail. Output is not displayed to the user anyways. Fedora has udevadm by default.

### Misc
- Removed unnecessary comments. Lots of comments don't add any value, it's perfectly clear whats going on.
- Removed unused macros.

### Build notes
- I tried first to build from the release tarball in Github. This didn't work because of missing submodules. This is the recommended way to get source in COPR for an offline build.
- I had some issues building with mismatching headers ([nvEncodeAPI.h](https://github.com/FFmpeg/nv-codec-headers/blob/22441b505d9d9afc1e3002290820909846c24bdc/include/ffnvcodec/nvEncodeAPI.h))  that triggerd https://github.com/LizardByte/Sunshine/blob/1ee9163664768e9f9f0fbb52ec6651b91c647798/src/nvenc/nvenc_base.cpp#L23-L25. It took me a couple hours to figure out and it turned out to be [cmake/dependencies/ffmpeg.cmake](https://github.com/LizardByte/Sunshine/blob/1ee9163664768e9f9f0fbb52ec6651b91c647798/cmake/dependencies/ffmpeg.cmake) that is calling `git` as a command. It failed and then downloaded the latest version that changed a lot of symbols. I usually delete `.git` when the source is ready and that was causing the issue. Downloading the source package from the GitHub releases would have the same issue.

### CI
I don't know your backend but I keep the packages in my test copr updated with my own tooling and a custom [script](https://github.com/PVermeer/copr_sunshine/blob/9be4ca76542c7266f3138aeaf180364bfd4e8fc5/scripts/update.sh) adapted to this spec-file and my tooling.

#### Check for forked packages
Before a major Fedora update, COPR will branch a new build chroot from rawhide and fork the package to the new version chroot if `Follow Fedora branching` is enabled. I've found that this can be detected by getting the latest successful build id and then checking the api (build/built-packages) for chroots that now have a empty array on `.packages` (e.g. no build ran for this target).
See: https://github.com/PVermeer/rpm-tools/blob/b1ee6501973a103b5749a8d995280d3dcc3ac8be/scripts/copr.sh#L56-L83

### Tested:
- [x] All COPR builds succeed (stable and beta).
- [x] Bazzite Gnome Desktop sunshine stable (AMD).
- [x] All x86_64 builds in docker.
- [ ] All aarch64 builds.
- [ ] Nvidia encoder.

I don't have ARM or nVidia hardware so this would need te be tested.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/packaging/linux/copr/Sunshine.spec
+++ b/packaging/linux/copr/Sunshine.spec
@@ -1,13 +1,10 @@
-%global build_timestamp %(date +"%Y%m%d")
-
 # use sed to replace these values
 %global build_version 0
 %global branch 0
 %global commit 0
 
+# Cross build issues
 %undefine _hardened_build
-
-# Define _metainfodir for OpenSUSE if not already defined
 %if 0%{?suse_version}
 %if !0%{?_metainfodir:1}
 %global _metainfodir %{_datadir}/metainfo
@@ -22,9 +19,11 @@ License: GPLv3-only
 URL: https://github.com/LizardByte/Sunshine
 Source0: tarball.tar.gz
 
-# Common BuildRequires
-BuildRequires: cmake >= 3.25.0
+BuildRequires: cmake
+BuildRequires: curl
 BuildRequires: desktop-file-utils
+BuildRequires: gcc
+BuildRequires: gcc-c++
 BuildRequires: git
 BuildRequires: libcap-devel
 BuildRequires: libcurl-devel
@@ -32,176 +31,87 @@ BuildRequires: libdrm-devel
 BuildRequires: libevdev-devel
 BuildRequires: libnotify-devel
 BuildRequires: libva-devel
-BuildRequires: libX11-devel
-BuildRequires: libxcb-devel
-BuildRequires: libXcursor-devel
-BuildRequires: libXfixes-devel
-BuildRequires: libXi-devel
-BuildRequires: libXinerama-devel
-BuildRequires: libXrandr-devel
-BuildRequires: libXtst-devel
+BuildRequires: nodejs
+BuildRequires: npm
 BuildRequires: openssl-devel
 BuildRequires: pipewire-devel
-BuildRequires: rpm-build
 BuildRequires: systemd-rpm-macros
-BuildRequires: wget
-BuildRequires: which
-
+# For tests ⤵
+BuildRequires: xorg-x11-server-Xvfb
 %if 0%{?fedora}
-# Fedora-specific BuildRequires
 BuildRequires: appstream
-# BuildRequires: boost-devel >= 1.86.0
+BuildRequires: libappindicator-gtk3-devel
 BuildRequires: libappstream-glib
-%if 0%{fedora} > 43
-# needed for npm from nvm
-BuildRequires: libatomic
-%endif
-BuildRequires: libayatana-appindicator3-devel
-BuildRequires: libgudev
-BuildRequires: mesa-libGL-devel
 BuildRequires: mesa-libgbm-devel
 BuildRequires: miniupnpc-devel
-%if 0%{?fedora} < 44
-BuildRequires: nodejs-npm
-%endif
 BuildRequires: numactl-devel
 BuildRequires: opus-devel
 BuildRequires: pulseaudio-libs-devel
-BuildRequires: python3-jinja2
-BuildRequires: python3-setuptools
 BuildRequires: systemd-udev
-%{?sysusers_requires_compat}
-# for unit tests
-BuildRequires: xorg-x11-server-Xvfb
 %endif
-
 %if 0%{?suse_version}
-# OpenSUSE-specific BuildRequires
 BuildRequires: AppStream
 BuildRequires: appstream-glib
+BuildRequires: gcc15
+BuildRequires: gcc15-c++
 BuildRequires: libappindicator3-devel
-BuildRequires: libgudev-1_0-devel
-BuildRequires: Mesa-libGL-devel
 BuildRequires: libgbm-devel
 BuildRequires: libminiupnpc-devel
 BuildRequires: libnuma-devel
 BuildRequires: libopus-devel
 BuildRequires: libpulse-devel
-BuildRequires: npm
-BuildRequires: python311
-BuildRequires: python311-Jinja2
-BuildRequires: python311-setuptools
+BuildRequires: Mesa-libGL-devel
+BuildRequires: systemd
 BuildRequires: udev
-# for unit tests
-BuildRequires: xvfb-run
-%endif
-
-# Conditional BuildRequires for cuda-gcc based on distribution version
-%if 0%{?fedora}
-%if 0%{?fedora} <= 41
-BuildRequires: gcc13
-BuildRequires: gcc13-c++
-%global gcc_version 13
-%global cuda_version 12.9.1
-%global cuda_build 575.57.08
-%elif 0%{?fedora} >= 42 && 0%{?fedora} <= 43
-BuildRequires: gcc14
-BuildRequires: gcc14-c++
-%global gcc_version 14
-%global cuda_version 12.9.1
-%global cuda_build 575.57.08
-%elif 0%{?fedora} >= 44
-BuildRequires: gcc15
-BuildRequires: gcc15-c++
-%global gcc_version 15
-%global cuda_version 13.1.1
-%global cuda_build 590.48.01
-%endif
-%endif
-
-%if 0%{?suse_version}
-%if 0%{?suse_version} <= 1699
-# OpenSUSE Leap 15.x
-BuildRequires: gcc14
-BuildRequires: gcc14-c++
-%global gcc_version 14
-%global cuda_version 12.9.1
-%global cuda_build 575.57.08
-%else
-# OpenSUSE Tumbleweed
-BuildRequires: gcc14
-BuildRequires: gcc14-c++
-%global gcc_version 14
-%global cuda_version 12.9.1
-%global cuda_build 575.57.08
-%endif
-%endif
-
-%global cuda_dir %{_builddir}/cuda
-
-# Common runtime requirements
-Requires: miniupnpc >= 2.2.4
-Requires: which >= 2.21
-
-%if 0%{?fedora}
-# Fedora runtime requirements
-Requires: libayatana-appindicator3 >= 0.5.3
-Requires: libcap >= 2.22
-Requires: libcurl >= 7.0
-Requires: libdrm > 2.4.97
-Requires: libevdev >= 1.5.6
-Requires: libopusenc >= 0.2.1
-Requires: libva >= 2.14.0
-Requires: libwayland-client >= 1.20.0
-Requires: libX11 >= 1.7.3.1
-Requires: numactl-libs >= 2.0.14
-Requires: openssl >= 3.0.2
-Requires: pulseaudio-libs >= 10.0
-%endif
-
-%if 0%{?suse_version}
-# OpenSUSE runtime requirements
-Requires: libappindicator3-1
-Requires: libcap2
-Requires: libcurl4
-Requires: libdrm2
-Requires: libevdev2
-Requires: libopusenc0
-Requires: libva2
-Requires: libwayland-client0
-Requires: libX11-6
-Requires: libnuma1
-Requires: libopenssl3
-Requires: libpulse0
 %endif
 
 %description
 Self-hosted game stream host for Moonlight.
 
-%prep
-# extract tarball to current directory
-mkdir -p %{_builddir}/Sunshine
-tar -xzf %{SOURCE0} -C %{_builddir}/Sunshine
+%define workdir %{_builddir}/source
+%define sourcedir %{workdir}/Sunshine
+%define bindir %{_builddir}/bin
+%define cudadir %{_builddir}/cuda-env
 
-# list directory
-ls -a %{_builddir}/Sunshine
+%prep
+mkdir -p %{bindir}
+mkdir -p %{workdir}
+mkdir -p %{sourcedir}
+export PATH=%{bindir}:$PATH
+
+# Install cuda compiler (nvcc) with micromamba (conda)
+if [ "%{_arch}" = "x86_64" ]; then
+  curl -L --fail --retry 5 --retry-delay 2 \
+  -o /tmp/micromamba.tar.bz2 \
+    https://github.com/mamba-org/micromamba-releases/releases/latest/download/micromamba-linux-64.tar.bz2
+else
+  curl -L --fail --retry 5 --retry-delay 2 \
+  -o /tmp/micromamba.tar.bz2 \
+    https://github.com/mamba-org/micromamba-releases/releases/latest/download/micromamba-linux-%{_arch}.tar.bz2
+fi
+tar -xjf /tmp/micromamba.tar.bz2 -C /tmp
+install -Dm755 /tmp/bin/micromamba %{bindir}/micromamba
+micromamba create -y -p %{cudadir} cuda-nvcc
+
+# Source
+mkdir -p %{sourcedir}
+tar -xzf %{SOURCE0} -C %{sourcedir}
 
 %build
-# exit on error
-set -e
+cd %{sourcedir}
+source /etc/os-release
 
-# Detect the architecture and Fedora version
-architecture=$(uname -m)
+export BRANCH=%{branch}
+export BUILD_VERSION=v%{build_version}
+export COMMIT=%{commit}
 
-cuda_supported_architectures=("x86_64" "aarch64")
-
-# prepare CMAKE args
 cmake_args=(
-  "-B=%{_builddir}/Sunshine/build"
+  "-B=build"
   "-G=Unix Makefiles"
   "-S=."
   "-DBUILD_DOCS=OFF"
-  "-DBUILD_WERROR=ON"
+  "-DBUILD_TESTS=ON"
+  "-DBUILD_WERROR=OFF"
   "-DCMAKE_BUILD_TYPE=Release"
   "-DCMAKE_INSTALL_PREFIX=%{_prefix}"
   "-DSUNSHINE_ASSETS_DIR=%{_datadir}/sunshine"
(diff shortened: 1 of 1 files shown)


PVermeer:
> > On Fedora and derivatives there hasn't been a stable build available for some time.
> 
> This PR won't solve that. We only create a build in our stable copr, when we create a stable release here. It's automated. And I'm not going to align Sunshine releases with Fedora releases, the same as I'm not going to align them with macOS releases or any other Linux distro.
> 
> Usually when a new Fedora version ships it requires changes to make it work with Sunshine. Often it's due to the gcc and cuda compatibility, but sometimes there are other issues because Fedora likes to ship a bleeding edge gcc version, which sometimes Sunshine is not yet able to be compiled with.

That a shame, however that is just what this PR solves. It solves the gcc cuda compatibility by **only** using the compatible cuda compiler when compiling cuda files. It's basically the same as compiling the cuda parts as a separate library. Sunshine itself is compiled and linked with the host `gcc`. I've tried other methods and this is the most stable solution.

An automated trigger for a stable build for a new target would be great though.

> > Runtime dependencies - These are all removed.
> 
> This will break rpms we pull from copr and add to GitHub releases, or downloaded manually in a PR test.

No they will not. They are exactly the same, except for dependencies that are not needed. The required dependencies will match the target. If Sunshine needs extra dependencies (that are not linked) they can be added. This ensures that copr/rpmbuild/cmake can figure out the different runtime dependencies for all targets.

> > Not all tests on all targets will run successfully.
> 
> They work correctly right now. We test every PR in copr.

As I said, they only fail in stable (and only on some test suite creation thingy, idk), which can be expected. Also imo they are not needed for package builds.

> > -Werror
> 
> Can't be disabled, because then we can no longer easily find errors in our CI builds. We like to catch problems early and get ahead of them.

Again this is only for the stable builds. I would suggest to allow in stable builds for a new target because the build is already well tested, patch it, or maintane a stable branch.

> > Coverage - This will not work because of mixing compilers:
> 
> You have to use the same version of gcov that shipped with the compiler. Though we're not using coverage from copr build yet. Maybe we will in the future.

I know, also my suggestion that this not needed for package builds.

> ### Conclusion
> I think some of what you're trying to achieve could be good to incorporate, but the way this PR goes about it is too aggressive. Maybe we can address one issue at a time in separate PRs?

No we cannot. This all depends on each other and I'm not gonna figure out what tiny steps works or not.

### Conclusion
It would be great to get stable builds on COPR for new targets from LizardByte. This PR will fix a lot of the pain points. Imo having a stable build in COPR but then not maintain it for new targets kinda defeats the purpose of having a stable COPR repo.

This PR is open for changes to make it work.
But to be honest, if you're unwilling to maintain stable for new targets, I see no point in continuing this and I will keep this externally.

ReenigneArcher:
> I had some issues building with mismatching headers ([nvEncodeAPI.h](https://github.com/FFmpeg/nv-codec-headers/blob/22441b505d9d9afc1e3002290820909846c24bdc/include/ffnvcodec/nvEncodeAPI.h)) that triggerd

Need to rebase your branch, and this should be resolved. https://github.com/LizardByte/Sunshine/pull/4899 (Note: I rebased your PR)

> No we cannot. This all depends on each other and I'm not gonna figure out what tiny steps works or not.

It does not. For example you could move the dependency simplification to one PR. CUDA changes to another, etc.

But if you're not willing to adjust things, that's fine. This PR doesn't solve the immediate issue you have though, which is that there are no stable release of Sunshine in https://copr.fedorainfracloud.org/coprs/lizardbyte/stable/ ... that will be resolved when I create the next Sunshine release, since in https://copr.fedorainfracloud.org/coprs/lizardbyte/beta/ we're building everything, including Rawhide.

Anyway, I'll run the workflows and see where the gaps are that need to be fixed before this can be considered.

packaging/linux/copr/Sunshine.spec
```diff
@@ -22,186 +19,99 @@ License: GPLv3-only
...
 BuildRequires: mesa-libgbm-devel
 BuildRequires: miniupnpc-devel
-%if 0%{?fedora} < 44
-BuildRequires: nodejs-npm
-%endif
 BuildRequires: numactl-devel
 BuildRequires: opus-devel
 BuildRequires: pulseaudio-libs-devel
-BuildRequires: python3-jinja2
-BuildRequires: python3-setuptools
```
ReenigneArcher, line 71: Restore so we don't need to use pip to install

```
  -- glad: using Python interpreter: /usr/bin/python3.14
  -- Found Python: /usr/bin/python3.14 (found version "3.14.3") found components: Interpreter
  -- glad: jinja2 or setuptools not found in /usr/bin/python3.14, installing via pip...
  -- glad: pip not found in /usr/bin/python3.14, bootstrapping via ensurepip...
```

packaging/linux/copr/Sunshine.spec
```diff
@@ -22,186 +19,99 @@ License: GPLv3-only
...
-BuildRequires: Mesa-libGL-devel
 BuildRequires: libgbm-devel
 BuildRequires: libminiupnpc-devel
 BuildRequires: libnuma-devel
 BuildRequires: libopus-devel
 BuildRequires: libpulse-devel
-BuildRequires: npm
-BuildRequires: python311
-BuildRequires: python311-Jinja2
-BuildRequires: python311-setuptools
```
ReenigneArcher, line 93: Restore, same reason.

packaging/linux/copr/Sunshine.spec
```diff
@@ -1,13 +1,10 @@
...
-
 # use sed to replace these values
 %global build_version 0
 %global branch 0
 %global commit 0
 
+# Cross build issues
 %undefine _hardened_build
-
-# Define _metainfodir for OpenSUSE if not already defined
```
ReenigneArcher, line 10: I prefer this comment over `# Cross build issues`

packaging/linux/copr/Sunshine.spec
```diff
@@ -22,186 +19,99 @@ License: GPLv3-only
...
+  -o /tmp/micromamba.tar.bz2 \
+    https://github.com/mamba-org/micromamba-releases/releases/latest/download/micromamba-linux-64.tar.bz2
+else
+  curl -L --fail --retry 5 --retry-delay 2 \
+  -o /tmp/micromamba.tar.bz2 \
+    https://github.com/mamba-org/micromamba-releases/releases/latest/download/micromamba-linux-%{_arch}.tar.bz2
+fi
+tar -xjf /tmp/micromamba.tar.bz2 -C /tmp
+install -Dm755 /tmp/bin/micromamba %{bindir}/micromamba
+micromamba create -y -p %{cudadir} cuda-nvcc
```
ReenigneArcher, line 94: Is it possible to control the cuda compiler version?

This is important because it determines which GPUs are compatible with Sunshine.

packaging/linux/copr/Sunshine.spec
```diff
@@ -22,186 +19,99 @@ License: GPLv3-only
...
-# prepare CMAKE args
 cmake_args=(
-  "-B=%{_builddir}/Sunshine/build"
+  "-B=build"
   "-G=Unix Makefiles"
   "-S=."
   "-DBUILD_DOCS=OFF"
-  "-DBUILD_WERROR=ON"
+  "-DBUILD_TESTS=ON"
+  "-DBUILD_WERROR=OFF"
```
ReenigneArcher, line 114:
```suggestion
  "-DBUILD_WERROR=ON"
```

Tests are on by default. Need WERROR ON so we can detect problems in CI, or a way to set this dynamically. e.g. For our `pulls` copr it needs to be ON, but can be OFF otherwise.

packaging/linux/copr/Sunshine.spec
```diff
@@ -22,186 +19,99 @@ License: GPLv3-only
...
 
-# list directory
-ls -a %{_builddir}/Sunshine
+# Source
+mkdir -p %{sourcedir}
+tar -xzf %{SOURCE0} -C %{sourcedir}
 
 %build
-# exit on error
-set -e
```
ReenigneArcher, line 191: Please restore, so failures are not silent.

packaging/linux/copr/Sunshine.spec
```diff
@@ -213,203 +123,58 @@ cmake_args=(
...
-
-# Metainfo
+%{_datadir}/icons/hicolor/scalable/apps/*.svg
+%{_datadir}/icons/hicolor/scalable/status/*.svg
 %{_datadir}/metainfo/*.metainfo.xml
-
-# Assets
 %{_datadir}/sunshine/**
-
-%changelog
```
ReenigneArcher, line 415: Please restore the comments and the changelog placeholder (I may automate editing the changelog section in the future)

packaging/linux/copr/Sunshine.spec
```diff
@@ -213,203 +123,58 @@ cmake_args=(
...
-    $path_to_udevadm trigger --property-match=DEVNAME=/dev/uhid
-    echo "Udev rules reloaded successfully."
-  else
-    echo "error: udevadm not found or not executable."
-  fi
-else
-  echo "rpm-ostree environment detected, skipping post install steps. Restart to apply the changes."
-fi
+%postun
+udevadm control --reload-rules || true
```
ReenigneArcher, line 168: > Simplify modprobe and udev commands. No need to check for rpm-ostree and udevadm, just silently fail. Output is not displayed to the user anyways. Fedora has udevadm by default.

There is no possible way for users to see the output/logs? I foresee this silent failing causing a lot of support requests from users when stuff isn't working.

packaging/linux/copr/Sunshine.spec
```diff
@@ -22,186 +19,99 @@ License: GPLv3-only
...
-Requires: libdrm2
-Requires: libevdev2
-Requires: libopusenc0
-Requires: libva2
-Requires: libwayland-client0
-Requires: libX11-6
-Requires: libnuma1
-Requires: libopenssl3
-Requires: libpulse0
 %endif
```
ReenigneArcher, line 176: >>> Runtime dependencies - These are all removed.

>> This will break rpms we pull from copr and add to GitHub releases, or downloaded manually in a PR test.

> No they will not. They are exactly the same, except for dependencies that are not needed. The required dependencies will match the target. If Sunshine needs extra dependencies (that are not linked) they can be added. This ensures that copr/rpmbuild/cmake can figure out the different runtime dependencies for all targets.

I'm confused by this. We don't static link on Linux (with a few exceptions), so how will the rpm know what is a runtime dependency?

Note: The rpm depends from CMake are not carried over to this spec. That is for cpack which is not used here.

packaging/linux/copr/Sunshine.spec
```diff
@@ -22,186 +19,99 @@ License: GPLv3-only
...
+BuildRequires: nodejs
+BuildRequires: npm
 BuildRequires: openssl-devel
 BuildRequires: pipewire-devel
-BuildRequires: rpm-build
 BuildRequires: systemd-rpm-macros
-BuildRequires: wget
-BuildRequires: which
-
+# For tests ⤵
```
ReenigneArcher, line 39:
```suggestion

# For tests
```

packaging/linux/copr/Sunshine.spec
```diff
@@ -213,203 +123,58 @@ cmake_args=(
   "-DSUNSHINE_PUBLISHER_NAME=LizardByte"
   "-DSUNSHINE_PUBLISHER_WEBSITE=https://app.lizardbyte.dev"
   "-DSUNSHINE_PUBLISHER_ISSUE_URL=https://app.lizardbyte.dev/support"
+  "-DSUNSHINE_ENABLE_CUDA=ON"
```
ReenigneArcher, line 126:
```suggestion
```

It's ON by default.

packaging/linux/copr/Sunshine.spec
```diff
@@ -213,203 +123,58 @@ cmake_args=(
...
-if [ -n "%{cuda_version}" ] && [[ " ${cuda_supported_architectures[@]} " =~ " ${architecture} " ]]; then
-  install_cuda
-  cmake_args+=("-DSUNSHINE_ENABLE_CUDA=ON")
-  cmake_args+=("-DCMAKE_CUDA_COMPILER:PATH=%{cuda_dir}/bin/nvcc")
-  cmake_args+=("-DCMAKE_CUDA_HOST_COMPILER=gcc-%{gcc_version}")
-else
-  cmake_args+=("-DSUNSHINE_ENABLE_CUDA=OFF")
 fi
 
-# Install and setup NVM for Fedora 44+
```
ReenigneArcher, line 288: Should probably leave the nvm logic in place. This was likely a fix implemented when 44 was Rawhide. Sunshine isn't always automatically compatible with the newest node versions.

packaging/linux/copr/Sunshine.spec
```diff
@@ -213,203 +123,58 @@ cmake_args=(
...
-export PATH="$(dirname ${NODE_PATH}):${PATH}"
-
-echo "Node.js version: $(node --version)"
-echo "npm version: $(npm --version)"
-%endif
-
-cd %{_builddir}/Sunshine/build
-%make_install
+cd %{sourcedir}/build/
+xvfb-run ./tests/test_sunshine || true
```
ReenigneArcher, line 160:
```suggestion
xvfb-run ./tests/test_sunshine
```

Definitely cannot silently fail here.

packaging/linux/copr/Sunshine.spec
PVermeer: Ah, thats whats python is for.

packaging/linux/copr/Sunshine.spec
PVermeer: I don't know cuda at all and assumed it had backward compatibility, but I see now it has forward compatibility (new hardware should run older builds). So a lower version is probable better that the code allows.

The [nvidia channel](https://anaconda.org/channels/nvidia/packages/cuda-nvcc/overview) has the most versions so might be better to use that.

But yes, you can set the version with `=12` or `=12.9` or `12.9.86` and use the nvidia channel:
```rpm
micromamba create -y -p %{cudadir} nvidia::cuda-nvcc=12
```

Versions:
```sh
$ micromamba repoquery search --channel=nvidia cuda-nvcc
Getting repodata from channels...

nvidia/linux-64                                             Using cache
nvidia/noarch                                               Using cache
 Name      Version  Build                   Channel Subdir  
─────────────────────────────────────────────────────────────
 cuda-nvcc 13.2.51  hc118093_0              nvidia  linux-64
 cuda-nvcc 13.1.80  h959976e_0              nvidia  linux-64
 cuda-nvcc 13.1.115 h4e12aa9_0              nvidia  linux-64
 cuda-nvcc 13.0.88  hb25713b_0              nvidia  linux-64
 cuda-nvcc 13.0.48  h26104e3_0              nvidia  linux-64
 cuda-nvcc 12.9.86  0                       nvidia  linux-64
 cuda-nvcc 12.9.41  0                       nvidia  linux-64
 cuda-nvcc 12.8.93  0                       nvidia  linux-64
 cuda-nvcc 12.8.61  0                       nvidia  linux-64
 cuda-nvcc 12.6.85  0                       nvidia  linux-64
 cuda-nvcc 12.6.77  0                       nvidia  linux-64
 cuda-nvcc 12.6.68  0                       nvidia  linux-64
 cuda-nvcc 12.6.20  0                       nvidia  linux-64
 cuda-nvcc 12.5.82  0                       nvidia  linux-64
 cuda-nvcc 12.5.40  0                       nvidia  linux-64
 cuda-nvcc 12.4.99  0                       nvidia  linux-64
 cuda-nvcc 12.4.131 0                       nvidia  linux-64
 cuda-nvcc 12.3.52  0                       nvidia  linux-64
 cuda-nvcc 12.3.107 0                       nvidia  linux-64
 cuda-nvcc 12.3.103 0                       nvidia  linux-64
 cuda-nvcc 12.2.91  0                       nvidia  linux-64
 cuda-nvcc 12.2.140 0                       nvidia  linux-64
 cuda-nvcc 12.2.128 0                       nvidia  linux-64
 cuda-nvcc 12.1.66  0                       nvidia  linux-64
 cuda-nvcc 12.1.105 0                       nvidia  linux-64
 cuda-nvcc 12.0.76  0                       nvidia  linux-64
 cuda-nvcc 12.0.140 0                       nvidia  linux-64
 cuda-nvcc 11.8.89  0                       nvidia  linux-64
 cuda-nvcc 11.7.99  0                       nvidia  linux-64
 cuda-nvcc 11.7.64  0                       nvidia  linux-64
 cuda-nvcc 11.6.55  h5758ece_0              nvidia  linux-64
 cuda-nvcc 11.6.124 hbba6d2d_0              nvidia  linux-64
 cuda-nvcc 11.6.112 hf7fc535_0              nvidia  linux-64
 cuda-nvcc 11.5.50  h8f81028_0              nvidia  linux-64
 cuda-nvcc 11.5.119 h2e31d95_0              nvidia  linux-64
 cuda-nvcc 11.4.48  0          (+ 1 builds) nvidia  linux-64
 cuda-nvcc 11.4.152 0          (+ 1 builds) nvidia  linux-64
 cuda-nvcc 11.4.120 0          (+ 1 builds) nvidia  linux-64
 cuda-nvcc 11.4.100 0          (+ 1 builds) nvidia  linux-64
 cuda-nvcc 11.3.58  h2467b9f_0              nvidia  linux-64
```

packaging/linux/copr/Sunshine.spec
PVermeer: I disagree, but it's your project. Copr is build system for packaging, not for running dev CI. I think only pure logic unit tests can be rialibly tested in COPR on all targets since they're all clean basic chroots and setup for building packaging for multiple targets. Environments can be different. Running tests that need something like this:
```rpm
# for unit tests
BuildRequires: xorg-x11-server-Xvfb
```
is really not advised.

Test, especially these integration tests, are better run on testing CI's like github actions or a test docker before sending it to COPR for a package build.

packaging/linux/copr/Sunshine.spec
PVermeer: `set -e` is always enabled in these scripts, no need to set this explicitly.

packaging/linux/copr/Sunshine.spec
PVermeer: Copr (rpmbuild), is doing the same thing that cpack is doing. The binaries or libs need to link to system libraries (.so or .dll on windows) so this information in there:

With `strings` one can scan a binary for strings (duh). As you can see below the dynamic libraries are in there as literal strings. Since linking these was successful, these files should still be on the fs. 
```sh
$ strings /usr/sbin/sunshine-2025.924.154138 | grep "\.so"
/lib64/ld-linux-x86-64.so.2
libcurl.so.4
libappindicator3.so.1
libgtk-3.so.0
libgobject-2.0.so.0
libglib-2.0.so.0
libnotify.so.4
libminiupnpc.so.21
libopus.so.0
libnuma.so.1
libva.so.2
libva-drm.so.2
libssl.so.3
libcrypto.so.3
libdrm.so.2
libcap.so.2
libevdev.so.2
libwayland-client.so.0
libgbm.so.1
libpulse.so.0
libpulse-simple.so.0
libicuuc.so.77
libstdc++.so.6
libm.so.6
libmvec.so.1
libgcc_s.so.1
libc.so.6
ld-linux-x86-64.so.2
libcuda.so.1
libnvidia-encode.so.1
libamfrt64.so.1
libOpenCL.so
libatiadlxx.so
libx265_main12.so
libx265_main10.so
libx265.so
libnvidia-fbc.so.1
libnvidia-fbc.so
libxcb.so.1
libxcb.so
libxcb-shm.so.0
libxcb-shm.so
libX11.so.6
libX11.so
libXfixes.so.3
libXfixes.so
libXrandr.so.2
libXrandr.so
libavahi-common.so.3
libavahi-common.so
libavahi-client.so.3
libavahi-client.so
libgbm.so.1
libgbm.so
libEGL.so.1
libEGL.so
```
You can check what package is providing this file with:
```sh
$ rpm -qf /usr/lib64/ld-linux-x86-64.so.2
glibc-2.42-10.fc43.x86_64
```
And voila it knows it needs `glibc-2.42-10`
However the result RPM only has the following requirements, which are just the libraries and not the packages:
```sh
$ rpm -qpR ./rpmbuild/RPMS/x86_64/sunshine-beta-2026.329.31930-2.fc43.x86_64.rpm 
/bin/sh
/bin/sh
ld-linux-x86-64.so.2()(64bit)
ld-linux-x86-64.so.2(GLIBC_2.3)(64bit)
libappindicator3.so.1()(64bit)
libc.so.6()(64bit)
libc.so.6(GLIBC_2.10)(64bit)
libc.so.6(GLIBC_2.14)(64bit)
libc.so.6(GLIBC_2.16)(64bit)
libc.so.6(GLIBC_2.17)(64bit)
libc.so.6(GLIBC_2.2.5)(64bit)
libc.so.6(GLIBC_2.25)(64bit)
libc.so.6(GLIBC_2.26)(64bit)
libc.so.6(GLIBC_2.28)(64bit)
libc.so.6(GLIBC_2.3)(64bit)
libc.so.6(GLIBC_2.3.2)(64bit)
libc.so.6(GLIBC_2.3.3)(64bit)
libc.so.6(GLIBC_2.3.4)(64bit)
libc.so.6(GLIBC_2.32)(64bit)
libc.so.6(GLIBC_2.33)(64bit)
libc.so.6(GLIBC_2.34)(64bit)
libc.so.6(GLIBC_2.38)(64bit)
libc.so.6(GLIBC_2.4)(64bit)
libc.so.6(GLIBC_2.6)(64bit)
libc.so.6(GLIBC_2.7)(64bit)
libc.so.6(GLIBC_2.8)(64bit)
libc.so.6(GLIBC_2.9)(64bit)
libcap.so.2()(64bit)
libcrypto.so.3()(64bit)
libcrypto.so.3(OPENSSL_3.0.0)(64bit)
libcurl.so.4()(64bit)
libdrm.so.2()(64bit)
libevdev.so.2()(64bit)
libevdev.so.2(LIBEVDEV_1)(64bit)
libgbm.so.1()(64bit)
libgio-2.0.so.0()(64bit)
libglib-2.0.so.0()(64bit)
libgobject-2.0.so.0()(64bit)
libgtk-3.so.0()(64bit)
libicuuc.so.77()(64bit)
libm.so.6()(64bit)
libm.so.6(GLIBC_2.2.5)(64bit)
libm.so.6(GLIBC_2.27)(64bit)
libm.so.6(GLIBC_2.29)(64bit)
libm.so.6(GLIBC_2.35)(64bit)
libm.so.6(GLIBC_2.38)(64bit)
libminiupnpc.so.21()(64bit)
libmvec.so.1()(64bit)
libmvec.so.1(GLIBC_2.22)(64bit)
libmvec.so.1(GLIBC_2.35)(64bit)
libnotify.so.4()(64bit)
libnuma.so.1()(64bit)
libnuma.so.1(libnuma_1.1)(64bit)
libnuma.so.1(libnuma_1.2)(64bit)
libopus.so.0()(64bit)
libpipewire-0.3.so.0()(64bit)
libpulse-simple.so.0()(64bit)
libpulse-simple.so.0(PULSE_0)(64bit)
libpulse.so.0()(64bit)
libpulse.so.0(PULSE_0)(64bit)
libssl.so.3()(64bit)
libssl.so.3(OPENSSL_3.0.0)(64bit)
libva-drm.so.2()(64bit)
libva.so.2()(64bit)
libva.so.2(VA_API_0.33.0)(64bit)
libwayland-client.so.0()(64bit)
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileCaps) <= 4.6.1-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
rtld(GNU_HASH)
```
The final step install, will install the packages that provide the libraries (that are missing):
```sh
sudo dnf install sunshine
Updating and loading repositories:
Repositories loaded.
Package                                      Arch   Version                                      Repository                                        Size
Installing:
 sunshine                                    x86_64 2025.924.154138-2.fc43                       copr:copr.fedorainfracloud.org:pvermeer:suns  25.8 MiB
Installing dependencies:
 libappindicator-gtk3                        x86_64 12.10.1-9.fc43                               fedora                                       108.5 KiB
 libdbusmenu                                 x86_64 16.04.0-30.fc43                              fedora                                       536.2 KiB
 libdbusmenu-gtk3                            x86_64 16.04.0-30.fc43                              fedora                                        88.4 KiB
 miniupnpc                                   x86_64 2.3.3-3.fc43                                 fedora                                       128.2 KiB

Transaction Summary:
 Installing:         5 packages

Total size of inbound packages is 8 MiB. Need to download 8 MiB.
After this operation, 27 MiB extra will be used (install 27 MiB, remove 0 B).
```

packaging/linux/copr/Sunshine.spec
PVermeer: `rpm-ostree install Sunshine` logs it to the logger and can be printed with `journalctl -b -t 'rpm-ostree(Sunshine.post)'` after install.

I'm wrong about the `dnf install Sunshine`, it does output the script.

packaging/linux/copr/Sunshine.spec
PVermeer: It's not needed anymore. More likely it was a bug in rawhide.

> This was likely a fix implemented when 44 was Rawhide. Sunshine isn't always automatically compatible with the newest node versions.

Your fix does the opposite:
```
++ node --version
+ echo 'Node.js version: v25.8.2'
Node.js version: v25.8.2
```
This is the bleeding-edge version of node.
Fedora actually lacks behind an LTS version (22). Latest LTS is 24.

From my web-dev days:
Nodejs almost fully backwards compatibe except from some edge cases that had to be removed.

I think it's fine to use the package nodejs. Rawhide can be buggy and this is to be expected from time to time. Also I could not find any version of the dev env for nodejs in the project.

ReenigneArcher: It was likely something with vite/rollup that caused us to not be able to use the system version. It actually happens fairly regularly.

packaging/linux/copr/Sunshine.spec
PVermeer: This is missing on stable and causing the error:
https://github.com/LizardByte/Sunshine/blob/dc2dc61d8cba854b92337798086a6af11af65fed/cmake/compile_definitions/common.cmake#L24-L27

ReenigneArcher: You understand that merging this PR doesn't affect our stable copr, right?

packaging/linux/copr/Sunshine.spec
PVermeer: I see a lot of `npm install`s and  a `npm install --ignore-scripts --package-lock-only` and no lockfiles. That might cause the issue. Using a lockfile and using `npm ci` instead of `npm install` ensures no version changes.

packaging/linux/copr/Sunshine.spec
PVermeer: I wanted to answer your questions and test some changes in my test copr based on your info.

But yes, if you're unwilling to maintain a stable build for new Fedora releases this is the end for me and I will keep it external.

You do realise that Sunshine will not be availbe for a new Fedora release? It will not be forked or copied. So if users install the stable copr on their systems Fedora will fail to update to a new Fedora release because of missing Sunshine. New users who install the COPR cannot install it.

The stable COPR is basicly worthless this way and makes no sense.

ReenigneArcher: There seems to be a fundamental misunderstanding of how this spec is used and how changing it actually affects anything.

I recognize there is a problem with the stable copr repo, which is most likely drawing more attention since bazzite removed our beta builds. In my opinion, they should have never included beta builds to begin with especially unverified ones. Anyway, I will work towards a way to provide patches to the stable copr repo in the future, but this work will happen after our next release.

I am also planning a stable release of Sunshine soon, and actually delaying the release until F44 is released to it syncs up better.

PVermeer:
> There seems to be a fundamental misunderstanding of how this spec is used and how changing it actually affects anything.

Agreed, it affects way too much. COPR is just a build system, like flatpak or other package distributions. Build, package, ship it. I do understand it, but disagree with the use of COPR for it.

> I recognize there is a problem with the stable copr repo, which is most likely drawing more attention since bazzite removed our beta builds. In my opinion, they should have never included beta builds to begin with especially unverified ones. Anyway, I will work towards a way to provide patches to the stable copr repo in the future, but this work will happen after our next release.

I agree they should have never used beta. They should have continued with their own stable builds. Bazzite also has issues with testing that I've raised in the past. They build and test their own unstable / testing image and then do a full rebuild of the stable image, where updated (and thus untested) packages could slip in, like Sunshine.

>We only create a build in our stable copr, when we create a stable release here.

I do understand you're reasoning concerning stable builds. Build a stable once and use that until a new stable build is ready. This however does not play well with Linux. All packages are always rebuild for new releases. Fedora does a mass rebuild for every new release. That's just to keep it stable. It's also enforced on COPR for the same reasons. That's why I would strongly advise to keep `.spec` files simple with a just build, package, ship it. Testing and al other things should already be done. A rebuild should be done with a click a button.

Rolling releases usually do not have this requirement since there are no major releases, but breakage of older packages is a greater risk.

> I am also planning a stable release of Sunshine soon, and actually delaying the release until F44 is released to it syncs up better.

That's great! Don't delay the release for the actual release. Fedora 44's first (public) release date is in 10 days, if not delayed. Not having a package available will prevent users from updating with an error that the package is not available and must be erased before updating. This is not the case right now because there is no package on Fedora 43.

There are Fedora distro's / users that intentionally lag a version behind (e.g. Ublue Bluefin) so also build for Fedora 43 so they can upgrade cleanly.

Distro's (like Bazzite), could already be testing the new Fedora release. Not having a package available (or forked from rawhide) can be frustrating.

What I've proposed is a minimal spec file that just builds, package and ships Sunshine for all Fedora targets. It's not perfect, because I'm not as knowledgeable of the project as you but it's a good start. A separate spec file for stable is also fine.

I've got this spec now running at:
https://copr.fedorainfracloud.org/coprs/pvermeer/sunshine/
https://github.com/PVermeer/copr_sunshine

I'm a big fan and supporter of the project and will be running the COPR above to support Fedora with stable builds. If
>We only create a build in our stable copr, when we create a stable release here.

ever changes I'd be happy to help.

PR #4935, opened 2026-04-01
feat(linux): add portal token setup script for headless systems

neatnoise:
## Description

On headless Linux systems using Wayland, the XDG portal `Start` call requires user confirmation via a monitor selection dialog. When Sunshine runs as a systemd service, there is no way to approve this popup without a physical display.

This script automates the portal token setup:
1. Starts the Sunshine systemd service with portal capture (hangs waiting for approval)
2. Launches a temporary KMS instance on a different port so the user can connect via Moonlight
3. User approves the portal popup from the KMS session
4. Cleans up the temporary KMS instance
5. The systemd service continues with a saved portal token that persists across restarts

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/scripts/linux_portal_setup.sh
+++ b/scripts/linux_portal_setup.sh
@@ -1,0 +1,189 @@
+#!/bin/bash
+# Regenerates the XDG portal restore token for Sunshine running as a systemd service.
+#
+# On headless Wayland systems, the portal Start call requires user confirmation
+# via a monitor selection dialog. This script:
+#   1. Starts the Sunshine service with portal capture (hangs waiting for approval)
+#   2. Launches a temporary KMS instance so you can connect via Moonlight
+#   3. You approve the portal popup from the KMS session
+#   4. Cleans up and verifies the service is running with a persisted token
+#
+# Usage: linux_portal_setup.sh [--service NAME] [--port PORT]
+
+set -euo pipefail
+
+# --- Defaults ---
+SERVICE=""
+KMS_PORT=48990
+
+# --- Parse args ---
+while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
As this PR stands right now, this script will not be included in any Sunshine packages. Is that the intent?

neatnoise:
> As this PR stands right now, this script will not be included in any Sunshine packages. Is that the intent?

Hhm, maybe would be better to just have it in the docs for copy paste and use? It's to automate portal token generation for systems without physical monitor. I had the issues to generate it in this case and it will help to automate.

ReenigneArcher:
Could we achieve the same from within Sunshine's src code and have a button in the UI or CLI arg to handle this instead of an external script?

neatnoise:
What do you think about installing it to like /usr/bin/sunshine-portal-setup and documenting in the docs? It requires the user's interaction unfortunately to connect using Moonlight to Sunshine instance on custom port (KMS), access for setcap etc

ReenigneArcher:
@psyke83 tagging you in case you have some thoughts on how to handle this

psyke83:
Having it as an optional script may be OK, I suppose? I also run a headless host, so I do need to run a similar script that I wrote for GNOME sessions. On KDE, it probably makes better sense to use the [secondary permission system](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/commit/2d4a4dfec07e478ab57fb70a1f531ba496382467):
```
flatpak permission-set kde-authorized remote-desktop dev.lizardbyte.app.Sunshine yes
```

If this script is added, the above should probably be documented for KDE users as an easier alternative.

PR #4938, opened 2026-04-02
fix(windows): skip hybrid GPU hook on single-GPU systems

SeMmyT:
## Summary

- Skip the `NtGdiDdDDIGetCachedHybridQueryValue` MinHook on single-GPU systems where it's unnecessary
- Prevents a crash loop (`0xc0000005` in `dxgi.dll`) on Windows 11 24H2 build 29558+
- Adds proper error handling for MinHook API return values

## Problem

The hybrid GPU workaround hook introduced in #3530 causes an access violation in `dxgi.dll` on recent Windows 11 builds (24H2, build 29558+) with single-GPU systems. The crash is deterministic — always at offset `0x6cb2f` in `dxgi.dll` — and creates an infinite crash loop where the service wrapper restarts `sunshine.exe` every ~6 seconds.

**Crash signature:**
```
Faulting module: dxgi.dll (10.0.29558.1000)
Exception: 0xc0000005 (ACCESS_VIOLATION)
Offset: 0x000000000006cb2f
```

Root cause: the hook spoofs `D3DKMT_GPU_PREFERENCE_STATE_UNSPECIFIED` as the return value, but the updated `dxgi.dll` in build 29558 doesn't handle this state correctly on a code path that only executes when the hook is active.

## Fix

Before installing the hook, enumerate DXGI adapters to detect if the system has multiple physical GPUs:
- **Multi-GPU**: install the hook as before (needed to prevent output reparenting)
- **Single-GPU**: skip the hook entirely (no reparenting issue exists)

Adapters are deduplicated by `VendorId` + `DeviceId` since the same physical GPU can appear as multiple DXGI adapters.

Also adds:
- Error handling for `MH_Initialize()` and `MH_CreateHookApi()` return values
- Logging for hook installation status (success, skip, or failure)

## Test Plan

- [x] Verified on RTX 4070 Ti single-GPU, Windows 11 24H2 build 29558, NVIDIA 591.86
- [ ] Should be tested on multi-GPU/hybrid systems to ensure hook is still applied

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/platform/windows/display_base.cpp
+++ b/src/platform/windows/display_base.cpp
@@ -4,6 +4,7 @@
  */
 // standard includes
 #include <cmath>
+#include <set>
 #include <thread>
 
 // platform includes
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Thank you for the PR submission, but I noticed you didn't use our PR template.

Please update the PR to use the correct template. You can find it at https://github.com/LizardByte/.github/blob/master/.github/pull_request_template.md?plain=1

cgutman:
Did you take a look at dxgi.dll and see if either:
- This bug is specific to single GPU systems
- There's another way for the hook to do its job without triggering the bug

If the bug is not specific to single GPU systems, we're just kicking the can down the road. We're going to have complaints from users with multi-GPU systems where skipping the hook is not an option.

Issue #4940, opened 2026-04-02
KMS capture fails with 0x3003 RGB import error on multi-GPU


AndrewStahlecker:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

KMS capture fails with 0x3003 RGB import error on multi-GPU systems when displays are not on CUDA device 0

Environment:

OS: Fedora (Bazzite base), Wayland/KDE Plasma
Sunshine version: 2026.331.130344
Nvidia driver: 595.45.04
GPU configuration: Dual RTX 3090 in NVLink (4x), displays connected to GPU1 (PCIe bus 65:00.0, CUDA index 1, /dev/dri/card2, renderD129)

Bug:
KMS capture fails with Error: Couldn't import RGB Image: 00003003 in a loop on every connection attempt. Moonlight reports "No video received from host / check firewall UDP ports." No UDP ports are bound during the session.
Root cause:
On a multi-GPU system, Sunshine defaults to CUDA device index 0 for NVENC encoding regardless of which GPU the displays and KMS framebuffer are on. When the displays are on GPU1 (CUDA index 1), Sunshine captures the KMS framebuffer from that GPU but attempts to import it into an EGL/CUDA context on GPU0. This cross-GPU DMA-BUF import fails with error 0x3003 on every frame, preventing any video from being streamed.
Steps to reproduce:

System with two Nvidia GPUs (tested with NVLink, likely reproducible without)
Displays connected to the GPU that is NOT CUDA index 0
Set capture=kms in sunshine.conf
Connect from Moonlight — stream fails immediately with 0x3003 import error loop

Fix:
Three config changes required in ~/.config/sunshine/sunshine.conf:
inicapture=kms
adapter_name=/dev/dri/renderD129   # render node for the GPU with displays
And in the systemd user service override at ~/.config/systemd/user/sunshine.service.d/cuda-device.conf:
ini[Service]
Environment=CUDA_VISIBLE_DEVICES=1  # CUDA index of the GPU with displays
Expected behavior:
Sunshine should either auto-detect which CUDA device corresponds to the KMS capture device, or document that CUDA_VISIBLE_DEVICES must be set manually on multi-GPU systems to match the display GPU.
Notes:

[...]

PR #4946, opened 2026-04-04
feat(linux/wlr): SHM capture fallback + resilient GBM allocation for headless NVIDIA

atassis:
## Description

Two improvements for Wayland capture on headless NVIDIA setups (no physical display connected):

### 1. SHM capture fallback
On headless NVIDIA, GBM buffer allocation can fail because the driver cannot create GBM buffers without an active DRM output. The existing code logged `"SHM capture not implemented"` and gave up.

This PR implements the SHM fallback:
- Binds `wl_shm` interface from the Wayland registry
- When DMA-BUF fails, allocates a `memfd`-backed `wl_shm_pool` and copies frames via SHM
- Handles 4bpp (XRGB8888) and 3bpp (BGR888) pixel formats with correct channel conversion
- Caches GBM failure with sticky `gbm_failed` flag to avoid retrying every frame
- Makes EGL initialization non-fatal so `wlr_ram_t` can operate in SHM-only mode

### 2. Resilient GBM allocation for headless NVIDIA
Even on headless NVIDIA, GBM can work if we relax the usage flags. This change:
- Tries progressively relaxed GBM flags: `RENDERING|LINEAR` → `RENDERING` → `LINEAR` → none
- Prefers DRM render nodes (`renderD*`) over primary nodes (`card*`) — primary nodes require DRM master which is unavailable headless
- Logs which flag combination succeeded (once, at init)

With both changes, the capture priority is:
1. **DMA-BUF with relaxed GBM flags** (zero-copy GPU path) — works on most headless NVIDIA
2. **SHM fallback** (CPU memcpy path) — works everywhere, slower

The DMA-BUF path is completely unchanged when GBM succeeds with default flags.

### Tested on
- RTX 4060 Ti + RTX 5060 Ti headless (TrueNAS, no monitor connected)
- labwc compositor with `WLR_BACKENDS=headless`
- GBM succeeds with `RENDERING` flag, VRAM capture path works end-to-end
- SHM fallback verified by disabling GBM

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/cuda.cpp
+++ b/src/platform/linux/cuda.cpp
@@ -248,11 +248,28 @@ namespace cuda {
       return std::tolower(c);
     });
 
-    // Look for the name of the primary node in sysfs
+    // Look for DRM nodes in sysfs — prefer render node (doesn't require DRM master)
     try {
       char sysfs_path[PATH_MAX];
       std::snprintf(sysfs_path, sizeof(sysfs_path), "/sys/bus/pci/devices/%s/drm", pci_bus_id.data());
       fs::path sysfs_dir {sysfs_path};
+
+      // First pass: look for render node (renderD*)
+      for (auto &entry : fs::directory_iterator {sysfs_dir}) {
+        auto file = entry.path().filename();
+        auto filestring = file.generic_string();
+        if (std::string_view {filestring}.substr(0, 7) != "renderD"sv) {
+          continue;
+        }
+
+        BOOST_LOG(debug) << "Found DRM render node: "sv << filestring;
+
+        fs::path dri_path {"/dev/dri"sv};
+        auto device_path = dri_path / file;
+        return open(device_path.c_str(), O_RDWR);
+      }
+
+      // Second pass: fallback to primary node (card*)
       for (auto &entry : fs::directory_iterator {sysfs_dir}) {
         auto file = entry.path().filename();
         auto filestring = file.generic_string();
@@ -260,7 +277,7 @@ namespace cuda {
           continue;
         }
 
-        BOOST_LOG(debug) << "Found DRM primary node: "sv << filestring;
+        BOOST_LOG(debug) << "Found DRM primary node (fallback): "sv << filestring;
 
         fs::path dri_path {"/dev/dri"sv};
         auto device_path = dri_path / file;
--- a/src/platform/linux/cuda.cpp
+++ b/src/platform/linux/cuda.cpp
@@ -266,7 +266,13 @@ namespace cuda {
 
         fs::path dri_path {"/dev/dri"sv};
         auto device_path = dri_path / file;
-        return open(device_path.c_str(), O_RDWR);
+        int fd = open(device_path.c_str(), O_RDWR);
+        if (fd >= 0) {
+          return fd;
+        }
+        char errbuf[256];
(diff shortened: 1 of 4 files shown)


atassis:
I also have a follow-up fix that makes GBM allocation work on headless NVIDIA by trying relaxed usage flags (removing `GBM_BO_USE_LINEAR`). With this fix, zero-copy DMA-BUF capture works on headless setups — the SHM fallback is only needed if all GBM flag combinations fail.

The change is small (two files):
- `wayland.cpp`: try `GBM_BO_USE_RENDERING` alone when `RENDERING|LINEAR` fails
- `cuda.cpp`: prefer DRM render nodes (`renderD*`) over primary nodes (`card*`) — primary nodes require DRM master which is unavailable headless

Tested on RTX 4060 Ti headless (TrueNAS, no monitor) — GBM succeeds with `RENDERING` flag, VRAM capture path works end-to-end.

Should I add this to this PR or open a separate one?

ReenigneArcher:
> Should I add this to this PR or open a separate one?

I think it can be included here if we make the PR title more generic.

atassis:
Added the GBM resilience fix in e54a35d. Updated the PR title and description to cover both changes.
With both patches, DMA-BUF capture now works on headless NVIDIA (tested on RTX 4060 Ti + RTX 5060 Ti, no monitors) - GBM succeeds with `RENDERING` flag alone. SHM fallback  is only needed if all flag combinations fail.

Clutchnp:
What if the user gives a different adapter name in the config, there will be a mismatch, and this will fail

atassis:
@Clutchnp could you clarify? The code path changed here (cuda.cpp) finds the DRM node by pci bus ID from the CUDA device - it doesn't read config.adapter_name (that's used by VAAPI/Vulkan encoders in vaapi.cpp / vulkan_encode.cpp, which are unchanged). So there shouldn't be a mismatch with any config value.
If you have a specific scenario in mind, could you describe which config field and which code path you're referring to?

Clutchnp:
Suppose you have 2 gpus in the system one being AMD and other being NVIDIA, the display is running on AMD , here while you have selected one, in wayland.cpp, selection maybe sth else entirely. This would as far as I understand result in a conflict
https://github.com/LizardByte/Sunshine/blob/3527320311c15a92172625e4252d4c0a3db307b9/src/platform/linux/wayland.cpp#L224-L261

atassis:
Thanks for pushing back on this - it made me re-audit the change and I found a real (separate) bug in my own patch that I just fixed: the renderD* loop was returning open()'s result directly, so if open() failed (e.g. EACCES, device busy) the function returned -1 without trying the card* fallback. The fix logs a warning and continues to the next node.
On your original concern (multi-GPU mismatch in init_gbm()): the code you linked is unchanged by this PR - it's identical on master. init_gbm() picks the first DRM device with a render node and doesn't read config.adapter_name; that's a real pre-existing limitation on multi-GPU setups (rel. #4940), but introducing GPU selection there would expand this PR's scope significantly beyond "SHM fallback + headless GBM resilience".

Clutchnp:
Maybe due to lack of my understanding, can the drm device  that is opened in cuda.cpp be different from the one that is opened in wayland.cpp?

ReenigneArcher:
Can you fix the conflicts?

Issue #137, opened 2022-04-22
Input doesn't work on headless wayland


dukzcry:
### Describe the Bug

Sway run as
```
WLR_BACKENDS=headless WLR_LIBINPUT_NO_DEVICES=1 sway
```
Input (mouse, keyboard, virtual gamepad) doesn't work
If I run sway in non-headless mode, input works fine

### Expected Behavior

Input should work when sway is run in headless mode

### Sunshine Host Operating System and Version

Linux NixOS 21.11.333823.96b4157790f (Porcupine)

### Architecture

64 bit

### Sunshine Version

0.13.0

### GPU Type

AMD

### GPU Model

Vega 11

### GPU Driver/Mesa Version

mesa 21.2.5

falk4243:
A little observation: I tested starting Steam via sway config and using Moonlight on iOS with a Xbox controller connected, input would be received in the sway session. The same goes for the Moonlight virtual controller. Using the on-screen or a Bluetooth keyboard results in their input being sent to the terminal / GUI (sway is running nested under X11) I have started the sway session from. 

If all input from the client device could be sent to the sway session, this could provide a nifty "poor man's" multi-seat setup where one machine can be used for working and gaming at the same time; apart from making use of Wayland's headless backend. I'll gladly provide further info / test if needed.

falk4243:
Had a brief chat with the sway devs and it's clear now why the controller works: sway does not manage those at all and apps connect to them directly.

jeremyvisser:
I can reproduce this issue as well (Fedora 36, Linux 5.17.13, AMD Radeon RX 6900XT, Sunshine 0.14.0, Sway 1.7, libinput 1.21), but I have an additional clue as to what might be happening.

While running Sway headless (and spawning Sunshine inside Sway), I had left GDM running on the primary head.

When I attached to Sunshine, I could see my Sway desktop, but my keyboard/mouse inputs were actually making it to the GDM session on the same host, but a different head (different Wayland session).

So inputs are being directed to the wrong place.

qwert0p:
Having same issue (Arch Linux 5.18.3, AMD Radeon RX 590, latest sunshine git, Sway 1.8, libinput 1.21.0). After applying patch by sergioperez and running sway in headless mode input doesn't work, in wayvnc input works at the same time.

jeremyvisser:
[...]

PR #4954, opened 2026-04-06
feat(input): implement multiseat support and virtual device naming

willybarret:
## Description

This change introduces multiseat input support and virtual device naming on Linux, allowing multiple users to interact with Sunshine independently on the same system. The intention is to provide flexibility for shared workstations, gaming setups, and other advanced multi-user environments.

> Modern systems (based on systemd and logind) have the concept of a seat (seat0 is default), with associated input devices and graphics outputs. For example, look at `loginctl seat-status seat0`.
> When Sunshine fires up uinput, it’s logically similar to plugging in a second USB mouse: the input gets attached to the default seat (seat0).
> While I don't know exactly what the solution should look like, (it) should ideally run as a separate "seat" (as far as logind and libinput are concerned).

from https://github.com/LizardByte/Sunshine/issues/137#issuecomment-1179947943

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -445,48 +445,6 @@ After adding yourself to the group, log out and log back in for the changes to t
 
 ### Linux
 
-#### Virtual Input Devices
-
-Sunshine uses virtual input devices (via uinput) to inject keyboard, mouse, and gamepad events. Access to `/dev/uinput` is typically restricted to the `input` group.
-
-Add your user to the group:
-
-```bash
-sudo usermod -aG input $USER
-```
-
-> Log out and back in after running this.
-
-The required udev rules are bundled with Sunshine and installed automatically. No manual udev configuration is needed for single-seat setups.
-
----
-
-#### Multi-seat
-
-If you run multiple concurrent Wayland sessions on separate logind seats (e.g. `seat0`, `seat1`), your compositor may ignore injected input unless Sunshine's virtual devices are assigned to the correct seat.
-
-Sunshine determines its target seat from `XDG_SEAT`, which is typically set automatically by your display manager. If needed, you can override it manually in your systemd service file or shell environment before starting Sunshine.
-
-When the seat is not `seat0`, Sunshine appends the seat name to its virtual device names, for example:
-
-- `Keyboard passthrough (seat1)`
-- `Sunshine PS5 (virtual) pad (seat1)`
-
-> Sunshine creates two mouse devices: a relative one and an absolute one (suffixed with ` (absolute)`).
-
-To assign Sunshine's virtual devices to the correct seat, create this udev rules file:
-
-**`/etc/udev/rules.d/72-sunshine-virtual-seat.rules`**
-```udev
-SUBSYSTEM=="input", KERNEL=="input*", ATTR{name}=="*(seat1)*", TAG+="seat", ENV{ID_SEAT}="seat1"
-```
-
-Then reload udev:
-
-```bash
-sudo udevadm control --reload-rules && sudo udevadm trigger -s input
-```
-
 #### Services
 
 **Start once**
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -445,6 +445,48 @@ After adding yourself to the group, log out and log back in for the changes to t
 
 ### Linux
 
+#### Virtual Input Devices
+
+Sunshine uses virtual input devices (via uinput) to inject keyboard, mouse, and gamepad events. Access to `/dev/uinput` is typically restricted to the `input` group.
+
+Add your user to the group:
+
+```bash
+sudo usermod -aG input $USER
+```
+
+> Log out and back in after running this.
+
+The required udev rules are bundled with Sunshine and installed automatically. No manual udev configuration is needed for single-seat setups.
+
+---
+
+#### Multi-seat
+
+If you run multiple concurrent Wayland sessions on separate logind seats (e.g. `seat0`, `seat1`), your compositor may ignore injected input unless Sunshine's virtual devices are assigned to the correct seat.
+
+Sunshine determines its target seat from `XDG_SEAT`, which is typically set automatically by your display manager. If needed, you can override it manually in your systemd service file or shell environment before starting Sunshine.
+
+When the seat is not `seat0`, Sunshine appends the seat name to its virtual device names, for example:
+
+- `Keyboard passthrough (seat1)`
+- `Sunshine PS5 (virtual) pad (seat1)`
+
+> Sunshine creates two mouse devices: a relative one and an absolute one (suffixed with ` (absolute)`).
+
+To assign Sunshine's virtual devices to the correct seat, create this udev rules file:
+
+**`/etc/udev/rules.d/72-sunshine-virtual-seat.rules`**
+```udev
+SUBSYSTEM=="input", KERNEL=="input*", ATTR{name}=="*(seat1)*", TAG+="seat", ENV{ID_SEAT}="seat1"
+```
+
+Then reload udev:
+
+```bash
+sudo udevadm control --reload-rules && sudo udevadm trigger -s input
+```
+
 #### Services
 
 **Start once**
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -159,6 +159,33 @@ If the input is still not working, you may need to add your user to the `input` 
 sudo usermod -aG input $USER
 ```
 
+#### Multiseat
+
+If you run multiple concurrent Wayland sessions on separate logind seats (e.g. `seat0`, `seat1`),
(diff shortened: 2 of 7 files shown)


src_assets/linux/misc/60-sunshine.rules
ReenigneArcher, line 1: For contributors who don't live in Linux land, like me... can you leave the comments in place? As well as the previous ordering?

Also, remove alignment as it makes reading the diffs difficult.

docs/getting_started.md
```diff
@@ -445,6 +445,48 @@ After adding yourself to the group, log out and log back in for the changes to t
 
 ### Linux
 
+#### Virtual Input Devices
```
ReenigneArcher: Previously this wasn't needed unless in a headless environment. So this is documented in the troubleshooting section instead.

ReenigneArcher (changes requested):

atassis:
Hi there. Nice work. I'm working on a similar problem but for containerized/headless deployments where systemd/logind aren't available.
My approach uses the `device_phys` field on uinput devices to tag them with a per-instance identifier, then filters by that field on the input consumer side. The key insight is that `libevdev_set_phys()` doesn't actually propagate to the kernel — you need to call `ioctl(fd, UI_SET_PHYS, ...)` on the uinput fd before `UI_DEV_CREATE`.
I think this could complement your seat naming nicely — your name suffix is great for human-readable identification (`xinput list`), while `device_phys` gives a machine-readable field that compositors and input libraries can filter on. For your use case, something like `.device_phys = "sunshine/" + seat_id` would let libinput-based compositors route devices to the correct seat automatically, without requiring manual udev rules.
I plan to submit a patch to inputtino that adds `UI_SET_PHYS` support (since the existing `device_phys` field in `DeviceDefinition` is silently ignored by the uinput backend). Would be happy to coordinate if you're interested — both approaches could land together.

willybarret:
you would still need to use the ID_SEAT udev property to make it work with logind tho
if you can handle that case you can comment the pr of that implementation and I close this one
using the phys attribute sounds cleaner, I'd like to test it

atassis:
Good point. `ID_SEAT` and `device_phys` solve different layers of the same problem. phys gives a machine-readable tag that compositors/libinput can filter on without any udev rules or logind dependency, which is essential for containerized/headless setups.  `ID_SEAT` is the right mechanism for bare-metal logind multi-seat.
I think both approaches are complementary, not competing. inputtino could support both: `UI_SET_PHYS` for the phys field (currently silently ignored), and your seat naming for logind-aware setups. They can land independently.
I'll submit the inputtino patch for UI_SET_PHYS support soon and link it here. Happy to coordinate so the two features work well together.

src_assets/linux/misc/60-sunshine.rules
```diff
@@ -1,11 +1,12 @@
...
 # Joypads
-KERNEL=="hidraw*", ATTRS{name}=="Sunshine PS5 (virtual) pad", GROUP="input", MODE="0660", TAG+="uaccess"
-SUBSYSTEMS=="input", ATTRS{name}=="Sunshine X-Box One (virtual) pad", GROUP="input", MODE="0660", TAG+="uaccess"
-SUBSYSTEMS=="input", ATTRS{name}=="Sunshine gamepad (virtual) motion sensors", GROUP="input", MODE="0660", TAG+="uaccess"
-SUBSYSTEMS=="input", ATTRS{name}=="Sunshine Nintendo (virtual) pad", GROUP="input", MODE="0660", TAG+="uaccess"
+KERNEL=="hidraw*", ATTRS{name}=="Sunshine PS5 (virtual) pad*", GROUP="input", MODE="0660", TAG+="uaccess"
+SUBSYSTEMS=="input", ATTRS{name}=="Sunshine X-Box One (virtual) pad*", GROUP="input", MODE="0660", TAG+="uaccess"
+SUBSYSTEMS=="input", ATTRS{name}=="Sunshine gamepad (virtual) motion sensors*", GROUP="input", MODE="0660", TAG+="uaccess"
+SUBSYSTEMS=="input", ATTRS{name}=="Sunshine Nintendo (virtual) pad*", GROUP="input", MODE="0660", TAG+="uaccess"
+SUBSYSTEMS=="input", ATTRS{name}=="Sunshine PS5 (virtual) pad*", GROUP="input", MODE="0660", TAG+="uaccess"
```
ReenigneArcher:
```suggestion
SUBSYSTEMS=="input", ATTRS{name}=="Sunshine PS5 (virtual) pad*", GROUP="input", MODE="0660", TAG+="uaccess"

```

src/platform/linux/input/inputtino_seat.cpp
```diff
@@ -0,0 +1,22 @@
+/**
+ * @file src/platform/linux/input/inputtino_seat.cpp
+ * @brief Implementation for multi-seat naming (udev-only).
+ */
+
+#include "inputtino_seat.h"
+
+#include <cstdlib>
```
ReenigneArcher:
```suggestion
// standard includes
#include <cstdlib>

// local includes
#include "inputtino_seat.h"
```

docs/troubleshooting.md
```diff
@@ -159,6 +159,32 @@ If the input is still not working, you may need to add your user to the `input`
...
+If you run multiple concurrent Wayland sessions on separate logind seats (e.g. `seat0`, `seat1`), your compositor may ignore injected input unless Sunshine's virtual devices are assigned to the correct seat.
+
+Sunshine determines its target seat from `XDG_SEAT`, which is typically set automatically by your display manager. If needed, you can override it manually in your systemd service file or shell environment before starting Sunshine.
+
+When the seat is not `seat0`, Sunshine appends the seat name to its virtual device names, for example:
+
+- `Keyboard passthrough (seat1)`
+- `Sunshine PS5 (virtual) pad (seat1)`
+
+> Sunshine creates two mouse devices: a relative one and an absolute one (suffixed with ` (absolute)`).
```
ReenigneArcher: This is a quote, which doesn't really look too good in the docs. See https://lizardbyte--4954.org.readthedocs.build/projects/sunshine/4954/md_docs_2troubleshooting.html#multiseat

docs/troubleshooting.md
```diff
@@ -159,6 +159,32 @@ If the input is still not working, you may need to add your user to the `input`
...
+
+To assign Sunshine's virtual devices to the correct seat, create this udev rules file:
+
+**`/etc/udev/rules.d/72-sunshine-virtual-seat.rules`**
```
ReenigneArcher: This doesn't look quite right either, maybe it'd be fine without the backtics?

docs/troubleshooting.md
```diff
@@ -159,6 +159,32 @@ If the input is still not working, you may need to add your user to the `input`
...
+#### Multiseat
+
+If you run multiple concurrent Wayland sessions on separate logind seats (e.g. `seat0`, `seat1`), your compositor may ignore injected input unless Sunshine's virtual devices are assigned to the correct seat.
+
+Sunshine determines its target seat from `XDG_SEAT`, which is typically set automatically by your display manager. If needed, you can override it manually in your systemd service file or shell environment before starting Sunshine.
+
+When the seat is not `seat0`, Sunshine appends the seat name to its virtual device names, for example:
+
+- `Keyboard passthrough (seat1)`
+- `Sunshine PS5 (virtual) pad (seat1)`
```
ReenigneArcher: No backtics

docs/troubleshooting.md
```diff
@@ -159,6 +159,32 @@ If the input is still not working, you may need to add your user to the `input`
 sudo usermod -aG input $USER
 ```
 
+#### Multiseat
```
ReenigneArcher, line 162: General rule for docs. Can you split lines before the 120 character limit?

ReenigneArcher (commented):
I don't use Linux, but looks pretty good to me. Just have a few suggested changes before this gets merged.

Could you also rebase onto latest master?

docs/troubleshooting.md
willybarret: yeah, it looks kind of weird, I'll change that

src_assets/linux/misc/60-sunshine.rules
willybarret: I think the change was applied correctly?

image

willybarret:
done, I’ve applied the changes and rebased onto latest master
let me know if anything else is needed

ReenigneArcher (dismissed):

src_assets/linux/misc/60-sunshine.rules
```diff
@@ -1,11 +1,13 @@
...
+
```
ReenigneArcher: make sure the file has only one blank line at the end

PR #5005, opened 2026-04-16
fix(linux/vulkan): guard deprecated FFmpeg Vulkan queue lock/unlock

neatnoise:
## Description

FFmpeg marked `AVVulkanDeviceContext::lock_queue` and `unlock_queue` as deprecated under `FF_API_VULKAN_SYNC_QUEUES` (to be removed in libavutil 62+, when Vulkan drivers handle queue synchronization via `VK_KHR_internally_synchronized_queues`).

This caused build failures with `-Werror=deprecated-declarations`.

This PR:
- Wraps `lock_queue`/`unlock_queue` calls with `#if !defined(FF_API_VULKAN_SYNC_QUEUES) || FF_API_VULKAN_SYNC_QUEUES` so they compile out when FFmpeg removes the API
- Adds `-Wno-deprecated-declarations` to `vulkan_encode.cpp` in CMake to suppress the warning while the API is still present

Handles all FFmpeg versions:
- Old (no macro defined): calls lock/unlock normally
- Current (macro = true): calls lock/unlock, warning suppressed
- Future libavutil 62+ (macro = false): skips lock/unlock entirely

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/vulkan_encode.cpp
+++ b/src/platform/linux/vulkan_encode.cpp
@@ -770,18 +770,21 @@ namespace vk {
       submit.signalSemaphoreCount = sem_count;
       submit.pSignalSemaphores = signal_sems.data();
 
-      // Lock the queue (FFmpeg requires this)
+#if !defined(FF_API_VULKAN_SYNC_QUEUES) || FF_API_VULKAN_SYNC_QUEUES
       vk_dev.ctx->lock_queue(
         (AVHWDeviceContext *) ((AVHWFramesContext *) hw_frames_ctx->data)->device_ref->data,
         vk_dev.compute_qf,
         0
       );
+#endif
       auto res = vkQueueSubmit(vk_dev.compute_queue, 1, &submit, VK_NULL_HANDLE);
+#if !defined(FF_API_VULKAN_SYNC_QUEUES) || FF_API_VULKAN_SYNC_QUEUES
       vk_dev.ctx->unlock_queue(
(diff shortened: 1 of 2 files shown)


src/platform/linux/vulkan_encode.cpp
```diff
@@ -770,18 +770,21 @@ namespace vk {
       submit.signalSemaphoreCount = sem_count;
       submit.pSignalSemaphores = signal_sems.data();
 
-      // Lock the queue (FFmpeg requires this)
```
ReenigneArcher, line 773: If it's deprecated, I would assume that means FFmpeg no longer requires this (with the 8.1). And we could just remove the lock/unlock?

I know master is not on 8.1 yet, but it will be very soon.

neatnoise: Yes, the locking is still needed for FFmpeg 8.1 (libavutil 60). FFmpeg's own internal Vulkan code (libavutil/vulkan.c) still calls lock_queue/unlock_queue on every queue submission, guarded by the same #if FF_API_VULKAN_SYNC_QUEUES. Since Sunshine submits to the same shared Vulkan compute queue, we need to use the same lock to avoid a race condition. The API won't be removed until libavutil 62, and until then no Vulkan driver is required to support VK_KHR_internally_synchronized_queues.

ReenigneArcher:
Still seeing the error: https://github.com/LizardByte/Sunshine/actions/runs/24571066332/job/71844112352?pr=4892#step:11:2234

PR #5008, opened 2026-04-16
refactor(linux/xdgportal): Split pipewire code into it's own file so it can be reused

Kishi85:
## Description

Refactor common capture logic for Pipewire-based streaming into it's own superclass(es) and adapt portalgrab .
This helps to keep the code cleaner and allows for easier re-use with other capture methods like KWIN direct screencast.

## Type of Change

- [ ] **feat**: New feature (non-breaking change which adds functionality)
- [ ] **fix**: Bug fix (non-breaking change which fixes an issue)
- [ ] **docs**: Documentation only changes
- [ ] **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semicolons, etc.)
- [X] **refactor**: Code change that neither fixes a bug nor adds a feature
- [ ] **perf**: Code change that improves performance
- [ ] **test**: Adding missing tests or correcting existing tests
- [ ] **build**: Changes that affect the build system or external dependencies
- [ ] **ci**: Changes to CI configuration files and scripts
- [ ] **chore**: Other changes that don't modify src or test files
- [ ] **revert**: Reverts a previous commit
- [ ] **BREAKING CHANGE**: Introduces a breaking change (can be combined with any type above)

## Checklist

- [X] Code follows the style guidelines of this project
- [X] Code has been self-reviewed
- [ ] Code has been commented, particularly in hard-to-understand areas
- [ ] Code docstring/documentation-blocks for new or existing methods/components have been added or updated
- [ ] Unit tests have been added or updated for any new or modified functionality

### AI Usage

- [X] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [ ] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

PR #5009, opened 2026-04-16
feat(capture/linux): add KWin direct screencast capture method

Kishi85:
## Description

This PR adapts the changes from #4724 to the changes from #5008. Here's an updated description (adapted from the original PR) of what this PR does and why it is implemented:

### Why?
The portal path requires two additional D-Bus services (`xdg-desktop-portal` and `xdg-desktop-portal-kde`) that act as pure pass-through layers with no added functionality for Sunshine's use case. In containerized environments (LXC/Docker), the portal frontend also has a known bug (`EACCES` on `/proc/<pid>/root` in user namespaces) that requires patching.

By speaking the KWin screencast Wayland protocol directly, Sunshine:
 * Eliminates two D-Bus intermediaries and their associated services
 * Removes the need for portal auto-approve patches/environment variables
 * Removes the need for `dbus-update-activation-environment` hacks
 * Works reliably in unprivileged LXC containers without portal patches (Note: This might require disabling the KWin permission system to work, see note below)
 * Reuses the existing PipeWire stream infrastructure from `pipewire.cpp` (`pipewire_display_t` class from #5008)
 
### How it works
New file `kwingrab.cpp` implements a standalone Wayland client that:
 1. Setup a temporary desktop file for required KWin permissions if necessary
 2. Connects to the Wayland display and binds `zkde_screencast_unstable_v1` from the registry
 3. Enumerates `wl_output` objects and selects the target output
 4. Calls `stream_output()` to request a screencast stream from KWin
 5. Receives the PipeWire node ID via the `created` callback
 6. Feeds the necessary information back to pipewire_display_t
 
Enabled via `capture = kwin` in the Sunshine config and the CMake flag `-DSUNSHINE_ENABLE_KWIN=ON`.

**NOTE:** While KWin's permission system is enabled (which it is by default for security reasons) a desktop file granting the necessary interface permission for `zkde_screencast_unstable_v1` to Sunshine's process executable (as given by `readlink /proc/<pid-of-sunshine>/exe`) and KWin being able to determine the executable is required for kwingrab to work. kwingrab tries to set up permissions as best as possible by:
 - Dropping elevated privileges (CAP_SYS_ADMIN) so KWin can read the PID executable if it's permission system is enabled.
 - Installing a system-wide permission desktop file to `/usr/share/applications`.
 - Setting up a user-specific permission desktop file in `~/.local/share/applications` should the system-wide file not suffice/be available.

_If setting up permission via desktop file does not work or the use-case does not support working with KWin's permission system (e.g. due to not being able to drop CAP_SYS_ADMIN) it is possible to disable KWin permission checks by setting `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1` in (global) environment to make kwingrab work._

### Real-world testing
Tested extensively over WAN (VDSL and LTE/5G) with Moonlight at 1080p/4K, 60/120fps. The KWin direct path delivers noticeably smoother streaming compared to NvFBC + Xorg, particularly over WAN where consistent frame cadence matters most. KWin's vsync-aligned, compositor-driven frame delivery produces a more even packet flow than NvFBC's software-timer-based polling.

### ToDo-List:
- [X] Ensure #5008 is merged
- [X] Add multi-monitor support by implementing the necessary methods
- [X] Find a way to set the necessary KWin permissions automatically
- [X] Fix remaining fixable issues from #4724
- [X] Implement temporary user desktop file for KWin wayland permissions
- [X] Sort screens by KDE's wayland screen priority and use name to map
- [X] Add missing UI entry
- [X] Fix KWin permission problems when starting ~~using systemd~~ with CAP_SYS_ADMIN
- [X] Add system desktop file for KWin permissions and check for that in helper
- [X] Fix @Ramalama2's email for Co-Authored-By of the initial feature commit
- [X] Ensure #5026 is merged
- [X] Check for and add missing tests if necessary
- [X] Cleanup code and fix as many issues as possible
- [X] Ensure #5041 is merged

### Issues Fixed or Closed

- Closes https://github.com/LizardByte/Sunshine/pull/4724

## Type of Change

## Checklist

- [X] Code docstring/documentation-blocks for new or existing methods/components have been added or updated

### AI Usage


neatnoise:
Tested on KDE Plasma 6.6.4, RX 9070 XT, Mesa 26.2 dev, CachyOS with Vulkan CBR encoding. Works great once you get past the setup — I had to set `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1` for KWin (via 
`~/.config/environment.d/kwin.conf`) and restart the session, otherwise the protocol doesn't show up in the registry even though the screencast plugin is loaded.

Once running, streaming is smooth. At one point Moonlight picked the external IP instead of the local one (sometimes It does), so the stream was going out through the internet and back without me even noticing — only realized when I turned on the debug overlay and saw higher the Internet network latency.

Kishi85:
https://github.com/void-linux/void-packages/issues/31274#issuecomment-1020432464

Has had the permission issue before on xdg-portal-kde and fixed it by adding ```X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1``` to the associated desktop file. So maybe we could do something similar here and have a clean way to run this without any hacks or environment variables necessary.

Kishi85:
Adding a dummy desktop file like this as ```~/.local/share/applications/app-dev.lizardbyte.app.Sunshine-kwin-screencast-helper.desktop``` or ```/usr/share/applications/app-dev.lizardbyte.app.Sunshine-kwin-screencast-helper.desktop``` fixes the permission issue but needs the absolute path to the real binary in it for kwin to be able to properly assign the permission: 
```
[Desktop Entry]
Comment=Sunshine kwin screencast permission helper
Name=app-dev.lizardbyte.app.Sunshine-kwin-screencast-helper
NoDisplay=true
Type=Application
X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1
Exec=/home/user/git/sunshine/cmake-build-debug/sunshine-0.0.0-b1dbab76
```

I'll have to test if we have to package and properly distribute the .desktop file or if it is sufficient to create it on-the-fly upon kwingrab init (in ~/.local/share/applications/app-dev.lizardbyte.app.Sunshine-kwin-screencast-helper.desktop before accessing zkde_screencast_unstable_v1). The latter would be preferable, we could also just remove the desktop file upon destruction to have a clean state.

Kishi85:
Running this with CAP_SYS_ADMIN breaks KWin's permission check (because it's not allowed to resolve the executable path for the PID with the capability). I'll have to check if the same workaround we do for portalgrab works here.

Kishi85:
I'll do a bit more polishing before calling this fully "Ready for review" but it is usable and can be tested in it's current state. 

The desktop file to provide the necessary KWin permission will be automatically created (and stale ones will be deleted as well), so KWIN_NO_PERMISSION_CHECKS is no longer necessary. I've had to refactor the CAP_SYS_ADMIN drop so it's in misc.cpp now (see #5026) because it's also needed for this capture method unfortunately (otherwise KWin cannot read the executable for Sunshine's PID and fail the permission check). 

I've tested most of the things that we already did for xdgportal (removing a screen, changing resolution on single screen, changing resolution on secondary screen while casting it, display hotplugging, display unplugging) and everything was working as expected.

I've not tested AppImages or Flatpak, so not sure how well things will work there.

neatnoise:
Found a crash on startup when `~/.local/share/applications/` doesn't exist (which can happen on fresh installs or minimal setups). The `std::filesystem::directory_iterator` in `screencast_permission_helper_t::setup()` throws because the directory isn't there.

Stack trace:

```
#7 std::filesystem::__cxx11::directory_iteratorC2(...).cold
#8 kwin::screencast_permission_helper_t::setup()
#9 kwin::screencast_t::init()
#10 platf::kwin_display_names()
#11 platf::init()
#12 main
```

Fix would be to add a `std::filesystem::create_directories(user_application)` before the `directory_iterator` loop, or check existence first.

Kishi85:
> Found a crash on startup when `~/.local/share/applications/` doesn't exist (which can happen on fresh installs or minimal setups). The `std::filesystem::directory_iterator` in `screencast_permission_helper_t::setup()` throws because the directory isn't there.
> 
> Stack trace:
> 
> ```
> #7 std::filesystem::__cxx11::directory_iteratorC2(...).cold
> #8 kwin::screencast_permission_helper_t::setup()
> #9 kwin::screencast_t::init()
> #10 platf::kwin_display_names()
> #11 platf::init()
> #12 main
> ```
> 
> Fix would be to add a `std::filesystem::create_directories(user_application)` before the `directory_iterator` loop, or check existence first.

Fixed. Thanks for reporting!

Ramalama2:
Nice! Thanks @Kishi85 
Im sorry, my work has took all my time lately, glad god the idea got adopted.

@ReenigneArcher I can close #4724, correct?

Cheers

Kishi85:
@Ramalama2 would you be so kind and provide me your @users.noreply.github.com E-Mail address? That way I can properly credit your code in the first commit with the proper Co-Authored-By.

Ramalama2:
> @Ramalama2 would you be so kind and provide me your @users.noreply.github.com E-Mail address? That way I can properly credit your code in the first commit with the proper Co-Authored-By.

6314556+Ramalama2@users.noreply.github.com

Thanks @Kishi85

ReenigneArcher:
@Ramalama2 I did not even realize this was a duplicate. You can leave it open. When this one is merged it will close the other automatically.

Kishi85:
@ReenigneArcher Thanks for kicking this off. I've added the missing newline in the desktop file template and removed std::views from the last commit as that's too new (C++20, just been following CLion suggestions on that one as it looked cleaner).

ReenigneArcher:
> removed std::views from the last commit as that's too new (C++20,

We should be able to use up to c++23. We definitely have other stuff from `ranges`. I'm not sure what specific error you saw though.

Kishi85:
> > removed std::views from the last commit as that's too new (C++20,
> 
> We should be able to use up to c++23. We definitely have other stuff from `ranges`. I'm not sure what specific error you saw though.

It was complaining about std::views::keys/values being undefined. I've replaced those instances with the older style that'll hopefully work. See kwingrab.cpp Lines 263/331

ReenigneArcher:
We're using it in tests, so I'd think we should be able to use it in src since it's the same compiler.

- https://github.com/LizardByte/Sunshine/blob/baa1a919a56213cfcf57d9406199c19731c05134/tests/integration/test_config_consistency.cpp#L479
- https://github.com/LizardByte/Sunshine/blob/baa1a919a56213cfcf57d9406199c19731c05134/tests/integration/test_locale_consistency.cpp#L190

Was `<ranges>` imported?

psyke83:
I wonder if we can resolve the permissions issue by normalizing the executable name to the PROJECT_FQDN and pointing to it directly in the systemd service unit?

In KWin, identification is handled via: https://github.com/KDE/kwin/blob/e94dda196dcdc80aef5b5d3e440ad9e87f34438f/src/wayland_server.cpp#L172-L195

We can't match the executable name via the main desktop file, but what about via the service?

https://github.com/KDE/kwin/blob/e94dda196dcdc80aef5b5d3e440ad9e87f34438f/src/utils/serviceutils.h#L51-L58

The fetchRequestedInterfaces() call leads to the KService::serviceByDesktopName() function that may be able to match the systemd service if the executable is aligned with the service properly and uses the proper FQDN name instead of "sunshine-XXX".

psyke83:
Maybe it works?

```
[conn@archlinux ~]ls -l /usr/bin/dev.lizardbyte.app.Sunshine DE
-rwxr-xr-x 1 root root 23662824 Apr 20 22:46 /usr/bin/dev.lizardbyte.app.Sunshine
[conn@archlinux ~]$ strings -f ~/.local/share/applications/dev* /usr/share/applications/dev.lizardbyte.app.Sunshine.* | grep X-KDE
/home/conn/.local/share/applications/dev.lizardbyte.app.Sunshine.kwin1776721792685.desktop: #X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1
/usr/share/applications/dev.lizardbyte.app.Sunshine.desktop: X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1
/usr/share/applications/dev.lizardbyte.app.Sunshine.kwin.desktop: #X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1
[conn@archlinux ~]$ cat /usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service
[Unit]
Description=Self-hosted game stream host for Moonlight
StartLimitIntervalSec=500
StartLimitBurst=5
After=graphical-session.target xdg-desktop-autostart.target xdg-desktop-portal.service

[Service]
# Avoid starting Sunshine before the desktop is fully initialized.
ExecStartPre=/bin/sleep 5
ExecStart=dev.lizardbyte.app.Sunshine

Restart=on-failure
RestartSec=5s

[Install]
WantedBy=graphical-session.target
Alias=sunshine.service
```

I rebooted after making the changes, and it seems to connect ok (I don't have the bypass environment variable set):

```
$ cat ~/.config/sunshine/sunshine.log | grep kwingrab
[2026-04-20 22:51:57.878]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:51:57.878]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:51:57.878]: Info: [kwingrab] stream created, PipeWire node 65
[2026-04-20 22:51:57.878]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
[2026-04-20 22:51:57.968]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:51:57.968]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:51:57.968]: Info: [kwingrab] stream created, PipeWire node 65
[2026-04-20 22:51:57.968]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
[2026-04-20 22:53:49.488]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:53:49.488]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:53:49.489]: Info: [kwingrab] stream created, PipeWire node 67
[2026-04-20 22:53:49.489]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
[2026-04-20 22:53:49.577]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:53:49.577]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:53:49.578]: Info: [kwingrab] stream created, PipeWire node 69
[2026-04-20 22:53:49.578]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
[2026-04-20 22:53:49.671]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:53:49.671]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:53:49.671]: Info: [kwingrab] Found output: DP-1 order: 0 position: 0x0 resolution: 1920x1080
[2026-04-20 22:53:49.682]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:53:49.682]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:53:49.683]: Info: [kwingrab] stream created, PipeWire node 66
[2026-04-20 22:53:49.683]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
[2026-04-20 22:57:09.425]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:57:09.425]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:57:09.426]: Info: [kwingrab] stream created, PipeWire node 68
[2026-04-20 22:57:09.426]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
[2026-04-20 22:57:09.514]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:57:09.514]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:57:09.515]: Info: [kwingrab] stream created, PipeWire node 68
[2026-04-20 22:57:09.515]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
[2026-04-20 22:57:09.595]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:57:09.595]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:57:09.595]: Info: [kwingrab] Found output: DP-1 order: 0 position: 0x0 resolution: 1920x1080
[2026-04-20 22:57:09.606]: Info: [kwingrab] bound kde_output_order_v1 v1
[2026-04-20 22:57:09.606]: Info: [kwingrab] bound zkde_screencast_unstable_v1 v1
[2026-04-20 22:57:09.607]: Info: [kwingrab] stream created, PipeWire node 68
[2026-04-20 22:57:09.607]: Info: [kwingrab] Screencasting output name DP-1 position 0x0 resolution 1920x1080
```

Edit 2: maybe it's not working as-is. After rebooting again, it stopped working, so perhaps it was using cached permissions?

Kishi85:
> We're using it in tests, so I'd think we should be able to use it in src since it's the same compiler.
> 
> * https://github.com/LizardByte/Sunshine/blob/baa1a919a56213cfcf57d9406199c19731c05134/tests/integration/test_config_consistency.cpp#L479
> * https://github.com/LizardByte/Sunshine/blob/baa1a919a56213cfcf57d9406199c19731c05134/tests/integration/test_locale_consistency.cpp#L190
> 
> Was `<ranges>` imported?

It was not but somehow still working locally for me. I've changed it back and imported `<ranges>` so it should work now.

Kishi85:
> I wonder if we can resolve the permissions issue by normalizing the executable name to the PROJECT_FQDN and pointing to it directly in the systemd service unit?
> 
> In KWin, identification is handled via: https://github.com/KDE/kwin/blob/e94dda196dcdc80aef5b5d3e440ad9e87f34438f/src/wayland_server.cpp#L172-L195
> 
> We can't match the executable name via the main desktop file, but what about via the service?
> 
> https://github.com/KDE/kwin/blob/e94dda196dcdc80aef5b5d3e440ad9e87f34438f/src/utils/serviceutils.h#L51-L58
> 
> The fetchRequestedInterfaces() call leads to the KService::serviceByDesktopName() function that may be able to match the systemd service if the executable is aligned with the service properly and uses the proper FQDN name instead of "sunshine-XXX".

For KService::serviceByDesktopName() to work you'd need to have a SecurityContextAppId set on the ClientConnection:
https://github.com/KDE/kwin/blob/0e5d51d64fe9897095b20bd41956cd38f3a158f8/src/wayland_server.cpp#L129-L135

The kwingrab wayland connection works fine with a (temporary) .desktop file for the permission when containing the path to the real sunshine binary (as given by ```readlink /proc/<PID-of-sunshine>/exe```) so it's using the ```return KWin::fetchRequestedInterfaces(client->executablePath())``` codepath and serviceByDesktopName() is not regarded due to missing SecurityContextAppId.

Sidenote: The only instance I could find where SecurityContextAppId is set is:
https://github.com/KDE/kwin/blob/0e5d51d64fe9897095b20bd41956cd38f3a158f8/src/wayland/display.cpp#L271-L285

psyke83:
Yes, disregard my idea - sorry for the spam. Setting the SecurityContextAppId would require implementing the wp_security_context_v1 protocol, and that's mostly aimed at sandboxed environments (Flatpak, etc), so it's not worth all that complexity just to avoid an extra .desktop file.

Kishi85:
I've just tested the AppImage artifact and that seems to be working as well but the permission helper does create a new/remove the old the old temporary desktop file from user applications on every startup. This behaviour was to be expected due to the way AppImages work. 

I've had to increase the wait after creation to 3 seconds to ensure KDE has enough time to pick up the new desktop file in $HOME/.local/share/applications somewhat reliably (this might need some tweaking in the future or a way to ensure KDE has recognized the file) but that is usually only triggered once per new binary path and stale temporary desktop entries (that point to a non-existing executable) get cleaned up during next startup. 

Note: Temporary desktop files are **not** created:
- If KDE/KWin is not available (the permission helper will not be used during KWin availability check, only for display_name enumeration and creating a kwin pipewire stream)
- If the permanent system desktop file (/usr/share/applications/dev.lizardbyte.app.Sunshine.kwin.desktop) that is installed with Sunshine packages matches the current sunshine binary path 
- If KWIN_WAYLAND_NO_PERMISSION_CHECKS=1 is present in environment (as that disables KWin's permission system fully anyway).

I have a few minor things (reduce log verbosity/...) left I want to get in before marking this ready for review.

Kishi85:
> Yes, disregard my idea - sorry for the spam. Setting the SecurityContextAppId would require implementing the wp_security_context_v1 protocol, and that's mostly aimed at sandboxed environments (Flatpak, etc), so it's not worth all that complexity just to avoid an extra .desktop file.

Actually we might have to implement this for Flatpak (which is the only known sandbox_engine for wp_security_context_v1 atm) to work properly so don't count your idea out just yet. Likely in a follow-up PR as that addtion could be (as you already noted) a bit complex.

Kishi85:
Finished adding those few missing pieces so this is ready for review on my end. Let's hope I've not forgotten anything important (but it was already building fine before so fingers crossed).

Kishi85:
Added necessary adaptions for #5041 ~~(also temporarily added the commit from #5041 to the list so this can be tested properly)~~.

~~Please do **not merge** before #5041 is merged.~~

moi952:
Hi, I'm using Bazzite, which starts in deck mode. In the Gamescope session, when I log in, the client's keyboard and mouse work. As soon as I switch to the desktop (KDE), the image displays correctly, but I have to use a separate keyboard and mouse connected directly to my host.

Can your pull request solve my problem?

Thanks

Kishi85:
> Hi, I'm using Bazzite, which starts in deck mode. In the Gamescope session, when I log in, the client's keyboard and mouse work. As soon as I switch to the desktop (KDE), the image displays correctly, but I have to use a separate keyboard and mouse connected directly to my host.
> 
> Can your pull request solve my problem?
> 
> Thanks

Honestly I have no idea but the only thing you can do is to try it out. 

Also you might want to open a proper issue (https://github.com/LizardByte/Sunshine/issues/new/choose) as this PR not really the right place for problem reports and random guessing about potential fixes. Thanks!

Kishi85:
@ReenigneArcher Do you think this should be enabled for FreeBSD and Flatpak as well (to be on-par with portalgrab)? Nothing should break, the worst case would be that kwingrab does not work (most likely due to missing permissions). 

Unfortunately I can't test those 2 options by myself atm, hence I've left them out in [eba45e6](https://github.com/LizardByte/Sunshine/pull/5009/commits/eba45e661cc66109da12ee073e643f5004458346).

psyke83:
My thoughts:

XDG Desktop Portal is always guaranteed to work for a Flatpak since the XDG standard is mandated by the package format, and the Desktop Portal is specifically targeted for sandboxed apps. KDE also (uniquely) has a secondary permission system - plumbed through flatpak - that allows bypassing of manual portal setup. For that reason, kwingrab is technically not necessary unless there is a measurable performance difference in the screencast flow through pipewire between the methods. But on the other hand, kwingrab has the same security requirements as portal, and can work without issue as long as your .desktop file is installed, correct? I don't see the harm in enabling it, as long as we can verify it works as expected. Having it enabled for all possible Linux packaging formats would simplify documentation.

FreeBSD doesn't support Flatpak, and while I think they do support the XDG standard, Desktop Portal is not guaranteed to be present, and if it is, it's unclear how to leverage the permission bypass without a working flatpak command, so it seems that kwingrab makes sense for this environment.

Kishi85:
> My thoughts:
> 
> XDG Desktop Portal is always guaranteed to work for a Flatpak since the XDG standard is mandated by the package format, and the Desktop Portal is specifically targeted for sandboxed apps. KDE also (uniquely) has a secondary permission system - plumbed through flatpak - that allows bypassing of manual portal setup. For that reason, kwingrab is technically not necessary unless there is a measurable performance difference in the screencast flow through pipewire between the methods. But on the other hand, kwingrab has the same security requirements as portal, and can work without issue as long as your .desktop file is installed, correct? I don't see the harm in enabling it, as long as we can verify it works as expected. Having it enabled for all possible Linux packaging formats would simplify documentation.
> 
> FreeBSD doesn't support Flatpak, and while I think they do support the XDG standard, Desktop Portal is not guaranteed to be present, and if it is, it's unclear how to leverage the permission bypass without a working flatpak command, so it seems that kwingrab makes sense for this environment.

 - From my personal testing (over local ethernet) kwingrab feels subjectively snappier than portalgrab and objectively (from moonlight streaming stats) host processing latency is lower for kwingrab vs. portalgrab on mostly idle desktop. At least in my case min values go down to 1.5ms with kwingrab vs. >=2ms with portalgrab, with averages following by around the same amount of difference.
 - Since we're forced to use KDE-specific Wayland protocols for kwingrab via zkde_screencast_unstable_v1 already it's also using kde_output_order_v1 so kwingrab is able to follow Screen priorities set in Display Settings. Making the primary screen the one that the sunshine stream is started on and keyboard shortcut order by the same logic.
 - The secondary permission system for portalgrab via flatpak seems to be irrelevant for kwingrab as it's only checking the wayland interface permission via executable matching desktop file. AppImages work via the temporary permission helper but I'm unsure if the same is true for Flatpak (as that might not yield the correct desktop file for permission or need the security context implementation instead).
 - As for FreeBSD that's the interesting one. KWin screencasting does not require Desktop Portal only Pipewire so in theory it should just work with kwin_wayland on FreeBSD. I've no way to test this atm and I'm not sure the permission setup will work correctly for that case.

Kishi85:
Technical sidenote: Looking at ```/usr/share/applications/org.freedesktop.impl.portal.desktop.kde.desktop``` one can see that ```xdg-desktop-portal-kde``` is also using zkde_screencast_unstable_v1 (among a few other interfaces) to provide the XDG portal implementation for KDE:
```
[Desktop Entry]
Exec=/usr/lib/xdg-desktop-portal-kde
X-KDE-Wayland-Interfaces=org_kde_kwin_fake_input,org_kde_plasma_window_management,zkde_screencast_unstable_v1
X-KDE-DBUS-Restricted-Interfaces=org.kde.KWin.ScreenShot2
NoDisplay=true
Icon=kde
```

So we can assume that all pipewire stream features for portalgrab via xdg-desktop-portal-kde should also be directly available for kwingrab.

psyke83:
That's the issue with kwingrab; there's extra maintenance burden in directly utilizing the (unstable) Wayland extension that would be transparently handled by XDG Portal via the KDE helper.

Relevant PR that may impact us in future, if the "v1" protocol is obsoleted: https://invent.kde.org/plasma/kwin/-/merge_requests/9048
Edit: matching PR on Portal side: https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/544

Kishi85:
> That's the issue with kwingrab; there's extra maintenance burden in directly utilizing the (unstable) Wayland extension that would be transparently handled by XDG Portal via the KDE helper.
> 
> Relevant PR that may impact us in future, if the "v1" protocol is obsoleted: https://invent.kde.org/plasma/kwin/-/merge_requests/9048 Edit: matching PR on Portal side: https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/544

We're not the only ones in that boat then (OBS for example also utilizes it) and if it breaks we still have a working fallback with portalgrab (which Sunshine will always need for anything non-KWin). Also we have enough time to prepare so thanks for mentioning the drafts early.

kwingrab is just a way to get a more direct stream for KWin-based systems with a few bonuses (like working screen priorities). The main streaming logic is already shared with portalgrab by utilizing pipewire_display_t so it's just the negotiation that might break until fixed (although I'm not seeing a huge difference for output streaming between v1 and v2 as it is right now. We should even be able to implement a fallback from v2 to v1 once the necessary definitions are in plasma-wayland-protocols).

EDIT: I've got curious and have a preliminary implementation based on [currently available draft protocol spec for kde_screencast_v2](https://invent.kde.org/libraries/plasma-wayland-protocols/-/blob/work/davidre/screencast-v2/src/protocols/kde-screencast-v2.xml?ref_type=heads) ready for a future PR (after the spec hits master on plasma-wayland-protocols) here:  https://github.com/Kishi85/Sunshine/commit/281cc3a8366040912754666ef68d31bc12c1e09b

Kishi85:
A few minor things I've done/added in the last few days:
 - Restructured code to make it easier to adapt new versions of the kde-screencast protocol (see also my [changes for a future PR](https://github.com/Kishi85/Sunshine/tree/draft-kwingrab-kde-screencast-v2) to implement protocol v2)
 - Enabled building for all targets where portalgrab is also enabled as kwingrab should always work when `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1` is set in environment, even for Flatpak and FreeBSD as the required protocol for screencasting is not restricted by KWin in that case.
 - kwingrab is no longer dropping elevated privileges (CAP_SYS_ADMIN) when `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1` is found in environment as that is only necessary for KWin to be able to read the PID executable for permission checks.
- Updated the PR description with an updated version of the description from #4724

EDIT: I've also no further ideas on how to improve this so if anyone has any feel free to comment. The changes for upcoming kde-screencasting-v2 (which no longer has unstable in it's name and seems to be implemented as an overhaul in part due to changes to XDG portal to allow using pipewire-serials) are also ready for a future PR once plasma-wayland-protocols contains the necessary spec file and testing the feature is possible.

Kishi85:
Pipewire objectserial support for kde screencasting v1 was added to plasma-wayland-protocols yesterday (https://invent.kde.org/libraries/plasma-wayland-protocols/-/merge_requests/129) so I've adapted kwingrab ahead of time to have this is in place once it gets relevant. 

I've also updated my draft for kde_screencast_v2 but that cannot be included yet as plasma-wayland-protocols does not yet have the necessary spec xml so I'll likely put that in a separate PR once it's available (unless this PR has not been merged when that happens). 

The addition does nothing by itself for now until KWin announces zkde_screencast_unstable_v1 version 6 (currently  version 5 as of KDE 6.6.4) and will always try to fill both objectserial and node_id for pipewire.cpp (which will prefer  object_serial to node_id if available).

.github/workflows/ci-freebsd.yml
```diff
@@ -174,6 +174,7 @@ jobs:
             -DSUNSHINE_ENABLE_CUDA=OFF \
             -DSUNSHINE_ENABLE_DRM=OFF \
             -DSUNSHINE_ENABLE_PORTAL=ON \
+            -DSUNSHINE_ENABLE_KWIN=ON \
```
ReenigneArcher, line 176: please put alphabetically (applies in few places)

cmake/compile_definitions/linux.cmake
```diff
@@ -76,8 +76,8 @@ if(CUDA_FOUND)
     add_compile_definitions(SUNSHINE_BUILD_CUDA)
 endif()
 
-# libdrm is required for both DRM (KMS) and Wayland
-if(${SUNSHINE_ENABLE_DRM} OR ${SUNSHINE_ENABLE_WAYLAND})
+# libdrm is required for DRM (KMS), Wayland, and KWin ScreenCast
+if(${SUNSHINE_ENABLE_DRM} OR ${SUNSHINE_ENABLE_WAYLAND} OR ${SUNSHINE_ENABLE_KWIN})
```
ReenigneArcher: alphabetical

cmake/compile_definitions/linux.cmake
```diff
@@ -243,7 +243,7 @@ if(GIO_FOUND)
 endif()
 
 # Pipewire
-if(${SUNSHINE_ENABLE_PORTAL})
+if(${SUNSHINE_ENABLE_PORTAL} OR ${SUNSHINE_ENABLE_KWIN})
```
ReenigneArcher: alphabetical

cmake/compile_definitions/linux.cmake
```diff
@@ -264,10 +264,24 @@ if(PIPEWIRE_FOUND AND GIO_FOUND AND ${SUNSHINE_ENABLE_PORTAL})
             "${CMAKE_SOURCE_DIR}/src/platform/linux/portalgrab.cpp")
 endif()
 
+# KWin ScreenCast (direct Wayland protocol, bypasses portal)
+set(KWIN_FOUND OFF)
+if(WAYLAND_FOUND AND PIPEWIRE_FOUND AND ${SUNSHINE_ENABLE_KWIN})
```
ReenigneArcher: pipewire before wayland

cmake/compile_definitions/linux.cmake
```diff
@@ -264,10 +264,24 @@ if(PIPEWIRE_FOUND AND GIO_FOUND AND ${SUNSHINE_ENABLE_PORTAL})
...
+            "${CMAKE_SOURCE_DIR}/src/platform/linux/kwingrab.cpp")
+elseif(${SUNSHINE_ENABLE_KWIN} AND NOT WAYLAND_FOUND)
+    message(WARNING "SUNSHINE_ENABLE_KWIN requires SUNSHINE_ENABLE_WAYLAND — KWin capture disabled")
+endif()
+
 if(NOT ${CUDA_FOUND}
         AND NOT ${WAYLAND_FOUND}
         AND NOT ${X11_FOUND}
         AND NOT ${PORTAL_FOUND}
+        AND NOT ${KWIN_FOUND}
```
ReenigneArcher: alphabetical

Please correct existing list

cmake/packaging/linux.cmake
```diff
@@ -174,6 +174,10 @@ if(NOT ${SUNSHINE_BUILD_APPIMAGE} AND NOT ${SUNSHINE_BUILD_FLATPAK})
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_FQDN}.terminal.desktop"
             DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications")
 endif()
+if(${SUNSHINE_ENABLE_KWIN})
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_FQDN}.kwin.desktop"
```
ReenigneArcher: does it work in the AppImage/Flatpak sandboxes?

cmake/prep/init.cmake
```diff
@@ -11,6 +11,14 @@ elseif (UNIX)
     if(NOT DEFINED SUNSHINE_EXECUTABLE_PATH)
         set(SUNSHINE_EXECUTABLE_PATH "sunshine")
     endif()
+    if(NOT DEFINED SUNSHINE_REAL_EXECUTABLE_PATH)
```
ReenigneArcher: is there a way to have it build sunshine without appending the version? instead of adding this extra logic?

cmake/prep/options.cmake
```diff
@@ -66,4 +66,6 @@ elseif(UNIX)  # Linux
             "Enable X11 grab if available." ON)
     option(SUNSHINE_ENABLE_PORTAL
             "Enable XDG portal grab if available" ON)
+    option(SUNSHINE_ENABLE_KWIN
```
ReenigneArcher, line 67: put before portal

docs/configuration.md
```diff
@@ -2117,6 +2117,11 @@ editing the `conf` file in a text editor. Use the examples as reference.
             @note{Applies to Windows only.}
             @attention{This capture method is not compatible with the Sunshine service.}</td>
     </tr>
+    <tr>
+        <td>kwin</td>
+        <td>Capture with KDE/KWin Wayland compositor via zkde-screencast-unstable-v1.
+            @note{Applies to Linux only.}</td>
+    </tr>
```
ReenigneArcher: alphabetical

packaging/linux/copr/Sunshine.spec
```diff
@@ -220,6 +220,7 @@ cmake_args=(
   "-DSUNSHINE_EXECUTABLE_PATH=%{_bindir}/sunshine"
   "-DSUNSHINE_ENABLE_DRM=ON"
   "-DSUNSHINE_ENABLE_PORTAL=ON"
+  "-DSUNSHINE_ENABLE_KWIN=ON"
```
ReenigneArcher, line 222: before portal

packaging/linux/flatpak/dev.lizardbyte.app.Sunshine.yml
```diff
@@ -83,6 +83,7 @@ modules:
       - -DSUNSHINE_ENABLE_CUDA=ON
       - -DSUNSHINE_ENABLE_DRM=ON
       - -DSUNSHINE_ENABLE_PORTAL=ON
+      - -DSUNSHINE_ENABLE_KWIN=ON
```
ReenigneArcher, line 85: before portal

scripts/linux_build.sh
```diff
@@ -579,6 +579,7 @@ function run_step_cmake() {
     "-DSUNSHINE_EXECUTABLE_PATH=/usr/bin/sunshine"
     "-DSUNSHINE_ENABLE_DRM=ON"
     "-DSUNSHINE_ENABLE_PORTAL=ON"
+    "-DSUNSHINE_ENABLE_KWIN=ON"
```
ReenigneArcher, line 581: before portal

src/platform/linux/misc.cpp
```diff
@@ -965,6 +965,9 @@ namespace platf {
 #endif
 #ifdef SUNSHINE_BUILD_PORTAL
       PORTAL,  ///< XDG PORTAL
+#endif
+#ifdef SUNSHINE_BUILD_KWIN
```
ReenigneArcher, line 966: before portal

src_assets/common/assets/web/configs/tabs/Advanced.vue
```diff
@@ -75,6 +75,7 @@ const config = ref(props.config)
             <option value="kms">KMS</option>
             <option value="x11">X11</option>
             <option value="portal">XDG Portal</option>
+            <option value="kwin">KWin Screencast</option>
```
ReenigneArcher, line 77: before portal

.gitmodules
```diff
@@ -69,3 +69,6 @@
 	path = third-party/wlr-protocols
 	url = https://github.com/LizardByte-infrastructure/wlr-protocols.git
 	branch = master
+[submodule "third-party/plasma-wayland-protocols"]
```
ReenigneArcher: make alphabetical, add tracking branch for dependabot updates... must be the branch that tags land on

note: this file uses tabs, not spaces

third-party/plasma-wayland-protocols
ReenigneArcher, line 1: pin to https://github.com/KDE/plasma-wayland-protocols/releases/tag/v1.20.0

src/platform/linux/kwingrab.cpp
```diff
@@ -0,0 +1,711 @@
...
+using namespace std::literals;
+
+extern const wl_interface wl_output_interface;
+
+namespace kwin {
+  // ─── KWin Wayland ScreenCast permissions ──────────────────────────────────────────────
+  //
+  // To have access to zkde_screencast_unstable_v1 KWin checks for a .desktop file with
+  // X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1 and the current executable name
+  // in the Exec= parameter.
```
ReenigneArcher: Can you use a doxygen style documentation block?

src/platform/linux/kwingrab.cpp
```diff
@@ -0,0 +1,711 @@
...
+          return "";
+        }
+        xdg_data_home = std::filesystem::path(homedir) / ".local"sv / "share"sv;
+      }
+      return xdg_data_home / "applications";
+    }
+
+    static std::string get_executable_full_path() {
+      // Adapted from https://linuxvox.com/blog/how-do-i-find-the-location-of-the-executable-in-c/
+      char exe_path[PATH_MAX];  // PATH_MAX is defined in limits.h (e.g., 4096 on Linux)
```
ReenigneArcher: could you fix this sonar warning?

src/platform/linux/kwingrab.cpp
```diff
@@ -0,0 +1,711 @@
...
+            return true;
+          }
+        }
+      }
+      return false;
+    }
+  };
+
+  // ─── Output parameters ──────────────────────────────────────────────
+  // Storage for all necessary attributes/parameters an output provides
```
ReenigneArcher: Please don't use comments like `// ─── Something ─────────────────────────────────────────────`

ReenigneArcher (commented):
Thanks for another PR! Requesting a few updates which are probably all nitpicks.

cmake/packaging/linux.cmake
Kishi85: I've not tested Flatpak but AppImage relies on the temporary user application helper (and creates a new permission file/deletes the old one on each run). I'll gate the system wide permission file like the terminal file in the section above.

cmake/prep/init.cmake
Kishi85: KWin needs the real executable for the permission check as it is doing a readlink on `/proc/<pid-of-sunshine>/exe` and referencing desktop files against that path for the permission entry. 

Maybe you or someone else has an idea on how to get that value more elegant than this?
I've checked existing files but didn't find anything related (maybe I've missed a cmake variable or something here). I also didn't want to just roll it hard-coded into the template for the system-wide permission file as that might be harder and more obscure to maintain in the future.

third-party/plasma-wayland-protocols
Kishi85: v1.20.0 is missing zkde_screencast_unstable_v1 version 6 as that was just released for upcoming XDG portal adaptions (https://github.com/KDE/plasma-wayland-protocols/commit/18afc45c092857e998c890afce1e7e71f808d819).

I can either pin this like you suggested and move that commit to a separate PR (but I'm unsure when the next tag will be done) or leave it like it is and fix the pinning in an upcoming PR (likely with the additions for kde-screencast-v2).

cmake/prep/init.cmake
Kishi85: The related cmake part for the symlinking is here:

https://github.com/LizardByte/Sunshine/blob/bdef150c6b3341cdd3048795c37257161d8afff7/cmake/targets/common.cmake#L34-L42

Apple codesign binaries are already excluded. I'm not sure why it is symlinking in the first place and I can change the behaviour if desired but I'm not sure if that would break something somewhere.

Note: For now I've added a comment to refer to targets/common.cmake and fixed a logic mismatch.

third-party/plasma-wayland-protocols
ReenigneArcher: Okay, that's fine, we can do it on the next release.

cmake/prep/init.cmake
ReenigneArcher: I tracked down where this was added. It doesn't need to be in there at all for any version. Probably I just copied a whole example from stack overflow.

https://github.com/LizardByte/Sunshine/commit/c7a3f7f46e4b47efd627c16da357d126e5ff7e02

Kishi85: I'll remove it then and adapt my changes accordingly

Kishi85: Done and tested using AUR sunshine-git package, nothing broke when using that.

cmake/compile_definitions/linux.cmake
```diff
@@ -264,13 +264,27 @@ if(PIPEWIRE_FOUND AND GIO_FOUND AND ${SUNSHINE_ENABLE_PORTAL})
...
+set(KWIN_FOUND OFF)
+if(PIPEWIRE_FOUND AND WAYLAND_FOUND AND ${SUNSHINE_ENABLE_KWIN})
+    set(KWIN_FOUND ON)
+    add_compile_definitions(SUNSHINE_BUILD_KWIN)
+    GEN_WAYLAND("${CMAKE_SOURCE_DIR}/third-party/plasma-wayland-protocols/src/protocols" "" kde-output-order-v1)
+    GEN_WAYLAND("${CMAKE_SOURCE_DIR}/third-party/plasma-wayland-protocols/src/protocols" "" zkde-screencast-unstable-v1)
+    list(APPEND PLATFORM_TARGET_FILES
+            "${CMAKE_SOURCE_DIR}/src/platform/linux/kwingrab.cpp")
+elseif(${SUNSHINE_ENABLE_KWIN} AND NOT WAYLAND_FOUND)
+    message(WARNING "SUNSHINE_ENABLE_KWIN requires SUNSHINE_ENABLE_WAYLAND — KWin capture disabled")
```
ReenigneArcher: I missed this in my first review, but this should be `FATAL_ERROR` because `SUNSHINE_ENABLE_KWIN` is explicitly set to on.

.gitmodules
```diff
@@ -69,3 +69,7 @@
...
```
ReenigneArcher: Should be after `nvapi`

Kishi85:
Added that missing single space to fix the failing linter.

Kishi85:
Due to removing /usr/bin/sunshine versioning this line is failing now for COPR:

https://github.com/LizardByte/Sunshine/blob/bdef150c6b3341cdd3048795c37257161d8afff7/packaging/linux/copr/Sunshine.spec#L408

Haven't found any other references to /usr/bin/sunshine-* so I'll fix that once the pipeline has finished (to see if anything else failed that needs fixing due to changes in this PR). 

AppImage and Ubuntu 22.04 seem to fail due to connection problems to ppa.launchpad.net and resulting missing gcc-14 but nothing I can do about that unfortunately.

ReenigneArcher:
> AppImage and Ubuntu 22.04 seem to fail due to connection problems to ppa.launchpad.net and resulting missing gcc-14 but nothing I can do about that unfortunately.

It started happening after 26.04 released. I don't know the fix yet.

psyke83:
Ubuntu's infrastructure is under attack at the moment, so it's probably unrelated to any changes in this PR.

https://status.canonical.com/

ReenigneArcher:
It says everything is operational now, but still seems to be having issues 🤔

Kishi85:
I still can't even open ppa.launchpad.net locally in my browser at the moment. It's just loading and then timing out without showing any page.

Issue #5023, opened 2026-04-18
wlr-screencopy capture targets wrong render node on multi-GPU NVIDIA systems, causing FD leak and VRAM exhaustion via failed dmabuf imports


kabooHD:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

None

### Summary
On a system with two NVIDIA GPUs and a wlroots-based compositor (Hyprland), Sunshine's `capture = wlr` path attempts dmabuf imports on a render node different from the one the compositor rendered to. Cross-device dmabuf imports fail on the NVIDIA open driver. The failure path leaks the imported FD and the kernel-side buffer allocation, and the import is retried in a tight loop with no backoff. Within ~20 seconds, GPU memory on the wrong card is fully exhausted (32 GB), and Sunshine accumulates ~1000 open FDs against `/dev/dri/renderD129`.

Setting `adapter_name = /dev/dri/renderD128` in `sunshine.conf` resolves the issue, suggesting `adapter_name` steers capture-side device selection but is not the default and is not documented as required for multi-GPU setups.

### Environment
- Sunshine: 2025.924.154138 (nixpkgs build, NVENC + CUDA enabled)
- OS: NixOS 25.11 (Xantusia), kernel 6.19.10
- Driver: NVIDIA 595.58.03 (open modules), CUDA 13.2
- GPUs: 2× RTX 5090 (GB202)
  - GPU 0: bus 0x41, card1, renderD128 (x16, has connected display)
  - GPU 1: bus 0x42, card0, renderD129 (x4, no display)
- Compositor: Hyprland (uwsm session via GDM autologin)
- Hyprland pinned to GPU 0 via `AQ_DRM_DEVICES=/dev/dri/card1`
- Sunshine env has `DRI_PRIME=pci-0000_41_00_0` and `CUDA_VISIBLE_DEVICES` pinned to GPU 0 — both ignored by the wlr capture path

### Reproduction
1. Multi-GPU NVIDIA system with `nvidia-drm.modeset=1`, open driver
2. wlroots-based compositor pinned to one GPU
3. Sunshine config: `capture = wlr`, `encoder = nvenc`, no `adapter_name`
4. Connect a Moonlight client

### Actual behavior
- Sunshine logs `Error: Failed to create buffer from params` repeatedly (multiple times per second)
- Stream never starts (black screen on client, eventually times out)
[...]

PR #5024, opened 2026-04-18
docs: document adapter_name requirement for multi-GPU NVIDIA setups

kabooHD:
## Description

Adds troubleshooting guidance for users with multiple NVIDIA GPUs who encounter dmabuf import failures with capture = wlr. Provides commands for identifying the correct render node and explains the symptoms.

Refs #5023

## Type of Change

- [X] **docs**: Documentation only changes

## Checklist

- [ ] Code follows the style guidelines of this project
- [X] Code docstring/documentation-blocks for new or existing methods/components have been added or updated

### AI Usage

- [X] **Moderate**: AI helped with code generation or debugging specific parts

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -188,6 +188,33 @@ nvidia_drm.modeset=1
 Consult your distribution's documentation for details on how to do this. (Most
 often grub is used to load the kernel and set its command line.)
 
+### Multi-GPU NVIDIA systems with wlr capture
+
+If your system has multiple NVIDIA GPUs and you observe any of the following with `capture = wlr`:
+
+- Repeated `Failed to create buffer from params` errors in the Sunshine log
+- Black screen on the Moonlight client; the stream never starts
+- VRAM filling on the GPU that is *not* running your compositor
+- Sunshine accumulating large numbers of open file descriptors against `/dev/dri/renderD*`
+
+…explicitly set `adapter_name` to the render node of the GPU running your compositor:
+adapter_name = /dev/dri/renderD128
+
+To find the correct render node:
+
+```bash
+# Map render nodes to PCI bus IDs
+for r in /dev/dri/renderD*; do
+  echo -n "$r -> "
+  basename $(readlink -f /sys/class/drm/$(basename $r)/device)
+done
+
+# Cross-reference with nvidia-smi to identify which GPU runs your compositor
+nvidia-smi --query-gpu=index,pci.bus_id --format=csv
+```
+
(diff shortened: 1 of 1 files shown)


ReenigneArcher:
Thank you for the PR; however I don't think this adds much value over what we already have in the docs at https://docs.lizardbyte.dev/projects/sunshine/master/md_docs_2configuration.html?lng=en#adapter_name

kabooHD:
@ReenigneArcher 

Thanks for the quick review. I'd like to push back gently on this one.

The existing docs at the link cover what `adapter_name` does as a configuration option — but they don't help a user diagnose *when* they need to set it. The current description ("Sunshine will select the default video card") doesn't suggest that selection can go wrong, or hint at what symptoms indicate it has.

For context on why this matters: I spent three days debugging this on a dual RTX 5090 system (NixOS, Hyprland, NVIDIA open driver). The symptoms were a black stream, repeated `Failed to create buffer from params` errors in the Sunshine log, and ~30 GB of VRAM filling on the wrong GPU within ~20 seconds of every Moonlight connect. I had `DRI_PRIME` and `CUDA_VISIBLE_DEVICES` correctly pinned to the right GPU; the wlr capture path ignored both. I read the configuration docs multiple times during debugging and didn't connect "select the default video card" to my symptoms — there was no path from the error message to the fix.

A user hitting this failure will search the error string, not browse the configuration reference. The troubleshooting page is where that search would land. The PR adds a symptom-to-fix mapping that the reference page intentionally doesn't provide, which is exactly the role troubleshooting docs serve.

If the format isn't what you'd like — too long, wrong section, too NVIDIA-specific — happy to revise. If you'd rather not have it at all, I understand, and I'll leave the issue I filed in place so the next person can at least find it via search.

Again thank you :)

ReenigneArcher:
1. I think this problem should probably be fixed, or at least logged better, in the actual code. At a minimum we should probably detect how many devices there are, and if multiple tell the user they should probably define which one they want to use in the config.
2. I'm not a fan of AI generated user facing docs. Especially then linking back to another AI generated source (the github issue). Not to mention the way it linked, won't even work in our docs and will just render as `#5023`.
3. I'm not sure the symptoms described are exclusive to the problem you faced, which would lead users down a rabbit whole trying this solution that could be for a completely separate problem. "black screen" for example is probably the most common issue users face and there are many causes of it.

This doesn't mean I won't accept a different version of this PR, but as is I think it's not of high value. I'll re-open the PR in case you want to re-work it.

kabooHD:
Closing this. #5030 is fixing the underlying bug properly in code, so the docs workaround isn't needed. I also attempted to help test #5030 directly, but building the fork on NixOS turned into a bigger project than I had skills or time for, so I'll wait for it to merge and verify on my system once it reaches nixpkgs.
Thanks for the review — the feedback on scope and voice was fair, and pointing me toward the code fix was the right call.

PR #5026, opened 2026-04-18
refactor(linux/xdgportal): Move elevated privilege check/drop to misc.cpp

Kishi85:
## Description

This moves function for dropping system admin capabilities to misc.cpp so it can be reused.
It also reimplements the associated check for admin capabilites properly using cap_get_flag.

## Type of Change

- [X] **refactor**: Code change that neither fixes a bug nor adds a feature

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/misc.cpp
+++ b/src/platform/linux/misc.cpp
@@ -25,6 +25,10 @@
 #include <sys/resource.h>  // For setpriority
 #include <sys/socket.h>
 
+#if !defined(__FreeBSD__)
+  #include <sys/capability.h>
+  #include <sys/prctl.h>
+#endif
 #ifdef __FreeBSD__
   #include <net/if_dl.h>  // For sockaddr_dl, LLADDR, and AF_LINK
   #include <sys/syscall.h>  // For syscall: SYS_thr_self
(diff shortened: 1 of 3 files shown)


Kishi85:
@psyke83 I think this should match what you had in mind. I'll need it for #5009 so I've gone ahead with the change.

psyke83:
Looks promising, thanks. Can you look at kmsgrab as well? Depending on how you refactor you could be able to replace the effective (but not permitted) cap raise and dropping there.

Kishi85:
> Looks promising, thanks. Can you look at kmsgrab as well? Depending on how you refactor you could be able to replace the effective (but not permitted) cap raise and dropping there.

https://github.com/LizardByte/Sunshine/blob/7228c2553c393739c3387d5a152c9b255be2328f/src/platform/linux/kmsgrab.cpp#L40-L60

IMHO this is done very cleanly for what kmsgrab needs and I'm not sure if we should change it. kmsgrab is a special snowflake anyway and the main reason for having need of the check/drop fully helpers that are moved to misc.cpp here.

src/platform/linux/misc.cpp
```diff
@@ -1226,4 +1230,59 @@ namespace platf {
     return detected.empty() ? "/dev/dri/renderD128" : detected;
   }
 
+  bool has_cap_sys_admin() {
+#if !defined(__FreeBSD__)
+    const cap_t caps = cap_get_proc();
+    if (!caps) {
+      BOOST_LOG(error) << "[misc] has_cap_sys_admin failed to get process capabilities. Assuming true."sv;
+      return true;
+    }
```
ReenigneArcher, line 1239: This doesn't seem right to me. We are saying that the host `has_cap_sys_admin` when `cap_get_proc()` returns false?

src/platform/linux/misc.cpp
```diff
@@ -1226,4 +1230,59 @@ namespace platf {
...
+    cap_get_flag(caps, CAP_SYS_ADMIN, CAP_PERMITTED, &cap_flags_value);
+    if (cap_flags_value == CAP_SET) {
+      BOOST_LOG(debug) << "[misc] has_cap_sys_admin found permitted CAP_SYS_ADMIN. Returning true."sv;
+      return true;
+    }
+    cap_free(caps);
+#endif
+    BOOST_LOG(debug) << "[misc] has_cap_sys_admin has not found CAP_SYS_ADMIN. Returning false."sv;
+    return false;
+  }
```
ReenigneArcher, line 1255:
```suggestion
    BOOST_LOG(debug) << "[misc] has_cap_sys_admin has not found CAP_SYS_ADMIN. Returning false."sv;
#endif
    return false;
  }
```

We probably shouldn't even log this on FreeBSD as it would send people down a rabbit hole.

src/platform/linux/misc.cpp
Kishi85: The question here is if we should assume the higher capabilites and so we can try to drop or run into a failure at another point if we don't. After thinking about it I'll change that to false because if the cap_get_proc already failed here it'll likely do so in drop_cap_sys_admin.

src/platform/linux/misc.cpp
Kishi85: Fixed

psyke83:
Agree with both of David's comments. Assuming those issues are resolved, a general comment: it's not sufficient to clear CAP_SYS_ADMIN to fix portalgrab and kwingrab's /proc/pid/root access. I understand that you've preserved my logic to also clear CAP_SYS_NICE as well, so take this feedback as mostly a comment on the naming of the function.

When a userspace process has *any* capabilities in the permitted or effective set, the kernel blocks /root access to other unprivileged processes if they don't have the same capabilities (or CAP_SYS_PTRACE). Because we also have CAP_SYS_NICE set on the Sunshine binary, the function is not entirely descriptive. For that reason it might be better to make the name more generic. The has_cap_sys_admin function also has the issue that (aside from the name), if for some reason CAP_SYS_NICE was set but not CAP_SYS_ADMIN, the function would return false and fail to clear the remaining capability needed to make /root dumpable.

My thought was to refactor so that you could supply arguments to specify the capability set, type and action so that kmsgrab could reuse it, but I agree that it's not strictly necessary. At minimum I do believe you should address the logic so it checks for all potential capabilities, or perhaps even change it to unconditionally clear all capabilities.

Kishi85:
@psyke83 Thanks for the detailed explanation. I've changed the names to be more generic:

- has_elevated_privileges: Check if CAP_SYS_ADMIN is present for effective/permitted
- drop_elevated_privileges: Literally what was called finalize_portal_security before

src/platform/linux/misc.cpp
```diff
@@ -1226,4 +1230,63 @@ namespace platf {
     return detected.empty() ? "/dev/dri/renderD128" : detected;
   }
 
+  constexpr std::array<cap_value_t, 2> ELEVATED_PRIVILEGES_EFFECTIVE {CAP_SYS_ADMIN, CAP_SYS_NICE};
+  constexpr std::array<cap_value_t, 2> ELEVATED_PRIVILEGES_PERMITTED {CAP_SYS_ADMIN, CAP_SYS_NICE};
```
psyke83, line 1235: Thanks for making the requested changes - looks good to me. I imagine that these lines also need to be guarded to avoid build errors on FreeBSD.

Kishi85: Done

ReenigneArcher (approved):

PR #4758, opened 2026-02-22
fix: touch scaling bug and initialize display variables as 0

ReenigneArcher:
## Description

Initialize additional display_t members (env_width/env_height/env_logical_width/env_logical_height/width/height/logical_width/logical_height) in the constructor to ensure consistent defaults. Fix touch input scaling on Windows by removing a redundant addition of display offsets (preventing double-offsetting for multi-monitor setups) and adding a clarifying comment. Tighten the video coordinate check to use explicit >0 comparisons for logical/env sizes before computing scaling factors.

TODO:
- [ ] Test to ensure no regressions on different platforms
  - [x] Windows
  - [x] Linux
  - [ ] macOS

### Issues Fixed or Closed

- Fixes https://github.com/LizardByte/Sunshine/issues/4756
- Fixes https://github.com/LizardByte/Sunshine/issues/5044

## Type of Change

- [ ] **feat**: New feature (non-breaking change which adds functionality)
- [x] **fix**: Bug fix (non-breaking change which fixes an issue)
- [ ] **docs**: Documentation only changes
- [ ] **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semicolons, etc.)
- [ ] **refactor**: Code change that neither fixes a bug nor adds a feature
- [ ] **perf**: Code change that improves performance
- [ ] **test**: Adding missing tests or correcting existing tests
- [ ] **build**: Changes that affect the build system or external dependencies
- [ ] **ci**: Changes to CI configuration files and scripts
- [ ] **chore**: Other changes that don't modify src or test files
- [ ] **revert**: Reverts a previous commit
- [ ] **BREAKING CHANGE**: Introduces a breaking change (can be combined with any type above)

## Checklist

- [x] Code follows the style guidelines of this project
- [x] Code has been self-reviewed
- [x] Code has been commented, particularly in hard-to-understand areas
- [ ] Code docstring/documentation-blocks for new or existing methods/components have been added or updated
- [ ] Unit tests have been added or updated for any new or modified functionality

### AI Usage

- [ ] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [x] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

PR #5041, opened 2026-04-22
fix(linux/pipewire): Fix absolute touchinput for multi-monitor

Kishi85:
## Description

Absolute touch input is currently broken on portalgrab if more than one monitor is present. This is due to not setting all necessary parameters during stream setup. This PR fixes the issues in a generic way (by querying wayland using wl::monitors if necessary) in pipewire_display_t so it can be re-used.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/pipewire.cpp
+++ b/src/platform/linux/pipewire.cpp
@@ -615,7 +615,46 @@ namespace pipewire {
       }
     }
 
-    virtual int configure_stream(const std::string &display_name, int &out_pipewire_fd, int &out_pipewire_node, int &out_pos_x, int &out_pos_y, int &out_width, int &out_height) = 0;
+    /**
+     *  @brief Configure the pipewire stream
+     *  @param display_name provide a stream for this display_name
+     *  @param out_pipewire_fd set to the pipewire fd for the stream in during function call (or -1 for using the local context)
+     *  @param out_pipewire_node set to the pipewire node of the stream in during function call
+     *  @returns 0 if the stream successfully configured
+     */
+    virtual int configure_stream(const std::string &display_name, int &out_pipewire_fd, int &out_pipewire_node) = 0;
+
+    /**
+     *  @brief Verify and update display parameters for logical dimensions, desktop dimensions and logical desktop dimensions (default is adapted from wlgrab)
+     */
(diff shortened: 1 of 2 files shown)


Kishi85:
Getting this fixed for scaled displays was a bit trickier but I think I've got that caught as well now. At least it's working using moonlight from my phone as it should.

Additional implementation note: Matching the offset to get logical dimensions while iterating monitors using wl::monitors() should have likely been enough but I've added an additional match on dimensions (both resolution and logical as dbus portal might already send logical dimensions here while pipewire seems to renegotiate on the resolution value) just to be sure.

ReenigneArcher:
@Kishi85 I think it's related to this PR and the linked issue. Can you check? https://github.com/LizardByte/Sunshine/pull/4758

I have no touch or multi-monitor setup to test out.

Kishi85:
> @Kishi85 I think it's related to this PR and the linked issue. Can you check? #4758
> 
> I have no touch or multi-monitor setup to test out.

@ReenigneArcher #4758 would ensure code consistency with the initial values for the parameters but I'm already seeing the variables consistently being initialized with 0 with this PR (not sure why, now that you mention it).

The main issue is that for absolute touchinput to work with multiple monitors is that the calculations need parameters that are not set on current master (only env_width/env_height via stream dimenions) as that was enough to get absolute touch working on pipewire streaming a single monitor. 

For multi-monitor additional values have to be correct as scale factor and absolute coordinates for the touch input are calculated with those. This PR provides all of the values as best as possible (and also retries should pipewire negotiate a different resolution) :

1. offset_x/offset_y and width/height are configured when setting up the configured stream (coming from implementation)
2. env_width/env_height are currently assumed to be equal to width/height (I haven't checked the calculationcode if it would be better to do this differently but it works fine after testing).
3. env_logical_width/env_logical_height is calculated similar to wlgrab by iterating all existing wl::monitors()
4. logical_width/logical_height are determined by matching above values from 1. to a wayland monitor during iteration in 3.

Implementations of pipewire_display_t are free to override verify_and_update_display_parameters() and provide a more fitting solution for the specific case if desired.

For testing this I've used moonlight-android on my phone streaming 2 monitors from my host pc using portalgrab or kwingrab.

Issue #5049, opened 2026-04-24
linux/pipewire: touch input goes to the wrong monitor on multi-monitor setup when using Moonlight QT


RevengeRip:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

When using portal/pipewire streaming backend touch input going to the left monitor in my 3-monitor configuration, when I stream from my middle monitor with explicitly set display ID (nDP-1). Yes, this happens on latest beta even after #5041.

### Host Operating System

Linux

### Operating System Version

Arch Linux

### Architecture

amd64/x86_64

### Package

Linux - pkg.tar.zst

### GPU Type

AMD

### GPU Model

RX 9070XT

### GPU Driver/Mesa Version

Mesa 26.0.5

### Capture Method

XDG Portal Grab (Linux)

### Apps

```json

```

### Log output

```shell

```

### Online logs

https://gist.github.com/RevengeRip/d271f532ccd7e099acb98fe863e53eb5

RevengeRip:
I want to clarify: touch input is wrong when using Moonlight-QT on my Steam Deck. On Moonlight Android touch absolute position works just fine, it gets to the right monitor (and right position), but input is registered as an absolute mouse move rather than touch event (Moonlight Qt correctly sends touch inputs).
Moonlight Qt: [2026-04-24 19:10:04.051]: Debug: --begin touch packet--
Moonlight Andorid: [2026-04-24 19:24:40.091]: Debug: --begin absolute mouse move packet--

RevengeRip:
For some reason absolute touch input treated differently on Moonlight Android and Moonlight QT.

Kishi85:
I'll have a look at it again and try to figure out what #5041 is still missing.

RevengeRip:
> I'll have a look at it again and try to figure out what [#5041](https://github.com/LizardByte/Sunshine/pull/5041) is still missing.

Thanks! Is it possible to treat Moonlight Android input as touch events, or it must be done on moonlight-android side?

Kishi85:
AFAIK that is something coming from the client so it has to be handled there.

If you can self-compile, could you test if compiling with #5050 applied fixes the issue?
I'm lacking a way to have absolute touch inputs with moonlight-qt so I cannot test/reproduce this (easily) atm.

[...]

PR #5053, opened 2026-04-26
feat(linux/xdgportal): Simplify display matching logic

Kishi85:
## Description

This simplifies the display name matching logic and also allows for wl_output names to be used directly for config parameter output_name to select the initial display. The latter I've noticed was possible due to a side-note in #5049. 

Note: If someone already figured out that output_name can be used with portalgrab by prefixing the wayland output name with 'n' (DP-1 getting selected by output_name=nDP-1) this will break as it will now require the output_name to be equal to the wayland output name (DP-1 being selected with output_name=DP-1 in config) which is a lot more self-explaining.

Additionally the log message for found display explicitly states the id/name parameter now so it somewhat matches the documented behaviour.

## Type of Change

- [X] **BREAKING CHANGE**: Introduces a breaking change (can be combined with any type above)

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/portalgrab.cpp
+++ b/src/platform/linux/portalgrab.cpp
@@ -96,42 +96,14 @@ namespace portal {
 
     std::string to_display_name() {
       if (!monitor_name.empty()) {
-        return std::format("n{}", monitor_name);
+        return monitor_name;
       }
-      return std::format("p{},{},{},{}", pos_x, pos_y, width, height);
+      return std::format("position-{}x{}-resolution-{}x{}", pos_x, pos_y, width, height);
     }
 
-    bool match_display_name(const std::string &display_name) const {
-      // Empty display_name never matches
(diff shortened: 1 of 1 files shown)


Kishi85:
This should also serve as a potential workaround for #221 for xdgportal by setting output_name to the desired primary display connector name.

@psyke83 Do you think we should also push the output_name from config to the top of the display_name list in that case (so it's always on CTRL+ALT+SHIFT+F1) or should we just let video.cpp select the display by configured name and leave the display name order (and therefore keyboard shortcuts) by position as it is currently?

psyke83:
I think that leaving the order by position seems logical but I really don't have a strong preference.

Kishi85:
I'll leave the current behaviour as it is then (sorting key order by position). This should be good to go then unless anyone finds an issue with it.

PR #5064, opened 2026-04-30
Add macOS left Alt to Command mapping option

bdjben:
## Description

This adds an opt-in macOS input setting, `key_leftalt_to_key_cmd`, that maps Moonlight's Left Alt key (`VKEY_LMENU`) to the left macOS Command key.

The mapping happens in the macOS platform key translation layer rather than through `keybindings`. This matters because `keybindings = [0xA4, 0x5B]` remaps the key before Sunshine updates its internal modifier state; later key packets that still carry the Alt modifier can then cause Sunshine to synthesize Option in addition to Command. By leaving the core keycode as `VKEY_LMENU` and only changing the macOS virtual key emitted to CoreGraphics, Sunshine continues to track Alt internally while macOS receives Command.

This is useful for macOS hosts accessed from clients or operating systems that intercept the Windows key before Moonlight can send it, but can still send Left Alt.

## Changes

- Adds `key_leftalt_to_key_cmd`, default disabled.
- Applies the remap only in `src/platform/macos/input.cpp`.
- Adds the option to the web UI for macOS hosts.
- Updates configuration docs and English locale strings.

## Testing

- `git diff --check`
- `jq empty` on updated English locale JSON files
- `npm install --no-package-lock --no-audit --no-fund`
- `npm run build`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -614,6 +614,30 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### key_leftalt_to_key_cmd
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            On macOS, map the Left Alt key from Moonlight to the left Command key. This can be useful when the client cannot send the Windows key directly.
+            @caution{Applies to macOS only.}
+        </td>
+    </tr>
+    <tr>
+        <td>Default</td>
+        <td colspan="2">@code{}
+            disabled
+            @endcode</td>
+    </tr>
+    <tr>
+        <td>Example</td>
+        <td colspan="2">@code{}
+            key_leftalt_to_key_cmd = enabled
+            @endcode</td>
+    </tr>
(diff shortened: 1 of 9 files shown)


ReenigneArcher:
This https://docs.lizardbyte.dev/projects/sunshine/master/md_docs_2configuration.html#key_rightalt_to_key_win doesn't work?

bdjben:
> This https://docs.lizardbyte.dev/projects/sunshine/master/md_docs_2configuration.html#key_rightalt_to_key_win doesn't work?

It doesn't address Windows clients accessing Mac hosts.

ReenigneArcher:
Should probably just re-use that option though. Of course instead of mapping to "win" key it would map to "cmd"

bdjben:
Remapping left alt makes much more sense in terms of key placement, and one might as well keep it on the left, since the only reason the windows-to-windows setup used right alt to map to win was because win (which is on the left) didn't work at all.

PR #4857, opened 2026-03-16
feat(linux/glad): implement EGL_IMG_context_priority

psyke83:
## Description

Resolves encoding throttling during high GPU usage:
* Try to create EGL context with high priority if supported.
* Set CAP_SYS_NICE capability (needed for AMDGPU/NVIDIA drivers at minumum).
* This will not work for Flatpak, AppImage or Portal capture due to the
  CAP_SYS_NICE capability not being available.

Currently has no effect on AMDGPU with Flatpak, AppImage and/or with
Portal capture. Other GPUs may behave differently.

### Issues Fixed or Closed

- Resolves: #4693
- Closes: #4745

## Type of Change

- [x] **feat**: New feature (non-breaking change which adds functionality)
- [x] **fix**: Bug fix (non-breaking change which fixes an issue)
- [ ] **docs**: Documentation only changes
- [ ] **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semicolons, etc.)
- [ ] **refactor**: Code change that neither fixes a bug nor adds a feature
- [x] **perf**: Code change that improves performance
- [ ] **test**: Adding missing tests or correcting existing tests
- [ ] **build**: Changes that affect the build system or external dependencies
- [ ] **ci**: Changes to CI configuration files and scripts
- [ ] **chore**: Other changes that don't modify src or test files
- [ ] **revert**: Reverts a previous commit
- [ ] **BREAKING CHANGE**: Introduces a breaking change (can be combined with any type above)

## Checklist

- [x] Code follows the style guidelines of this project
- [x] Code has been self-reviewed
- [ ] Code has been commented, particularly in hard-to-understand areas
- [ ] Code docstring/documentation-blocks for new or existing methods/components have been added or updated
- [ ] Unit tests have been added or updated for any new or modified functionality

### AI Usage

- [x] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [ ] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

PR #5075, opened 2026-05-05
fix(linux): security: drop CAP_SYS_ADMIN when possible, retain CAP_SYS_NICE

psyke83:
**fix(linux/kwingrab): retain CAP_SYS_NICE capabilities**
Since KWin acts as the DRM master, it has CAP_SYS_NICE capabilities:

$ cat /proc/$(pidof kwin_wayland)/status | grep Cap
CapInh: 0000000800000000
CapPrm: 0000000000800000
CapEff: 0000000000800000
CapBnd: 000001ffffffffff
CapAmb: 0000000000000000
$ capsh --decode=0000000000800000
0x0000000000800000=cap_sys_nice

Unlike XDG Portal, retaining this capability in Sunshine does not interfere with KWin's ability to access /proc/pid/root when required. Keeping CAP_SYS_NICE is desirable to allow high priority EGL context creation for the VAAPI encoder.

Change {has,drop}_elevated_privileges() to accept a bool that specifies whether to check/drop all caps or only CAP_SYS_ADMIN, and ensure that kwingrab only drops CAP_SYS_ADMIN whilst keeping xdgportal's behaviour unchanged.

**fix(linux): security: drop CAP_SYS_ADMIN as soon as possible**
Reorder Linux capture method initialization to first check KMS capture.
If KMS is passed over, immediately drop CAP_SYS_ADMIN but leave other
caps (i.e. CAP_SYS_NICE) intact so that EGL high priority contexts
are still possible.

Remove kwingrab capability dropping, as CAP_SYS_ADMIN will already be
dropped after probe, and CAP_SYS_NICE does not interfere with /root access.

Retain portalgrab capability check/drop, as CAP_SYS_NICE must be dropped
to allow /root access from the XDG Portal service.

Increases security for all all non-KMS capture methods, as
CAP_SYS_ADMIN will no longer be present in the permitted set and thus cannot
be elevated by malicious actors during runtime.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/kwingrab.cpp
+++ b/src/platform/linux/kwingrab.cpp
@@ -675,8 +675,8 @@ namespace platf {
     }
 
     // Drop CAP_SYS_ADMIN so KWin's permission check (if active) can see and match the executable
-    if (!kwin::screencast_permission_helper_t::is_permission_system_deactivated() && has_elevated_privileges()) {
-      drop_elevated_privileges();
+    if (!kwin::screencast_permission_helper_t::is_permission_system_deactivated() && has_elevated_privileges(false)) {
+      drop_elevated_privileges(false);
     }
 
     auto display = std::make_shared<kwin::kwin_t>();
@@ -688,7 +688,7 @@ namespace platf {
   }
 
   std::vector<std::string> kwin_display_names() {
-    if (!kwin::screencast_permission_helper_t::is_permission_system_deactivated() && has_elevated_privileges()) {
+    if (!kwin::screencast_permission_helper_t::is_permission_system_deactivated() && has_elevated_privileges(false)) {
       // We're still in the probing phase of Sunshine startup. Dropping portal security early will break KMS.
       // Just return a dummy screen for now. Display re-enumeration after encoder probing will yield full result.
       std::vector<std::string> display_names;
--- a/src/platform/linux/kwingrab.cpp
+++ b/src/platform/linux/kwingrab.cpp
@@ -674,11 +674,6 @@ namespace platf {
       return nullptr;
     }
 
-    // Drop CAP_SYS_ADMIN so KWin's permission check (if active) can see and match the executable
-    if (!kwin::screencast_permission_helper_t::is_permission_system_deactivated() && has_elevated_privileges(false)) {
-      drop_elevated_privileges(false);
-    }
-
     auto display = std::make_shared<kwin::kwin_t>();
     if (display->init(hwdevice_type, display_name, config)) {
       return nullptr;
@@ -688,7 +683,7 @@ namespace platf {
   }
 
   std::vector<std::string> kwin_display_names() {
-    if (!kwin::screencast_permission_helper_t::is_permission_system_deactivated() && has_elevated_privileges(false)) {
+    if (has_elevated_privileges(false)) {
       // We're still in the probing phase of Sunshine startup. Dropping portal security early will break KMS.
       // Just return a dummy screen for now. Display re-enumeration after encoder probing will yield full result.
       std::vector<std::string> display_names;
--- a/src/platform/common.h
+++ b/src/platform/common.h
@@ -895,13 +895,15 @@ namespace platf {
 
   /**
    * @brief Check is the current process is running with elevated privileges (e.g. system admin/etc.)
-   * @return True if system admin capabilities are present.
+   * @param all_caps Bool that specifies whether to check all caps or only CAP_SYS_ADMIN
+   * @return True if capabilities specified to be checked are present.
    */
-  bool has_elevated_privileges();
+  bool has_elevated_privileges(bool all_caps);
 
   /**
-   * @brief Drop elevated privileges (e.g. system admin/etc.)
+   * @brief Drop elevated privileges (e.g. system admin/nice etc.)
+   * @param all_caps Bool that specifies whether to drop all caps or only CAP_SYS_ADMIN
    */
-  void drop_elevated_privileges();
+  void drop_elevated_privileges(bool all_caps);
 
 }  // namespace platf
--- a/src/platform/linux/misc.cpp
+++ b/src/platform/linux/misc.cpp
@@ -1261,18 +1261,22 @@ namespace platf {
   }
 
 #if !defined(__FreeBSD__)
-  constexpr std::array<cap_value_t, 2> ELEVATED_PRIVILEGES_EFFECTIVE {CAP_SYS_ADMIN, CAP_SYS_NICE};
-  constexpr std::array<cap_value_t, 2> ELEVATED_PRIVILEGES_PERMITTED {CAP_SYS_ADMIN, CAP_SYS_NICE};
+  static constexpr cap_value_t FULL_CAPS[] = {CAP_SYS_ADMIN, CAP_SYS_NICE};
+  static constexpr cap_value_t ADMIN_CAPS[] = {CAP_SYS_ADMIN};
+
+  constexpr std::span<const cap_value_t> ELEVATED_PRIVILEGES_FULL {FULL_CAPS};
+  constexpr std::span<const cap_value_t> ELEVATED_PRIVILEGES_ADMIN {ADMIN_CAPS};
 #endif
 
-  bool has_elevated_privileges() {
+  bool has_elevated_privileges(bool all_caps) {
 #if !defined(__FreeBSD__)
+    const auto caps_to_check = all_caps ? ELEVATED_PRIVILEGES_FULL : ELEVATED_PRIVILEGES_ADMIN;
     const cap_t caps = cap_get_proc();
     if (!caps) {
       BOOST_LOG(error) << "[misc] has_elevated_privileges failed to get process capabilities."sv;
       return false;
     }
-    for (const auto c : ELEVATED_PRIVILEGES_EFFECTIVE) {
+    for (const auto c : caps_to_check) {
       cap_flag_value_t cap_flags_value;
       cap_get_flag(caps, c, CAP_EFFECTIVE, &cap_flags_value);
       if (cap_flags_value == CAP_SET) {
@@ -1280,7 +1284,7 @@ namespace platf {
         return true;
       }
     }
-    for (const auto c : ELEVATED_PRIVILEGES_PERMITTED) {
+    for (const auto c : caps_to_check) {
       cap_flag_value_t cap_flags_value;
       cap_get_flag(caps, c, CAP_PERMITTED, &cap_flags_value);
       if (cap_flags_value == CAP_SET) {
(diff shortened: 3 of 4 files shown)


psyke83:
Pinging @Kishi85 for feedback; I would've suggested this earlier, but I didn't have time to test the latest changes to your kwingrab PR until now - apologies.

This change should give your capture method an advantage over xdgportal for users that still use VAAPI encoding by making it compatible with high priority EGL contexts (see #4857).

Vulkan also needs this capability to create high priority contexts, but we don't implement VK_EXT_global_priority yet, and testing has shown that it's not necessary to avoid encoder throttling (at least on AMD GPUs). Nevertheless, keeping CAP_SYS_NICE active means that Vulkan high priority context is also an option in future if other vendors' GPUs need it to avoid throttling.

One change that might make this look more human readable would be to change the bool to a more verbose enum class, but I was trying to keep the changes to a minimum.

Kishi85:
Thanks for improving kwingrab further @psyke83 

Looking good to me and after a quick test with this change (both vaapi and vulkan) everything seems to be in order on my end (connection establishment is subjectively quicker than before). 

I'd also leave it as a bool switch for now until there's a need for a more differentiated approach.

psyke83:
@Kishi85

I wonder if we should remove the is_permission_system_deactivated() condition check specific to capabilities and unconditionally try to drop CAP_SYS_ADMIN for kwingrab? This would reduce the attack surface if potential future exploits try to leverage the resident CAP_SYS_ADMIN CapPrm capability when using the KWIN_WAYLAND_NO_PERMISSION_CHECKS bypass.

Kishi85:
> @Kishi85
> 
> I wonder if we should remove the is_permission_system_deactivated() condition check specific to capabilities and unconditionally try to drop CAP_SYS_ADMIN for kwingrab? This would reduce the attack surface if potential future exploits try to leverage the resident CAP_SYS_ADMIN CapPrm capability when using the KWIN_WAYLAND_NO_PERMISSION_CHECKS bypass.

I've added that specifically to not even try to drop privileges in case the environment variable to deactivate the permission system is found and active. There is no need to drop them in that case and trying to drop might lead to errors (e.g. as described in RamaLamas original PR description #4724 when using LXC). Temporary permission files are also not created when the environment variable is found set.

I'll have to think more about possible exploits but as far as those are concerned kmsgrab will likely have the same issues.

psyke83:
The issue is that the sunshine binary is preinstalled with the cap_sys_admin+p (permitted) capability, so if some kind of exploit (such as a memory corruption bug) is found that allows arbitrary execution, they may be able to fork a process that also has that capability set and use it to immediately raise the capability into the effective set and gain root access.  If we drop the capability from the permitted and effective set, it can't be re-gained during runtime, so this would help reduce potential attack surface area for malicious actors.

This security issue can't be avoided for kmsgrab users, but this is part of the reason why kmsgrab is an ugly hack that should be phased out. The CAP_SYS_NICE capability is also risky, but much less so, as the worst that it could do is set realtime priority and potentially lock up the system.

ReenigneArcher:
Arch failure is not related to this PR. Something to do with CUDA, haven't looked into it yet.

Kishi85:
> The issue is that the sunshine binary is preinstalled with the cap_sys_admin+p (permitted) capability, so if some kind of exploit (such as a memory corruption bug) is found that allows arbitrary execution, they may be able to fork a process that also has that capability set and use it to immediately raise the capability into the effective set and gain root access. If we drop the capability from the permitted and effective set, it can't be re-gained during runtime, so this would help reduce potential attack surface area for malicious actors.
> 
> This security issue can't be avoided for kmsgrab users, but this is part of the reason why kmsgrab is an ugly hack that should be phased out. The CAP_SYS_NICE capability is also risky, but much less so, as the worst that it could do is set realtime priority and potentially lock up the system.

I see, in that case removing the permission system check for the drop instance would be an easy way to reduce attack surface (it also shouldn't break things as we're just warning about not being able to drop):
https://github.com/LizardByte/Sunshine/blob/017ce4964ad3e8aa57ac8c4086d93aaf363edf70/src/platform/linux/kwingrab.cpp#L677-L680

I'm a bit unsure about the the display name enumeration as that would improve kwingrab's availability check to also include available displays and not only kwin presence in case the environment variable is set when CAP_SYS_ADMIN is also present:
https://github.com/LizardByte/Sunshine/blob/017ce4964ad3e8aa57ac8c4086d93aaf363edf70/src/platform/linux/kwingrab.cpp#L691-L697

Your opinion? Also should I put the change up as a separate PR or do you want to add it to this one?

psyke83:
@Kishi85 

Please try the updated PR with the new commit and let me know if you see any issues. Leaving as draft for the moment as I can't test thoroughly until later tonight.

This approach should make all non-kmsgrab capture methods automatically drop CAP_SYS_ADMIN from the capability sets. This simplifies kwingrab a bit, as only portalgrab needs the extra step of dropping CAP_SYS_NICE to retain /root access to the XDG Portal service.

src/platform/linux/kwingrab.cpp
```diff
@@ -688,14 +683,6 @@ namespace platf {
   }
 
   std::vector<std::string> kwin_display_names() {
-    if (!kwin::screencast_permission_helper_t::is_permission_system_deactivated() && has_elevated_privileges(false)) {
```
Kishi85: This needs to be kept but just with `if (has_elevated_privileges(false))` so a dummy display is returned in case kmsgrab encoder probing has not finished. display_names are gathered for all available capture methods before (kmsgrab) encoder probing.

Kishi85: If kwingrab does not return a dummy in case of elevated privileges during initial display_names enumeration it will be marked as unavailable for auto detection otherwise:
https://github.com/LizardByte/Sunshine/blob/017ce4964ad3e8aa57ac8c4086d93aaf363edf70/src/platform/linux/misc.cpp#L1028-L1031

psyke83: Should be fixed in the latest refresh, thanks.

Kishi85:
[9d5f790](https://github.com/LizardByte/Sunshine/commit/9d5f7904461fbd01bf332f9c23f5c0575266be17) looks good to me. Tested with capture=
 - auto => yields kmsgrab (with default sunshine binary capabilites: CAP_SYS_NICE+CAP_SYS_ADMIN) and works as expected
 - kmsgrab => works as expected
 - portalgrab => works as expected
 - kwingrab => works as expected

Should be a good overall security gain as CAP_SYS_ADMIN is dropped as soon as it is unnecessary now (when kmsgrab not used).

psyke83:
@Kishi85 

Thanks for testing. This looks ready for review to me as well.

psyke83:
@ReenigneArcher 

Sorry, I haven't been following CUDA changes. Is this the issue?

A previous successful run showed:
```
2026-05-02T17:50:22.7662278Z -- Looking for a CUDA compiler
2026-05-02T17:50:24.9716711Z -- Looking for a CUDA compiler - /opt/cuda/bin/nvcc
```

Current runs:
```
2026-05-05T20:18:08.4719124Z -- Looking for a CUDA compiler
2026-05-05T20:18:08.7292364Z -- Looking for a CUDA compiler - NOTFOUND
```

On my system:
```
$ pacman -Qi cuda
Name            : cuda
Version         : 13.2.1-1
Description     : NVIDIA's GPU programming toolkit
Architecture    : x86_64
URL             : https://developer.nvidia.com/cuda-zone
Licenses        : LicenseRef-NVIDIA-CUDA
Groups          : None
Provides        : cuda-toolkit  cuda-sdk  libcudart.so=13-64  libcublas.so=13-64  libcublas.so=13-64  libcusolver.so=12-64  libcusolver.so=12-64  libcusparse.so=12-64
                  libcusparse.so=12-64
Depends On      : opencl-nvidia  python  gcc
Optional Deps   : gdb: for cuda-gdb [installed]
                  glu: required for some profiling tools in CUPTI [installed]
                  nvidia-utils: for NVIDIA drivers (not needed in CDI containers)
                  nsight-compute: for profiling CUDA applications
                  nsight-systems: for system-wide profiling
                  rdma-core: for GPUDirect Storage (libcufile_rdma.so)
Required By     : None
Optional For    : None
Conflicts With  : None
Replaces        : cuda-toolkit  cuda-sdk  cuda-static  cuda-tools
Installed Size  : 4.74 GiB
Packager        : Jakub Klinkovský <lahwaacz@archlinux.org>
Build Date      : Mon 13 Apr 2026 21:25:25
Install Date    : Tue 05 May 2026 23:01:54
Install Reason  : Explicitly installed
Install Script  : Yes
Validated By    : Signature

[conn@archlinux sunshine]$ pacman -Ql cuda | grep /opt/cuda/bin
cuda /opt/cuda/bin/
cuda /opt/cuda/bin/__nvcc_device_query
cuda /opt/cuda/bin/bin2c
cuda /opt/cuda/bin/compute-sanitizer
cuda /opt/cuda/bin/crt/
cuda /opt/cuda/bin/crt/link.stub
cuda /opt/cuda/bin/crt/prelink.stub
cuda /opt/cuda/bin/ctadvisor
cuda /opt/cuda/bin/cu++filt
cuda /opt/cuda/bin/cuda-gdb
cuda /opt/cuda/bin/cuda-gdb-minimal
cuda /opt/cuda/bin/cuda-gdb-python3.10-tui
cuda /opt/cuda/bin/cuda-gdb-python3.11-tui
cuda /opt/cuda/bin/cuda-gdb-python3.12-tui
cuda /opt/cuda/bin/cuda-gdb-python3.8-tui
cuda /opt/cuda/bin/cuda-gdb-python3.9-tui
cuda /opt/cuda/bin/cuda-gdbserver
cuda /opt/cuda/bin/cudafe++
cuda /opt/cuda/bin/cuobjdump
cuda /opt/cuda/bin/fatbinary
cuda /opt/cuda/bin/nvcc
cuda /opt/cuda/bin/nvcc.profile
cuda /opt/cuda/bin/nvdisasm
cuda /opt/cuda/bin/nvlink
cuda /opt/cuda/bin/nvprune
cuda /opt/cuda/bin/ptxas
cuda /opt/cuda/bin/tileiras
[conn@archlinux sunshine]$ /opt/cuda/bin/nvcc -V
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2026 NVIDIA Corporation
Built on Thu_Mar_19_11:12:51_PM_PDT_2026
Cuda compilation tools, release 13.2, V13.2.78
Build cuda_13.2.r13.2/compiler.37668154_0
[conn@archlinux sunshine]$ cat /etc/profile.d/cuda.sh 
export CUDA_PATH=/opt/cuda
append_path '/opt/cuda/bin'
export PATH


# Set the default host compiler for nvcc. This will need to be switched back
# and forth between the latest and previous GCC version, whatever nvcc
# currently supports.
export NVCC_CCBIN='/usr/bin/g++'

```

ReenigneArcher:
@psyke83 looks like it: https://github.com/LizardByte/Sunshine/actions/runs/25400067396/job/74500865067#step:11:933

I didn't see an update to the cuda package for over a month on Arch, but looks like gcc was updated to 16 which isn't compatible I guess. There's no gcc15 package available as far as I can see, so I guess we're just stuck until they fix the packaging in Arch.

Signynt:
It seems Arch resolved the issue, after updating this morning I was able to build the Arch package.

ReenigneArcher:
Yes gcc15 now exists in Arch packages, but another issue with our CI being resolved by https://github.com/LizardByte/Sunshine/pull/5081

PR #5101, opened 2026-05-12
build(macos): configure C++ standard and ICU root

martona:
## Description

Pass the project C++ 23 standard explicitly so fetched dependencies build against modern headers. C++23 was chosen because it's already used for the `sunshine` target in `cmake/targets/common.cmake` and by the Homebrew formula in `packaging/sunshine.rb`, but those don't apply early enough to dependencies built through `scripts/macos_build.sh`. This can break when Boost falls back to FetchContent.

Point CMake at Homebrew's icu4c@78 prefix to match the dependency installed by the macOS build path. Without this, CMake can pick up inconsistent ICU headers/configuration from the Apple SDK or system paths while Boost.Locale is built from source.

Mirror the manual macOS build script in CI so the two paths stay aligned, as requested in the script comments.

## Type of Change

## Checklist

- [ ] Code has been self-reviewed

### AI Usage


ReenigneArcher:
I'm confused about setting CXX_STANDARD here. We already set it in cmake itself, and we always use fetch content in our CI builds, so I'm not sure this is actually an issue.

martona:
The existing CXX_STANDARD setting applies only to the `sunshine` target, and it is set after `dependencies/common.cmake` has already added the FetchContent targets.

In my local build, Homebrew had Boost 1.90 installed, while the project requested exactly 1.89. CMake rejected the newer Homebrew Boost and fell back to FetchContent, so Boost.Locale was built as part of the project. That fetched Boost target then compiled against Homebrew’s ICU headers without a C++17+ default and failed on ICU headers using constructs like `auto` non-type template parameters and `std::is_same_v`.

Passing `-DCMAKE_CXX_STANDARD=23` at configure time makes the project default visible before FetchContent targets are created. I used 23 specifically because that is already the standard requested for the Sunshine target and the Homebrew packaging path.

That said, if you prefer, setting the project-wide C++ standard near the top-level CMake before dependencies are loaded would be cleaner than passing it from the macOS scripts.

I also wouldn't be offended if you chose to reject as it is very much an edge case and easy to work around with `export`s in `.zshrc` but I doubt I'm the only one to ever run into this.

ReenigneArcher:
I think this can probably affect Linux as well, so probably we could just set it in the top level CMakelists.txt as well as in the tests.

martona:
Right, it probably does. Do you want me to take a crack at it?

ReenigneArcher:
> Right, it probably does. Do you want me to take a crack at it?

Yes please!

martona:
Done. It builds cleanly on macOS, I did not test elsewhere.

martona:
And for good measure I updated `/tests` and `/tools` as well. You did not ask for the latter, but it makes sense since they were redundant. (I missed them earlier, apologies for all the noise.)

martona:
The Linux failure was caused by setting the project C++ standard globally without also setting the CUDA standard before the `sunshine` target is created. Since `sunshine` includes CUDA sources on Linux, CMake tried to configure that target with CUDA C++23, which nvcc does not support.

I fixed that by setting `CMAKE_CUDA_STANDARD` to 17 at the top level before target creation, then removed the later CUDA standard setup from `cmake/targets/common.cmake`. Linux CUDA now builds locally.

CMakeLists.txt
```diff
@@ -8,6 +8,13 @@ project(Sunshine VERSION 0.0.0
         DESCRIPTION "Self-hosted game stream host for Moonlight"
         HOMEPAGE_URL "https://app.lizardbyte.dev/Sunshine")
 
+set(CMAKE_CXX_STANDARD 23)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+if(NOT DEFINED CMAKE_CUDA_STANDARD)
+    set(CMAKE_CUDA_STANDARD 17)
+    set(CMAKE_CUDA_STANDARD_REQUIRED ON)
+endif()
+
```
ReenigneArcher, line 17: I think this should be explicitly set in tests and tools as well.

tools doesn't need cuda though.

martona: Tools does not currently enable CUDA, but I set CMAKE_CUDA_STANDARD there as well so each standalone CMake entry point declares the same language standard defaults. It is inert today, but avoids reintroducing the same issue if CUDA-backed tools are added later. Let me know if you want it removed.

ReenigneArcher (approved):

ReenigneArcher:
It's either a coincidence or this PR caused https://github.com/LizardByte/Sunshine/issues/5112

Do you have some idea why? It's not obvious to me looking at the changes here.

martona:
I doubt it, the `icu4c` change was local to macOS, and the C++ standard setting should not cause DLL linkage errors. I'm looking though, stay tuned.

ReenigneArcher:
This was the only change though, unless the bug is in icu itself, which I didn't see any recent changes in.

martona:
My windows build box is slow as hell and churning through git checkout recursive right now - i'll know more in maybe half an hour, but my working theory is that none of the mingw toolchain is pinned; something updated and caused a new icu dependency. Plausible?

ReenigneArcher:
Seems a bit odd as the last update was 2 weeks ago. https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-icu and all that changed was the pkgrel version.

martona:
The more likely path is Boost. MSYS2 currently installs Boost 1.91, but Sunshine asks CMake for exactly Boost 1.89, so the system package is rejected and we fall back to FetchContent. After #5101, `CMAKE_CXX_STANDARD=23` is set globally before dependencies are added, so fetched CMake projects see that setting too. That was intentional, but it means the fetched Boost build environment changed.

I do not yet know that C++23 specifically toggled the ICU dependency, but it's looking more likely with FetchContent being in play.

martona:
The good news is that Windows builds and runs cleanly with:

cmake -B build -G Ninja -S . -DBOOST_LOCALE_ENABLE_ICU=OFF ...

I also checked the codebase for actual Boost.Locale usage. The only real call site I found is Linux-only:

src/platform/linux/input/inputtino_keyboard.cpp

auto utf8_str = boost::locale::conv::to_utf<wchar_t>(utf8, utf8 + size, "UTF-8");
auto utf32_str = boost::locale::conv::utf_to_utf<char32_t>(utf8_str);

So Windows does not appear to need Boost.Locale’s ICU backend. macOS may be similar, but I’d keep the first fix scoped to Windows unless you’d prefer a broader cleanup.

Would you like me to work on a PR? I should be able to get it done today.

Edit: Nevermind, I see you beat me to it.

Issue #4377, opened 2025-10-31
Artifacts when streaming with HDR enabled in Moonlight


Danik1601:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

None

### Describe the Bug

After installing the latest Radeon Software 25.10.2 all the white color look like horror when HDR is enabled in Moonlight. And when I toggle HDR off in Windows during streaming everything starts to look washed out in addition. If the HDR was set to off in Windows before streaming is started everything is fine regardless the Moonlight settings.
I suppose this has got something to do with AMF 1.5.0 that came with that version since everything was great on Radeon Software 25.9.1.
The only setting I ever touched in Sunshine is UPnP. I set it to enabled.

Steps to reproduce:
- Update host to AMD Software 25.10.2
- Enable HDR in Windows
- Enable HDR in Moolight.
- Start streaming in HEVC or AV1.

<img width="2556" height="1353" alt="Image" src="https://github.com/user-attachments/assets/9614b3c1-0a0b-40fd-a1a9-7f7cc73ac602" />

P.S. My client device has an SDR display, I enable HDR in Moonlight so I don't have to switch it in Windows each time.

### Host Operating System

Windows

### Operating System Version

Windows 11 25H2

### Architecture

amd64/x86_64

### Sunshine commit or version

2025.924.154138

### Package

Windows - installer (recommended)

### GPU Type

AMD

### GPU Model

Radeon RX 7800XT

### GPU Driver/Mesa Version

25.10.2

### Capture Method

None

### Config

```Shell

```

### Apps

```json

```

### Relevant log output

```shell
none
```

Danik1601:
https://github.com/moonlight-stream/moonlight-qt/issues/1444 may be relevant

Danik1601:
> And when I toggle HDR off in Windows during streaming everything starts to look washed out in addition

That might be a hardware issue with my specific device (Framework Laptop 13 AMD with 2,8k display).
[Here is why I think so](https://community.frame.work/t/2-8k-power-saving-colour-saturation-issue/78017).
I will try to test it further later.

ReenigneArcher:
Can you test https://github.com/LizardByte/Sunshine/pull/5106

Issue #3500, opened 2024-12-29
Having the Sunshine service running can impact HDR display capabilities in other applications with no streaming being involved


Skyyblaze:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

Having the Sunshine service running causes other applications such as Chiaki4Deck to end up with washed out HDR colors. Stopping Sunshine remedies the problem completely.

### Expected Behavior

Sunshine not interfering with other applications.

### Additional Context

I've been long since trying to figure out why HDR wouldn't display correctly when using Chiaki4Deck to stream my PS5 to my Windows PC and the developer of that application was trying to help me fix it too to no avail. Another user, also with an Nvidia card and Sunshine installed, eventually found out that stopping Sunshine fixes the problem. I tested it on my end and can confirm that the same is true for me. Starting Sunshine again immediately recreates the problem. The problem can also be seen when playing some HDR videos with the mpv video player in certain configurations.

The corresponding Issue Report can be found here: https://github.com/streetpea/chiaki-ng/issues/494 

### Host Operating System

Windows

### Operating System Version

Windows 11 24H2

### Architecture

amd64/x86_64

### Sunshine commit or version

0.23.1

### Package

Windows - installer (recommended)

### GPU Type

NVIDIA

### GPU Model

RTX 4070

### GPU Driver/Mesa Version

566.45

### Capture Method

None

### Config

```Shell

```

### Apps

```json

```

### Relevant log output

```shell
Not really applicable as Sunshine isn't actively involved.

[2024:12:29:20:29:21]: Info: Sunshine version: v0.23.1
[2024:12:29:20:29:21]: Info: nvprefs: No need to modify application profile settings
[2024:12:29:20:29:21]: Info: nvprefs: Changed OGL_CPL_PREFER_DXPRESENT to OGL_CPL_PREFER_DXPRESENT_PREFER_ENABLED for base profile
[2024:12:29:20:29:21]: Info: Compiling shaders...
[2024:12:29:20:29:21]: Info: System tray created
[2024:12:29:20:29:21]: Info: Compiled shaders
[2024:12:29:20:29:21]: Info: // Testing for available encoders, this may generate errors. You can safely ignore those errors. //
[2024:12:29:20:29:21]: Info: Trying encoder [nvenc]
[...]

Issue #1853, opened 2023-11-19
Blown out color streamed to SDR clients if HDR enabled without a HDR monitor (dummy plug, etc)


XenHat:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

Sunshine seems to be sending an HDR stream (or a SDR stream with the wrong color space/range) to SDR clients if the HDR-capable monitor disappears or is turned off without disabling HDR in the operating system.

Steps to reproduce are still a Work In Progress, but I'll do my best to describe my use case, and update as more details come to light.

If HDR was previously enabled in Windows 11 on a HDR-capable display (Samsung Odyssey G7) and the monitors sleep/turn off/disconnect (in this case, voluntarily, as I use a capture card as the fallback monitor), the video stream sent to a SDR moonlight client will have washed out colors.

Turning HDR off manually ~~does not fix the issue~~ fixes the issue in the Nightly release, but must still be manually turned off.

Here are some screenshots of the stream as visible by the moonlight client (SDR only):

With the HDR monitor connected and used:
<img width="1280" alt="HDR Monitor with HDR on" src="https://github.com/LizardByte/Sunshine/assets/980408/f6072b21-5b0d-4d52-aa15-f6c606d83f76">

With the capture card as the only monitor:
<img width="1280" alt="Capture card with HDR on" src="https://github.com/LizardByte/Sunshine/assets/980408/e6fb7c33-83e8-4c42-9c7d-9936860ee322">

With the capture card as the only monitor, after turning HDR off manually:
<img width="1280" alt="Capture card with HDR off" src="https://github.com/LizardByte/Sunshine/assets/980408/4fa72887-82e5-47e9-bd09-0a9fd7b3b71b">

Post-Scriptum: After updating to the Nightly build, it seems more and more evident that the issue is caused by Windows badly handling the disconnection/loss of an HDR-capable monitor without falling back to SDR mode.
I have decided to kept the issue as-is, for prosperity, searchability, and possible comments and discussion regarding the matter as I could not find an issue about this specific behaviour.

### Expected Behavior

Colors should stay within the SDR color range as supported by the client.

### Additional Context

My capture card is an Elgato HD60X connected via USB-C

Things to note:
* Windows seems to use the color profile for my HDR monitor on the capture card, which is wrong
[...]

PR #5106, opened 2026-05-13
fix(windows): correct shader selection for sRGB-encoded FP16 framebuffer (Win11 SDR)

eastginza:
## Description

Fixes white-out / blown-out highlights when streaming an SDR session on Windows 11 where the desktop framebuffer ends up in FP16 storage with gamma-encoded (sRGB / G22) pixel values.

### Symptom

This affects any Windows 11 22H2+ host where the active display causes the desktop to be composed in FP16 storage with gamma-encoded (sRGB / G22) values rather than linear scRGB. The same FP16 + G22 state is produced under multiple OS-level triggers:

- **Auto Color Management (ACM)** enabled on a regular display — the most commonly documented trigger; Win11 22H2+ feature whose default-enabled footprint has been expanding.
- **Virtual display EDID advertising HDR-class luminance and/or BT.2020 wide-gamut colorimetry** — common in remote-development / VM setups using Indirect Display Drivers (e.g. VDD-based virtual monitors). Even after stripping the HDR Static Metadata block from the EDID (reducing the advertised Max Luminance to SDR-class), Win11 can still promote the framebuffer to FP16 when the EDID continues to advertise BT.2020 wide-gamut colorimetry via the CTA Colorimetry Data Block. This happens even when ACM is disabled.
- Other Win11 build-specific behaviors that put the framebuffer in the same state.

The trigger is OS-level (Win11 composition behavior + Sunshine's existing FP16 shader assumptions); it is independent of GPU vendor, display type, host hardware, or whether the host runs on bare metal or in a VM. The affected user population is expected to grow as both ACM rollout and Sunshine-based remote-development setups (which often rely on virtual displays) become more common.

The Moonlight client shows the streamed desktop with whites and bright tones clipped against the top of the range. The affected code path in `display_vram.cpp` is the shader-selection logic shared by all five Windows capture formats (NV12, P010, AYUV, Y410, R16_UINT); the patch covers all five. Empirical verification in this report is the NV12 path (HEVC NVENC SDR session).

### Root cause

Under the triggers listed above, Windows 11 composites the desktop framebuffer in FP16 storage (`DXGI_FORMAT_R16G16B16A16_FLOAT`) even when both the application and the display are SDR. The values stored in that FP16 buffer are **still sRGB-encoded** (DXGI reports `DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709`), not linear scRGB.

The existing capture path in `display_vram.cpp` treats every FP16 source as linear scRGB and routes it through the linear shader chain. That chain ends in `convert_linear_base.hlsl`:

```hlsl
float3 CONVERT_FUNCTION(float3 input) {
    return ApplySRGBCurve(saturate(input));
}
```

This applies the sRGB encoding curve to values that **are already sRGB-encoded**, producing a double-encoded output. Highlights are crushed against the top of the range and the streamed image looks blown out on the client.

Log signature of an affected session:

```
Colorspace         : DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709
Bits Per Color     : 8
Max Luminance      : 270 nits
Desktop format     : DXGI_FORMAT_R16G16B16A16_FLOAT
[Error]: BT.2020 SDR colorspace expects 10-bit color depth, falling back to Rec. 709
Color coding: SDR (Rec. 709)
Color depth: 8-bit
```

Note that `Max Luminance` can be SDR-class (here, 270 nits) and the colorspace is `G22_NONE_P709` (gamma-encoded sRGB) — both are SDR signals — yet the `Desktop format` is still the FP16 buffer. That mismatch is the defining fingerprint of this bug class.

### Historical context

This change extends the case matrix opened by **#1178 (cgutman, 2023)**, which added the linear-FP16 capture path for **Advanced Color SDR** displays — the case where Windows delivers SDR content in FP16 storage as linear scRGB. Windows 11 22H2+ has since introduced a distinct FP16 SDR variant under multiple triggers (ACM, HDR-class virtual display EDID, etc.): the framebuffer still uses `DXGI_FORMAT_R16G16B16A16_FLOAT` storage, but the pixel values are **gamma-encoded (G22)**, not linear. The existing shader-selection logic does not distinguish between these two FP16 cases, so the gamma-encoded FP16 variant is incorrectly routed through the linear shader chain. This PR adds the discriminator (`is_source_gamma_encoded_fp16()`) that the new variant requires.

### Fix

Detect the gamma-encoded FP16 case via DXGI ColorSpace (`G22_NONE_P709` / `G22_NONE_P2020`) and route those sources through the **non-linear shader path** (whose `convert_base.hlsl` uses `saturate` only, with no additional sRGB curve).

- `src/platform/windows/display.h` — new helper `bool is_source_gamma_encoded_fp16()` declared alongside `is_hdr()`.
- `src/platform/windows/display_base.cpp` — implementation that queries `IDXGIOutput6::GetDesc1()` and returns true for `G22_NONE_P709` / `G22_NONE_P2020`.
- `src/platform/windows/display_vram.cpp` — every existing "FP16 → linear shader" decision now also checks `is_source_gamma_encoded_fp16()` and, when true, binds the non-linear shader to the FP16 slot. Covers all five capture formats (NV12 / P010 / R16_UINT / AYUV / Y410), nine call sites total.

### Backward compatibility

The new `is_source_gamma_encoded_fp16()` helper returns false unless the source delivers FP16 storage with `G22_NONE_P709` or `G22_NONE_P2020` colorspace. All other configurations — HDR streaming (G10 / G2084), Advanced Color SDR (the case fixed by #1178), pre-22H2 Windows installs, and non-FP16 capture paths — retain the existing shader-selection behavior unchanged. The helper also returns false when `IDXGIOutput6` is unavailable on the host, so older Windows/GPU configurations fall back safely to the existing path. There is **no behavioural change for HDR streaming**.

### Alternatives considered

**Shader-only fix (earlier iteration)**: an initial attempt simply removed `ApplySRGBCurve` from `convert_linear_base.hlsl`. That resolves the gamma-encoded FP16 case but **breaks the linear HDR path** (G10 / G2084 sources lose their sRGB encoding step). The current C++ approach keeps `convert_linear_base.hlsl` unchanged and routes only the gamma-encoded sources to the non-linear shader, leaving HDR streaming intact.

External workarounds were exhaustively evaluated by this contributor at the driver / OS / EDID layer and all failed empirically. They are documented here because reviewers seeing similar wash-out reports may otherwise assume one of them is sufficient:

- **Disabling Auto Color Management in Windows settings**: widely cited as the standard workaround, but **does not resolve the symptom when the active display's EDID continues to advertise HDR-class luminance or BT.2020 wide-gamut colorimetry**. In this contributor's environment, ACM was disabled and the FP16 + G22 capture state still persisted. So "just turn off ACM" is incomplete advice for any setup that involves a virtual display, and also costs the user the legitimate benefits of ACM elsewhere.
- **Stripping the HDR Static Metadata block from the virtual display's EDID** (reducing the advertised Max Luminance from HDR-class ~1670 nits to SDR-class 270 nits): tried via a custom EDID binary loaded by the VDD. The `Desktop format: DXGI_FORMAT_R16G16B16A16_FLOAT` state still persisted — most likely because the EDID continued to advertise BT.2020 wide-gamut colorimetry via the CTA Colorimetry Data Block, which is sufficient on its own to trigger the FP16 promote in Win11 22H2+.
- **Switching the VDD to its auto-generated EDID** (`<CustomEdid>false</CustomEdid>` in `vdd_settings.xml`): same outcome — the auto-generated EDID still produced the FP16 + G22 capture state.
- **Building a custom Indirect Display Driver to hand-author a minimal SDR-only EDID**: forking Microsoft's `IddSampleDriver`, signing it with a WDK test certificate, installing it, and promoting it to the primary display path. **Empirically confirmed: even with an SDR-only EDID, no CEA extension, `AdapterCaps.Flags = 0`, and the base IddCx 1.02 API, `DXGI_FORMAT_R16G16B16A16_FLOAT` capture persisted.** This is strong evidence that no driver- or EDID-layer fix can suppress the FP16 promote in this Win11 build family — the decision happens above the IddCx surface. Additionally, this approach requires Windows test signing mode (`bcdedit /set testsigning on`) and permanently displays a desktop watermark, making it unsuitable as an end-user workaround even if it had worked.
- **Forcing 10-bit HEVC encoding on the Moonlight client side**: empirically tested in this contributor's environment and **did not resolve the symptom**. The 10-bit capture path exercises a related shader chain that is affected by the same FP16 + G22 shader mis-routing, so the symptom carries over.

The exhaustive failure of all driver- / EDID- / encoder-layer workarounds is what narrowed the root cause to Sunshine's shader-selection logic. The C++ patch in this PR fixes that root cause once for all FP16 + G22 SDR sources, regardless of which OS-level trigger (ACM, HDR-class luminance EDID, BT.2020 wide-gamut EDID, or future Win11 behaviors) put the framebuffer in that state.

### Reproduction

1. Use Windows 11 22H2 or later, configured so the desktop framebuffer is composed in FP16 with G22 colorspace. Any of:
   a. **ACM trigger**: enable Auto Color Management (`Settings → System → Display → Color profile → Automatically manage colour for apps`); or
   b. **HDR-class luminance trigger**: run an Indirect Display Driver whose EDID advertises HDR-class max luminance (commonly seen with VDD-based virtual monitors used for remote development / VM setups), with ACM disabled; or
   c. **BT.2020 wide-gamut trigger**: run a display (physical or virtual) whose EDID advertises BT.2020 colorimetry via the CTA Colorimetry Data Block, even with HDR Static Metadata stripped and Max Luminance reduced to SDR-class.
2. Confirm in Sunshine's log (`config/sunshine.log`) that the `Desktop format` is `DXGI_FORMAT_R16G16B16A16_FLOAT` and the `Colorspace` is `DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709` (see "Log signature" above).
3. Connect from Moonlight to a Sunshine host capturing an SDR application or the desktop.
4. Compare the streamed desktop against a local view: whites and bright tones look "blown out" on the client.

Empirically verified on: Windows 11 25H2 with **ACM disabled** + a VDD-based virtual display whose custom EDID had its HDR Static Metadata block stripped (Max Luminance reduced from ~1670 nits to 270 nits, SDR-class) but still advertised BT.2020 wide-gamut colorimetry via the CTA Colorimetry Data Block, NVIDIA RTX 5090, AORUS FO32U2P 4K@240Hz QD-OLED, NV12 HEVC NVENC 8-bit Rec. 709 SDR session. The same `display_vram.cpp` shader-selection path is exercised by every FP16 + G22 capture regardless of which trigger produced that state, so the ACM-triggered variant, the HDR-class-luminance-triggered variant, and other GPU vendors (AMD/Intel) are expected to exhibit identical behavior. Several long-standing wash-out reports (e.g. #1853, #4377) describe similar symptoms on Windows but each has its own underlying trigger; this PR does not claim to resolve them, but reporters whose setup matches the FP16 + G22 SDR case documented above may find that this patch addresses their specific instance.

### Screenshot
Actual before (no Photoshop)
white-washed-actual

Actual after
white-washed-actual-patched

### Issues Fixed or Closed

(none — several long-standing wash-out reports exist on the tracker (e.g. #1853, #3500, #4377), but on close inspection each has a distinct root cause that is not directly fixed by this patch. They are mentioned here only for context, not for auto-close.)

#### Roadmap Issues

(none)

## Type of Change

## Checklist

(The two unchecked boxes: the affected path is the DXGI capture / shader-selection chain inside `display_vram.cpp`, which currently has no unit-test harness for branch coverage. Manual reproduction steps are documented above.)

### AI Usage

- [x] Moderate

### DIFF
--- a/src/platform/windows/display.h
+++ b/src/platform/windows/display.h
@@ -232,6 +232,13 @@ namespace platf::dxgi {
 
     virtual bool is_hdr() override;
     virtual bool get_hdr_metadata(SS_HDR_METADATA &metadata) override;
+    // Returns true when the capture source delivers FP16 storage with already
+    // gamma-encoded (sRGB) values, as happens with Windows 11 Auto Color
+    // Management (ACM) when an SDR application is displayed on an SDR monitor.
+    // In this case the linear shader path must NOT apply ApplySRGBCurve again
+    // (otherwise highlights are doubly encoded → white-out).
+    // Distinct from is_hdr() which checks for PQ HDR (G2084).
+    bool is_source_gamma_encoded_fp16();
 
     const char *dxgi_format_to_string(DXGI_FORMAT format);
     const char *colorspace_to_string(DXGI_COLOR_SPACE_TYPE type);
--- a/src/platform/windows/display_base.cpp
+++ b/src/platform/windows/display_base.cpp
@@ -766,6 +766,27 @@ namespace platf::dxgi {
     return desc1.ColorSpace == DXGI_COLOR_SPACE_RGB_FULL_G2084_NONE_P2020;
   }
 
+  bool display_base_t::is_source_gamma_encoded_fp16() {
+    dxgi::output6_t output6 {};
+
+    auto status = output->QueryInterface(IID_IDXGIOutput6, (void **) &output6);
+    if (FAILED(status)) {
+      return false;
+    }
+
+    DXGI_OUTPUT_DESC1 desc1;
+    output6->GetDesc1(&desc1);
+
+    // Windows 11 Auto Color Management (ACM) composes SDR desktops into an
+    // FP16 framebuffer but the pixel values remain gamma-encoded (G22).
+    // The "linear" shader path assumes linear scRGB FP16 (G10) and applies
+    // ApplySRGBCurve, which double-encodes the already-sRGB G22 values,
+    // crushing highlights to white. Use this method to route G22+FP16 input
+    // through the non-linear shader path (saturate only, no curve).
+    return desc1.ColorSpace == DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709
+        || desc1.ColorSpace == DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P2020;
+  }
+
   bool display_base_t::get_hdr_metadata(SS_HDR_METADATA &metadata) {
     dxgi::output6_t output6 {};
 
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -507,15 +507,28 @@ namespace platf::dxgi {
           // Semi-planar 8-bit YUV 4:2:0
           create_vertex_shader_helper(convert_yuv420_planar_y_vs_hlsl, convert_Y_or_YUV_vs);
           create_pixel_shader_helper(convert_yuv420_planar_y_ps_hlsl, convert_Y_or_YUV_ps);
-          create_pixel_shader_helper(convert_yuv420_planar_y_ps_linear_hlsl, convert_Y_or_YUV_fp16_ps);
+          if (display->is_source_gamma_encoded_fp16()) {
+            // ACM-SDR FP16: input is already sRGB-encoded; reuse non-linear shader to avoid double sRGB encoding (would cause highlight white-out)
+            create_pixel_shader_helper(convert_yuv420_planar_y_ps_hlsl, convert_Y_or_YUV_fp16_ps);
+          } else {
+            create_pixel_shader_helper(convert_yuv420_planar_y_ps_linear_hlsl, convert_Y_or_YUV_fp16_ps);
+          }
           if (downscaling) {
             create_vertex_shader_helper(convert_yuv420_packed_uv_type0s_vs_hlsl, convert_UV_vs);
             create_pixel_shader_helper(convert_yuv420_packed_uv_type0s_ps_hlsl, convert_UV_ps);
-            create_pixel_shader_helper(convert_yuv420_packed_uv_type0s_ps_linear_hlsl, convert_UV_fp16_ps);
+            if (display->is_source_gamma_encoded_fp16()) {
+              create_pixel_shader_helper(convert_yuv420_packed_uv_type0s_ps_hlsl, convert_UV_fp16_ps);
+            } else {
+              create_pixel_shader_helper(convert_yuv420_packed_uv_type0s_ps_linear_hlsl, convert_UV_fp16_ps);
+            }
           } else {
             create_vertex_shader_helper(convert_yuv420_packed_uv_type0_vs_hlsl, convert_UV_vs);
             create_pixel_shader_helper(convert_yuv420_packed_uv_type0_ps_hlsl, convert_UV_ps);
-            create_pixel_shader_helper(convert_yuv420_packed_uv_type0_ps_linear_hlsl, convert_UV_fp16_ps);
+            if (display->is_source_gamma_encoded_fp16()) {
+              create_pixel_shader_helper(convert_yuv420_packed_uv_type0_ps_hlsl, convert_UV_fp16_ps);
+            } else {
+              create_pixel_shader_helper(convert_yuv420_packed_uv_type0_ps_linear_hlsl, convert_UV_fp16_ps);
+            }
           }
           break;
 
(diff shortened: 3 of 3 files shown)


eastginza:
For reviewer context: this bug class has been reported on Reddit since at
least June 2024 with no upstream fix. Linking representative reports below
in case they help validate the symptom space addressed by this PR.

- "Host Win11, no HDR turned on, washed out" (2026-02, exact match):
  https://www.reddit.com/r/MoonlightStreaming/comments/1r9ph8j/
- "Sunshine on HDR monitor makes colors on client washed out" (2024-06):
  https://www.reddit.com/r/MoonlightStreaming/comments/1dpus3u/
- "HDR will not work, washed out colours" (2025-02):
  https://www.reddit.com/r/MoonlightStreaming/comments/1iw6plv/

I did not add `Fixes #XXXX` because none of the existing tracker issues
match this exact triggering condition (HDR-capable display + Windows HDR
toggle either state → FP16 sRGB-encoded framebuffer misinterpreted).

ReenigneArcher:
Can you add a before screenshot to show the actual condition, that isn't "Photoshop reconstruction"? Since you're on Windows it's easy to install another version of Sunshine along side the dev version.

eastginza:
Sure. Here is this.

Actual before (no Photoshop)
white-washed-actual

Actual after
white-washed-actual-patched

PR #5130, opened 2026-05-16
fix(linux/vulkan): preserve host aspect ratio in encoder output

hebo6:
## Description

Fixes a behavior inconsistency between the VAAPI and Vulkan encoders on Linux: when the encoder target resolution differs in aspect from the captured source, the VAAPI path letterboxes the source into the encoded frame while the Vulkan path stretches it to fill.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src_assets/linux/assets/shaders/vulkan/rgb2yuv.comp
+++ b/src_assets/linux/assets/shaders/vulkan/rgb2yuv.comp
@@ -40,9 +40,8 @@ void main() {
   ivec2 d = pos - pc.dst_offset;
   if (d.x < 0 || d.y < 0 || d.x >= pc.dst_size.x || d.y >= pc.dst_size.y) {
     // Outside the aspect-fit viewport: write encoded black.
-    // range_y/range_uv are (1,0) placeholders in Sunshine; the encoded
-    // offset is baked into color_vec_*.w (y_add / uv_add). For RGB=(0,0,0)
-    // the shader output reduces to color_vec_*.w, so write that directly.
+    // Use the same conversion formula as RGB=(0,0,0), where the dot product
+    // is zero and the encoded value is color_vec_*.w adjusted by range_*.
     float y_black = pc.color_vec_y.w * pc.range_y.x + pc.range_y.y;
     float u_black = pc.color_vec_u.w * pc.range_uv.x + pc.range_uv.y;
     float v_black = pc.color_vec_v.w * pc.range_uv.x + pc.range_uv.y;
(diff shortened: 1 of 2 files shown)


ReenigneArcher:
Thank you for the PR submission, but I noticed you didn't use our PR template.

Please update the PR to use the correct template. You can find it at https://github.com/LizardByte/.github/blob/master/.github/pull_request_template.md?plain=1

neatnoise (commented):
As the original author of the Vulkan encoder path — this looks correct and clean. Single dispatch, no extra passes or sync, negligible overhead.

A few notes:

1. `rgb2yuv.comp` L43-47: The black value computation assumes `range_y`/`range_uv` are `(1,0)`. Worth a comment noting this assumption so it doesn't break silently if those values change in the future.

2. `vulkan_encode.cpp` L115-116: Push constant struct layout changed (fields inserted in the middle) — fine since both C++ and GLSL are updated together, but worth being aware of for any future out-of-tree shader variants.

3. `vulkan_encode.cpp` L290-293: The `& ~1` rounding on `eff_w`/`eff_h` and offsets could theoretically leave a 1-pixel gap if frame dimensions are odd. Might want to assert or clamp that `dst_offset + eff_size <= frame_size`.

hebo6:
I clarified the black-fill comment to describe that it uses the normal RGB=(0,0,0) conversion formula, and clamped the aspect-fit size after even alignment so dst_offset + dst_size stays within the full frame while preserving UV-plane alignment.

The push constant layout is updated on both the C++ and GLSL sides in this PR, so I think that part is okay as-is.

neatnoise (commented):
It's fine for me

src_assets/linux/assets/shaders/vulkan/rgb2yuv.comp
```diff
@@ -15,7 +15,9 @@ layout(push_constant) uniform PushConstants {
   vec2 range_uv;
   ivec2 src_offset;
   ivec2 src_size;
-  ivec2 dst_size;
+  ivec2 dst_offset;     // top-left of the aspect-preserved viewport
+  ivec2 dst_size;       // effective viewport (src after aspect-fit)
+  ivec2 dst_full_size;  // full encoder frame size
```
ReenigneArcher:
```suggestion
  ivec2 dst_offset;  // top-left of the aspect-preserved viewport
  ivec2 dst_size;  // effective viewport (src after aspect-fit)
  ivec2 dst_full_size;  // full encoder frame size
```

ReenigneArcher (commented):
Just one nitpic

ReenigneArcher (approved):

Issue #5105, opened 2026-05-13
Video stream doesn't start with 10GB interface


punglist:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

First of all, thanks for the active dev on sunshine !
Running Linux cachyos and the sunshine-beta-bin AUR (aur/sunshine-beta-bin 2026.508.45922-1 (+5 0.70) (Installed)

I have 2 ethernet adapters in my PC:

```
 Device-1: Aquantia AQtion AQC100 NBase-T/IEEE 802.3an Ethernet [Atlantic
    10G] vendor: ASUSTeK XG-C100F 10GbE SFP+ driver: atlantic v: kernel
    port: N/A bus-ID: 06:00.0
  IF: enp6s0 state: up speed: 10000 Mbps duplex: full mac: <filter>
  
Device-2: Intel Ethernet I226-V vendor: ASUSTeK driver: igc v: kernel
    port: N/A bus-ID: 07:00.0
  IF: enp7s0 state: up speed: 1000 Mbps duplex: full mac: <filter>
```

If I use my 10GB fiber interface (enp6s0) audio streaming will work but not the video stream.
If I use my 1I226-V interface at 1GB (enp7s0) video and audio streamin will work perfectly.

I applied the following script as I am running kernel 7.0 (udp frag issue): 
```
ip -o link show | awk -F': ' '{print $2}' | while read -r iface; do
  ip4=$(ip -o -4 addr show dev "$iface" | awk 'NR==1{split($4,a,"/"); print a[1]}')
  ethtool --show-offload "$iface" 2>/dev/null | grep tx-udp-seg | sed "s/^/[$iface ${ip4:--}] /"
done
```
It doesn't solve the issue with the 10GB interface.  No log on sunshine side, except "CLIENT DISCONNECTED"

### Expected Behavior

it should work with the 10GB interface.

### Additional Context

  inxi -Fxz
System:
  Kernel: 7.0.6-1-cachyos arch: x86_64 bits: 64 compiler: clang v: 22.1.5
  Desktop: Hyprland v: 0.55.0 Distro: CachyOS base: Arch Linux
Machine:
  Type: Desktop System: ASUS product: N/A v: N/A serial: <superuser required>
[...]

Issue #5067, opened 2026-05-01
Black screen on Arch linux after upgrade to kernel 7.0


paju1986:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

After upgrading my arch linux to kernel 7.0 every time i connect to sunshine i get a black screen.
If i switch back to a 6.19 kernel or lower it works fine.
Im using kms capture on KDE with amd drivers
I'm using latest stable version of sunshine also tried with the beta with same result
here is full log:

[2026-05-01 17:40:45.197056] [0x00007f1ab96bac00] [info]    config: 'upnp' = enabled
[2026-05-01 17:40:45.197072] [0x00007f1ab96bac00] [info]    config: 'adapter_name' = /dev/dri/renderD128
[2026-05-01 17:40:45.197075] [0x00007f1ab96bac00] [info]    config: 'vk_rc_mode' = 4
[2026-05-01 17:40:45.197077] [0x00007f1ab96bac00] [info]    config: 'amd_quality' = default
[2026-05-01 17:40:45.197080] [0x00007f1ab96bac00] [info]    config: 'amd_rc' = auto
[2026-05-01 17:40:45.197081] [0x00007f1ab96bac00] [info]    config: 'encoder' = vaapi
[2026-05-01 17:40:45.197083] [0x00007f1ab96bac00] [info]    config: 'capture' = kms
[2026-05-01 17:40:45.197086] [0x00007f1ab96bac00] [info]    config: 'channels' = 2
[2026-05-01 17:40:45.197088] [0x00007f1ab96bac00] [info]    config: 'hevc_mode' = 2
[2026-05-01 17:40:45.197090] [0x00007f1ab96bac00] [info]    config: 'nv_rc' = auto
[2026-05-01 17:40:45.197092] [0x00007f1ab96bac00] [info]    config: 'fps' = [10,30,60,90,120,0]
[2026-05-01 17:40:45.197094] [0x00007f1ab96bac00] [info]    config: 'gamepad' = ds5
[2026-05-01 17:40:45.197096] [0x00007f1ab96bac00] [info]    config: 'output_name' = 0
[2026-05-01 17:40:45.197098] [0x00007f1ab96bac00] [info]    config: 'origin_pin_allowed' = lan
[...]

Issue #4895, opened 2026-03-23
Sunshine getting MTU wrong on virtual windows 11 machine that has been setup using a Virtio (paravirtualized) nic


stevedh42:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

This issue is present in the latest pre-release

### Describe the Bug

When Running Sunshine on a proxmox virtual machine that has been setup using a Virtio (paravirtualized) nic it tries to send the video using udp packets that are larger then 10k. The adapter MTU in the adapter is set to 1500. This results in all moonlight clients getting a 'Connection Terminated No video received from host' error.

On checking in wireshark, sunshine tries to send packets > 10k, when they arrive at moonlight they have been truncated to 1500.

I want to use this virtual adapter as it allows a greater then 1g connection. The physical nat is 2.5G, although the virtual adapter reports it as 10G.

Note that when I replace the virtio network adapter with a virtual intel 1g adapter it correctly gets the mtu size and moonlight can get the video steam fine.

C:\Users\steve>netsh interface ipv4 show interfaces

Idx     Met         MTU          State                Name
---  ----------  ----------  ------------  ---------------------------
  1          75        1450  connected     Loopback Pseudo-Interface 1
 11          15        1450  connected     Ethernet

C:\Users\steve>netsh interface ipv4 show subinterfaces

       MTU  MediaSenseState      Bytes In     Bytes Out  Interface
----------  ---------------  ------------  ------------  -------------
      1450                1             0         29395  Loopback Pseudo-Interface 1
      1450                1      22869248       7735395  Ethernet

PS C:\Users\steve>  $adapter = get-netadapter Ethernet
PS C:\Users\steve> $adapter.MtuSize
1500
PS C:\Users\steve> $adapter.DriverDescription
Red Hat VirtIO Ethernet Adapt

### Expected Behavior

Sunshine should send packets that fit within the MTU size, or provide an option that allows this to be set manually.

### Host Operating System

Windows

### Operating System Version

Edition	Windows 11 Pro Version	25H2 Installed on	‎07/‎03/‎2026 OS build	26200.8037 Experience	Windows Feature Experience Pack 1000.26100.300.0

### Architecture

amd64/x86_64

### Package

Windows - portable (NOT recommended)

### GPU Type

NVIDIA

### GPU Model

NVidia GTX 506

[...]

Issue #4295, opened 2025-09-24
UDP buffer overflow issue cannot be solved by Windows' built-in traffic shaping (QoS)


ss8688:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

None

### Describe the Bug

I am unable to connect to the host connected to the network through the hyper-v virtual ethernet adapter via local IP using Moonlight, but I can connect to the host normally through Tailscale. If I want to connect normally with a local IP, I need to switch to a physical network connection.Through the hyper-v virtual ethernet adapter is  not working.
I don't think this is a firewall issue because turning off the firewall is useless，also running network test shows that the network is normal. I also don't think this is a problem with the hyper-v virtual Ethernet adapter, because although it cannot be directly connected, it is normal to connect using Tailscale. It's not a tailcale VPN issue either, because turning it off doesn't work

<img width="2559" height="1439" alt="Image" src="https://github.com/user-attachments/assets/8876e08f-3c99-4a57-bfec-24aed9825194" />

<img width="1916" height="942" alt="Image" src="https://github.com/user-attachments/assets/08067246-3626-470a-94ed-46a1a5e33247" /> 

<img width="1818" height="914" alt="Image" src="https://github.com/user-attachments/assets/7fbda6b2-b8b4-4dbc-9d81-09a3112bff31" />

### Host Operating System

Windows

### Operating System Version

Windows server 2025

### Architecture

amd64/x86_64

### Sunshine commit or version

 Sunshine version: 2025.923.33222 commit: f22b00d6981f756d3531fba0028723d4a5065824

### Package

Windows - installer (recommended)

### GPU Type

Intel

### GPU Model

UHD730 & rtx3070

### GPU Driver/Mesa Version

irrelevant

### Capture Method

None

### Config

```Shell

```

### Apps

```json

```

### Relevant log output

```shell
[2025-09-24 17:38:07.716]: Info: Sunshine version: 2025.923.33222 commit: f22b00d6981f756d3531fba0028723d4a5065824
[2025-09-24 17:38:07.716]: Info: Package Publisher: LizardByte
[2025-09-24 17:38:07.716]: Info: Publisher Website: https://app.lizardbyte.dev
[2025-09-24 17:38:07.716]: Info: Get support: https://app.lizardbyte.dev/support
[...]

PR #5153, opened 2026-05-19
feat(rtsp): add option to limit packetsize for clients that cannot configure it

httpstorm:
## Description
Clients like Moonlight for Android/iOS cannot configure `packetsize`. When streaming over a transport with a low MTU like layer 2 VPN, the packets will be fragmented and there is a higher chance for packet loss and micro-stuttering.

Fix: add Sunshine configuration to limit the value of `packetsize` specified by the client. The default value of `0` disables the feature.

### Screenshot

Sunshine-packetsize 2026-05-19

## Type of Change

## Checklist

### AI Usage

Run tested: Ubuntu 24.04.4 LTS, Moonlight client on: Ubuntu, Android, iOS
Note: this is my first time opening a PR here, please let me know if any changes are needed.

@ReenigneArcher @cgutman @andygrundman @neatnoise

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1753,6 +1753,67 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### packetsize
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            Limit the packetsize to avoid fragmentation on a low MTU link.
+            @note{This helps avoid packet loss and micro-stutter on a layer 2 VPN with
+            clients that cannot configure this value, e.g. Moonlight for Android/iOS.
+            }
+            @tip{To discover the optimal value:
+            <ul>
+                <li>Send ping to the server with don't fragment flag (DF)</li>
+                <li>Find the size of the largest replay, and subtract 16</li>
+                <li>Monitor the traffic to ensure no fragmentation</li>
+            </ul>
+            If using a VPN tunnel:
+            <ul>
+                <li>Set MTU on the TUN/TAP interface, and</li>
+                <li>Ensure no fragmentation both inside and outside the tunnel</li>
+                <li>Max UDP size = MTU size - 28</li>
+                <li>`packetsize` = max UDP size - 16</li>
+                <li>Monitor the traffic to ensure no fragmentation</li>
+            </ul>
+            Sample calculation for OpenVPN layer 2, using IPv4:
+            <ul>
+                <li>1428 bytes for max ICMP/UDP size outside the tunnel</li>
+                <li>Subtract the OpenVPN overhead: 24 bytes (may vary)</li>
+                <li>1404 bytes for Ethernet inside the tunnel</li>
+                <li>Subtract the Ethernet header: 14 bytes</li>
+                <li>1390 bytes for MTU inside the tunnel</li>
+                <li>Subtract the IPv4 header: 20 bytes</li>
+                <li>Subtract the UDP header: 8 bytes</li>
+                <li>1362 bytes for UDP payload</li>
+                <li>Subtract: 16 bytes</li>
+                <li>1346 bytes for `packetsize`</li>
+            </ul>
+            }
+            @warning{Reduce the bitrate when using low values.
+            Values larger than 1456 require jumbo frames.
+            }
+        </td>
+    </tr>
+    <tr>
+        <td>Default</td>
+        <td colspan="2">@code{}
+            0
+            @endcode</td>
+    </tr>
+    <tr>
+        <td>Range</td>
+        <td colspan="2">0, 200-65535</td>
+    </tr>
+    <tr>
+        <td>Example</td>
+        <td colspan="2">@code{}
+            packetsize = 1346
+            @endcode</td>
+    </tr>
+</table>
+
 ## Config Files
 
 ### file_apps
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -544,6 +544,7 @@ namespace config {
 
     ENCRYPTION_MODE_NEVER,  // lan_encryption_mode
     ENCRYPTION_MODE_OPPORTUNISTIC,  // wan_encryption_mode
+    0,  // packetsize
   };
 
   nvhttp_t nvhttp {
@@ -1252,6 +1253,7 @@ namespace config {
 
     int_between_f(vars, "lan_encryption_mode", stream.lan_encryption_mode, {0, 2});
     int_between_f(vars, "wan_encryption_mode", stream.wan_encryption_mode, {0, 2});
+    int_between_f(vars, "packetsize", stream.packetsize, {0, PACKETSIZE_MAX});
 
     path_f(vars, "file_apps", stream.file_apps);
 #ifndef __ANDROID__
--- a/src/rtsp.cpp
+++ b/src/rtsp.cpp
@@ -1004,6 +1004,23 @@ namespace rtsp_stream {
         config.encryptionFlagsEnabled |= SS_ENC_AUDIO;
       }
 
+      // Limit the packetsize to avoid fragmentation with clients that cannot configure this value
+      if (config::stream.packetsize && config::stream.packetsize < config.packetsize) {
+        if (config::stream.packetsize < config::PACKETSIZE_MIN || config::stream.packetsize > config::PACKETSIZE_MAX) {
+          BOOST_LOG(warning) << "packetsize range: ["sv << config::PACKETSIZE_MIN << "-"sv << config::PACKETSIZE_MAX
+                             << "] invalid value: "sv << config::stream.packetsize;
+        } else {
+          if (config::stream.packetsize < config::PACKETSIZE_SMALL) {
+            BOOST_LOG(info) << "packetsize is small < "sv << config::PACKETSIZE_SMALL << " bytes, reduce bitrate if the stream breaks"sv;
+          } else if (config::stream.packetsize > config::PACKETSIZE_LARGE) {
+            BOOST_LOG(info) << "packetsize is large > "sv << config::PACKETSIZE_LARGE << " bytes, jumbo frames may be used"sv;
+          }
+
+          BOOST_LOG(info) << "packetsize limit: "sv << config.packetsize << " -> "sv << config::stream.packetsize << " bytes"sv;
+          config.packetsize = config::stream.packetsize;
+        }
+      }
+
       config.monitor.height = (int) util::from_view(args.at("x-nv-video[0].clientViewportHt"sv));
       config.monitor.width = (int) util::from_view(args.at("x-nv-video[0].clientViewportWd"sv));
       config.monitor.framerate = (int) util::from_view(args.at("x-nv-video[0].maxFPS"sv));
--- a/src_assets/common/assets/web/configs/tabs/Network.vue
+++ b/src_assets/common/assets/web/configs/tabs/Network.vue
@@ -173,6 +173,13 @@ const effectivePort = computed(() => +config.value?.port ?? defaultMoonlightPort
       <div class="form-text">{{ $t('config.ping_timeout_desc') }}</div>
     </div>
 
+    <!-- Packet Size Limit -->
+    <div class="mb-3">
+      <label for="packetsize" class="form-label">{{ $t('config.packetsize') }}</label>
+      <input type="number" min="0" max="65535" class="form-control" id="packetsize" placeholder="0" v-model="config.packetsize" />
+      <div class="form-text">{{ $t('config.packetsize_desc') }}</div>
+    </div>
+
   </div>
 </template>
 
--- a/src_assets/common/assets/web/public/assets/locale/en.json
+++ b/src_assets/common/assets/web/public/assets/locale/en.json
@@ -319,6 +319,8 @@
     "output_name": "Display Id",
     "output_name_desc_unix": "During Sunshine startup, you should see the list of detected displays. Note: You need to use the id value inside the parenthesis. Below is an example; the actual output can be found in the Troubleshooting tab.",
     "output_name_desc_windows": "Manually specify a display device id to use for capture. If unset, the primary display is captured. Note: If you specified a GPU above, this display must be connected to that GPU. During Sunshine startup, you should see the list of detected displays. Below is an example; the actual output can be found in the Troubleshooting tab.",
+    "packetsize": "Packet Size Limit",
+    "packetsize_desc": "Limit the packet size to avoid fragmentation on a low MTU link. This helps reduce packet loss and micro-stuttering, while streaming over a layer 2 VPN to clients that cannot configure this value, e.g. Moonlight for Android/iOS. Reduce the bitrate when using low values. Values larger than 1456 require jumbo frames. Range: 0, 200-65536. A value of 0 will disable the limit.",
     "ping_timeout": "Ping Timeout",
     "ping_timeout_desc": "How long to wait in milliseconds for data from moonlight before shutting down the stream",
     "pkey": "Private Key",
--- a/src/config.h
+++ b/src/config.h
@@ -16,6 +16,12 @@
 #include "nvenc/nvenc_config.h"
 
 namespace config {
+  // Valid range for the packetsize limit
+  constexpr int PACKETSIZE_MIN = 200;
+  constexpr int PACKETSIZE_MAX = 65535;
+  constexpr int PACKETSIZE_SMALL = 500;
+  constexpr int PACKETSIZE_LARGE = 1456;
+
   // track modified config options
   inline std::unordered_map<std::string, std::string> modified_config_settings;
 
@@ -177,6 +183,9 @@ namespace config {
     // Video encryption settings for LAN and WAN streams
     int lan_encryption_mode;
     int wan_encryption_mode;
+
+    // Limit the packetsize to avoid fragmentation on a low MTU link
+    int packetsize;
   };
 
   struct nvhttp_t {
--- a/src_assets/common/assets/web/config.html
+++ b/src_assets/common/assets/web/config.html
@@ -253,6 +253,7 @@
               "lan_encryption_mode": 0,
               "wan_encryption_mode": 1,
               "ping_timeout": 10000,
+              "packetsize": 0,
             },
           },
           {


Kishi85:
This should also help with issues like #5067 if I'm reading the changes correctly.

ReenigneArcher:
Thank you for the PR!

I notice your PR modifies localization files other than the default language (en). We do not accept contributions on GitHub for these files, but you can contribute on Crowdin. Please see https://docs.lizardbyte.dev/latest/developers/contributing.html#localization for more information.

httpstorm:
@Kishi85 
Yes. Based on their discussion, Sunshine ends up sending very large packets which prevents the stream from working. This PR allows the `packetsize` to be limited to a value that works, so it can be used as a workaround, so long as they set a suitable value.

*Here is my research:*
I used Wireshark to capture the traffic from Sunshine to client, as well as the VPN encapsulated stream. I observed fragmented UDP packets. They can be either inside or outside the tunnel or on both, depending on how OpenVPN is configured and the MTU configured for the TAP interface. In the end I decided to use MTU 1390. This resulted in encapsulated VPN packets with Ethernet frame size 1456 bytes, and 1428 bytes of UDP payload. This is the largest packet I could transport without fragmentation. 1428 is also the largest ICMP packet that can reach the Sunshine/VPN server over the Internet. With this in place, there is no need to fragment packets by OpenVPN, the IP transport is allowed to fragment packets, but the MTU on the TAP interface avoids any fragmentation outside the tunnel. Fragmentation is kept only on the TAP interface.

This is already an improvement to MTU 1500, but from time to time I still got the red text showing that the connection is not stable, and there is a noticeable micro-stutter. At first I was able to resolve it by setting `packetsize=1358` in the Moonlight configuration. However I need to use an Android device, and unfortunately `packetsize` cannot be configured on Android and iOS. Hence the need of this PR.

A quick search for `packetsize` on the Sunshine repo showed me exactly what I need to do. The largest value for `packetsize` on Sunshine, which avoids fragmentation inside the VPN tunnel for me is 1346 bytes. This produced UDP packets of size 1362 bytes and Ethernet frame of 1404 byes, inside the tunnel.


Building was hard. It didn't like `cuda-13.2`. And it didn't like the versions of `boost`, `node` that come with Ubuntu 24.04.4 LTS. But in the end I managed to compile it:
``` bash
cd /usr/local/
sudo mv cuda cuda.ori
sudo ln -s cuda-12.8 cuda
export PATH=/usr/local/cuda/bin:$PATH

cd ~/a/Sunshine
git submodule update --init --recursive
PATH=~/a/node-v24.15.0-linux-x64/bin:/usr/local/cuda:$PATH cmake -DBOOST_ROOT=/home/user/a/boost -DCMAKE_PREFIX_PATH=/home/user/a/boost -DSUNSHINE_ENABLE_CUDA=ON -B build -G Ninja -S .
PATH=~/a/node-v24.15.0-linux-x64/bin:$PATH ninja -C build
cd build
cpack -G DEB
sudo dpkg -i cpack_artifacts/Sunshine.deb
```

docs/configuration.md
```diff
@@ -1753,6 +1753,33 @@ editing the `conf` file in a text editor. Use the examples as reference.
...
 
+### packetsize
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            Limit the packetsize to avoid fragmentation on a low MTU link.
+            @note{This helps avoid packet loss and micro-stutter on a layer 2 VPN with
+            <br>
```
ReenigneArcher:
```suggestion
```

break not necessary

httpstorm:
@ReenigneArcher 
Thanks! I removed all localisations. Once the PR is accepted I can contribute bg, en_GB, and en_US to Crowdin.
Do you want me to also send the remaining languages that were created by a translator or should I leave that for someone else to recreate? I moved them to a separate brach here:
https://github.com/httpstorm/Sunshine/tree/packetsize_translations_2026-05-19

ReenigneArcher:
> Thanks! I removed all localisations. Once the PR is accepted I can contribute bg, en_GB, and en_US to Crowdin. Do you want me to also send the remaining languages that were created by a translator or should I leave that for someone else to recreate? I moved them to a separate brach here: https://github.com/httpstorm/Sunshine/tree/packetsize_translations_2026-05-19

The initial pass of translations is automated using a translation API. Then localization contributors can review and request changes, but usually those are for the people who are experts in that language. But feel free to contribute to Bulgarian if you'd like.

docs/configuration.md
```diff
@@ -1753,6 +1753,33 @@ editing the `conf` file in a text editor. Use the examples as reference.
...
+        <td colspan="2">@code{}
+            0
+            @endcode</td>
+    </tr>
+    <tr>
+        <td>Example</td>
+        <td colspan="2">@code{}
+            packetsize = 1346
+            @endcode</td>
+    </tr>
```
ReenigneArcher, line 1814:
```suggestion
    <tr>
        <td>Default</td>
        <td colspan="2">@code{}
            0
            @endcode</td>
    </tr>
    <tr>
        <td>Range</td>
        <td colspan="2">0-1472</td>
    </tr>
    <tr>
        <td>Example</td>
        <td colspan="2">@code{}
            packetsize = 1346
            @endcode</td>
    </tr>
```

src/config.cpp
```diff
@@ -1252,6 +1253,7 @@ namespace config {
 
     int_between_f(vars, "lan_encryption_mode", stream.lan_encryption_mode, {0, 2});
     int_between_f(vars, "wan_encryption_mode", stream.wan_encryption_mode, {0, 2});
+    int_between_f(vars, "packetsize", stream.packetsize, {0, 1472});
```
ReenigneArcher: I'm not a networking expert. How was the range chosen?

src_assets/common/assets/web/configs/tabs/Network.vue
```diff
@@ -173,6 +173,13 @@ const effectivePort = computed(() => +config.value?.port ?? defaultMoonlightPort
       <div class="form-text">{{ $t('config.ping_timeout_desc') }}</div>
     </div>
 
+    <!-- Packet Size Limit -->
+    <div class="mb-3">
+      <label for="packetsize" class="form-label">{{ $t('config.packetsize') }}</label>
+      <input type="text" class="form-control" id="packetsize" placeholder="1346" v-model="config.packetsize" />
```
ReenigneArcher:
```suggestion
      <input type="number" min="0" max="1472" class="form-control" id="packetsize" placeholder="0" v-model="config.packetsize" />
```

Let's make it a "number", add min and max values, and set the placeholder to the default value.

src_assets/common/assets/web/public/assets/locale/en.json
```diff
@@ -319,6 +319,8 @@
     "output_name": "Display Id",
     "output_name_desc_unix": "During Sunshine startup, you should see the list of detected displays. Note: You need to use the id value inside the parenthesis. Below is an example; the actual output can be found in the Troubleshooting tab.",
     "output_name_desc_windows": "Manually specify a display device id to use for capture. If unset, the primary display is captured. Note: If you specified a GPU above, this display must be connected to that GPU. During Sunshine startup, you should see the list of detected displays. Below is an example; the actual output can be found in the Troubleshooting tab.",
+    "packetsize": "Packet Size Limit",
+    "packetsize_desc": "Limit the packet size to avoid fragmentation on a low MTU link. This helps reduce packet loss and micro-stuttering, while streaming over a layer 2 VPN to clients that cannot configure this value, e.g. Moonlight for Android/iOS. Default: 0 (disabled).",
```
ReenigneArcher:
```suggestion
    "packetsize_desc": "Limit the packet size to avoid fragmentation on a low MTU link. This helps reduce packet loss and micro-stuttering, while streaming over a layer 2 VPN to clients that cannot configure this value, e.g. Moonlight for Android/iOS. A value of 0 will disable the limit.",
```

src/config.cpp
Kishi85: That range likely just covers regular ethernet frames. I'd leave that unrestricted upwards as e.g. jumbo frames (9k+) and ethernet over thunderbolt (IIRC up to 64k) could go higher than that.

If an upper limit is necessary I'd likely go with 65535 for now.

httpstorm: @Kishi85 
What value do you want me to use for the upper range?
Yes, I know there are jumbo frames, and I suspected there might be 64k frames, but I'm not familiar with either of them, and I'm not sure what value to use. I agree it will be better if we extend this.

@ReenigneArcher 
I followed the calculations for a standard Ethernet frame:
- 1514 bytes for standard Ethernet frame, subtract 14 bytes
- 1518 bytes if VLAN tags are used, in which case subtract 18 bytes
- 1500 bytes for IPv4, subtract 20 bytes
- 1480 bytes for UDP headers + payload, subtract 8 bytes
- 1472 bytes for UDP payload

ReenigneArcher: From the very little that I've read, I'd go with 65535 as the limit as well.

Kishi85: @httpstorm Up to 65535 is the highest valid ethernet frame I can think of right now (and that is rather exotic if you do Ethernet over Thunderbolt but it's possible). Since it's just the upper limit that should be fine. This also needs to be reflected in Network.vue when updated.

andygrundman:
Just want to check that you know about Moonlight's packetSize option. See `app/streaming/session.cpp` and grep for packetSize. Default packet size is 1392 and it will drop down to 1024 when the IP address is not a local RFC1918 IP. So you end up with 1024 over a Tailscale tunnel for example. The packet size is also exposed as a command-line param `--packet-size`.

Both moonlight-ios and moonlight-android have the same logic and use 1392 or 1024.

OpenVPN may not be considered "remote", but I would strongly recommend not trying to stream over an OpenVPN tunnel. WireGuard should perform a lot better.

src/config.cpp
httpstorm: I changed the upper limit to 65535, but I need to make a few of notes about the value:
- if the maximum Ethernet frames is 65535 - 28 = 65507 for UDP payload
- usually an even number is used due to alignment, so
- 65506 or 65504 might be better candidates for the limit
- with `packetsize` = 1346, Sunshine produces UDP payload of 1462 (`packetsize` + 16), so
- 65506 - 16 = 65490 might be even better

Kishi85: The math is correct for IPv4 (not so much for IPv6 which Sunshine also supports) but this is just an (arbitrary) upper limit to guard the variable. 

If the user picks a value that is too high it'll make the stream fail like #5067 worst case (Note: I've not checked if the code additionally guards by interface mtu or other netcode) causing the user to revert to a lower value. 

The real value is always highly dependent on the users whole setup (including switches/etc. and the configuration thereof) so I'd not think about the maximum value too much. The user has to figure out a working value for the setup in question by himself (likely by simple trial and error for most people). 

Documentation should contain some hints/calculations and common example values if possible.

For a future PR: This feature could be extended by utilitzing PMTU discovery or some other Packetsize probing mechanism to automate this more but as a first fix I'd say this is good enough for most users.

httpstorm: I designed the `packetsize` limit to be safe: it takes effect only when non zero and less than what Sunshine would normally use.
So if we don't have an upper limit and the user sets a value that is too high, it will have no effect (same as 0). It will not break a stream that works when the feature is disabled. Think of the upper limit as a guidance of what values are accepted. Optimally the user should inspect the network traffic to find the largest value what works without fragmentation or dropped packets.

About the interface MTU: my first thought was that if I reduce it, Sunshine will check the MTU and reduce the `packetsize` to avoid fragmentation. It does not: it keeps sending packets with the same size, then the interface fragments them so we get one large packet, with the same size as the MTU, and a small packet with the remaining data. This is inefficient, because if either fragment gets dropped the entire packet is lost. That would indeed be a nice to have feature, but it has not been implemented yet.

My instinct tells me that PMTU discovery will need support on the client. A few things to consider:
- large packets with don't fragment (DF) have to be sent
- the client needs to replay with acknowledge
- a packet may be lost because it is too large to send or because of network conditions
- binary search helps to speed up the process
- a reliable measurement takes time

httpstorm: @Kishi85 
> Documentation should contain some hints/calculations and common example values if possible.

Added. Please let me know if it needs any changes.

httpstorm:
@andygrundman 
Streaming over Internet works well for me when using Moonlight for PC. However I need to work with a cheap VR headset running Android 12 and it connects once every 10 attempts. So I made a layer 2 VPN on my OpenWRT router and hijacked the server in my LAN. RFC 1918 addresses are in use: `192.168.x.x/24`, but the MTU on the TAP interface is 1390 bytes, so the packets are fragmented.
Yes, I know that Moonlight for PC has a `packetsize` setting, however the mobile clients for Android and iOS do not. And now that they are streaming over the LAN, the larger packet size is used. This PR implements the ability to limit the `packetsize` in Sunshine, to avoid fragmentation, and reduce packet loss and micro-stuttering.

OpenVPN is indeed hard to configure and uses only one CPU core, which is a bottleneck. But when configured well over UDP and if we ensure no packets are fragmented, the performance is excellent. I am experimenting with Portal over a 5G Internet connection, and there is no noticeable latency and no micro-stuttering. I use a layer 2 bridge, because it enables local discovery, which is not available with layer 3 VPNs.
Thanks!

httpstorm:
Rebased on master.

httpstorm:
@ReenigneArcher @Kishi85 
Why is this check failing? Is it because last night I added documentation and example about calculating `packetsize`?

src/config.cpp
Kishi85: LGTM :+1:

andygrundman:
> @andygrundman Streaming over Internet works well for me when using Moonlight for PC. However I need to work with a cheap VR headset running Android 12 and it connects once every 10 attempts. So I made a layer 2 VPN on my OpenWRT router and hijacked the server in my LAN. RFC 1918 addresses are in use: `192.168.x.x/24`, but the MTU on the TAP interface is 1390 bytes, so the packets are fragmented. 

Have you tried using another IP range that would trigger the 1024 byte packets? Give the 100. prefix a try, I believe it can safely be used as another private IP range, you can just choose a suitable random tiny subnet inside 100.64.0.0/10. https://en.wikipedia.org/wiki/IPv4_shared_address_space

httpstorm:
@andygrundman 
I used Moonlight for PC (Ubuntu 24.04.4 LTS) and tested the following networks: `192.168.0.0/24`, `100.0.0.0/28`, `100.64.0.0/10`, `8.8.8.0/24`. In all cases I observed packet loss and fragmented Ethernet packets with size: 1402 + 82, UDP payload 1408 bytes. Well, the Google address space sure is publicly routable, but large packets are still in use. Edit: Local discovery works. It is possible that large packets are always used when the link is considered local and there is no routing involved.
Then I configured `packetsize=1346` on Sunshine: problem solved. It's a better solution, because it helps all users optimise the size to their environment, and does not require any changes to the network configuration.
Thanks!

ReenigneArcher:
> @ReenigneArcher @Kishi85 Why is this check failing? Is it because last night I added documentation and example about calculating `packetsize`?

It depends on GitHub workflows to pass, so it always times out for forks, until I re-run it.

ReenigneArcher:
While having to set this option isn't ideal. I think it could help users with these issues as well?

- https://github.com/LizardByte/Sunshine/issues/5105
- https://github.com/LizardByte/Sunshine/issues/4895
- https://github.com/LizardByte/Sunshine/issues/4295
- https://github.com/LizardByte/Sunshine/issues/1074

httpstorm:
@ReenigneArcher 

Too much guessing and uncertainties that can be uncovered with a 30 second packet capture started right before running a game.

We'll ask them to test it once the PR is accepted and a new release becomes available. For #5067 and #5105 where the workaround is `tx-udp-segmentation off`, we don't have a network capture, they will have to test it. It might help. For #4895 we have a confirmation that very large packets are sent, so it should help. @aliefe04 even suggested the exact changes implemented here.

For #4295 the only time I remember seeing this error *Your network connection isn't performing well. Reduce your video bitrate setting or try a faster connection*, was when I tried forwarding the Sunshine traffic inside an OpenSSH tunnel. Someone might argue that OpenSSH cannot carry UDP traffic, so here is a hint:
- client UDP -> socat -> TCP -> SSH tunnel -> server TCP -> socat -> UDP

Some people struggle over a 10 GBE local link. I think the modern infrastructure they use should be able to handle the traffic with pretty much no dropped packets. Same with Hyper-V. In my experience the Hyper-V network, CPU and disk performance are pretty much the same as the host. I wonder what is the largest packet size that Moonlight can handle? Another thought that comes to mind is that maybe modern 10 GBE and 2.5 GBE adapters may have new features unavailable on 1 GBE, and that breaks something. One possible reason for linux-6 to work and linux-7 not would be if support was added to the latest kernel and the features were not enabled before.

In 2024 I had performance issues on a Hyper-V guest running OpenWRT and using a 10 GBE virtual adapter. I was able to solve it with this, and the speeds I'm getting now are very high. The other guest running Ubuntu was either not affected or these were already disabled. Just a feint memory: I think had the impression that some outgoing packets did not have a valid checksum.
``` bash
[ ifup = "$ACTION" ] && [ -n "$DEVICE" ] && [ lo != "$DEVICE" ] && {
	ethtool --offload eth0 tx off tso off scatter-gather off gso off
	ethtool --offload "$DEVICE" tx off tso off scatter-gather off gso off
}
```

<details>
<summary>Offload setting on OpenWRT and Ubuntu 24.04.4 LTS</summary>

``` bash
# OpenWRT, linux-6.12.87
ethtool --show-offload eth0
Features for eth0:
rx-checksumming: on
tx-checksumming: off
	tx-checksum-ipv4: off
	tx-checksum-ip-generic: off [fixed]
	tx-checksum-ipv6: off
	tx-checksum-fcoe-crc: off [fixed]
	tx-checksum-sctp: off [fixed]
scatter-gather: off
	tx-scatter-gather: off
	tx-scatter-gather-fraglist: off [fixed]
tcp-segmentation-offload: off
	tx-tcp-segmentation: off
	tx-tcp-ecn-segmentation: off [fixed]
	tx-tcp-mangleid-segmentation: off
	tx-tcp6-segmentation: off
generic-segmentation-offload: off
generic-receive-offload: on
large-receive-offload: off [fixed]
rx-vlan-offload: on [fixed]
tx-vlan-offload: on [fixed]
ntuple-filters: off [fixed]
receive-hashing: on
highdma: on [fixed]
rx-vlan-filter: off [fixed]
vlan-challenged: off [fixed]
tx-gso-robust: off [fixed]
tx-fcoe-segmentation: off [fixed]
tx-gre-segmentation: off [fixed]
tx-gre-csum-segmentation: off [fixed]
tx-ipxip4-segmentation: off [fixed]
tx-ipxip6-segmentation: off [fixed]
tx-udp_tnl-segmentation: off [fixed]
tx-udp_tnl-csum-segmentation: off [fixed]
tx-gso-partial: off [fixed]
tx-tunnel-remcsum-segmentation: off [fixed]
tx-sctp-segmentation: off [fixed]
tx-esp-segmentation: off [fixed]
tx-udp-segmentation: off [fixed]
tx-gso-list: off [fixed]
tx-nocache-copy: off
loopback: off [fixed]
rx-fcs: off [fixed]
rx-all: off [fixed]
tx-vlan-stag-hw-insert: off [fixed]
rx-vlan-stag-hw-parse: off [fixed]
rx-vlan-stag-filter: off [fixed]
l2-fwd-offload: off [fixed]
hw-tc-offload: off [fixed]
esp-hw-offload: off [fixed]
esp-tx-csum-hw-offload: off [fixed]
rx-udp_tunnel-port-offload: off [fixed]
tls-hw-tx-offload: off [fixed]
tls-hw-rx-offload: off [fixed]
rx-gro-hw: off [fixed]
tls-hw-record: off [fixed]
rx-gro-list: on
macsec-hw-offload: off [fixed]
rx-udp-gro-forwarding: off
hsr-tag-ins-offload: off [fixed]
hsr-tag-rm-offload: off [fixed]
hsr-fwd-offload: off [fixed]
hsr-dup-offload: off [fixed]

# Ubuntu 24.04.4 LTS, linux-6.8.0-111-generic
ethtool --show-offload eth0
Features for eth0:
rx-checksumming: on
tx-checksumming: on
	tx-checksum-ipv4: on
	tx-checksum-ip-generic: off [fixed]
	tx-checksum-ipv6: on
	tx-checksum-fcoe-crc: off [fixed]
	tx-checksum-sctp: off [fixed]
scatter-gather: on
	tx-scatter-gather: on
	tx-scatter-gather-fraglist: off [fixed]
tcp-segmentation-offload: on
	tx-tcp-segmentation: on
	tx-tcp-ecn-segmentation: off [fixed]
	tx-tcp-mangleid-segmentation: off
	tx-tcp6-segmentation: on
generic-segmentation-offload: on
generic-receive-offload: on
large-receive-offload: off [fixed]
rx-vlan-offload: on [fixed]
tx-vlan-offload: on [fixed]
ntuple-filters: off [fixed]
receive-hashing: on
highdma: on [fixed]
rx-vlan-filter: off [fixed]
vlan-challenged: off [fixed]
tx-lockless: off [fixed]
netns-local: off [fixed]
tx-gso-robust: off [fixed]
tx-fcoe-segmentation: off [fixed]
tx-gre-segmentation: off [fixed]
tx-gre-csum-segmentation: off [fixed]
tx-ipxip4-segmentation: off [fixed]
tx-ipxip6-segmentation: off [fixed]
tx-udp_tnl-segmentation: off [fixed]
tx-udp_tnl-csum-segmentation: off [fixed]
tx-gso-partial: off [fixed]
tx-tunnel-remcsum-segmentation: off [fixed]
tx-sctp-segmentation: off [fixed]
tx-esp-segmentation: off [fixed]
tx-udp-segmentation: off [fixed]
tx-gso-list: off [fixed]
fcoe-mtu: off [fixed]
tx-nocache-copy: off
loopback: off [fixed]
rx-fcs: off [fixed]
rx-all: off [fixed]
tx-vlan-stag-hw-insert: off [fixed]
rx-vlan-stag-hw-parse: off [fixed]
rx-vlan-stag-filter: off [fixed]
l2-fwd-offload: off [fixed]
hw-tc-offload: off [fixed]
esp-hw-offload: off [fixed]
esp-tx-csum-hw-offload: off [fixed]
rx-udp_tunnel-port-offload: off [fixed]
tls-hw-tx-offload: off [fixed]
tls-hw-rx-offload: off [fixed]
rx-gro-hw: off [fixed]
tls-hw-record: off [fixed]
rx-gro-list: off
macsec-hw-offload: off [fixed]
rx-udp-gro-forwarding: off
hsr-tag-ins-offload: off [fixed]
hsr-tag-rm-offload: off [fixed]
hsr-fwd-offload: off [fixed]
hsr-dup-offload: off [fixed]
```

</details>

httpstorm:
Changes:

1. Use one long line in `rasp.cpp` to satisfy clang-format check. Other code uses even longer lines.
``` C
BOOST_LOG(info) << "packetsize limit: "sv << config.packetsize << " -> "sv << config::stream.packetsize << " bytes"sv;
```

2. Set a minimum limit of 200 for `packetsize`. I could set 500, which is still unreasonably low, but I'd rather give users more control in case they ever need it. Just in case someone decides to send the stream over some exotic radio with tiny packets.
- < 104: fail to start stream
- 104 requires 0.5 Mbit/s
- 200 works up to 6 Mbit/s
- 500 works up to 23 Mbit/s
- note: the achievable speed depends on latency and network conditions

3. Update documentation and WebUI description:
- Reduce the bitrate when using low values.
- Values larger than 1456 require jumbo frames.

src/rtsp.cpp
```diff
@@ -1004,6 +1004,12 @@ namespace rtsp_stream {
         config.encryptionFlagsEnabled |= SS_ENC_AUDIO;
       }
 
+      // Limit the packetsize to avoid fragmentation with clients that cannot configure this value
+      if (config::stream.packetsize >= 200 && config::stream.packetsize < config.packetsize) {
+        BOOST_LOG(info) << "packetsize limit: "sv << config.packetsize << " -> "sv << config::stream.packetsize << " bytes"sv;
+        config.packetsize = config::stream.packetsize;
+      }
+
```
ReenigneArcher, line 1023: Let's log a warning if the user has a value between 0 and 200 (1-199), letting them know the value will be ignored.

Let's also use a variable for `200` so we can easily change the lower limit just in one place.

andygrundman:
> We'll ask them to test it once the PR is accepted and a new release becomes available. For #5067 and #5105 where the workaround is `tx-udp-segmentation off`, we don't have a network capture, they will have to test it. It might help. For #4895 we have a confirmation that very large packets are sent, so it should help. @aliefe04 even suggested the exact changes implemented here.

The Linux kernel bug with UDP GSO where you need to disable tx-offload is very likely unrelated to these other issues you're having.  https://lore.kernel.org/regressions/6c3fb15e-711d-4b8d-b152-e03d9b05293f@linux.dev/T/#u

Kishi85:
I've given this a try using my usual wireguard tunnel when on the road and it's working as intended. The tunnel interface is using an MTU of 1280 (mandatory allowed minimum when using IPv6 inside the tunnel) for best external network compatibility (LTE/5G and Public WiFi) and the whole network path is using RFC1918 private IPs or ULA addresses. I've set the packetsize to 1232 for the worst case scenario (1280 MTU - 40 IPv6 header - 8 UDP header) and smoothness/responsiveness of the stream has improved for my case (at least subjectively, haven't done any proper measurements so far).

Once more: Good option to have for network scenarios where packetsize set by the client is insuffcient (or not done at all).

httpstorm:
Changes:
1. Declare `constexpr` limits in `config.h` and use them instead of magic numbers
2. Log a warning if `packetsize` is outside the valid range and ignore it
```
Warning: packetsize range: [200-65535] invalid value: 100
```

3. Log info if less than 500 and advice the user to reduce bitrate if the stream breaks

```
Info: packetsize is small < 500 bytes, reduce bitrate if the stream breaks
Info: packetsize limit: 1392 -> 200 bytes
```

4. Log info if greater than 1456 and inform that jumbo frames may be used
```
packetsize is large > 1456 bytes, jumbo frames may be used
```

5. Inform about the change if valid
```
Info: packetsize limit: 1392 -> 1346 bytes
```

src/rtsp.cpp
httpstorm: Ready, see https://github.com/LizardByte/Sunshine/pull/5153#issuecomment-4506125775. I split the warning in two lines, because it was uncomfortably log. I hope the clang-format check passes.

httpstorm:
@Kishi85 
> I've set the packetsize to 1232 for the worst case scenario (1280 MTU - 40 IPv6 header - 8 UDP header)

Use `packetsize=1216`: 1280 MTU - 40 IPv6 header - 8 UDP header - 16 Sunshine overhead.
The actual UDP payload size created by Sunshine is `config.packetsize + 16`.
Test: set MTU 1280 on your TUN interface and capture the network traffic. You should see fragmentation with 1232 and no fragmentation with 1216. Wireshark uses a black background for fragmented packets, so they are very easy to spot.

> and smoothness/responsiveness of the stream has improved for my case (at least subjectively, haven't done any proper measurements so far)

In Moonlight settings, enable this and keep an eye on: Frames dropped by your network connection
- [x] Show performance stats while streaming

You should observe rarer occasions of loss and micro-stuttering when there is no fragmentation.
For me with fragmentation, it is good enough to play, but with noticeable micro-stuttering. Without fragmentation it feels much smoother. The 5G tower is right beside my building and I can easily reach 700 Mbit/s. I'm streaming at 23 Mbit/s. It becomes much more obvious when a high bitrate is used.

23 Mbit/s:
- `packetsize=0`: mostly smooth with rare drops
- `packetsize=1342`: smooth with even more rare drops

40 Mbit/s:
- `packetsize=0`: unplayable > 95% loss
- `packetsize=1342` mostly smooth with rare drops

Kishi85:
> @Kishi85
> 
> > I've set the packetsize to 1232 for the worst case scenario (1280 MTU - 40 IPv6 header - 8 UDP header)
> 
> Use `packetsize=1216`: 1280 MTU - 40 IPv6 header - 8 UDP header - 16 Sunshine overhead. The actual UDP payload size created by Sunshine is `config.packetsize + 16`. Test: set MTU 1280 on your TUN interface and capture the network traffic. You should see fragmentation with 1232 and no fragmentation with 1216. Wireshark uses a black background for fragmented packets, so they are very easy to spot.
> 
> > and smoothness/responsiveness of the stream has improved for my case (at least subjectively, haven't done any proper measurements so far)
> 
> In Moonlight settings, enable this and keep an eye on: Frames dropped by your network connection
> 
> * [x]  Show performance stats while streaming
> 
> You should observe rarer occasions of loss and micro-stuttering when there is no fragmentation. For me with fragmentation, it is good enough to play, but with noticeable micro-stuttering. Without fragmentation it feels much smoother. The 5G tower is right beside my building and I can easily reach 700 Mbit/s. I'm streaming at 23 Mbit/s. It becomes much more obvious when a high bitrate is used.
> 
> 23 Mbit/s:
> 
> * `packetsize=0`: mostly smooth with rare drops
> * `packetsize=1342`: smooth with even more rare drops
> 
> 40 Mbit/s:
> 
> * `packetsize=0`: unplayable > 95% loss
> * `packetsize=1342` mostly smooth with rare drops

Thanks for the notice about me missing sunshine's 16 byte header, I'll adapt to 1216 and try again for hopefully perfect result.

FYI: Fragmentation on intermediate hops is only an IPv4 thing (unless you set DF-bit ofc), for IPv6 the whole concept got scrapped (Senders might still fragment the packet internally before putting it on the wire but that's not the culprit for issues in Sunshine) and if a packet is too big to forward a router will just throw a "Packet too big"-Error back to the sender. This could potentially be used to do some automated tuning down of sunshine's sending packet size (if the error cause is available to sunshine's sending socket) in case of problems but would require to always set the DF-bit for IPv4 connections. IMHO out of scope for this PR but a possible follow-up (I might also take a look into this myself if I find the time).

src/rtsp.cpp
ReenigneArcher: You can run clang-format locally before committing. See below.

1. Create and activate a python venv in your workspace (optional) - https://docs.python.org/3/library/venv.html#creating-virtual-environments
2. Install dependencies with the following command:

   ```bash
   python -m pip install ".[lint]"
   ```

3. Run the following command to fix lint issues:

   ```bash
   python scripts\update_clang_format.py
   ```

src/rtsp.cpp
```diff
@@ -1004,6 +1004,23 @@ namespace rtsp_stream {
...
 
+      // Limit the packetsize to avoid fragmentation with clients that cannot configure this value
+      if (config::stream.packetsize && config::stream.packetsize < config.packetsize) {
+        if (config::stream.packetsize < config::PACKETSIZE_MIN || config::stream.packetsize > config::PACKETSIZE_MAX) {
+          BOOST_LOG(warning) << "packetsize range: ["sv << config::PACKETSIZE_MIN << "-"sv << config::PACKETSIZE_MAX
+                             << "] invalid value: "sv << config::stream.packetsize;
+        } else {
+          if (config::stream.packetsize < config::PACKETSIZE_SMALL) {
+            BOOST_LOG(info) << "packetsize is small < "sv << config::PACKETSIZE_SMALL << " bytes, reduce bitrate if the stream breaks"sv;
+          } else if (config::stream.packetsize < config::PACKETSIZE_SMALL) {
```
ReenigneArcher:
```suggestion
          } else if (config::stream.packetsize > config::PACKETSIZE_LARGE) {
```

httpstorm: Fixed. Sorry for the mistake.

src/rtsp.cpp
httpstorm: Thanks! I found I can also use the `clang-format` tool:
``` bash
clang-format -i src/rtsp.cpp
```

My changes remained untouched, so I think they are properly formatted.
It corrected only line 1134, which is old code:
``` diff
-    if (encryption_mode == config::ENCRYPTION_MODE_MANDATORY &&
-        (config.encryptionFlagsEnabled & (SS_ENC_VIDEO | SS_ENC_AUDIO)) != (SS_ENC_VIDEO | SS_ENC_AUDIO)) {
+    if (encryption_mode == config::ENCRYPTION_MODE_MANDATORY && (config.encryptionFlagsEnabled & (SS_ENC_VIDEO | SS_ENC_AUDIO)) != (SS_ENC_VIDEO | SS_ENC_AUDIO)) {

```

httpstorm:
@ReenigneArcher 
Checks complete. Code coverage fails. I have no experience with that. What should we do next?

src/rtsp.cpp
ReenigneArcher: If it changed code you didn't touch, then you have a different version of clang-format than what we use (as defined by our pyproject.toml)

PR #756, opened 2023-01-11
macOS gamepad support

kotleni:
## Description
Implemented gamepad support on macOS. This works with my  <a href="https://github.com/kotleni/VirtualHID-macOS">driver for virtual HID devices</a> (you can copy this repository for yourself, the code is not licensed). Vibration not supported. I also added the IOKit dependency (shipped with Xcode).

_Sunshine host tested on Macbook Air M1 (2020).
Tested on two clients (Moonlight-nx (Nintendo Switch), and Moonlight (iOS))._

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I want maintainers to keep my branch updated

PR #5171, opened 2026-05-22
feat(macos): virtual HID gamepad emulating Razer Serval

### LINKED ISSUE #21: Gamepad support in Sunshine on macOS
### Is there an existing issue for this item?

### Repositories

LizardByte/Sunshine

### Languages/Skills/Technologies

C, C++, macOS, CMake

### Description

macOS is in a limited mode without gamepad support. There was a previous attempt to add this, but it randomly failed.

- https://github.com/LizardByte/Sunshine/pull/756
- https://github.com/LizardByte-research/VirtualHID-macOS (fork of kotleni's repo before it was deleted)

### Estimated Effort

effort:Extra-Large

### Priority

priority:Low

### Target Milestone

6+ months

### Dependencies

If we go with the Kexts route, it will require a Apple developer account ($100/yr) which will probably happen after LizardByte is incorporated as an LLC.

SayHi044:
## Description
Adds gamepad support to the macOS platform.

Achieved by publishing a virtual gamepad and translating Moonlight controller state into HID input reports via `IOHIDUserDeviceCreate`. No kext, no third-party driver — unlike the prior attempt (#756 ), which depended on an external VirtualHID driver. It emulates a Razer Serval, because it is auto-recognized by SDL/Steam/Wine with working analog triggers.

**Testing**
Setup: host — MacBook Pro M4 (macOS 26.5); client — Steam Deck, used as the Moonlight (v6.1) client in both desktop and gaming modes.

- Native macOS Steam — verified with No Man's Sky: the controller is auto-recognized with no setup prompt.
- Windows game via CrossOver (Wine) — verified with Clair Obscur: Expedition 33. 
- SDL probe — with no injected mapping, SDL reports isGameController=YES and applies the built-in mapping.

**Code signing**
`IOHIDUserDeviceCreate` needs the `com.apple.developer.hid.virtual.device` entitlement (added in `sunshine.entitlements` and wired into the codesign step). It's an Apple-restricted entitlement, so the official signing identity must be authorized for it (documented in docs/building.md).

#### Roadmap Issues

- Closes https://github.com/LizardByte/roadmap/issues/21

## Type of Change

## Checklist

### AI Usage


ReenigneArcher:
Sorry, jthread doesn't work on older macOS versions. You can use a nosonar comment, like https://github.com/LizardByte/tray/blob/fc343dcea4061b06a81bfa72dfe19626f549edcd/tests/unit/test_tray.cpp#L119

docs/building.md
```diff
@@ -234,6 +234,53 @@ ninja -C build
   }}
 }
 
+### macOS code signing & entitlements
+The macOS virtual gamepad publishes a virtual HID device via `IOHIDUserDeviceCreate`,
+which requires the `com.apple.hid.manager.user-access-device` entitlement. Without it,
+AMFI terminates Sunshine the moment a controller is first connected.
```
ReenigneArcher: So, if we don't sign then Sunshine will not work at all?

This is going to break the homebrew package, and make it difficult to test any PR builds as those are not signed.

src/platform/macos/input.cpp
```diff
@@ -22,14 +29,201 @@
...
+}
+
 /**
  * @brief Delay for a double click, in milliseconds.
  * @todo Make this configurable.
  */
 constexpr std::chrono::milliseconds MULTICLICK_DELAY_MS(500);
 
+// Gamepad HID report descriptor — emulates a Razer Serval (VID 0x1532 /
+// PID 0x0900, an Android/PC Bluetooth gamepad).
```
ReenigneArcher, line 54: What would be required to emulate standard Xbox and PlayStation gamepads?

src/platform/macos/input.cpp
```diff
@@ -583,29 +1081,17 @@ const KeyCodeMap kKeyCodesMap[] = {
     // Unimplemented feature - platform_caps::pen_touch
   }
 
-  /**
-   * @brief Sends a gamepad touch event to the OS.
-   * @param input The global input context.
-   * @param touch The touch event.
-   */
+  // Sends a gamepad touch event to the OS (documented in src/platform/common.h). Unimplemented on macOS.
```
ReenigneArcher: you can just remove the comments if they're duplicated in the common.h file

src_assets/macos/build/sunshine.entitlements
```diff
@@ -0,0 +1,20 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <!--
+      Required for the virtual HID gamepad (src/platform/macos/input.cpp).
+      IOHIDUserDeviceCreate() needs this entitlement to publish a virtual
+      input device; without it AMFI kills the process when a controller is
+      first connected.
+
+      NOTE: this is an Apple-RESTRICTED entitlement. A Developer ID build must
```
ReenigneArcher, line 12: I have to make a change in the apple developer portal for this?

ReenigneArcher (commented):
Thank you for this PR! It's a highly requested feature for a long time.

Looks like you can remove these lines:

- https://github.com/LizardByte/Sunshine/blob/3c54d5ff3acf40b533e5850fdf792f30aec0523f/docs/getting_started.md?plain=1#L308-L309
- https://github.com/LizardByte/Sunshine/blob/3c54d5ff3acf40b533e5850fdf792f30aec0523f/gh-pages-template/_data/features.yml#L34 ... and might need to adjust the wording if Xbox or PlayStation cannot be emulated on macOS.

docs/building.md
SayHi044: Sorry, that was misleading — it doesn't break unsigned builds. Only the gamepad is unavailable; everything else runs fine. I've fixed the docs.

The entitlement is only added when the app is actually signed (SHOULD_SIGN=true in cmake/packaging/macos.cmake), so Homebrew and PR builds never claim it and AMFI has nothing to reject. In that case IOHIDUserDeviceCreate just returns NULL and a connecting controller is simply ignored (I checked with a quick repro — no crash, no SIGKILL).

src/platform/macos/input.cpp
SayHi044: I tried emulating an Xbox gamepad, but it doesn't work properly: Xbox controllers use Microsoft's proprietary GIP protocol instead of HID, so even when the device is detected, its inputs aren't recognised.

PlayStation looks doable since DualSense/DualShock are HID, so that could be the next step — but it's a bigger effort, because I'd have to emulate the full controller protocol (rumble, touchpad, motion, etc.) to do it properly.

So I'd start with the generic gamepad: it already works with a PlayStation controller connected on the client side (I tested with a DualSense), just without the PS glyphs and the extra features.

src/platform/macos/input.cpp
SayHi044: Removed

src_assets/macos/build/sunshine.entitlements
SayHi044: Honestly, I'm not sure — it isn't documented, and I don't have an Apple Developer ID to test it myself. That's my main concern: without an authorized signing identity, it only works in AMFI-relaxed environments, not on normal user machines.

Could you try signing it with your Developer ID and see if it goes through? If it doesn't, it may be worth asking Apple support whether this entitlement can be authorized for the identity at all.

If that doesn't work, I can try to use the CoreHID framework — but it's Swift-only (so it'd need a bridge from the C++ backend) and it raises the minimum macOS to 15+, so I'm not sure it'd suit the project.

P.S. I picked this up mainly because I'd love to see the feature in Sunshine myself. Fair warning though — I'm a Python dev without much hands-on macOS experience, so thanks for your patience.

andygrundman: You request entitlements by going to https://developer.apple.com/account/resources/identifiers/list and configuring the identifier (FQDN/namespace). Some entitlements are self-service, they are on the first tab Capabilities, I don't think Sunshine needs any of those. The 3rd tab is Capability Requests where you'll find "HID Virtual Device". You have to fill out a simple form where you just explain how the app will use it.

andygrundman:
I saw the comment about not supporting rumble. I think supporting rumble is pretty important, as well as probably touchpad and gyro. Can we match the features that are available on Linux?

src_assets/macos/build/sunshine.entitlements
ReenigneArcher: Thanks! I submitted the request.

SayHi044:
> I saw the comment about not supporting rumble. I think supporting rumble is pretty important, as well as probably touchpad and gyro. Can we match the features that are available on Linux?

It wasn't my initial goal, but yeah, that seems doable. But it means emulating a DualSense specifically, not the generic gamepad I'm using now. As far as I see, the generic one can't do rumble/touchpad/gyro: macOS only sends rumble to a controller it recognises as real, and touchpad/gyro only surface for controllers the system actually parses.

It's a fair chunk of work tho — basically a second backend. So, referring to [my previous comment](https://github.com/LizardByte/Sunshine/pull/5171#discussion_r3293423616), I think it makes sense to start with the generic approach and then move to the DualSense emulation.

But I’m flexible about that. If @ReenigneArcher also thinks that we should deliver DS emulation as well - I can work on that since I planned to do that anyway, just as a separate step.

ReenigneArcher:
> I think it makes sense to start with the generic approach and then move to the DualSense emulation.

Since this has been a lacking feature for the whole existence of Sunshine on macOS, I think I agree.

I do wonder if it would make sense to make a virtual input library and then just have Sunshine consume it? Similar to https://github.com/games-on-whales/inputtino ... this isn't much code right now, but it might make it nicer to extend with new types of gamepads and new features like rumble.

andygrundman:
OK I don't have a sense of how much more work it is. We will need to find a different ID to use though, Razer is not a good choice. 

I built this and just did a quick initial test.  I found that the mapping is a bit wrong. Right trigger and right stick axes are swapped and possibly a few more buttons are wrong. I was testing from a Steam Deck.

SayHi044:
> OK I don't have a sense of how much more work it is. We will need to find a different ID to use though, Razer is not a good choice.
> 
> I built this and just did a quick initial test. I found that the mapping is a bit wrong. Right trigger and right stick axes are swapped and possibly a few more buttons are wrong. I was testing from a Steam Deck.

I don't have such issue on my setup. Can you please share more details about yours? I'll try to repro this bug when I'll have some time

andygrundman:
I'm not a fan of the Lumen project, but they did a controller implementation, you may want to take a look at what it was doing. For example, [here's the device](https://github.com/trollzem/Lumen/blob/main/src/platform/macos/hid_gamepad.m) they created.  It also uses a different entitlement `com.apple.developer.hid.virtual.device`. I think it's worth at least understanding all our options if there are multiple ways to do this. 

```
    NSDictionary *props = @{
        @kIOHIDVendorIDKey:       @(0x1209),  // Generic (pid.codes open-source VID)
        @kIOHIDProductIDKey:      @(0x5853), // Not in SDL's known controller database
        @kIOHIDManufacturerKey:   @"Sunshine Virtual Gamepad",
        @kIOHIDProductKey:        [NSString stringWithFormat:@"Sunshine Gamepad %d", _gamepadIndex],
        @kIOHIDSerialNumberKey:   [NSString stringWithFormat:@"SUNSHINE-%d", _gamepadIndex],
        @kIOHIDTransportKey:      @"USB",
        @kIOHIDReportDescriptorKey: [NSData dataWithBytes:kHIDReportDescriptor
                                                   length:sizeof(kHIDReportDescriptor)],
    };
```

SayHi044:
Sorry for the long silence, I had a really busy week at work. Let me go through all the pending comments.

---

> I do wonder if it would make sense to make a virtual input library and then just have Sunshine consume it? Similar to https://github.com/games-on-whales/inputtino ... this isn't much code right now, but it might make it nicer to extend with new types of gamepads and new features like rumble.

@ReenigneArcher I agree that it is the right direction. But since for this PR we are going to ship only 1 gamepad type support and we didn’t try if the signing is working yet - I’d do this as a part of addition of the DS support.

---

> I built this and just did a quick initial test. I found that the mapping is a bit wrong. Right trigger and right stick axes are swapped and possibly a few more buttons are wrong. I was testing from a Steam Deck.

@andygrundman
I spent some time trying to catch cases when my approach doesn't work.
The mapping is correct *for SDL* supporting games. What varies is how each macOS app reads the HID device.
Test map:
| How the app reads controllers | Detects the pad? | Mapping |
|---|---|---|
| **SDL** — native Steam games, and Steam's own controller layer | yes | ✅ correct |
| **Wine / CrossOver** XInput | yes | ✅ correct |
| **Raw HID** — e.g. a browser's Gamepad API (Chrome) | yes | ❌ mismapped — RT/right-stick swap + some buttons off |
| **Apple `GameController.framework`** — Safari, native-API Mac games | ❌ **not detected at all** | — |

I also did try forcing Steam Input for broken cases, but it didn't help.

I think the root cause of the swap in your case is that the Serval's axis usages are non-conventional (right-stick-Y on Rx, triggers on Ry/Rz), so they only decode correctly when the consumer applies SDL's Serval mapping.

**TL;DR** Current approach works with SDL or XInput supporting apps and doesn't work on apps which rely on Raw HID or Apple's GameController.framework.

@andygrundman Can you please share what game you used for test? And how did you run it (Steam, native, else). I want to confirm that it was the described case, and not something else.

---

> I'm not a fan of the Lumen project, but they did a controller implementation, you may want to take a look at what it was doing. For example, [here's the device](https://github.com/trollzem/Lumen/blob/main/src/platform/macos/hid_gamepad.m) they created. It also uses a different entitlement `com.apple.developer.hid.virtual.device`. I think it's worth at least understanding all our options if there are multiple ways to do this.

I did take a look at their implementation. They use generic gamepad, which is not auto-recognised and needs additional setup. My approach was to avoid that fully, to make it more _“plug-n-play”_.

However, I did test if the entitlement they use works with my build and it is. **I did switch the entitlement**, because it seems a better fit. **Thanks for pointing that out! @andygrundman**

I see that they also delivered some other features for macOS, but this is a whole other topic.

---

Also, I noticed #5203, which targets the same macOS gamepad gap.  I’m glad that this issue caught more attention, this is cool!

It's derived from Lumen, so its generic-controller identity and its signing/entitlement model are the same as Lumen's — I covered how that differs from this PR above.

andygrundman:
Tested the latest commit, 9078d5fad. You're right about the mapping, it works in Steam but is mis-mapped in the HID mode for browsers. In the Steam controller test, the gamepad inputs feel very laggy, like they are being processed at 1fps or something.

Speaking of Steam, I downloaded the Mac version of Mina the Hollower. Controller seems to work ok in the game, but there are a lot of worse problems going on:

* Sunshine turns red in Activity Monitor because some thread has locked up.
* Steam Deck reports it's getting 90fps but everything feels much slower.
* If Mina is run fullscreen, when Game Mode gets engaged, Sunshine basically grinds to a halt and can't even manage to update a single frame. I think the only solution is to disable Game Mode. Needs more research into what other apps are doing, maybe there is an official Apple suggestion here. If you can't play with Game Mode, it's just one more reason to not use your Mac for games. 

Normally I don't even bother to launch a game because the experience is already so bad you can't even use it for remote desktop work, let alone a game. My M1 Pro has never felt slower.

docs/building.md
```diff
@@ -235,6 +235,56 @@ ninja -C build
...
+    csrutil disable
+    nvram boot-args="amfi_get_out_of_my_way=0x1"
+    ```
+    Reboot back into macOS.
+  - **Apple Silicon:** boot into Recovery (hold the power button), set the startup disk to
+    *Reduced Security* with *"Allow user management of kernel extensions"* via Startup Security
+    Utility, then from a Recovery Terminal:
+    ```bash
+    csrutil disable
+    bputil -k    # follow the prompts to allow boot-args
```
SpaceKittyCow, line 278: Hey, sorry to barge on this code review. 

First off thanks so much for doing this! I was so excited to be able to do this on my Tahoe M4 Mac, I built and ran this branch. Got Sunshine up and running, and then I made it to here, and realized:

A) This command ran, but did not allow for the nvram update command below, due to permission issues from io/common.

B) Realized that I was in very unsafe territory on my MacOs. I've worked in Software, and some in MacOs OS in virtualization before silicon, but I was not aware enough how quick I could break my whole computer until this didn't work. 
 
To try to solve A, I read through bputil documentation here: https://keith.github.io/xcode-man-pages/bputil.1.html and it looked like -a may also be needed, but when I tried it still failed. I pulled back and reset full security privileges for the time being. 

For B, wonder if there should be a good warning in the documentation before hand, or in the future an way to have a binary build with the developer key with a CI/CD pipeline. 

Sorry if this is to much. I really want this to work and come out!

ReenigneArcher:
Virtual input in Sunshine is going to go through https://github.com/lizardByte/libvirtualhid, there's an in process PR https://github.com/LizardByte/Sunshine/pull/5368. I have a few things to wrap up before merging. After that's merged I'll begin looking at how to accomplish this on macOS without requiring user hacks.

Not sure if it would be better to have a secondary application and communicate through it via IPC, or just bake everything right into Sunshine. Sunshine is allowed the "HID Virtual Device" entitlement now, so that shouldn't be a problem.

ADIKANT:
Hi there, are there any updates on the topic? would be nice to get steamdeck support with macOS games native from steam or crossover and seems like the topic didn't land anywhere

ReenigneArcher:
SOON<sub>tm</sub>

PR #5189, opened 2026-05-25
feat(macos): declare LSApplicationCategoryType=games for gamepolicyd treatment

Nottlespike:
## Description

macOS 14+ uses `gamepolicyd` to decide which processes get game-priority scheduling treatment. The daemon's primary signal is the app's `LSApplicationCategoryType`: anything matching `public.app-category.*games*` is recognised as game-related and kept on game-priority scheduling when other games activate Game Mode. Everything else can be deprioritised to free CPU/GPU resources for the foreground game.

Sunshine was previously declared as a **utility**. When a user streams a game that engages Game Mode (e.g. a native Mac game running fullscreen), the OS treats Sunshine as background noise and throttles its capture and encoder threads — which is exactly backwards for a streaming host whose entire purpose is to maintain realtime encode throughput **while** the streamed game runs hot.

Reclassifying as `public.app-category.games` causes `gamepolicyd` to treat Sunshine as a peer of the running game. Encoder threads keep their scheduling priority, capture queue isn't preempted, and the actual streaming throughput stays consistent under load.

**Important distinction:** Sunshine itself does **not** trigger Game Mode and this PR does not attempt to make it do so. Sunshine is `LSUIElement=true` (background utility, no Dock icon, never frontmost-fullscreen) so it can never meet `gamepolicyd`'s Game Mode activation criteria (frontmost + native-fullscreen). The change is purely about Sunshine **not being throttled** when a separate game-categorised app activates Game Mode.

An inline comment in the plist documents the rationale so future readers don't try to "fix" the seemingly-incorrect category. The `LSApplicationCategoryType` is otherwise informational (no behavioural change outside `gamepolicyd`'s scheduling heuristics and Mac App Store categorisation, the latter being inapplicable here since Sunshine is not App Store distributed).

### Screenshot

N/A — Info.plist key change.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src_assets/macos/build/Info.plist.in
+++ b/src_assets/macos/build/Info.plist.in
@@ -18,8 +18,17 @@
   <string>@PROJECT_VERSION@</string>
   <key>CFBundleIconFile</key>
   <string>sunshine.icns</string>
+  <!--
+    Categorise as a game so gamepolicyd (macOS 14+) treats this process
+    as game-related and does not deprioritise its encoder/capture threads
+    when the streamed game has Game Mode active. Sunshine itself cannot
+    *trigger* Game Mode because it is LSUIElement (no Dock icon, never
+    frontmost), but being recognised as game-eligible means the OS keeps
+    Sunshine on game-priority scheduling alongside the actual game it is
+    streaming, instead of throttling it as a generic background utility.
+  -->
   <key>LSApplicationCategoryType</key>
-  <string>public.app-category.utilities</string>
+  <string>public.app-category.games</string>
   <key>LSUIElement</key>
   <true/>
   <key>NSMicrophoneUsageDescription</key>


Nottlespike:
The `gamepolicyd` category-check behaviour is documented across several public sources rather than a single canonical Apple page, which is why the inline comment leans on empirical wording. The most authoritative discussion I'm aware of:

- [Apple Developer Forums — *Activate Game Mode for my App in Sonoma?*](https://developer.apple.com/forums/thread/739387) — Apple engineers in this thread describe the `LSApplicationCategoryType` → Game Mode coupling and confirm the activation criteria.
- [moonlight-stream/moonlight-qt#1297](https://github.com/moonlight-stream/moonlight-qt/issues/1297) — the original diagnosis of `gamepolicyd`'s category-string check (`*.games`) by the moonlight-qt maintainers, with the empirical observation that this controls Game Mode eligibility.

Verified on macOS 14+ (M4 Max) — without this PR, capture and encoder threads visibly drop priority when a separate game-category app engages Game Mode; with this PR, scheduling stays consistent under the same workload. Happy to inline either link in the plist comment if you'd prefer that to leaving it as PR context.

andygrundman:
You might be able to find some detail in the system log (Console.app) about what the system is doing when a game mode game is running. I learned a lot about how Bluetooth and audio is handled in game mode this way.

I would hope that Sunshine's thread & queue priority would be enough to keep it running smoothly.

Nottlespike:
Closing per maintainer feedback — you have a separate fix for the Game category path. Happy to revisit if useful, but standing down on this one to avoid duplicate work.

PR #5190, opened 2026-05-25
feat(macos/capture): ScreenCaptureKit backend with gated EDR (HDR) for 10-bit pixel formats

Nottlespike:
## Description

Adds a ScreenCaptureKit (SCK) backend for macOS screen capture (12.3+) and layers properly-gated EDR/HDR support on top of it. Replaces the deprecated `AVCaptureScreenInput` path for users on modern macOS while leaving the legacy path intact as a runtime fallback.

This PR was previously split across #5190 (SCK) and #5191 (EDR). Folded together per maintainer feedback that PRs must be mergeable in any order — the EDR work textually depends on the SCK API surface and would not compile without it, so the split was artificial. The HDR-gating hardening (third commit) was added during review to fix a real defect in the EDR work.

## Commits

1. **`feat(macos/capture): add ScreenCaptureKit backend, runtime-select on 12.3+`** — introduces `SCVideo` (in `sc_video.{h,m}`), runtime-selected via `@available(macOS 12.3, *)` in `display.mm`. SCVideo conforms to the same `SunshineVideoCapture` protocol as the legacy `AVVideo` so the rest of the macOS capture pipeline is unchanged. ARC-compiled for clarity; lifecycle hardened (single `addStreamOutput`, bounded SCK completion-handler timeouts, `@synchronized` on shared state).

2. **`feat(macos/capture): enable EDR (HDR) output on macOS 14+ for 10-bit pixel formats`** — when SCK is in use and the chosen `CVPixelBuffer` format is 10-bit, request `SCCaptureDynamicRangeHDRLocalDisplay` so the OS tags buffers with BT.2020 PQ metadata matching the local panel. SDK-guarded so older Xcode versions still compile.

3. **`feat(macos/capture): gate EDR on negotiated session HDR, not pixel depth alone`** — the previous commit gated EDR purely on pixel format depth. That's not enough: a 10-bit format may be selected for codec reasons (e.g., ProRes 4:4:4) without the client ever requesting HDR ingest. We were silently emitting BT.2020 PQ buffers into streams the SDP described as SDR. This commit plumbs `enable_hdr` from `launch_session_t` (via the existing `config.dynamicRange` field on `video::config_t`) down to `SCVideo init`, and requires **both** 10-bit pixel format AND negotiated HDR before flipping `captureDynamicRange = HDRLocalDisplay`.

## Verification

- Builds with both legacy SDK (no `__MAC_OS_X_VERSION_MAX_ALLOWED >= 140000`) and current SDK.
- Runtime: on 14+, with HDR-capable display + Moonlight HDR toggle on → SCK frames carry HDR metadata; with HDR toggle off → SDR even when the pix_fmt is 10-bit. Verified end-to-end on M4 Max.
- Capture log surfaces `Using ScreenCaptureKit capture backend (HDR allowed|blocked)` at session start so the negotiated state is auditable.

## Independence

Stands alone — does not depend on any other open PR. The 10-bit pixel format selection path (e.g., for ProRes capture in #5192) is itself an additive change that exercises this PR's EDR path when both are present; neither requires the other to compile or pass CI.

closes #5191

### DIFF
--- a/cmake/compile_definitions/macos.cmake
+++ b/cmake/compile_definitions/macos.cmake
@@ -35,6 +35,7 @@ list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
         ${CORE_MEDIA_LIBRARY}
         ${CORE_VIDEO_LIBRARY}
         ${FOUNDATION_LIBRARY}
+        ${SCREEN_CAPTURE_KIT_LIBRARY}
         ${VIDEO_TOOLBOX_LIBRARY})
 
 set(APPLE_PLIST_TEMPLATE "${SUNSHINE_SOURCE_ASSETS_DIR}/macos/build/Info.plist.in")
@@ -55,6 +56,16 @@ set(PLATFORM_TARGET_FILES
         "${CMAKE_SOURCE_DIR}/src/platform/macos/nv12_zero_device.cpp"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/nv12_zero_device.h"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/publish.cpp"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/sc_video.h"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/sc_video.m"
         "${CMAKE_SOURCE_DIR}/third-party/TPCircularBuffer/TPCircularBuffer.c"
         "${CMAKE_SOURCE_DIR}/third-party/TPCircularBuffer/TPCircularBuffer.h"
         ${APPLE_PLIST_FILE})
+
+# sc_video.m is written against ARC for clarity (SCK APIs are async/
+# block-heavy and benefit from ARC). The rest of the macOS Obj-C
+# sources remain MRC; objects flowing across the boundary follow the
+# standard +1-retain alloc/init convention so both modes interoperate.
+set_source_files_properties(
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/sc_video.m"
+        PROPERTIES COMPILE_FLAGS "-fobjc-arc")
--- a/cmake/dependencies/macos.cmake
+++ b/cmake/dependencies/macos.cmake
@@ -10,6 +10,14 @@ FIND_LIBRARY(CORE_MEDIA_LIBRARY CoreMedia)
 FIND_LIBRARY(CORE_VIDEO_LIBRARY CoreVideo)
 FIND_LIBRARY(FOUNDATION_LIBRARY Foundation)
 FIND_LIBRARY(VIDEO_TOOLBOX_LIBRARY VideoToolbox)
+# ScreenCaptureKit is the modern (macOS 12.3+) replacement for the
+# deprecated AVCaptureScreenInput-based capture path. Sunshine's
+# sc_video.{h,m} is unconditionally compiled into the macOS target;
+# fail configure with a clear message rather than failing the build
+# later on header lookup when the SDK doesn't ship the framework
+# (e.g., when building with an Xcode older than 13.3 / SDK older than
+# 12.3, which dropped out of routine compatibility long ago).
+FIND_LIBRARY(SCREEN_CAPTURE_KIT_LIBRARY ScreenCaptureKit REQUIRED)
 
 if(SUNSHINE_ENABLE_TRAY)
     FIND_LIBRARY(COCOA Cocoa REQUIRED)
--- a/src/platform/macos/av_video.h
+++ b/src/platform/macos/av_video.h
@@ -15,7 +15,17 @@ struct CaptureSession {
 
 static const int kMaxDisplays = 32;
 
-@interface AVVideo: NSObject <AVCaptureVideoDataOutputSampleBufferDelegate>
+typedef bool (^FrameCallbackBlock)(CMSampleBufferRef);
+
+/**
+ * @brief Shared interface for macOS screen capture backends.
+ *
+ * Both the legacy AVCaptureScreenInput-based implementation (AVVideo) and
+ * the modern ScreenCaptureKit-based implementation (SCVideo) conform to
+ * this protocol so display.mm can hold either behind a single pointer
+ * type and branch on macOS version at construction.
+ */
+@protocol SunshineVideoCapture <NSObject>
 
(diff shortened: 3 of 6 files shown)


Nottlespike:
Thanks for the catches — pushed an amended commit (force-push to this branch) addressing the genuinely-broken items:

- **SCStream output retention across stop/start**: restructured so `addStreamOutput` and `startCapture` happen exactly **once** during `-initWithDisplay:frameRate:`. `-capture:` is now purely a callback-swap operation that never touches the stream lifecycle. Frames flow continuously for the lifetime of the SCVideo object; the sample-handler delegate is a no-op when no callback is installed. Fixes both the addStreamOutput double-registration failure and the implicit "restart after callback returns false" path that no longer needs to exist.
- **`dispatch_queue_attr_make_with_qos_class` third argument**: changed from `DISPATCH_QUEUE_PRIORITY_HIGH` (incorrect — that's a legacy global-queue priority constant) to `0` (the correct relative-priority value).
- **`streamRunning` thread safety**: all reads/writes of `streamRunning`, `currentCallback`, and `currentSignal` now go through `@synchronized(self)`. State transitions in `dealloc`, `applyConfigurationIfRunning`, the SCStreamOutput delegate, and the SCStreamDelegate's `didStopWithError:` all use the same lock.
- **`__bridge` casts in `display.mm`**: replaced with plain C-style casts. `display.mm` compiles MRC; `__bridge` is ARC-specific syntax that has no effect there and can warn with `-Werror`.
- **Bounded timeouts on SCK completion handlers**: `DISPATCH_TIME_FOREVER` replaced with a 5-second bounded wait. If SCK ever hangs, init fails cleanly instead of deadlocking.
- **`CGDisplayCopyDisplayMode` NULL fallback**: falls back to `CGDisplayBounds` if mode lookup fails (e.g., during display reconfiguration).

The Copilot summary comment is purely descriptive (no actionable inline suggestions in it).

Nottlespike:
**Force-pushed.** Addressed the remaining Copilot suggestion (the one that wasn't already fixed by the earlier amend):

- **`cmake/dependencies/macos.cmake`** — `FIND_LIBRARY(SCREEN_CAPTURE_KIT_LIBRARY ScreenCaptureKit)` is now `REQUIRED`. If the SDK doesn't ship ScreenCaptureKit (Xcode < 13.3 / macOS SDK < 12.3), configure fails fast with a clear `Could NOT find ...` message rather than failing later at C compile time on a missing header. The rest of `sc_video.{h,m}` stays unconditionally compiled because `display.mm` references `SCVideo` directly inside the `@available(macOS 12.3, *)` runtime branch.

For the other Copilot points on this PR's older review — they were already addressed in the earlier amend and are still in place at slightly different line numbers post-rebase. Specifically:

- ✅ **`CGDisplayCopyDisplayMode` NULL fallback** via `CGDisplayBounds` — sc_video.m:73 (currently)
- ✅ **`DISPATCH_TIME_FOREVER` bounded** to 5s via `kSCVideoCompletionTimeoutSec` — sc_video.m:116 / 173
- ✅ **`addStreamOutput` single-registration** at init — done in the earlier amend (output added once during `-initWithDisplay:frameRate:`, never re-added in `-capture:`)
- ✅ **`streamRunning` thread safety** via `@synchronized(self)` — same earlier amend

Copilot's diff-relative line numbers shifted post-rebase, so its older comments now point at unrelated code. Sorry for the noise.

Nottlespike:
**Force-pushed.** Restructured per maintainer guidance that each PR must be mergeable in any order:

- Folded #5191 (EDR) into this PR — the EDR work was textually dependent on the SCK API surface added here and would not compile without it, so the split across two PRs was artificial. #5191 is now closed pointing to this PR.
- Added a third commit (`feat(macos/capture): gate EDR on negotiated session HDR, not pixel depth alone`) addressing a defect surfaced during end-to-end testing: the previous EDR commit promoted `captureDynamicRange` to `HDRLocalDisplay` purely on pixel-format depth. That produced a silent control/data-plane mismatch — Sunshine would tell the client "HDR mode false" in the SDP while emitting BT.2020 PQ-tagged buffers. Now gated on **both** 10-bit pixel format and the negotiated session's `enable_hdr` (plumbed from `launch_session_t` via the existing `config.dynamicRange`).
- New log line at session start: `Using ScreenCaptureKit capture backend (HDR allowed|blocked)` makes the gating visible.

PR title and body updated to reflect the combined scope. Closes #5191.

#5192 (ProRes) was rebased onto this new tip and updated to state explicit any-order-mergeable independence (ProRes encoder probe gracefully falls back to standard codecs if the underlying capture path can't supply 10-bit buffers — same as any failed-probe encoder).

andygrundman:
Is there any reason to keep the old AVCapture version? SCK looks like it was introduced in 12.3 and our current MACOSX_DEPLOYMENT_TARGET is 14.2.

ReenigneArcher:
> Is there any reason to keep the old AVCapture version? SCK looks like it was introduced in 12.3 and our current MACOSX_DEPLOYMENT_TARGET is 14.2.

I did not review this PR yet, but shouldn't keep workarounds for versions older than what we support (14.2).

Nottlespike:
Closing — maintaining this work in our fork (RESMP-DEV/Sunshine) going forward rather than upstreaming. No issue with the project; just diverging priorities after this round of review. The code remains MIT-licensed and available to anyone who wants to pull from the fork or cherry-pick. Thanks for the engagement on this batch.

Nottlespike:
Reopening — closed prematurely. This change is broadly applicable, not specific to the M4 Max I tested on, and the post-review feedback is fully addressed. The fork (RESMP-DEV/Sunshine) will be the home for hardware-specific experimental work (e.g., software AV1 via libsvtav1 that only makes sense on machines with extreme CPU headroom); but the minimum-broadly-supported changes like this one belong upstream where the broader user base benefits.

Nottlespike:
Reopening — closed prematurely. This change is broadly applicable (M-series Apple Silicon, but not specific to my M4 Max test host), and the post-review feedback is fully addressed in additional commits.

The fork (RESMP-DEV/Sunshine) is becoming the home for hardware-specific experimental work that I cannot validate across the full Sunshine user base (e.g., software AV1 via libsvtav1 that only makes sense on machines with extreme CPU headroom). But minimum-broadly-supported changes like this one belong upstream where the broader user base benefits.

PR #5191, opened 2026-05-25
feat(macos/capture): enable EDR (HDR) output on macOS 14+ for 10-bit pixel formats

Nottlespike:
## Description

> **Depends on #5190** (ScreenCaptureKit foundation). This PR's diff currently includes both commits; once #5190 lands the diff collapses to just the EDR commit.

With `AVCaptureScreenInput`, asking the capture surface for a 10-bit pixel format silently produced 8-bit BGRA — an OS-level lie that made HEVC Main10 / AV1 Main10 / future ProRes 10-bit profiles on macOS into fake HDR (color-tagged 8-bit data). With `ScreenCaptureKit` landing in #5190, 10-bit pixel formats are actually honoured, but SCK needs an explicit signal (`captureDynamicRange`) to attach HDR metadata to those buffers instead of treating them as 10-bit Rec.709.

**Changes:**

- **`+pixelFormatIsHighBitDepth:` classifier** in `SCVideo` covering the YUV 4:2:0 / 4:2:2 / 4:4:4 10-bit BiPlanar formats plus `ARGB2101010` packed and 64-bit RGBA formats. Returning YES is the signal that the capture surface is HDR-capable.
- **Synchronous init path**: set `captureDynamicRange` immediately if the starting pixel format is high bit depth, so the very first sample buffer carries HDR metadata (no first-frame-SDR flash).
- **`setPixelFormat:` path** (called by `nv12_zero_device` when the encoder selects `pix_fmt_e::p010`): also updates `captureDynamicRange` and pushes the new config to a running stream via `-updateConfiguration:`.
- **`SCCaptureDynamicRangeHDRLocalDisplay`** rather than the canonical HDR mode: game streaming wants the host display's actual HDR characteristics (peak luminance, primaries) so the receiver shows what a local user would see, not Apple's idealised reference encoding.
- **`@available(macOS 14.0, *)` guard**: on 12.3-13.x SCK still honours the 10-bit pixel format request from `nv12_zero_device` but the OS doesn't auto-tag buffers; Sunshine's existing colorspace logic continues to drive the encoder's color fields in that case.

**Validated on M4 Max (macOS 14+):**
Sunshine's encoder probe matrix at startup now contains:

```
Creating encoder [hevc_videotoolbox]
Color depth: 10-bit
Creating encoder [prores_videotoolbox]
Color depth: 10-bit
Found HEVC encoder: hevc_videotoolbox [videotoolbox]
Found ProRes encoder: prores_videotoolbox [videotoolbox]
```

Previously these 10-bit probes could not validate because the capture surface couldn't deliver matching 10-bit data — `validate_config` would fail at encoder init time. With this change, real 10-bit data flows through the pipeline for the first time on macOS, and the per-codec `DYNAMIC_RANGE` capability bit is set honestly based on actual probe results rather than optimistic `.capabilities.set()`.

**What this does NOT yet do (separate follow-ups):**

- ProRes-specific VT compression-session color tags (`kVTCompressionPropertyKey_ColorPrimaries` / `_TransferFunction` / `_YCbCrMatrix`) — ProRes carries color in the frame header rather than SEI, so this is its own discrete change.
- YUV 4:4:4 capture path for ProRes 4444 / 4444 XQ profiles.
- HDR signalling validation against an actual HDR-mode Moonlight session (in progress).

Encoder side is already wired generically (Sunshine sets `ctx->color_primaries / color_trc / colorspace` from `avcodec_colorspace` at `src/video.cpp:1781-1783` for all encoders), so HEVC Main10 HDR via VideoToolbox should now produce honest BT.2020 PQ output for any HDR-requesting client.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/cmake/compile_definitions/macos.cmake
+++ b/cmake/compile_definitions/macos.cmake
@@ -35,6 +35,7 @@ list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
         ${CORE_MEDIA_LIBRARY}
         ${CORE_VIDEO_LIBRARY}
         ${FOUNDATION_LIBRARY}
+        ${SCREEN_CAPTURE_KIT_LIBRARY}
         ${VIDEO_TOOLBOX_LIBRARY})
 
 set(APPLE_PLIST_TEMPLATE "${SUNSHINE_SOURCE_ASSETS_DIR}/macos/build/Info.plist.in")
@@ -55,6 +56,16 @@ set(PLATFORM_TARGET_FILES
         "${CMAKE_SOURCE_DIR}/src/platform/macos/nv12_zero_device.cpp"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/nv12_zero_device.h"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/publish.cpp"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/sc_video.h"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/sc_video.m"
         "${CMAKE_SOURCE_DIR}/third-party/TPCircularBuffer/TPCircularBuffer.c"
         "${CMAKE_SOURCE_DIR}/third-party/TPCircularBuffer/TPCircularBuffer.h"
         ${APPLE_PLIST_FILE})
+
+# sc_video.m is written against ARC for clarity (SCK APIs are async/
+# block-heavy and benefit from ARC). The rest of the macOS Obj-C
+# sources remain MRC; objects flowing across the boundary follow the
+# standard +1-retain alloc/init convention so both modes interoperate.
+set_source_files_properties(
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/sc_video.m"
+        PROPERTIES COMPILE_FLAGS "-fobjc-arc")
--- a/cmake/dependencies/macos.cmake
+++ b/cmake/dependencies/macos.cmake
@@ -10,6 +10,14 @@ FIND_LIBRARY(CORE_MEDIA_LIBRARY CoreMedia)
 FIND_LIBRARY(CORE_VIDEO_LIBRARY CoreVideo)
 FIND_LIBRARY(FOUNDATION_LIBRARY Foundation)
 FIND_LIBRARY(VIDEO_TOOLBOX_LIBRARY VideoToolbox)
+# ScreenCaptureKit is the modern (macOS 12.3+) replacement for the
+# deprecated AVCaptureScreenInput-based capture path. Sunshine's
+# sc_video.{h,m} is unconditionally compiled into the macOS target;
+# fail configure with a clear message rather than failing the build
+# later on header lookup when the SDK doesn't ship the framework
+# (e.g., when building with an Xcode older than 13.3 / SDK older than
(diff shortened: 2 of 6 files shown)


Nottlespike:
Most of the inline suggestions are duplicated from #5190 and have been addressed there (this PR's diff includes #5190's commit; once that lands the diff reduces to just the EDR commit). Specifically resolved in the amended SCK commit:

- SCStream output single-registration at init (`addStreamOutput`)
- `dispatch_queue_attr_make_with_qos_class` priority arg → `0`
- `streamRunning` / `currentCallback` / `currentSignal` thread safety via `@synchronized(self)`
- `CGDisplayCopyDisplayMode` NULL fallback to `CGDisplayBounds`
- Bounded 5-second timeouts on the SCK completion-handler semaphores

EDR-specific:

- **SDK compile-time guards**: added `#if defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 140000` around the `captureDynamicRange` / `SCCaptureDynamicRange*` block. The `@available(macOS 14.0, *)` runtime check is preserved inside the guarded block. On older SDKs the entire block is preprocessed away; on newer SDKs running on macOS 12.3-13.x, the runtime check skips it.

Force-pushed both branches.

Nottlespike:
**Force-pushed (rebased on new #5190 tip).** Same `SCREEN_CAPTURE_KIT_LIBRARY REQUIRED` addition is included via the rebase.

Re Copilot's latest inline pass: all six suggestions it surfaced are already fixed in the rebased branch, just at shifted line numbers post-amend. Quick map:

| Copilot says | Reality |
|---|---|
| `frameWidth/Height` can be 0 if `CGDisplayCopyDisplayMode` returns NULL | `CGDisplayBounds` fallback at `sc_video.m:73` |
| `DISPATCH_TIME_FOREVER` for `getShareableContent` (line 98) | Bounded to `kSCVideoCompletionTimeoutSec` (5s) at `sc_video.m:116` |
| `addStreamOutput` re-added on restart (line 210) | Added exactly once during `-initWithDisplay:frameRate:` (see lines 162–170); `-capture:` only swaps callbacks |
| `streamRunning` data race (lines 190, 300) | All reads/writes under `@synchronized(self)` |
| `@available(macOS 14.0, *)` compile-time SDK guard (lines 223, 231) | `#if defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 140000` at `sc_video.m:221` |
| `sc_video.m` unconditional inclusion (cmake lines 60, 71) | `FIND_LIBRARY(... REQUIRED)` in this revision — fails configure fast if SCK SDK absent |

The Copilot bot doesn't re-evaluate inline comments after force-pushes; the suggestions are valid critiques of the pre-amend code that's no longer present in the branch.

Nottlespike:
Closing and folding into #5190. Per the 'each PR must be mergeable in any order' guidance, the EDR work is textually dependent on the SCK API surface in #5190 (it would not compile against the legacy AVCaptureScreenInput path), so splitting it across two PRs is artificial. The EDR commit will be squashed onto #5190 along with the HDR-gating hardening (gating EDR on negotiated `enable_hdr` rather than just pixel-format depth).

PR #5192, opened 2026-05-25
feat(macos): experimental ProRes VideoToolbox encoder (opt-in, additive protocol)

Nottlespike:
## Description

Adds experimental, opt-in macOS ProRes support via VideoToolbox (`prores_videotoolbox`). Disabled by default; requires both a config opt-in (`prores_mode`) and a custom client that requests the ProRes codec (Moonlight does not).

## Independence

Compiles, links, and passes CI standalone. Does not block, and is not blocked by, any other open PR.

Functional interaction with #5190 (SCK backend with EDR): ProRes capture requires 10-bit (or wider) pixel buffers. The legacy `AVCaptureScreenInput` path only supports 8-bit BGRA, so on a tree that has #5192 but not #5190, the ProRes encoder probe will fail to initialize and Sunshine falls back to the standard codec list — same behavior as any other encoder that fails probe. When both PRs land, ProRes works end-to-end through SCK's 10-bit buffers. The two PRs can merge in either order; there is no textual dependency.

## What this changes

- `pix_fmt_e` gains `nv24` and `p410` (4:4:4 BiPlanar 8-bit / 10-bit) so the encoder pipeline can request the buffer types ProRes needs.
- `nv12_zero_device` maps the new formats to their `kCVPixelFormatType_*` equivalents at the platform/macOS layer.
- `prores_videotoolbox` encoder entry added in `src/video.cpp` with the correct color/dynamic-range probe knobs (`encoderCscMode=3` BT.709 full-range, `chromaSamplingType=1` 4:4:4, `dynamicRange=1`) so the probe succeeds on Apple Silicon.
- Config: `prores_mode` (0=disabled default / 1=accept client request / 2=force) and `prores_profile` (proxy/lt/standard/hq/4444/xq).
- UI: `prores_mode` and `prores_profile` both live on the VideoToolbox encoder tab; documented as constant-quality (slider is advisory; real stream rate is dictated by profile and content complexity — hundreds of Mbps at 4K HQ/4444 is normal).
- Tests: `ProResConfigTest` fixture covers the config defaults / forced-mode invariants without leaking global state.

## Bot/Sonar follow-up

Open SonarCloud findings on this PR are pre-existing globals (`S5421`) that can't be made `const` because they're mutated at runtime, and one pre-existing `S6045` (heterogeneous hashing) in `src/config.h` that's out of scope. See thread replies for the per-finding justifications.

## Why "experimental"

ProRes is a constant-quality codec; bitrate-targeted streaming doesn't apply, and Moonlight will warn about network bandwidth even on lossless links. Suited for local-LAN testing and capture workflows, not general streaming.


Nottlespike:
Force-pushed addressing all the inline suggestions:

- **`is_known_video_format` upper bound** (`src/video.h:33`): switched from hardcoded `video_format <= SUNSHINE_FORMAT_PRORES` to `video_format < static_cast<int>(SUNSHINE_FORMAT_COUNT)`. Future format additions now require updating only `SUNSHINE_FORMAT_COUNT` (and the constant declarations the `static_assert` block already pins), not the range check too.

- **`require_prores` from immutable source** (`src/video.cpp:2882`): now computed from `config::video.prores_mode >= 2` instead of the runtime-mutable `active_prores_mode`. `adjust_encoder_constraints` can still zero `active_prores_mode` when an encoder lacks ProRes support, but `require_prores` stays anchored to user intent so the "force ProRes" policy is consistently enforced across all encoder candidates in the search loop. Inline comment explains the dependency.

- **Test fixture for global config state** (`tests/unit/test_video.cpp`): converted `ProResConfigTest` from free `TEST(...)` cases with per-test manual save/restore into a `ProResConfigTest : testing::Test` fixture using `SetUp` / `TearDown`. Every test now starts from the documented defaults regardless of execution order or whether an earlier test fails mid-body. Closes the ordering/flakiness concern.

- **Vue numeric `v-model`** (`VideotoolboxEncoder.vue:46-50`): switched to `v-model.number="config.prores_mode"` with `:value="0"` / `:value="1"` / `:value="2"` so the model stays numeric instead of becoming `'0' | '1' | '2'` after user interaction. Avoids the subtle type confusion if any downstream JS does `===` comparisons against numeric literals.

Compile-checked locally on both edits. Skipped the `apply_config` API-surface point (`src/config.h:34`) as that's the existing implementation's scope; happy to revisit if you have a specific shape in mind.

Nottlespike:
**Force-pushed.** Significant updates after end-to-end runtime verification on M4 Max:

### Stacking

This PR now explicitly stacks on top of #5190 (ScreenCaptureKit backend) and #5191 (EDR / HDR for 10-bit pixel formats). Reason: `prores_videotoolbox` cannot accept any pix_fmt that `AVCaptureScreenInput` is capable of producing — its supported input formats are 4:2:2 / 4:4:4 BiPlanar (or `AV_PIX_FMT_VIDEOTOOLBOX`-wrapped CVPixelBuffers of those types). SCK is the only macOS capture API that can deliver these. When #5190 and #5191 merge, this PR's diff collapses to just the ProRes-specific changes.

### New 4:4:4 BiPlanar capture path (added in this revision)

Wiring ProRes end-to-end required extending Sunshine's pix_fmt plumbing to support 4:4:4 BiPlanar formats, which were never needed before because H.264/HEVC/AV1 VideoToolbox are 4:2:0 only:

- **`src/platform/common.h`**: added `nv24` and `p410` to `pix_fmt_e` + `from_pix_fmt`.
- **`src/video.cpp` `map_pix_fmt`**: added cases for `AV_PIX_FMT_NV24` and `AV_PIX_FMT_P410`.
- **`src/video.cpp` `videotoolbox` encoder declaration**: filled the previously-NONE 4:4:4 pix_fmt slots with `AV_PIX_FMT_NV24` and `AV_PIX_FMT_P410`; added `YUV444_SUPPORT` to the encoder flags.
- **`src/platform/macos/display.mm`**: extended `make_avcodec_encode_device` to route `nv24` / `p410` through `nv12_zero_device` alongside the existing `nv12` / `p010` paths.
- **`src/platform/macos/nv12_zero_device.cpp`**: extended `init` to set the matching CVPixelBufferType (`kCVPixelFormatType_444YpCbCr*BiPlanarVideoRange`) per pix_fmt.

H.264 / HEVC VideoToolbox don't gain anything new functionally — they remain 4:2:0 only on Apple Silicon hardware — but the 4:4:4 probe runs against them harmlessly and falls through with their `YUV444` capability bit `false`, which is correct.

### ProRes encoder probe (also added in this revision)

The original probe ran ProRes through the same SDR 8-bit `config_max_ref_frames` / `config_autoselect` templates the H.264 / HEVC encoders use, and ProRes correctly refuses to open against those (ProRes has no 8-bit input path, no BT.601 path, no 4:2:0 path). The probe now uses a ProRes-specific config:

- `dynamicRange = 1` → 10-bit pix_fmt instead of 8-bit
- `encoderCscMode = 3` → BT.709 full range instead of BT.601
- `chromaSamplingType = 1` → 4:4:4 (P410) instead of 4:2:0 (P010)

Any successful ProRes probe inherently uses 10-bit input, so `DYNAMIC_RANGE` is promoted eagerly here rather than relying on the subsequent `test_hdr_and_yuv444` lambda, which gates on `PASSED` and only sets `DYNAMIC_RANGE` itself.

### Build-deps dependency

This PR's runtime verification requires [`LizardByte/build-deps#693`](https://github.com/LizardByte/build-deps/pull/693) (the `--enable-encoder=...,prores_videotoolbox` FFmpeg configure flag) to actually compile the encoder symbol into the static libavcodec.a. The Sunshine PR will pass code review without it, but the runtime probe needs the build-deps change to succeed.

### Verified end-to-end

Startup log on M4 Max with all changes:

```
Found H.264 encoder: h264_videotoolbox [videotoolbox]
Found HEVC encoder: hevc_videotoolbox [videotoolbox]
Found ProRes encoder: prores_videotoolbox [videotoolbox]
```

Encoder probe creates `prores_videotoolbox` cleanly at 10-bit Rec.709 4:4:4 P410, validates, and the `Found ProRes encoder` line lands. Was previously failing with "Couldn't open" at every config permutation until both the capture-side 4:4:4 pipeline and the probe config were in place.

Nottlespike:
**Force-pushed** addressing the 7 SonarCloud code smells:

**Fixed (2):**

- `src/platform/macos/nv12_zero_device.cpp:64` — added `using enum pix_fmt_e;` inside the function so the case labels can drop the `pix_fmt_e::` prefix (S6177).
- `src/nvenc/nvenc_base.cpp:459` — extracted the nested ternary (`videoFormat == H264 ? "H.264 " : … : "ProRes " : " "`) into an IIFE'd `switch` statement (S3358). Easier to read, easier to extend.

**Defending (4 + 1):**

- `src/video.cpp:1224 / 1226`, `src/video.h:368 / 370` — Sonar flags `active_hevc_mode` / `active_prores_mode` (and their `extern` declarations) as "global variables should be const" (S5421). These cannot be made `const` because they are mutated at runtime in `adjust_encoder_constraints` when the chosen encoder doesn't support the requested codec mode (e.g. HEVC mode 3 gets demoted to 0 if the encoder lacks `DYNAMIC_RANGE`). The existing `active_hevc_mode` and `active_av1_mode` follow the same mutable pattern from prior PRs; `active_prores_mode` simply mirrors it. Making these const would require a behavioural refactor that's out of scope for the ProRes feature PR.
- `src/config.h:34` — Sonar wants `apply_config()`'s `std::unordered_map<std::string, std::string>` parameter to use transparent hashing (S6045) for `string_view` lookups. The API was added in an earlier commit on this branch (`d238fa63`) and is a single internal-use entrypoint for testing; the heterogeneous-hash refactor is a project-wide style change that affects more than this PR.

Happy to address either set if reviewers prefer otherwise.

Nottlespike:
Re Copilot's three latest inline comments:

- **`src/video.cpp:2945 / 2978` — `require_prores` from mutable `active_prores_mode`** — Already addressed earlier in the branch's history. The current line (post-rebase, around `src/video.cpp:2921`) is:

  ```cpp
  const bool require_prores = config::video.prores_mode >= 2;
  ```

  Derives from the immutable user-intent source (`config::video.prores_mode`), not the runtime-mutable `active_prores_mode`. `adjust_encoder_constraints` can still zero `active_prores_mode` when an encoder lacks ProRes support, but `require_prores` stays anchored to user intent so the "force ProRes" policy is consistently enforced across all encoder candidates in the search loop. The inline comment in the code explains the rationale. Copilot's reference to line 2945 / 2978 is from a stale diff representation; the actual `require_prores =` line is unique and already uses the immutable source.

- **`src/config.h:34` — `apply_config()` in public header** — Same as my Sonar reply above: the API was added in an earlier commit on this branch and is currently a single internal-use entrypoint for testing. The transparent-heterogeneous-hashing refactor is a project-wide style change worth its own PR, not in scope here.

Force-pushed after rebasing onto the latest #5191 tip (which includes the new `SCREEN_CAPTURE_KIT_LIBRARY REQUIRED` configure check from #5190). No code changes specific to this PR in the rebase — diff is identical to before for the ProRes-only portion.

Nottlespike:
**Force-pushed.** Rebased onto the new combined #5190 tip (`9030b66a`) which now bundles SCK + EDR + HDR gating into a single PR.

Updated PR body to make the any-order mergeability explicit: this PR compiles and passes CI standalone. The functional interaction with #5190 is that ProRes capture needs 10-bit buffers; on a tree with #5192 but without #5190 the ProRes encoder probe fails and Sunshine falls back to the standard codec list (same path as any other failed-probe encoder), preserving existing behavior. When both land, ProRes works end-to-end through SCK's 10-bit path.

No code changes from the prior tip beyond the rebase context — same ProRes commit content (`687b6c93` ← `47d4b0de`), same UI relocation commit (`39eb20f7` ← `f47b3953`).

andygrundman:
I forgot to ask, how are you viewing this stream on the client side?

Nottlespike:
Closing — maintaining this work in our fork (RESMP-DEV/Sunshine) going forward rather than upstreaming. No issue with the project; just diverging priorities after this round of review. The code remains MIT-licensed and available to anyone who wants to pull from the fork or cherry-pick. Thanks for the engagement on this batch.

Nottlespike:
Reopening — closed prematurely. This change is broadly applicable, not specific to the M4 Max I tested on, and the post-review feedback is fully addressed. The fork (RESMP-DEV/Sunshine) will be the home for hardware-specific experimental work (e.g., software AV1 via libsvtav1 that only makes sense on machines with extreme CPU headroom); but the minimum-broadly-supported changes like this one belong upstream where the broader user base benefits.

Nottlespike:
Reopening — closed prematurely. This change is broadly applicable (M-series Apple Silicon, but not specific to my M4 Max test host), and the post-review feedback is fully addressed in additional commits.

The fork (RESMP-DEV/Sunshine) is becoming the home for hardware-specific experimental work that I cannot validate across the full Sunshine user base (e.g., software AV1 via libsvtav1 that only makes sense on machines with extreme CPU headroom). But minimum-broadly-supported changes like this one belong upstream where the broader user base benefits.

PR #5201, opened 2026-05-27
feat(macos/capture): ScreenCaptureKit backend with gated EDR for 10-bit pixel formats

Nottlespike:
Reopens #5190 (closed prematurely; GitHub blocked direct reopen due to regenerated branch SHAs). All review feedback from the original cycle is addressed in additional commits. See the original PR for full discussion; rebased on current master.


Nottlespike:
**Force-pushed.** Addressing the maintainer feedback I missed during the original review cycle:

> **@andygrundman**: "Is there any reason to keep the old AVCapture version? SCK looks like it was introduced in 12.3 and our current MACOSX_DEPLOYMENT_TARGET is 14.2."
> **@ReenigneArcher**: "shouldn't keep workarounds for versions older than what we support (14.2)."

Right call — the \`@available(macOS 12.3, *)\` branch and the entire \`AVVideo\` class were dead code on every supported build. New commit on the branch removes them:

- \`@available(macOS 12.3, *)\` check in \`display.mm\` → gone; SCK is the only branch.
- \`id<SunshineVideoCapture>\` → replaced with \`SCVideo *\` directly. The protocol existed to abstract over both classes; with only SCVideo left, the indirection is pure overhead.
- \`FrameCallbackBlock\` typedef, \`+displayNames\`, \`+getDisplayName:\` → moved from \`av_video.{h,m}\` into \`sc_video.{h,m}\` (they used CG/AppKit, not AVFoundation, so no AVFoundation dependency comes along).
- \`src/platform/macos/av_video.{h,m}\` → deleted (208 lines).
- Both removed from \`PLATFORM_TARGET_FILES\` in \`cmake/compile_definitions/macos.cmake\`.

Net diff: -208 / +75 lines, single concrete macOS capture backend, no runtime feature flags to chase.

Nottlespike:
Pushed `45382a2b` addressing Copilot inline feedback.

**Real bugs fixed (2):**

- **`sc_video.m:339` — dealloc signals pending semaphore.** Was clearing `currentCallback` but never signalling `currentSignal`. If the stream stopped without firing `-stream:didStopWithError:`, any caller still waiting on the semaphore returned by `-capture:` would stall forever. Snapshot the pending signal in the `@synchronized` block, then send it after clearing the callback so the waiter wakes up to observe nil callback and exits. Real concurrency bug — good catch.

- **`sc_video.m:44` — drop unused `streamOutputAdded` property.** Set in `-init` but never read anywhere. Removed both the `@property` declaration and the assignment. The "register output exactly once at init" invariant is now structural (the call is in `-init`, not gated on a flag) so the dead state can't drift.

**Defended (2) — with rationale:**

- **`av_video.h:47` — interface re-declares protocol properties.** Stylistic, but explicit declaration on the conforming class makes the type-of-property visible to readers without having to chase the protocol header, especially for the `IsTrayPresented` / `pixelFormat` boundary where attributes (`atomic`, `readonly`) matter. The protocol is authoritative; the class declaration is intentionally redundant for readability. Will fix if you'd prefer the leaner form.

- **`cmake/dependencies/macos.cmake:20` — `REQUIRED` SCK.** Sunshine's `MACOSX_DEPLOYMENT_TARGET` is 14.2 (well above SCK's 12.3 minimum). On any SDK old enough to lack ScreenCaptureKit, the rest of `sc_video.m` won't compile either, so failing at configure time gives a clearer error than letting the build break mid-compile on missing headers. `REQUIRED` matches `MACOSX_DEPLOYMENT_TARGET ≥ 14.2`. (@andygrundman / @ReenigneArcher had also indicated SCK-only is the direction — see the comment chain on AVCapture removal.)

Build clean on macOS 26.5 / Apple clang 21. The bug fix for #1 is the substantive one — it eliminates a real wait-forever path during shutdown.

ReenigneArcher:
Thank you for the PR, but closing as another contributor is working on this.

PR #5202, opened 2026-05-27
feat(macos): experimental ProRes VideoToolbox encoder (opt-in, additive protocol)

Nottlespike:
Reopens #5192 (closed prematurely; GitHub blocked direct reopen due to regenerated branch SHAs). All review feedback from the original cycle is addressed in additional commits. See the original PR for full discussion; rebased on current master.


Nottlespike:
@andygrundman re: "how are you viewing this stream on the client side?" — testing with a custom Moonlight build that has the ProRes decode path wired (stock Moonlight has no ProRes decoder, which is why this PR is opt-in via \`prores_mode\` and the protocol negotiation is additive — a stock Moonlight client never sees ProRes signaled).

The intent is to leave the door open for custom clients that want lossless-ish desktop streaming on LAN — capture workflows, A/B tests against HEVC, etc. — without affecting the default streaming path at all.

Nottlespike:
Force-pushed (`eedcf8fd`) — rebased onto the new #5201 tip (`45382a2b`) which carries the dealloc-semaphore fix Copilot flagged on both PRs. No conflicts.

Re Copilot's specific findings on this PR:

- **`sc_video.m:339` — use-after-free risk on dealloc.** Addressed via the #5201 rebase. Pending `currentSignal` is now snapshotted and signalled in dealloc so waiting callers can never observe a freed `self`. Also note the existing `[stream stopCaptureWithCompletionHandler:]` + 2 s wait already establishes a teardown barrier before `dealloc` returns — the new signal mainly closes the corner case where `-capture:` is mid-wait at shutdown.

- **`video.cpp:2944` — `prores_mode` on non-macOS builds.** Real point. ProRes config option is already platform-gated by virtue of the `prores_videotoolbox` encoder only being registered on `#ifdef __APPLE__`, so setting `prores_mode=2` on Linux/Windows will fail at the encoder-lookup step. But agreed it's tidier to short-circuit before that — happy to add a platform guard in `adjust_encoder_constraints` if you'd like. Will fold into the next push if confirmed.

- **`rtsp.cpp:822` — SDP injection from `prores_profile`.** Config parsing in `src/config.cpp` already constrains `prores_profile` to a known enum set (`proxy`/`lt`/`standard`/`hq`/`4444`/`xq`); arbitrary strings can't reach the SDP emitter. Defense-in-depth would still be nice — I can add an explicit whitelist at emission. Will queue.

- **`display.mm:97` — C-style cast for Obj-C object through `void *`.** Consistent with the existing pattern in this file (`(SCVideo *) display`, `(id<SunshineVideoCapture>) display`, etc.). Switching to `__bridge` casts requires either ARC (this file is MRC) or `__bridge_retained` semantics that the surrounding code isn't structured for. Defendable as-is; would prefer to keep cast style consistent across the file.

- **`tests/unit/test_video.cpp:15` — duplicate `static_assert`s.** Will trim the duplicate.

Let me know on the platform guard / SDP whitelist / test cleanup preferences; happy to do them in a follow-up commit on this PR.

ReenigneArcher:
Closing since no Moonlight clients can decode this.

PR #5203, opened 2026-05-27
feat(macos/input): host-side virtual gamepad via IOHIDUserDevice

Nottlespike:
## Description

Replaces the macOS gamepad stubs in `src/platform/macos/input.cpp` with a working virtual gamepad implementation backed by `IOHIDUserDevice`. Moonlight clients' gamepad events have been silently dropped on the macOS host since the port shipped; this lights up the path end-to-end.

Approach: publish a virtual HID device with a complete Xbox-style report descriptor (16 buttons, 4-bit hat switch, two 8-bit triggers, four 16-bit signed stick axes). macOS's HID matching layer publishes the device to userspace; `GameController.framework`, SDL, Steam, and apps consuming raw IOKit HID all see it as a real USB controller.

VID `0x1209` / PID `0x5853` — pid.codes open-source VID range, intentionally not in SDL's known-controller database so it routes through generic GameController mapping.

## Required boot flag

`IOHIDUserDeviceCreateWithProperties` returns a NULL device unless `com.apple.developer.hid.virtual.device` is granted, and that entitlement is only honored when AMFI is bypassed at boot:

\`\`\`
sudo nvram boot-args=\"amfi_get_out_of_my_way=1\"   # plus any existing flags
# reboot
\`\`\`

SIP can stay on. The implementation probes once at `platf::input()` construction; without the flag, `alloc_gamepad` returns `-1` and the startup log emits a clear instruction line. This matches behavior on Linux when `uinput` is unavailable.

A DriverKit DEXT (no boot flag, requires Apple DriverKit entitlement approval + signed system-extension distribution) would be the longer-term shipping story but is multi-week work. Submitting the IOHIDUserDevice path first so the feature gap closes today.

## Changes

- **`src/platform/macos/hid_gamepad.{h,m}`** (new): MRC Obj-C wrapper around `IOHIDUserDevice`. Defines the gamepad HID report descriptor (Usage Page Generic Desktop / Usage Joystick to dodge SDL's GCController-only path on known controller IDs). Maps Sunshine's 32-bit `buttonFlags` to the 16-bit HID button field plus a 4-bit hat switch. Bounded 2 s teardown via `IOHIDUserDeviceSetCancelHandler`. Report dispatch on a dedicated `QOS_CLASS_USER_INTERACTIVE` serial queue.

- **`src/platform/macos/input.cpp` → `input.mm`**: renamed so the gamepad path holds `HIDGamepad` strong references directly without a C bridge. Mouse/keyboard injection paths are unchanged. `macos_input_t` gains a 16-slot `HIDGamepad *` array plus a `hid_gamepad_available` flag (probed once at init). `alloc_gamepad` finds the lowest free slot, creates a virtual device, returns the slot index. `free_gamepad` disconnects and releases. `gamepad_update` forwards state to the slot's `HIDGamepad`. `supported_gamepads` reports `macos_hid` or `macos_amfi_required` depending on probe result.

- **`cmake/dependencies/macos.cmake`**: \`FIND_LIBRARY(IO_KIT_LIBRARY IOKit)\` for the new symbols.

- **`cmake/compile_definitions/macos.cmake`**: add `hid_gamepad.{h,m}` to `PLATFORM_TARGET_FILES`, rename `input.cpp` → `input.mm`, link `IO_KIT_LIBRARY`.

## What this does NOT do

- **Rumble / haptic feedback.** \`feedback_queue_t\` is accepted but not forwarded — virtual HID has no force-feedback motor to drive. Easy follow-up if wanted.
- **DualSense touchpad / gyro / adaptive triggers / LED.**
- **Battery state.**
- **DriverKit DEXT path** for users who won't disable AMFI.

All four are additive to the existing \`HIDGamepad\` class and can ship in follow-up PRs.

## Attribution

The IOHIDUserDevice approach and the HID report descriptor are derived from [Lumen](https://github.com/trollzem/Lumen), an Apple-Silicon-focused Sunshine fork. Lumen is MIT-licensed. The HID report descriptor is copied verbatim with attribution in the source file header; the surrounding lifecycle/integration code is original to this PR.

## Verification

- Builds clean on macOS 26.5 / Apple clang 21 / Apple Silicon.
- \`_OBJC_CLASS_\$_HIDGamepad\` is present in the linked binary.
- AMFI-on probe correctly fails fast at startup and logs the instruction.
- End-to-end gamepad flow under AMFI-off (controller-driven game receiving inputs sourced from a Moonlight client over the network) verified locally on M4 Max.

## Independence

Stands alone. Does not depend on any other open PR. No interaction with the SCK / EDR / ProRes work (separate PRs).

### DIFF
--- a/cmake/compile_definitions/macos.cmake
+++ b/cmake/compile_definitions/macos.cmake
@@ -35,6 +35,7 @@ list(APPEND SUNSHINE_EXTERNAL_LIBRARIES
         ${CORE_MEDIA_LIBRARY}
         ${CORE_VIDEO_LIBRARY}
         ${FOUNDATION_LIBRARY}
+        ${IO_KIT_LIBRARY}
         ${VIDEO_TOOLBOX_LIBRARY})
 
 set(APPLE_PLIST_TEMPLATE "${SUNSHINE_SOURCE_ASSETS_DIR}/macos/build/Info.plist.in")
@@ -48,7 +49,9 @@ set(PLATFORM_TARGET_FILES
         "${CMAKE_SOURCE_DIR}/src/platform/macos/av_video.h"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/av_video.m"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/display.mm"
-        "${CMAKE_SOURCE_DIR}/src/platform/macos/input.cpp"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/hid_gamepad.h"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/hid_gamepad.m"
+        "${CMAKE_SOURCE_DIR}/src/platform/macos/input.mm"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/microphone.mm"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/misc.mm"
         "${CMAKE_SOURCE_DIR}/src/platform/macos/misc.h"
--- a/cmake/dependencies/macos.cmake
+++ b/cmake/dependencies/macos.cmake
@@ -9,6 +9,12 @@ FIND_LIBRARY(CORE_AUDIO_LIBRARY CoreAudio)
 FIND_LIBRARY(CORE_MEDIA_LIBRARY CoreMedia)
 FIND_LIBRARY(CORE_VIDEO_LIBRARY CoreVideo)
 FIND_LIBRARY(FOUNDATION_LIBRARY Foundation)
+# IOKit is needed for IOHIDUserDevice* (virtual gamepad device — hid_gamepad.m).
+# Actually creating devices at runtime requires the user to disable AMFI via
+# `nvram boot-args="amfi_get_out_of_my_way=1"`, but the symbols themselves
+# are unconditionally present and the host alloc_gamepad path probes
+# availability before relying on them.
+FIND_LIBRARY(IO_KIT_LIBRARY IOKit)
 FIND_LIBRARY(VIDEO_TOOLBOX_LIBRARY VideoToolbox)
 
 if(SUNSHINE_ENABLE_TRAY)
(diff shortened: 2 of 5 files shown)


Nottlespike:
Pushed `10f22b2d` addressing Copilot + Codex review feedback.

**Real bugs fixed (3):**

- **Codex P2: `alloc_gamepad` return contract.** Previously returned the slot index; `src/platform/common.h:831` documents "return 0 on success" and `src/input.cpp:897` calls `if (platf::alloc_gamepad(...))` treating any non-zero as failure. The implementation now keys HIDGamepad slots off `id.globalIndex` (which Sunshine allocates from a bitmask before calling us) and returns `0`/`-1` per the contract, matching Windows. Without this, only the first controller would have worked — every additional gamepad would be silently dropped.

- **Codex P2: `supported_gamepads` `is_enabled` mirrors probe.** When AMFI is enabled and `HIDGamepad isAvailable` returns NO, the entry was still pushed as `is_enabled=true`. The web UI would offer a controller type that immediately fails. Now `is_enabled` matches the probe so the UI surfaces the correct state.

- **Copilot: SIP → AMFI terminology fix.** `hid_gamepad.{h,m}` header docs said "Requires SIP to be disabled"; the actual gating mechanism is AMFI. Updated both files' doc blocks consistently. User-facing instruction strings already referenced `amfi_get_out_of_my_way`; this brings header docs in line.

**Defended (3) — included in commit body for record:**

- Copilot: `@kIOHIDVendorIDKey` is fine — IOKit headers `#define kIOHIDVendorIDKey "VendorID"` as a plain C string, so `@kIOHIDVendorIDKey` correctly expands to `@"VendorID"`. The dictionary construction is correct.
- Copilot: `strong` in MRC is treated as `retain` by clang. Matches Lumen's original for traceability.
- Copilot: NSLog vs BOOST_LOG — `hid_gamepad.m` is plain Obj-C by design; logs only fire during device create/teardown (max 16x per session). Switching to BOOST_LOG would require renaming to `.mm` and pulling Boost log headers into a hot-path file for negligible benefit.

All addressed in 3 files, +55/-26 lines. The fix for #1 is the important one — without it, multi-gamepad scenarios would fail silently. Thanks for the careful pass.

Kishi85:
It is just me or is this trying to achieve the same as #5171 but instead of doing the proper code signing way to get the necessary entitlement it's requiring a bootflag that works around it (potentially reducing system security)? Not sure if this is acceptable or practicable for most users.

ReenigneArcher:
Closing as this is a duplicate effort.

Issue #2463, opened 2024-04-23
Incorrect keyboard mapping using a QWERTZ (de_DE) keyboard using an Android Moonlight client


FlorianWolters:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

After connecting (using direct LAN connection) via the "Moonlight" app to the "Sunshine" instance running on my Windows PC, the keyboard behaves as it is QWERTY (en_US), e.g. pressing the `Z` and `? ß \` keys (and some others) on my keyboard generate incorrect output. I've tested this with a text editor.

- On the Windows host only one language "Germany" and one keyboard layout is "installed".
- The input works properly if using a German QWERTZ (de_DE) keyboard attached to the Windows PC that is running "Sunshine".

### Expected Behavior

I identify multiple approaches:

1. A setting `force_client_keyboard` is added to "Sunshine" that automatically detects the client keyboard and does not required any additional configuration (e.g. modifying `keybindings`).
2. The functionality behind the setting `always_send_scancodes` is updated in a way that it also works with non en_US keyboard. As it is now, deactivating this option seems to fix the key mappings, but it makes PC Gaming unusable, e.g. holding the SHIFT key does not work, using the NUM LOCK does not work as with a hardware keyboard, etc.
3. The documentation provides working `keybindings` settings for commonly used keyboard layouts. As it is now I was not able to manually map some of the incorrectly mapped keys.

### Additional Context

There are multiple (closed) issues that seem to be related to this issue:

- #493
- #780
- #989
- #1392
- https://github.com/loki-47-6F-64/sunshine/issues/82 

I did comment on some of these. Since I couldn't come up with a working solution yet, I've created this new separate issue. My comments can be found here:
- https://github.com/LizardByte/Sunshine/issues/493#issuecomment-1928138874
- https://github.com/LizardByte/Sunshine/issues/1392#issuecomment-1925362069
- https://github.com/loki-47-6F-64/sunshine/issues/82#issuecomment-1861836889

### Host Operating System

Windows

### Operating System Version

10.0.22631.3447

### Architecture

64 bit

### Sunshine commit or version

0.23.1

### Package

Windows - Winget (Third Party)

### GPU Type

Nvidia

### GPU Model

NVIDIA GeForce RTX 4070

### GPU Driver/Mesa Version

552.22

[...]

PR #5273, opened 2026-06-08
fix(linux/input): type printable ASCII via keycodes instead of Ctrl+Shift+U

OMGrant:
## Description

On Linux hosts, typing certain characters via Moonlight's UTF-8 text input path corrupts or clears the target text field instead of inserting the character. The most reliable trigger is `@` typed from a mobile soft keyboard (e.g. Android Gboard), which makes logging in over Moonlight on a Linux host effectively impossible.

### Root cause

`platf::keyboard::unicode()` (`src/platform/linux/input/inputtino_keyboard.cpp`) types **all** UTF-8 text using the IBus/GTK `Ctrl+Shift+U` + hex-codepoint entry method:

```
press CTRL, press SHIFT, press U, release U, <hex digits...>, release SHIFT, release CTRL
```

That method is only honored by IBus-enabled GTK apps. In Chrome, KDE/Qt apps, Wayland-native apps, games, and password fields it is **not** interpreted as Unicode entry — the leading `Ctrl`-modified chord (`Ctrl+U`, etc.) is interpreted as an application shortcut and clears/corrupts the field.

Mobile Moonlight clients commit symbols-page characters (like `@`) as UTF-8 text rather than as keycodes, so this path is exercised during ordinary typing.

### Evidence

Sunshine debug log (Plasma 6 Wayland host, US layout), typing `@` from Android/Gboard:

```
--begin unicode packet--
text [@]
--end unicode packet--
Unicode, typing U+40        # Ctrl+Shift+U path -> clears the field in Chrome/KDE
```

The same `@`, when the client happens to send it as a keycode instead, works perfectly:

```
keyCode [8032]  (VK_2)   modifiers [01]  (Shift)
```

So the keycode+Shift path is correct; only the UTF-8 text path is broken.

### Fix

Add a US-layout `char32_t -> {VK code, needs Shift}` map and inject printable ASCII (0x20–0x7E) as real keystrokes — exactly mirroring the working keycode path above. The `Ctrl+Shift+U` method is retained only as a fallback for genuine non-ASCII (emoji, accented characters, ...), where it remains the best available option.

For example, `@` now emits `Shift↓ · 2↓↑ · Shift↑` — identical to a hardware keyboard, and works in every app.

### Notes / testing

- Root cause and the corrected keystroke sequence were verified against the Sunshine debug logs above; the ASCII output now matches Sunshine's existing (working) keycode path.
- The fallback path for non-ASCII is unchanged.
- Scope is limited to `inputtino_keyboard.cpp`; no protocol or cross-platform changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/platform/linux/input/inputtino_keyboard.cpp
+++ b/src/platform/linux/input/inputtino_keyboard.cpp
@@ -170,23 +170,83 @@ namespace platf::keyboard {
     }
   }
 
+  /**
+   * US-layout map of printable ASCII -> {Moonlight/Windows VK code, needs Shift}.
+   *
+   * Lets us inject printable ASCII as real US-layout keystrokes instead of the
+   * IBus/GTK "Ctrl+Shift+U + hex codepoint" entry method used below. That method
+   * is only honored by IBus-enabled apps; in Chrome, KDE, Wayland-native apps and
+   * password fields the leading Ctrl chord is interpreted as a shortcut (e.g.
+   * Ctrl+U) and corrupts/clears the target field. Mobile soft keyboards (Gboard)
+   * commit symbols like '@' as UTF-8 text, so this path is hit in normal typing.
+   */
+  static const std::map<char32_t, std::pair<short, bool>> ascii_to_vk = {
+    {U' ', {0x20, false}},
+    {U'0', {0x30, false}}, {U'1', {0x31, false}}, {U'2', {0x32, false}}, {U'3', {0x33, false}}, {U'4', {0x34, false}},
+    {U'5', {0x35, false}}, {U'6', {0x36, false}}, {U'7', {0x37, false}}, {U'8', {0x38, false}}, {U'9', {0x39, false}},
+    {U')', {0x30, true}}, {U'!', {0x31, true}}, {U'@', {0x32, true}}, {U'#', {0x33, true}}, {U'$', {0x34, true}},
+    {U'%', {0x35, true}}, {U'^', {0x36, true}}, {U'&', {0x37, true}}, {U'*', {0x38, true}}, {U'(', {0x39, true}},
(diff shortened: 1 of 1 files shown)


OMGrant:
> Avoid hard-coding US keystrokes for text input

Good catch, thanks — you're right the fast path is US-specific (`Shift+2` is `@` on US, `"` on a German layout).

Context worth noting: Sunshine's Linux keyboard injection is already US-centric. inputtino emits fixed US-positioned evdev keycodes and there's no xkbcommon/keymap handling anywhere in `src/`, so the normal keycode path already mistypes on a non-US host layout (y/z on QWERTZ, cf. #2463). The `Ctrl+Shift+U` Unicode path was the one layout-independent method — but only in IBus apps; in Chrome/KDE/Wayland-native/password fields it corrupts the field (the bug this PR fixes).

So the trade-off: today non-US users get a correct `@` only in IBus apps (destructive elsewhere); with this patch US users get a correct `@` everywhere, and non-US users would get the US-positioned character.

The fully-correct fix is layout-aware injection — resolve each codepoint to keycode+level via libxkbcommon against the active keymap. That's a larger change (new dependency + host-layout detection, non-trivial on Wayland), arguably its own feature.

How would you like to scope this PR?

1. Gate the ASCII fast path to US layouts, keeping the Unicode fallback otherwise — no non-US regression, but needs layout detection Sunshine doesn't currently do.
2. Do the layout-aware fix properly via libxkbcommon — happy to take it on if you want it here.
3. Land as-is, documenting the US-layout assumption (consistent with the existing keycode path).

Happy to implement whichever direction you prefer.

ReenigneArcher:
https://docs.lizardbyte.dev/latest/developers/contributing.html

OMGrant:
Closing this in favor of a plain bug report: #5274.

Per the contributing guidelines, the fix is better implemented by the maintainers directly (ideally layout-aware via libxkbcommon rather than my US-layout shortcut), so I've moved the root-cause analysis, reproduction, and debug logs into the issue. Thanks @ReenigneArcher for the pointer to the policy.

PR #5284, opened 2026-06-12
fix(tray): fix compiling without system tray enabled

### LINKED ISSUE #5278: SUNSHINE_ENABLE_TRAY=OFF has no effect: SUNSHINE_TRAY compile definition is captured before the option is applied

### Describe the Bug

Building on Linux with `-DSUNSHINE_ENABLE_TRAY=OFF` does not actually
disable the tray. The compile definition `SUNSHINE_TRAY` is still emitted
as `1`, so `src/system_tray.cpp` is compiled with the tray code active.

On a source tree without the `tray` submodule this fails the build with:

    src/system_tray.cpp:42:12: fatal error: tray.h: No such file or directory

`CMakeCache.txt` correctly reflects the option:

    SUNSHINE_ENABLE_TRAY:BOOL=OFF

but the generated compile flags do not:

    $ grep -o 'SUNSHINE_TRAY=[0-9]' build/CMakeFiles/sunshine.dir/flags.make
    SUNSHINE_TRAY=1
    SUNSHINE_TRAY=1

### Root Cause

It is an ordering issue in the CMake scripts:

- `cmake/prep/constants.cmake:5` sets the default `set(SUNSHINE_TRAY 1)`
- `cmake/compile_definitions/common.cmake:128` appends the definition
  while the value is still 1:
  `list(APPEND SUNSHINE_DEFINITIONS SUNSHINE_TRAY=${SUNSHINE_TRAY})`
- `cmake/compile_definitions/common.cmake:163-166` only afterwards, in the
  `else()` branch (tray disabled), runs `set(SUNSHINE_TRAY 0)`

Because the `list(APPEND ...)` at line 128 runs before the value is lowered
to 0 at line 166, the definition is permanently baked as `SUNSHINE_TRAY=1`
regardless of the option.

### Suggested Fix

Move the `list(APPEND SUNSHINE_DEFINITIONS SUNSHINE_TRAY=${SUNSHINE_TRAY})`
to after the `if(SUNSHINE_ENABLE_TRAY)` / `else()` block, so it captures the
final value. (The same `if(SUNSHINE_ENABLE_TRAY)` pattern also exists in
`cmake/dependencies/macos.cmake`, worth checking it honors the option too.)

### Expected Behavior

With `-DSUNSHINE_ENABLE_TRAY=OFF`, the compiler should receive
`SUNSHINE_TRAY=0` and `system_tray.cpp` should compile without requiring
the `tray` submodule.

### Host Operating System
Linux (Void Linux)

### Sunshine release
v2026.608.233907

### Expected Behavior

_No response_

### Additional Context

_No response_

### Host Operating System

Linux

### Operating System Version

Void Linux

### Architecture

amd64/x86_64

### Package

n/a

### GPU Type

AMD

### GPU Model

RX 9060XT

### GPU Driver/Mesa Version

26.0.6_1

### Capture Method

KMS (Linux)

### Apps

```json

```

### Log output

```shell

```

### Online logs

_No response_

Kishi85:
## Description

Currently sunshine cannot be compiled with `SUNSHINE_ENABLE_TRAY=0` due to ordering issues. Fixed by moving one line to the end of the related cmake definition. 

Afterwards a linker error was left that had to be fixed in main.cpp due to `system_tray::process_tray_events()` not being defined when compiled without system tray. The code itself would have never been reached in that case but the linker requires it.

### Issues Fixed or Closed

 - Fixes #5278 

## Type of Change

## Checklist

### AI Usage


src/main.cpp
```diff
@@ -116,7 +116,9 @@ void mainThreadLoop(const std::shared_ptr<safe::event_t<bool>> &shutdown_event)
 
   // Main thread event loop
   BOOST_LOG(info) << "Starting main loop"sv;
+#if defined SUNSHINE_TRAY && SUNSHINE_TRAY >= 1
```
ReenigneArcher, line 119: Can you use `tray_is_enabled` instead of another macro?

Kishi85: Unfortunately not. This is logically already skipped due to the run loop check above but the linker will fail  in the end due to `system_tray::process_tray_events()` not being linkable even if it is never reached. So the only way to handle this is via macro unless you know of another option?

ReenigneArcher: I guess not easily without refactoring the tray code.

PR #5288, opened 2026-06-13
fix(graphics): add EGL_BAD_MATCH fallback for incompatible format modifiers

muracodex:
## Problem

When capturing from displays that report DMA-BUF format modifiers incompatible with the GPU's EGL implementation (e.g. USB/DisplayLink monitors), `eglCreateImage()` fails with `EGL_BAD_MATCH` (0x3009). This causes the error:

```
Couldn't import RGB Image: 00003009
```

The result is that streaming shows only static/interference instead of the actual display content.

## Root Cause

The `surface_descriptor_to_egl_attribs()` function includes the display's DMA-BUF format modifier when creating the EGL image. Some displays (particularly USB-connected ones) report modifiers that the GPU's EGL implementation cannot handle, causing `eglCreateImage()` to fail.

## Fix

In `import_source()`, when the initial `eglCreateImage()` fails with `EGL_BAD_MATCH`, retry the import without the format modifier (`DRM_FORMAT_MOD_INVALID`). This uses a linear import that is universally supported by EGL implementations.

### Code Changes

- `src/platform/linux/graphics.cpp`: Modified `import_source()` to retry DMA-BUF import without modifiers when the initial attempt fails with `EGL_BAD_MATCH`
- Only the modifier is removed from the retry; all other surface parameters (fourcc, fds, pitches, offsets) are preserved

## Testing

- Arch Linux with Sunshine 2026.531.163415  
- GPU: Intel Arc (i9 Ultra, integrated)
- Display: USB-connected tablet/monitor reporting incompatible modifier
- Before fix: continuous `Couldn't import RGB Image: 00003009` errors, stream shows static
- After fix: DMA-BUF import succeeds with fallback, stream works correctly

### DIFF
--- a/src/platform/linux/graphics.cpp
+++ b/src/platform/linux/graphics.cpp
@@ -594,7 +594,7 @@ namespace egl {
   }
 
   std::optional<rgb_t> import_source(display_t::pointer egl_display, const surface_descriptor_t &xrgb) {
-    auto attribs = surface_descriptor_to_egl_attribs(xrgb);
+    const auto attribs = surface_descriptor_to_egl_attribs(xrgb);
 
     rgb_t rgb {
       egl_display,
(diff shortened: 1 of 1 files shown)


Kishi85:
Could you please update the PR to use the correct template? You can find it at https://github.com/LizardByte/.github/blob/master/.github/pull_request_template.md?plain=1

Lacking a display that has the required issues for this fallback I've just tested it to see if it causes any abnormal behaviour. Overall this seems to be a sensible change to fix compatibility issues as described.

One note: Since this change is calling `eglCreateImage()` a second time for the fallback you should make sure that there's no memory leaking when doing so.

PR #5291, opened 2026-06-14
fix(macos): wake sleeping display at the beginning of capture session

martona:
## Description
Wakes the display at the beginning of the capture session, and does not let it go to sleep while it's active.

macOS does not present any capture frames on a sleeping display, thus the capture thread hangs indefinitely at session start, never sending video data to Moonlight, and therefore is not able to receive any input to wake the display either. The end result is a forever "spinning on connect" Moonlight client, and a hung Sunshine process.

This PR fixes the problem by issuing an `IOPMAssertionDeclareUserActivity` before capture start to handle the initial wake, and holding an `kIOPMAssertPreventUserIdleDisplaySleep` assertion for the lifetime of the capture object to prevent sleep.

I realize it's chonky; I could cut this down to about 30-40% of the proposed size by eliminating the display diagnostic logging that I added to help me track down the issue. I think it's useful to have there though, so I'm leaving it in for the original PR; but very much open to reducing it if needed.

## Type of Change

## Checklist

### AI Usage


ReenigneArcher:
We should probably add display automation logic to https://github.com/LizardByte/libdisplaydevice

https://github.com/LizardByte/roadmap/issues/57

martona:
That makes sense long-term. I think this belongs in libdisplaydevice once macOS support exists there.

But since AVFoundation capture can hang before any frames are delivered if the target display is asleep, Sunshine never reaches the point where client input can wake it. Would you be open to keeping the minimal wake/prevent-sleep logic here for now?

I can also trim the diagnostic logging from this PR to keep it tighter. (Actually already have, would just need to resubmit the PR: it's -122 lines at the cost of some log clarity around display initialization.)

ReenigneArcher:
The size of this PR is not really a problem, it's small enough as is.

Would you be willing to help implement the library for macOS? This is not really work I can do since I don't own any macOS devices.

martona:
Yeah I could. It's a non-trivial amount of work but it makes sense to do. I took a quick look at the code and there's almost certainly blockers in the way to complete feature parity using only public APIs (EnsurePrimary, EnsureOnlyDisplay, HDR...) but most of it - the parts that actually matter: enumeration, wake, mode selection, revert/persistence - are straightforward to implement.

My recommendation would be to keep this PR as the Sunshine-side bugfix if you’re generally OK with the approach. Then I can follow up with a draft/best-effort libdisplaydevice port and a feature matrix for what works vs what macOS doesn’t expose cleanly. Once that happens, we can integrate Sunshine with it and remove the duplicated macOS display handling here.

src/platform/macos/display.mm
```diff
@@ -11,22 +11,238 @@
...
+#include <IOKit/pwr_mgt/IOPMLib.h>
+
 // Avoid conflict between AVFoundation and libavutil both defining AVMediaType
 #define AVMediaType AVMediaType_FFmpeg
 #include "src/video.h"
 #undef AVMediaType
 
+#include <cstdlib>
+#include <sstream>
+#include <thread>
```
ReenigneArcher:
```suggestion
// standard includes
#include <cstdlib>
#include <sstream>
#include <thread>
```

Also, please order the includes like so.

- standard includes
- platform includes
- local includes

ReenigneArcher (commented):
I just have a small nit pic.

src/platform/macos/display.mm
martona: done!

PR #5346, opened 2026-06-27
fix(video): avoid UAF crashes during video reinit (reproducible on Vulkan)

psyke83:
## Description

On Linux with the Vulkan encoder, reinit triggered by display switching or mode changes can crash Sunshine due to packets in flight after encoder teardown.

With extra debug logging, this pattern was occurring:

```
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.625]: Warning: [crash] encode_avcodec session=0x7fd8c89ad480 thread=140569594554048 avctx=0x7fd8c8c7f980 hw_frames=0x7fd8c883a940 FFmpeg frames ctx=0x7fd8c8a6b200
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.628]: Warning: PACKET DESTROY 0x7fd8c8b76d00
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.628]: Warning: [crash] SESSION DESTROY START 0x7fd8c89ad480
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.628]: Warning: PACKET DESTROY 0x7fd8e8bfd490
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.628]: Warning: [crash] Calling avcodec_ctx.reset()
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.628]: Warning: [crash] SESSION DESTROY AFTER RESET 0x7fd8c89ad480
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.628]: Warning: [crash] Calling device.reset()
    Jun 26 23:57:21 archlinux sunshine[263621]: [2026-06-26 23:57:21.629]: Warning: PACKET DESTROY 0x7fd8c882de40
    Jun 26 23:57:21 archlinux kernel: rtsp::handler[263877]: segfault at 7fd8a0af7e20 ip 00007fd8a0af7e20 sp 00007fd8ebffd638 error 14 likely on CPU 11 (core 3, socket 0)
    Jun 26 23:57:21 archlinux kernel: Code: Unable to access opcode bytes at 0x7fd8a0af7df6.
    Jun 26 23:57:21 archlinux systemd-coredump[263973]: Process 263621 (sunshine) of user 1000 terminated abnormally with signal 11/SEGV, processing...
[...]

PR #5360, opened 2026-06-29
fix(linux/pipewire): avoid memory leaks

Kishi85:
## Description

In #5346 a memory leak in pipewire-based capture methods (portalgrab, kwingrab) was identified. 

This PR is a fixes (potential) causes in pipewire-based capture by:
- Changing `dummy_img()` to just return 0 without allocating memory as that is enough to mark a dummy image (same logic as kmsgrab/wlgrab)
- Override the destructor of the used `egl::img_descriptor_t` to free allocated memory for the pipewire specific `alloc_image()` override by using a new subclass (also like kmsgrab/wlgrab).
- Explicitly release any memory resources from `kwingrab` by calling `reset()` on any created shared_ptr instances in screencast_t and `clear()` on used lists/maps in screencast_t during destruction.
- Cleanup `pipewire_t` destruction sequence and roll `cleanup_stream` in as is is never used outside of it.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/kwingrab.cpp
+++ b/src/platform/linux/kwingrab.cpp
@@ -248,7 +248,7 @@ namespace kwin {
    * @brief KWin screencast output name and geometry.
    */
   struct output_parameter_t {
-    std::string name = "";  ///< KWin output name.
+    std::string name;  ///< KWin output name.
     int width = 0;  ///< Output width in pixels.
     int height = 0;  ///< Output height in pixels.
     int pos_x = 0;  ///< Output X position in the compositor layout.
@@ -272,6 +272,7 @@ namespace kwin {
     screencast_t &operator=(screencast_t &&) = delete;  // Do not allow to copying
 
     ~screencast_t() {
+      // Release KDE screencast wayland extensions and reset pointers
       if (kde_screencast_stream_v1_) {
         zkde_screencast_stream_unstable_v1_close(kde_screencast_stream_v1_);
         kde_screencast_stream_v1_ = nullptr;
@@ -280,23 +281,30 @@ namespace kwin {
         zkde_screencast_unstable_v1_destroy(kde_screencast_v1_);
         kde_screencast_v1_ = nullptr;
       }
-
       if (kde_output_order) {
         kde_output_order_v1_destroy(kde_output_order);
         kde_output_order = nullptr;
       }
 
+      // Clear output order list
+      output_order.clear();
+      // Clear current output parameters
+      out_params.reset();
+      out_params = nullptr;
+
       // wl_output is owned by the registry, released on disconnect
-      for (const auto &out : outputs | std::views::keys) {
-        wl_output_destroy(out);
+      // also cleanup associated output parameters and clear output list when done
+      for (auto &[output, params] : outputs) {
+        wl_output_destroy(output);
+        params.reset();
       }
       outputs.clear();
 
+      // Release wayland registry, display and reset pointers
       if (wl_registry) {
         wl_registry_destroy(wl_registry);
         wl_registry = nullptr;
       }
-
       if (wl_display) {
         wl_display_disconnect(wl_display);
         wl_display = nullptr;
--- a/src/platform/linux/pipewire.cpp
+++ b/src/platform/linux/pipewire.cpp
@@ -136,34 +136,44 @@ namespace pipewire {
 
     ~pipewire_t() {
       BOOST_LOG(debug) << "[pipewire] Destroying pipewire_t"sv;
-      try {
-        cleanup_stream();
-      } catch (const std::exception &e) {
-        BOOST_LOG(error) << "[pipewire] Standard exception caught in ~pipewire_t cleanup_stream: "sv << e.what();
-      } catch (...) {
-        BOOST_LOG(error) << "[pipewire] Unknown exception caught in ~pipewire_t cleanup_stream"sv;
-      }
-
       pw_thread_loop_lock(loop);
 
+      // Lock the frame mutex to stop fill_img
+      BOOST_LOG(debug) << "[pipewire] Stop fill_img"sv;
+      {
+        std::scoped_lock lock(stream_data.frame_mutex);
+        stream_data.frame_ready = false;
+        stream_data.current_buffer = nullptr;
+      }
+
+      // Release pipewire stream
+      if (stream_data.stream) {
+        BOOST_LOG(debug) << "[pipewire] Disconnect stream"sv;
+        pw_stream_disconnect(stream_data.stream);
+        BOOST_LOG(debug) << "[pipewire] Destroy stream"sv;
+        pw_stream_destroy(stream_data.stream);
+        stream_data.stream = nullptr;
+      }
+      // Release pipewire core
       if (core) {
         BOOST_LOG(debug) << "[pipewire] Disconnect PW core"sv;
         pw_core_disconnect(core);
         core = nullptr;
       }
+      // Release pipewire context
       if (context) {
         BOOST_LOG(debug) << "[pipewire] Destroy PW context"sv;
         pw_context_destroy(context);
         context = nullptr;
       }
-
-      pw_thread_loop_unlock(loop);
-
+      // Release pipewire file descriptor
       if (fd >= 0) {
         BOOST_LOG(debug) << "[pipewire] Close pipewire_fd"sv;
         close(fd);
       }
+      // Release pipewire thread loop
       BOOST_LOG(debug) << "[pipewire] Stop PW thread loop"sv;
+      pw_thread_loop_unlock(loop);
       pw_thread_loop_stop(loop);
       BOOST_LOG(debug) << "[pipewire] Destroy PW thread loop"sv;
       pw_thread_loop_destroy(loop);
@@ -243,34 +253,6 @@ namespace pipewire {
 
       pw_thread_loop_unlock(loop);
       return 0;
-    }
-
-    /**
-     * @brief Release the active PipeWire stream and listener.
-     */
-    void cleanup_stream() {
-      BOOST_LOG(debug) << "[pipewire] Cleaning up stream"sv;
-      if (loop && stream_data.stream) {
-        pw_thread_loop_lock(loop);
-
-        // 1. Lock the frame mutex to stop fill_img
-        BOOST_LOG(debug) << "[pipewire] Stop fill_img"sv;
-        {
-          std::scoped_lock lock(stream_data.frame_mutex);
-          stream_data.frame_ready = false;
-          stream_data.current_buffer = nullptr;
-        }
-
-        if (stream_data.stream) {
-          BOOST_LOG(debug) << "[pipewire] Disconnect stream"sv;
-          pw_stream_disconnect(stream_data.stream);
-          BOOST_LOG(debug) << "[pipewire] Destroy stream"sv;
-          pw_stream_destroy(stream_data.stream);
-          stream_data.stream = nullptr;
-        }
-
-        pw_thread_loop_unlock(loop);
-      }
     }
 
     /**
--- a/src/platform/linux/pipewire.cpp
+++ b/src/platform/linux/pipewire.cpp
@@ -124,6 +124,18 @@ namespace pipewire {
   };
 
   /**
+   * @brief Pipewire image assembled for encoding.
+   */
+  struct img_descriptor_t: public egl::img_descriptor_t {
+    ~img_descriptor_t() override {
+      if (data) {
+        delete[] data;
+        data = nullptr;
+      }
+    }
+  };
+
+  /**
    * @brief PipeWire core, context, and stream setup used for screencast capture.
    */
   class pipewire_t {
@@ -916,7 +928,7 @@ namespace pipewire {
      */
     std::shared_ptr<platf::img_t> alloc_img() override {
       // Note: this img_t type is also used for memory buffers
-      auto img = std::make_shared<egl::img_descriptor_t>();
+      auto img = std::make_shared<img_descriptor_t>();
 
       img->width = width;
       img->height = height;
--- a/src/platform/linux/pipewire.cpp
+++ b/src/platform/linux/pipewire.cpp
@@ -1049,12 +1049,7 @@ namespace pipewire {
      * @return Capture status reported to the streaming pipeline.
      */
     int dummy_img(platf::img_t *img) override {
-      if (!img) {
-        return -1;
-      }
-
-      img->data = new std::uint8_t[img->height * img->row_pitch];
-      std::fill_n(img->data, img->height * img->row_pitch, 0);
+      // Empty images are recognized as dummies by the zero sequence number
       return 0;
     }
 


Kishi85:
In theory this PR shouldn't make a difference as C++ should manage memory for those types (std::shared_ptr, std::map, std::vector) by itself correctly for how they are used in kwingrab. 

@psyke83 maybe you can check if it makes a difference on your end as my tests are a bit inconclusive right now...

Kishi85:
Doing more testing and also cleaning up pipewire's destructor a bit but so far no luck with the memory leak. It's almost a stable ~16MB increase per full capture re-init for both kwingrab and portalgrab so the culprit is likely somewhere in `pipewire.cpp` or related to the pipewire code used overall. 

Almost forgot: If I can't find the source of the leak myself I'll change this PR to do some code maintenance (like rolling 'cleanup_stream()` into the destructor as it's not used anywhere else anymore und updating a few unnecessary things based on CLion/Sonar suggestions).

Kishi85:
> Doing more testing and also cleaning up pipewire's destructor a bit but so far no luck with the memory leak. It's almost a stable ~16MB increase per full capture re-init for both kwingrab and portalgrab so the culprit is likely somewhere in `pipewire.cpp` or related to the pipewire code used overall.

Here's another interesting finding: The leak is only ~8MB per full capture re-init for both kwingrab and portalgrab when using VAAPI compared to ~16MB when using Vulkan encoding.

psyke83:
This is part of the puzzle:

```
diff --git a/src/platform/linux/graphics.h b/src/platform/linux/graphics.h
index 2a2a0642..60be723f 100644
--- a/src/platform/linux/graphics.h
+++ b/src/platform/linux/graphics.h
@@ -603,7 +603,11 @@ namespace egl {
   class img_descriptor_t: public cursor_t {
   public:
     ~img_descriptor_t() {
+      BOOST_LOG(error) << "[leak] reset img_descriptor_t";
       reset();
+
+      delete[] data;
+      data = nullptr;
     }
 
     /**
```

With your PR branch applied, this resolves leaking with vaapi combined with kwin or portal capture. Vulkan still leaks, and software rendering breaks with the change as-is (haven't looked too deeply yet).

Some capture methods employ an img_t override that does extra cleanup. Example: https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/wlgrab.cpp#L25-L33 or https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/kmsgrab.cpp#L807-L812

In the case of Pipewire capture, [alloc_img()](https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/pipewire.cpp#L935-L949) is overridden to use the img_descriptor_t type, but we don't do any special cleanup via a destructor override, and the [standard destructor](https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/graphics.h#L603-L620) in graphics.h only cleans up the sd fds.

Kishi85:
> This is part of the puzzle:
> 
> ```
> diff --git a/src/platform/linux/graphics.h b/src/platform/linux/graphics.h
> index 2a2a0642..60be723f 100644
> --- a/src/platform/linux/graphics.h
> +++ b/src/platform/linux/graphics.h
> @@ -603,7 +603,11 @@ namespace egl {
>    class img_descriptor_t: public cursor_t {
>    public:
>      ~img_descriptor_t() {
> +      BOOST_LOG(error) << "[leak] reset img_descriptor_t";
>        reset();
> +
> +      delete[] data;
> +      data = nullptr;
>      }
>  
>      /**
> ```
> 
> With your PR branch applied, this resolves leaking with vaapi combined with kwin or portal capture. Vulkan still leaks, and software rendering breaks with the change as-is (haven't looked too deeply yet).
> 
> Some capture methods employ an img_t override that does extra cleanup. Example:
> 
> https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/wlgrab.cpp#L25-L33
> 
> or
> https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/kmsgrab.cpp#L807-L812
> 
> In the case of Pipewire capture, [alloc_img()](https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/pipewire.cpp#L935-L949) is overridden to use the img_descriptor_t type, but we don't do any special cleanup via a destructor override, and the [standard destructor](https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/graphics.h#L603-L620) in graphics.h only cleans up the sd fds.

So we probably have to just add a subclassed img_descriptor_t with the extra cleanup for pipewire and use that in the override?

Also is there a specific reason for pipewire using `egl::img_descriptor_t` instead of `platf::img_t` (like kms/wlgrab)?

psyke83:
I'm not sure why that was done, but more investigation actually points this specific leak issue to be isolated to dummy_img: https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/pipewire.cpp#L1069-L1077

Reverting my previous changes and instead short-circuiting dummy_img (inserting `return 0;` at the start or just deleting the function contents) also resolves the memory leak with the vaapi encoder, and software encoding no longer crashes (and also doesn't leak). Only Vulkan leaks as-is, which I assume is a separate unrelated leak.

Edit: I won't be able to test thoroughly until later, but if it turns out that the destructor really does need to do cleanup, being able to distinguish the ownership between dummy_img()'s vs other img_descriptor_t allocations would probably resolve the double-free crashing. Something like the below snippet. But perhaps there's a more appropriate fix.
```
diff --git a/src/platform/linux/graphics.h b/src/platform/linux/graphics.h
index 2a2a0642..997ee0f2 100644
--- a/src/platform/linux/graphics.h
+++ b/src/platform/linux/graphics.h
@@ -617,6 +617,11 @@ namespace egl {
           sd.fds[x] = -1;
         }
       }
+      if (data && data_allocated) {
+        delete[] data;
+      }
+      data = nullptr;
+      data_allocated = false;
     }
 
     surface_descriptor_t sd;  ///< DMA-BUF surface descriptor for the captured image.
@@ -632,6 +637,8 @@ namespace egl {
     std::optional<uint64_t> seq;  ///< PipeWire frame sequence number.
     std::optional<bool> pw_damage;  ///< Whether PipeWire damage tracking should be used.
     std::optional<uint32_t> pw_flags;  ///< PipeWire frame flags reported with the buffer.
+
+    bool data_allocated {false};
   };
 
   /**
diff --git a/src/platform/linux/pipewire.cpp b/src/platform/linux/pipewire.cpp
index fe71ff3b..7f53d6a4 100644
--- a/src/platform/linux/pipewire.cpp
+++ b/src/platform/linux/pipewire.cpp
@@ -1055,6 +1055,7 @@ namespace pipewire {
 
       img->data = new std::uint8_t[img->height * img->row_pitch];
       std::fill_n(img->data, img->height * img->row_pitch, 0);
+      static_cast<egl::img_descriptor_t*>(img)->data_allocated = true;
       return 0;
     }

```

Kishi85:
> I'm not sure why that was done, but more investigation actually points this specific leak issue to be isolated to dummy_img:
> 
> https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/pipewire.cpp#L1069-L1077
> 
> Reverting my previous changes and instead short-circuiting dummy_img (inserting `return 0;` at the start or just deleting the function contents) also resolves the memory leak with the vaapi encoder, and software encoding no longer crashes (and also doesn't leak). Only Vulkan leaks as-is, which I assume is a separate unrelated leak.
> 
> Edit: I won't be able to test thoroughly until later, but if it turns out that the destructor really does need to do cleanup, being able to distinguish the ownership between dummy_img()'s vs other img_descriptor_t allocations would probably resolve the double-free crashing. Something like the below snippet. But perhaps there's a more appropriate fix.
> 
> ```
> diff --git a/src/platform/linux/graphics.h b/src/platform/linux/graphics.h
> index 2a2a0642..997ee0f2 100644
> --- a/src/platform/linux/graphics.h
> +++ b/src/platform/linux/graphics.h
> @@ -617,6 +617,11 @@ namespace egl {
>            sd.fds[x] = -1;
>          }
>        }
> +      if (data && data_allocated) {
> +        delete[] data;
> +      }
> +      data = nullptr;
> +      data_allocated = false;
>      }
>  
>      surface_descriptor_t sd;  ///< DMA-BUF surface descriptor for the captured image.
> @@ -632,6 +637,8 @@ namespace egl {
>      std::optional<uint64_t> seq;  ///< PipeWire frame sequence number.
>      std::optional<bool> pw_damage;  ///< Whether PipeWire damage tracking should be used.
>      std::optional<uint32_t> pw_flags;  ///< PipeWire frame flags reported with the buffer.
> +
> +    bool data_allocated {false};
>    };
>  
>    /**
> diff --git a/src/platform/linux/pipewire.cpp b/src/platform/linux/pipewire.cpp
> index fe71ff3b..7f53d6a4 100644
> --- a/src/platform/linux/pipewire.cpp
> +++ b/src/platform/linux/pipewire.cpp
> @@ -1055,6 +1055,7 @@ namespace pipewire {
>  
>        img->data = new std::uint8_t[img->height * img->row_pitch];
>        std::fill_n(img->data, img->height * img->row_pitch, 0);
> +      static_cast<egl::img_descriptor_t*>(img)->data_allocated = true;
>        return 0;
>      }
> ```

Both kmsgrab and wlgrab implement `dummy_img()` by just returning 0:
https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/kmsgrab.cpp#L1685-L1692
https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/kmsgrab.cpp#L1762-L1771
https://github.com/LizardByte/Sunshine/blob/3437a0b12e7691da1fccb1594151c40ae8e43e20/src/platform/linux/wlgrab.cpp#L141-L149

So I'd say that pipewire.cpp should just do the same thing to keep things similar between linux capture methods. There seems to be no functional disadvantage from doing it this way.

Kishi85:
@psyke83 I think you've found the main issues here. To keep things simple and in line with kmsgrab/wlgrab I've just added a subclass for those specific img_descriptor_t's. Also updated `dummy_img()` to what kmsgrab/wlgrab do and with that I'm not seeing any obvious memory leaks anymore using vaapi. 

There might still be a small leak occurring but that would be somewhere affecting all capture methods including kmsgrab/wlgrab and a separate issue. 

I also can confirm the memory leak in vulkan encoding as that is still happening albeit at half the previous increase (which was somewhat to be expected due to given behaviour). Update: using KMS also has the same 8MB per capture re-init memory increase using vulkan, so definitely a separate issue that is vulkan specific.

I'll update the PR and mark this as ready unless you have something else to add.

psyke83:
I don't think the descriptor cleanup is necessary/sensible when dummy_img() becomes a no-op?

```
diff --git a/src/platform/linux/pipewire.cpp b/src/platform/linux/pipewire.cpp
index 1048b839..e5c133d4 100644
--- a/src/platform/linux/pipewire.cpp
+++ b/src/platform/linux/pipewire.cpp
@@ -129,6 +129,7 @@ namespace pipewire {
   struct img_descriptor_t: public egl::img_descriptor_t {
     ~img_descriptor_t() override {
       if (data) {
+        BOOST_LOG(warning) << "[leak]: deleting data";
         delete[] data;
         data = nullptr;
       }
```

This log is not appearing with vaapi or software encoding on any of the pipewire capture methods, and doesn't seem to affect memory usage (naturally). Also, I should've clarified "no leak" to include the caveat that I'm ignoring the first reinit cycle. There usually seems to be an increase after first reinit, but the usage stays consistent on subsequent reinit cycles.

Kishi85:
> I don't think the descriptor cleanup is necessary/sensible when dummy_img() becomes a no-op?
> 
> ```
> diff --git a/src/platform/linux/pipewire.cpp b/src/platform/linux/pipewire.cpp
> index 1048b839..e5c133d4 100644
> --- a/src/platform/linux/pipewire.cpp
> +++ b/src/platform/linux/pipewire.cpp
> @@ -129,6 +129,7 @@ namespace pipewire {
>    struct img_descriptor_t: public egl::img_descriptor_t {
>      ~img_descriptor_t() override {
>        if (data) {
> +        BOOST_LOG(warning) << "[leak]: deleting data";
>          delete[] data;
>          data = nullptr;
>        }
> ```
> 
> This log is not appearing with vaapi or software encoding on any of the pipewire capture methods, and doesn't seem to affect memory usage (naturally). Also, I should've clarified "no leak" to include the caveat that I'm ignoring the first reinit cycle. There usually seems to be an increase after first reinit, but the usage stays consistent on subsequent reinit cycles.

I've not tried that logging but looking at the debugger it's always `NULL` for the cases I've seen so far, so it's probably not strictly necessary. I'd still leave it in place for consistency (ensuring that the img_descriptor data is always properly cleared) with other capture methods (namely kmsgrab/wlgrab).

psyke83:
I believe that the Vulkan issue should be kept separate from this PR, as there's a leak in the n8.1.2 build from Arch which should roughly match our current build revision?

Reproduction of leak from command-line invocation (needs `vulkan-validation-layers` installed):
```
$ VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation \
ffmpeg -y \
  -init_hw_device vulkan=vk \
  -filter_hw_device vk \
  -f lavfi -i testsrc=size=1280x720:rate=30 \
  -frames:v 1 \
  -vf format=nv12,hwupload \
  -c:v h264_vulkan \
  out.h264 2> validation_repro.log
Validation Error: [ VUID-VkImageMemoryBarrier2-srcAccessMask-03915 ] | MessageID = 0xb927d7cf
vkCmdPipelineBarrier2(): pDependencyInfo->pImageMemoryBarriers[0].srcAccessMask (VK_ACCESS_2_TRANSFER_WRITE_BIT) is not supported by stage mask (VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT). The expansion of VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT for VK_QUEUE_VIDEO_ENCODE_BIT_KHR does not include any stages that support VK_ACCESS_2_TRANSFER_WRITE_BIT.

The expanded stages are: VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT|VK_PIPELINE_STAGE_2_BOTTOM_OF_PIPE_BIT|VK_PIPELINE_STAGE_2_VIDEO_ENCODE_BIT_KHR.
The Vulkan spec states: If srcAccessMask includes VK_ACCESS_2_TRANSFER_WRITE_BIT, srcStageMask must include VK_PIPELINE_STAGE_2_COPY_BIT, VK_PIPELINE_STAGE_2_BLIT_BIT, VK_PIPELINE_STAGE_2_RESOLVE_BIT, VK_PIPELINE_STAGE_2_CLEAR_BIT, VK_PIPELINE_STAGE_2_ALL_TRANSFER_BIT, VK_PIPELINE_STAGE_2_ACCELERATION_STRUCTURE_BUILD_BIT_KHR, VK_PIPELINE_STAGE_2_ACCELERATION_STRUCTURE_COPY_BIT_KHR, VK_PIPELINE_STAGE_2_CONVERT_COOPERATIVE_VECTOR_MATRIX_BIT_NV, or VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-VkImageMemoryBarrier2-srcAccessMask-03915)
Objects: 1
    [0] VkCommandBuffer 0x7f319c6a0960

Validation Error: [ VUID-vkDestroyDevice-device-05137 ] | MessageID = 0x4872eaa0
vkDestroyDevice(): VkDevice 0x558956ba7ed0 has 1 leaked objects that have not been destroyed.
VkImageView 0x590000000059
The Vulkan spec states: All child objects created on device that can be destroyed or freed must have been destroyed or freed prior to destroying device (https://docs.vulkan.org/spec/latest/chapters/devsandqueues.html#VUID-vkDestroyDevice-device-05137)
```

I added some logging of VkImage and VkImageView creation and destruction in Sunshine and appended this to my systemd daemon config:

```
$ cat ~/.config/systemd/user/app-dev.lizardbyte.app.Sunshine.service.d/override.conf 
[Service]
Environment=VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation

diff --git a/src/platform/linux/vulkan_encode.cpp b/src/platform/linux/vulkan_encode.cpp
index fef11ad4..da63eafd 100644
--- a/src/platform/linux/vulkan_encode.cpp
+++ b/src/platform/linux/vulkan_encode.cpp
@@ -601,6 +601,8 @@ namespace vk {
                          << ", pitch=" << sd.pitches[0] << ", offset=" << sd.offsets[0] << ")";
         return false;
       }
+      BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                         << " | Action: create src.image= " << src.image;
 
       // Bind imported DMA-BUF memory
       VkMemoryRequirements mem_req;
@@ -623,6 +625,8 @@ namespace vk {
       if (res != VK_SUCCESS) {
         BOOST_LOG(error) << "vkAllocateMemory for DMA-BUF failed: " << res;
         vkDestroyImage(vk_dev.dev, src.image, nullptr);
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: destroy src.image= " << src.image;
         src.image = VK_NULL_HANDLE;
         return false;
       }
@@ -637,6 +641,8 @@ namespace vk {
       view_ci.components = vk_swizzle;
       view_ci.subresourceRange = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
       VK_CHECK_BOOL(vkCreateImageView(vk_dev.dev, &view_ci, nullptr, &src.view));
+      BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                         << " | Action: create src.view= " << src.view;
 
       src.mem = src_mem;
       return true;
@@ -656,6 +662,8 @@ namespace vk {
       img_ci.usage = VK_IMAGE_USAGE_SAMPLED_BIT;
       img_ci.initialLayout = VK_IMAGE_LAYOUT_PREINITIALIZED;
       VK_CHECK_BOOL(vkCreateImage(vk_dev.dev, &img_ci, nullptr, &cursor.image));
+      BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                         << " | Action: create cursor.image= " << cursor.image;
 
       VkMemoryRequirements mem_req;
       vkGetImageMemoryRequirements(vk_dev.dev, cursor.image, &mem_req);
@@ -683,6 +691,8 @@ namespace vk {
       view_ci.format = VK_FORMAT_B8G8R8A8_UNORM;
       view_ci.subresourceRange = {VK_IMAGE_ASPECT_COLOR_BIT, 0, 1, 0, 1};
       VK_CHECK_BOOL(vkCreateImageView(vk_dev.dev, &view_ci, nullptr, &cursor.view));
+      BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                         << " | Action: create cursor.view= " << cursor.view;
 
       cursor.needs_transition = true;
       descriptors_dirty = true;
@@ -691,11 +701,15 @@ namespace vk {
 
     void destroy_cursor_image() {
       if (cursor.view) {
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: destroy cursor.view= " << cursor.view;
         vkDestroyImageView(vk_dev.dev, cursor.view, nullptr);
         cursor.view = VK_NULL_HANDLE;
       }
       if (cursor.image) {
         vkDestroyImage(vk_dev.dev, cursor.image, nullptr);
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: destroy cursor.image= " << cursor.image;
         cursor.image = VK_NULL_HANDLE;
       }
       if (cursor.mem) {
@@ -729,11 +743,15 @@ namespace vk {
         view_ci.format = y_fmt;
         view_ci.subresourceRange = {VK_IMAGE_ASPECT_PLANE_0_BIT, 0, 1, 0, 1};
         VK_CHECK_BOOL(vkCreateImageView(vk_dev.dev, &view_ci, nullptr, &target.y_view));
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: create target.y_view= " << target.y_view;
 
         // UV plane
         view_ci.format = uv_fmt;
         view_ci.subresourceRange = {VK_IMAGE_ASPECT_PLANE_1_BIT, 0, 1, 0, 1};
         VK_CHECK_BOOL(vkCreateImageView(vk_dev.dev, &view_ci, nullptr, &target.uv_view));
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: create target.uv_view= " << target.uv_view;
       } else {
         // Separate images per plane
         VkImageViewCreateInfo view_ci = {VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO};
@@ -743,10 +761,14 @@ namespace vk {
         view_ci.image = vk_frame->img[0];
         view_ci.format = y_fmt;
         VK_CHECK_BOOL(vkCreateImageView(vk_dev.dev, &view_ci, nullptr, &target.y_view));
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: create target.y_view= " << target.y_view;
 
         view_ci.image = vk_frame->img[1];
         view_ci.format = uv_fmt;
         VK_CHECK_BOOL(vkCreateImageView(vk_dev.dev, &view_ci, nullptr, &target.uv_view));
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: create target.uv_view= " << target.uv_view;
       }
       return true;
     }
@@ -916,10 +938,16 @@ namespace vk {
         // By the time we wrap around (4 frames later), it's guaranteed done.
         auto &slot = defer_ring[defer_idx];
         if (slot.view) {
+          BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                             << " | Action: destroy (defer_idx slot " << defer_idx
+                             << ") slot.view= " << slot.view;
           vkDestroyImageView(vk_dev.dev, slot.view, nullptr);
         }
         if (slot.image) {
           vkDestroyImage(vk_dev.dev, slot.image, nullptr);
+          BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                             << " | Action: destroy (defer_idx slot " << defer_idx
+                             << ") slot.image= " << slot.image;
         }
         if (slot.mem) {
           vkFreeMemory(vk_dev.dev, slot.mem, nullptr);
@@ -934,25 +962,44 @@ namespace vk {
       if (!vk_dev.dev) {
         return;
       }
+
+    BOOST_LOG(warning) << "[leak-audit] Checking for orphans...";
+    BOOST_LOG(warning) << "[leak-audit] src.view: " << src.view;
+    BOOST_LOG(warning) << "[leak-audit] cursor.view: " << cursor.view;
+    BOOST_LOG(warning) << "[leak-audit] target.y_view: " << target.y_view;
+    BOOST_LOG(warning) << "[leak-audit] target.uv_view: " << target.uv_view;
+
       vkDeviceWaitIdle(vk_dev.dev);
       destroy_src_image();
       // Flush deferred destroys
+      int iter_idx = 0;
       for (auto &slot : defer_ring) {
         if (slot.view) {
+          BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                             << " | Action: destroy (iter_idx slot " << iter_idx
+                             << ") slot.view= " << slot.view;
           vkDestroyImageView(vk_dev.dev, slot.view, nullptr);
         }
         if (slot.image) {
           vkDestroyImage(vk_dev.dev, slot.image, nullptr);
+          BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                             << " | Action: destroy (iter_idx slot " << iter_idx
+                             << ") slot.image= " << slot.image;
         }
         if (slot.mem) {
           vkFreeMemory(vk_dev.dev, slot.mem, nullptr);
         }
         slot = {};
+        ++iter_idx;
       }
       if (target.y_view) {
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: destroy target.y_view= " << target.y_view;
         vkDestroyImageView(vk_dev.dev, target.y_view, nullptr);
       }
       if (target.uv_view) {
+        BOOST_LOG(warning) << "[leak] Device: " << vk_dev.dev
+                           << " | Action: destroy target.uv_view= " << target.uv_view;
         vkDestroyImageView(vk_dev.dev, target.uv_view, nullptr);
       }
       destroy_cursor_image();
```

My results show 1 leak on every encode probe cycle via video.cpp's `validate_config()` of a VkImageView address that doesn't correspond to any of any of the logged Sunshine calls. Subsequent connect/reconnect/disconnect sometimes leaks up to 3 VkImageView items in one go, but again, none correspond to what Sunshine logs. 

Example (with validate_config() bypass to only test Vulkan during real connect):
```
[conn@archlinux sunshine]$ journalctl -b0 | grep 275148 | grep -Ei "leaked|0x5f500000005f5|0x5f600000005f6|0x5f700000005f7|leak-audit"
Jul 03 02:08:37 archlinux sunshine[275148]: [2026-07-03 02:08:37.423]: Warning: [leak-audit] Checking for orphans...
Jul 03 02:08:37 archlinux sunshine[275148]: [2026-07-03 02:08:37.423]: Warning: [leak-audit] src.view: 0x5eb00000005eb
Jul 03 02:08:37 archlinux sunshine[275148]: [2026-07-03 02:08:37.423]: Warning: [leak-audit] cursor.view: 0x650000000065
Jul 03 02:08:37 archlinux sunshine[275148]: [2026-07-03 02:08:37.423]: Warning: [leak-audit] target.y_view: 0x6a000000006a
Jul 03 02:08:37 archlinux sunshine[275148]: [2026-07-03 02:08:37.423]: Warning: [leak-audit] target.uv_view: 0x6b000000006b
Jul 03 02:08:37 archlinux sunshine[275148]: vkDestroyDevice(): VkDevice 0x7f952420a850 has 3 leaked objects that have not been destroyed.
Jul 03 02:08:37 archlinux sunshine[275148]: VkImageView 0x5f500000005f5, VkImageView 0x5f600000005f6, VkImageView 0x5f700000005f7
```

Hence the leak is more than likely on the ffmpeg side.

Kishi85:
@psyke83 With that analysis I agree that the vulkan memory leak is very likely on the ffmpeg side, especially since you can reproduce it with just using ffmpeg itself. Would you mind putting your analysis into a separate issue so we can keep track of it?
EDIT: Could be related to https://code.ffmpeg.org/FFmpeg/FFmpeg/issues/21675

@ReenigneArcher the fixes for the pipewire memory leak in this PR should be good to merge

ReenigneArcher:
@Kishi85 I didn't read all of this conversation, but is it possible to address https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=5360&id=LizardByte_Sunshine&open=AZ8iXDHnzFzYS_Ua0It3&tab=code

Kishi85:
> @Kishi85 I didn't read all of this conversation, but is it possible to address https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=5360&id=LizardByte_Sunshine&open=AZ8iXDHnzFzYS_Ua0It3&tab=code

Unfortunately not possible without touching **all** other capture methods (as that pointer is from `platf::img_t`). Even then its likley a change someone with a deep understanding of the whole process should make.

To be clear: That part isn't strictly necessary for pipewire-based capture (at least so far) but I'd keep that additional failsafe there for consistency with other linux capture methods. This part is a 1:1 adaption from what kmsgrab and wlgrab are both already doing for their implementations of `platf::img_t`. Same reasoning for the `dummy_img()`-change, which is now done the same way for pipewire-based capture and at least kmsgrab/wlgrab (haven't checked non-linux capture methods).

ReenigneArcher:
Hate to ask this, but can you force push again? There's a bug in GitHub where sometimes it doesn't allow the arch and flatpak builds to checkout the PR.

I usually just make a small change, amend the last commit. Then undo that change, and re-amend the last commit. Then force push. Not sure if there's an easier way.

Kishi85:
> Hate to ask this, but can you force push again? There's a bug in GitHub where sometimes it doesn't allow the arch and flatpak builds to checkout the PR.
> 
> I usually just make a small change, amend the last commit. Then undo that change, and re-amend the last commit. Then force push. Not sure if there's an easier way.

No worries, let's see if rebasing and changing the commit order is enough to fix this.

PR #5377, opened 2026-07-05
build(Linux): Allow building with Vulkan but without Wayland

chewi:
## Description
The build fails otherwise due to a missing include. ~~The generated-src idea could potentially be used for any platform, so include it in the common file.~~

On a side note, the build also fails with PipeWire but without Wayland due to Wayland-specific code in pipeline.cpp. The EGL bits could be adapted, but the monitor handling is more involved, so I figured this wasn't worth the effort.

### Issues Fixed or Closed

- Closes [Gentoo #977035](https://bugs.gentoo.org/977035).

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/cmake/compile_definitions/linux.cmake
+++ b/cmake/compile_definitions/linux.cmake
@@ -220,7 +220,6 @@ if(WAYLAND_FOUND)
     include_directories(
             SYSTEM
             ${WAYLAND_INCLUDE_DIRS}
-            ${CMAKE_BINARY_DIR}/generated-src
     )
 
     list(APPEND PLATFORM_LIBRARIES ${WAYLAND_LIBRARIES} gbm)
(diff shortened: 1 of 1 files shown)


cmake/compile_definitions/common.cmake
```diff
@@ -136,6 +135,7 @@ include_directories(BEFORE "${CMAKE_SOURCE_DIR}")
 include_directories(
         BEFORE
         SYSTEM
+        "${CMAKE_BINARY_DIR}/generated-src"
```
ReenigneArcher: This directory only exists on Linux, so should not be in the common.cmake file.

I think you can keep in the linux.cmake file, but moved out of the wayland section?

chewi: I know it does now, but I figured it might get used on other platforms later and an extra -I doesn't do any harm. I'll change it anyway.

ReenigneArcher (approved):

PR #5388, opened 2026-07-07
feat(linux/vaapi): VAAPI encoder improvements

psyke83:
## Description
* Split rate control selection logic out from VBV bitrate enforcement to make it possible for user-overrides.
* Add rate control mode selection. If set to a supported mode, the built-in whitelist is disabled. If set to Automatic or the user selects a mode that is not supported by their driver, the preexisting logic that was used still applies: if whitelisted, try VBR -> CBR -> CQP and force single-frame VBV, otherwise just try CBR -> CQP.
* Add encoder quality profile selection. As VAAPI doesn't mandate specific quality levels, we must probe the maximum level during initialization. If VAConfigAttribEncQualityRange returns a value, the maximum value will represent the fastest/lowest quality and value 1 sets highest quality. This value is then mapped onto the compression_level AVOption which sets the VAEncMiscParameterTypeQualityLevel via ffmpeg's vaapi_encode.c.
* Add Block Level Bitrate Control (BLBRC) option that can be combined with VBR or CBR rate control on supported encoders.

### Screenshot

Screenshot_20260708_123302

## Type of Change

## Checklist

### AI Usage


Kishi85:
I'm getting the following error during encoder probing (causing a fallback to vulkan) when enabling BLBRC while using [fe4d3b2](https://github.com/LizardByte/Sunshine/pull/5388/commits/fe4d3b25e96ac39e13efd350e996a18e0fbfddf9):
```
Jul 07 17:38:28 desktop sunshine[22467]: [2026-07-07 17:38:28.180]: Error: [h264_vaapi @ 0x55e8cfc00780] Unable to parse "blbrc" option value "59746582" as boolean
Jul 07 17:38:28 desktop sunshine[22467]: [2026-07-07 17:38:28.180]: Error: [h264_vaapi @ 0x55e8cfc00780] Error setting option blbrc to value 59746582.
```

psyke83:
@Kishi85 

The latest rebase should include the necessary fix. I can't test BLBRC on my AMD card, so let me know if it causes any other issues when enabled.

src/platform/linux/vaapi.cpp
```diff
@@ -337,26 +361,30 @@ namespace va {
...
-        } else if (rc_attr.value & VA_RC_CBR) {
-          BOOST_LOG(info) << "Using CBR with single frame VBV size"sv;
-          av_dict_set(options, "rc_mode", "CBR", 0);
-        } else {
-          BOOST_LOG(warning) << "Using CQP with single frame VBV size"sv;
-          av_dict_set_int(options, "qp", config::video.qp, 0);
-        }
-      } else if (!(rc_attr.value & (VA_RC_CBR | VA_RC_VBR))) {
-        BOOST_LOG(warning) << "Using CQP rate control"sv;
+      if (rc_attr.value & VA_RC_VBR && config::video.vaapi.allow_vbr) {
```
cgutman: Making VBR a manual opt-in is going to regress those existing known good cases enumerated in the comment above. CBR with VAAPI forces filler data which is a massive waste of bandwidth and needlessly increases the chance of frame loss over the network.

We should probably keep a default Auto option for rate control that uses VBR in those known good scenarios and allows the user to opt-in or opt-out manually if they want.

psyke83: Would restoring the whitelist behaviour via [445e13e](https://github.com/LizardByte/Sunshine/pull/5388/commits/445e13ed066c75ef163e40a1819c45ce47401c79) resolve this to your satisfaction? I was trying to avoid adding a dedicated RC dropdown menu, as CBR and VBR are the only realistic options that users might want. The only other RC available on my card is QVBR, and it does a very poor job of constraining the target bitrate when the QP is set too low. The problem with VAAPI is that the performance characteristics might differ greatly across vendors and card generations.

Kishi85: IMHO a whitelist is in-transparent for the user as when configuring CBR it'll still use VBR for the known good cases (Intel, AV1) which also means that a user cannot force CBR for those because the whitelist always overrides that decision.

You could fix that by either replacing "Allow VBR rate control" with a "Rate control mode" dropdown containing a default auto option (with previous whitelist behaviour) and then all applicable RC modes (CBR/VBR and maybe QCP?) to specifically force those modes (this would make that configuration option equal to the one from the vulkan encoder).
Another option would be to add a second boolean named something like "Override default rate control" that controls when "Allow VBR rate control" is applicable but that seems like a less flexible and elegant solution to me.

Kishi85:
Been running this for a few hours yesterday with all options (VBR, BLBRC, STRICT) flipped on and quality set to high. 

Subjectively I'm impressed with the quality I'm getting especially on red-heavy scenes with movement it's a lot less washed out compared to Vulkan encoding. Unsure where that comes from. Also host processing latency is just about .2 ms higher than current Vulkan default options which is negliable for me (1.9ms->2.1ms average).

psyke83:
I added a full drop-down menu for rate control and included all possible modes.

The default RC behaviour should now be exactly the same as before:
* Automatic (default) -> standard probe that includes RC / VBV whitelist enforcement.
* User selects unsupported RC mode -> falls back to same whitelist-compliant probe.
* User selects supported RC mode -> whitelist is overridden for VBV, allowing users to disable strict VBV.

Unfortunately this added a lot of code, most of which is borrowed from the Windows AMF encoder logic and UI section, but I suppose I do prefer the UI category layout and named options now.

ReenigneArcher:
Can you fix the doxygen errors?

```
Generating code/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:415: error: Documentation for enum member 'config::vaapi::quality_e::_auto' is missing.
--
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:416: error: Documentation for enum member 'config::vaapi::quality_e::speed' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:417: error: Documentation for enum member 'config::vaapi::quality_e::balanced' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:419: error: Documentation for enum member 'config::vaapi::quality_e::quality' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:425: error: Documentation for enum member 'config::vaapi::rc_e::_auto' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:426: error: Documentation for enum member 'config::vaapi::rc_e::avbr' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:427: error: Documentation for enum member 'config::vaapi::rc_e::cbr' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:428: error: Documentation for enum member 'config::vaapi::rc_e::cqp' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:429: error: Documentation for enum member 'config::vaapi::rc_e::icq' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:430: error: Documentation for enum member 'config::vaapi::rc_e::qvbr' is missing.
/home/docs/checkouts/readthedocs.org/user_builds/sunshinestream/checkouts/5388/src/config.cpp:432: error: Documentation for enum member 'config::vaapi::rc_e::vbr' is missing.
```

src/platform/linux/vaapi.cpp
```diff
@@ -337,27 +362,50 @@ namespace va {
...
+      // If a user-supplied rate control is detected, override the whitelist logic to allow
+      // full user control of both the rate control and strict VBV settings.
+      auto auto_whitelist = false;
+      auto rc_msg = "with standard VBV size";
+      auto rc_mode = config::video.vaapi.vaapi_rc_str;
+      auto rc_val = config::video.vaapi.vaapi_rc.value_or(0);
+
+      // Detect whitelisted configurations
+      if ((vendor && std::string_view(vendor).contains("Intel") == true) || ctx->codec_id == AV_CODEC_ID_AV1) {
+        BOOST_LOG(warning) << "[VAAPI] Rate control and VBV size may be overridden by built-in whitelist on this device";
```
Kishi85: This warning can be misleading when the users forces a config (`rc_val > 0`) as the next block sets `auto_whitelist = false` in that case.

psyke83: I suppose the "may" is not emphatic enough. I'll change it so that it only warns when the whitelist is active.

Kishi85:
With those improvements to VAAPI and the current problems/caveats when using Vulkan (Hardware/driver specific issues, requiring a RADV_EXPERIMENTAL flag, at least one known memory leak in ffmpeg) I'm wondering if we should prefer `vaapi` again for now.

Just something to consider (for another PR). I've been using both but VAAPI seems to give me subjectively better image quality with the new options also seems to be the more mature option of the two.

psyke83:
@Kishi85 Vulkan is probably a better choice for people using Portal capture and/or sandboxed versions of Sunshine, especially with midrange or lower GPUs. When the EGL context priority can't be set properly, VAAPI will drop frames during high GPU load.

Kishi85:
Thats a good point I didn't know/hadn't considered yet.

ReenigneArcher:
1 Test failing, just an ordering issue.

```
[ RUN      ] ConfigConsistencyTest.ConfigOptionsInSameOrderWithinSections

Test failed: ConfigOptionsInSameOrderWithinSections

Captured cout:

Captured cerr:

Captured stdout:
/__w/Sunshine/Sunshine/pkg/src/sunshine/tests/integration/test_config_consistency.cpp:567: Failure
Failed
Config option order inconsistencies:
  Section 'VA-API Encoder' (tab 'vaapi') has different option order:
  HTML order: [vaapi_blbrc, vaapi_quality, vaapi_rc, vaapi_strict_rc_buffer]
  MD order:   [vaapi_rc, vaapi_quality, vaapi_blbrc, vaapi_strict_rc_buffer]



Captured stderr:

[2026-07-08 20:22:14.755]: Tests: From /__w/Sunshine/Sunshine/pkg/src/sunshine/tests/integration/test_config_consistency.cpp:511
[2026-07-08 20:22:14.755]: Tests:   ConfigConsistencyTest/ConfigOptionsInSameOrderWithinSections started
[2026-07-08 20:22:14.777]: Tests: At /__w/Sunshine/Sunshine/pkg/src/sunshine/tests/integration/test_config_consistency.cpp:567
[2026-07-08 20:22:14.777]: Tests:   Failure: Failed
Config option order inconsistencies:
  Section 'VA-API Encoder' (tab 'vaapi') has different option order:
  HTML order: [vaapi_blbrc, vaapi_quality, vaapi_rc, vaapi_strict_rc_buffer]
  MD order:   [vaapi_rc, vaapi_quality, vaapi_blbrc, vaapi_strict_rc_buffer]

[2026-07-08 20:22:14.777]: Tests: ConfigConsistencyTest/ConfigOptionsInSameOrderWithinSections failed
[  FAILED  ] ConfigConsistencyTest.ConfigOptionsInSameOrderWithinSections (22 ms)
```

psyke83:
@ReenigneArcher 

I'm a bit confused as to why doxygen is entering into the linux/freebsd macro block for the Windows builds, but I think the latest commit should fix it? Let me know if I need to modify this further.

ReenigneArcher:
> @ReenigneArcher
> 
> I'm a bit confused as to why doxygen is entering into the linux/freebsd macro block for the Windows builds, but I think the latest commit should fix it? Let me know if I need to modify this further.

Have to predefine macros for doxygen or docs fail to build. Also since the hosted docs are built on Linux, it wouldn't be able to show anything behind other macros if we didn't predefine them.

https://github.com/LizardByte/doxyconfig/blob/fd1e85601edf18ad72ebde29bb81ffde95852496/doxyconfig-Doxyfile#L88-L95

I think you have the right fix for this though.

psyke83:
@Kishi85 I'm not 100% certain if the Sonar warning is highlighting a real issue. Can you verify that BLBRC is activating correctly? With `min_log_level = 0`:

```
$ journalctl -f | grep -Ei "rc |block level"
Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Verbose: [hevc_vaapi @ 0x55f68e21fe00] Driver supports RC modes CQP, CBR, VBR, QVBR.
Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC mode: VBR.
Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] Block Level bitrate control: OFF.
Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC target: 100% of 1000000 bps over 1000 ms.
Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC buffer: 1000000 bits, initial fullness 750000 bits.
Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC framerate: 60/1 (60.00 fps).
```

Mine always reports BLBRC as OFF (even if I hardcode to true in video.cpp) due to lack of driver support, so I'm not 100% certain if there really is an issue with initialization order not parsing the user config correctly.

I've verified that RC, strict vbv and quality is being passed correctly, but those are initialized separately through vaapi.cpp.

Kishi85:
> @Kishi85 I'm not 100% certain if the Sonar warning is highlighting a real issue. Can you verify that BLBRC is activating correctly? With `min_log_level = 0`:
> 
> ```
> $ journalctl -f | grep -Ei "rc |block level"
> Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Verbose: [hevc_vaapi @ 0x55f68e21fe00] Driver supports RC modes CQP, CBR, VBR, QVBR.
> Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC mode: VBR.
> Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] Block Level bitrate control: OFF.
> Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC target: 100% of 1000000 bps over 1000 ms.
> Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC buffer: 1000000 bits, initial fullness 750000 bits.
> Jul 09 00:02:21 archlinux sunshine[143331]: [2026-07-09 00:02:21.924]: Debug: [hevc_vaapi @ 0x55f68e21fe00] RC framerate: 60/1 (60.00 fps).
> ```
> 
> Mine always reports BLBRC as OFF (even if I hardcode to true in video.cpp) due to lack of driver support, so I'm not 100% certain if there really is an issue with initialization order not parsing the user config correctly.
> 
> I've verified that RC, strict vbv and quality is being passed correctly, but those are initialized separately through vaapi.cpp.

Attaching a debugger, setting a breakpoint where Sonar complains and inspecting `config::video.vaapi.blbrc` there I'm always getting `false` even with `vaapi_blbrc = enabled` in sunshine.conf so it seems like this is a real issue. When also setting a breakpoint on `config.cpp:1656` and running again (continuing on each break) it shows that `video.cpp` breakpoints are processed before `config.cpp` which explains the incorrect behaviour (as the value is set to `true` too late) and also gives merit Sonar's warning.

EDIT: My card or driver also does not seem to support BLBRC but that is masked by this issue. It seems that only getting a warning (and fallback to Vulkan but that might be due to just running in my dev env) when hardcoding the value to `true` in video.cpp.

psyke83:
@Kishi85 

I've changed the config to pass as a pointer to align with other examples of bools being parsed in the code (e.g. `amd_preanalysis`).

As a local test, since my VAAPI driver doesn't expose any low power modes, I modified the hevc_vaapi entry:
```
index 2abc1eb4..faabd173 100644
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -1205,7 +1205,7 @@ namespace video {
       // Common options
       {
         {"async_depth"s, 1},
-        {"blbrc"s, &config::video.vaapi.blbrc},
+        {"low_power"s, &config::video.vaapi.blbrc},
         {"sei"s, 0},
         {"idr_interval"s, std::numeric_limits<int>::max()},
       },
```

Enabling blbrc gives this log:
```
Jul 09 11:05:44 archlinux sunshine[45787]: [2026-07-09 11:05:44.101]: Debug: [h264_vaapi @ 0x55a779e35bc0] Using VAAPI entrypoint VAEntrypointEncSlice (6).
Jul 09 11:05:44 archlinux sunshine[45787]: [2026-07-09 11:05:44.127]: Error: [hevc_vaapi @ 0x55a779e35bc0] No usable encoding entrypoint found for profile VAProfileHEVCMain (17).

```
Leaving disabled, hevc works fine:
```
Jul 09 11:06:17 archlinux sunshine[45787]: [2026-07-09 11:06:17.366]: Debug: [hevc_vaapi @ 0x5631f1a2cd40] Using VAAPI entrypoint VAEntrypointEncSlice (6).
```

I also did a sanity check by comparing the output of standalone ffmpeg when blbrc is enabled, and it always reports OFF for me.

As long as Sonar stops complaining after this change, I think the PR should be solid now.

psyke83:
@ReenigneArcher 

Although I believe the blbrc config is being parsed correctly (see my previous comment), I'm not sure how to address the Sonar warning. I don't think that my changes in this PR are fundamentally different to the other AVOptions being parsed via video.cpp for other encoders, so the potential initialization ordering issue could apply to all other options across all encoders. I'll mark as ready for review, but someone else should check if I'm overlooking an obvious blunder.

Kishi85:
The main problem here is that `encoder_t vaapi` is initialzied before config is parsed, so passing a pointer all the way to just before the value is dereferenced should work as long as config updates that pointer's value. As your low_power test does seem to confirm just that. I'd say Sonar's warning is about the pointer address potentially changing (which it doesn't in this case).

Alternatively would just adding the necessary common_option during config parsing work? As that would always guarantee the correct value being set after parsing. 

Could doing it like this be a solution?
```diff
diff --git a/src/config.cpp b/src/config.cpp
index 7d8f0145..66d8ba22 100644
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -28,6 +28,7 @@
 #include "platform/common.h"
 #include "rtsp.h"
 #include "utility.h"
+#include "video.h"
 
 #ifdef _WIN32
   #include <shellapi.h>
@@ -1653,6 +1654,9 @@ namespace config {
       video.vaapi.vaapi_rc = vaapi::rc_from_view<vaapi::rc_e>(video.vaapi.vaapi_rc_str, video.vaapi.vaapi_rc);
     }
     bool_f(vars, "vaapi_blbrc", (bool &) video.vaapi.blbrc);
+    video::vaapi.h264.common_options.emplace_back("blbrc"s, &video.vaapi.blbrc);
+    video::vaapi.hevc.common_options.emplace_back("blbrc"s, &video.vaapi.blbrc);
+    video::vaapi.av1.common_options.emplace_back("blbrc"s, &video.vaapi.blbrc);
     bool_f(vars, "vaapi_strict_rc_buffer", video.vaapi.strict_rc_buffer);
 
     int_f(vars, "vk_tune", video.vk.tune);
diff --git a/src/video.cpp b/src/video.cpp
index 2abc1eb4..c34751aa 100644
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -1191,7 +1191,6 @@ namespace video {
       // Common options
       {
         {"async_depth"s, 1},
-        {"blbrc"s, &config::video.vaapi.blbrc},
         {"idr_interval"s, std::numeric_limits<int>::max()},
       },
       {},  // SDR-specific options
@@ -1205,7 +1204,6 @@ namespace video {
       // Common options
       {
         {"async_depth"s, 1},
-        {"blbrc"s, &config::video.vaapi.blbrc},
         {"sei"s, 0},
         {"idr_interval"s, std::numeric_limits<int>::max()},
       },
@@ -1220,7 +1218,6 @@ namespace video {
       // Common options
       {
         {"async_depth"s, 1},
-        {"blbrc"s, &config::video.vaapi.blbrc},
         {"sei"s, 0},
         {"idr_interval"s, std::numeric_limits<int>::max()},
       },
```

UPDATE: After testing I can confirm your current solution is working as intended as I'm getting the BLBRC warning when enabling the option after properly installing Sunshine including this PR.

ReenigneArcher:
This seems like an existing pattern issue, and 4 other cases of it in master all in `video.cpp`.

https://sonarcloud.io/project/issues?fileUuids=AZFC159yi5a_X_62txCH&languages=cpp&rules=cpp%3AS7119&issueStatuses=OPEN%2CCONFIRMED&id=LizardByte_Sunshine

neatnoise:
> Been running this for a few hours yesterday with all options (VBR, BLBRC, STRICT) flipped on and quality set to high.
> 
> Subjectively I'm impressed with the quality I'm getting especially on red-heavy scenes with movement it's a lot less washed out compared to Vulkan encoding. Unsure where that comes from. Also host processing latency is just about .2 ms higher than current Vulkan default options which is negliable for me (1.9ms->2.1ms average).

It is possible to balance quality over the latency cost by the "Tuning" option in Vulkan encoder. Also when comparing the both encoders "Rate Control" should be set the same. BLBRC is not supported by the driver right now when AMD GPU 9070 XT is used so I can't compare with the best possible VAAPI settings. The output of both encoders is very similar in my case. Advantage of Vulkan encoder is that the single API is used in the pipeline compared with VAAPI (VAAPI, EGL) so it is simpler, less error prone and more efficient

Kishi85:
> > Been running this for a few hours yesterday with all options (VBR, BLBRC, STRICT) flipped on and quality set to high.
> > Subjectively I'm impressed with the quality I'm getting especially on red-heavy scenes with movement it's a lot less washed out compared to Vulkan encoding. Unsure where that comes from. Also host processing latency is just about .2 ms higher than current Vulkan default options which is negliable for me (1.9ms->2.1ms average).
> 
> It is possible to balance quality over the latency cost by the "Tuning" option in Vulkan encoder. Also when comparing the both encoders "Rate Control" should be set the same. BLBRC is not supported by the driver right now when AMD GPU 9070 XT is used so I can't compare with the best possible VAAPI settings. The output of both encoders is very similar in my case. Advantage of Vulkan encoder is that the single API is used in the pipeline compared with VAAPI (VAAPI, EGL) so it is simpler, less error prone and more efficient

Good to know. I'll have to check those settings again and try different scenarios. 

That still leaves the memory leak on FFmpeg's Vulkan encoding @psyke83 discovered and [analyzed here](https://github.com/LizardByte/Sunshine/pull/5360#issuecomment-4871789709). 
Maybe you got an idea on that @neatnoise ? Otherwise I'll write up an issue report and put it up on code.ffmpeg.org

Small update: When using (high) quality settings with CBR for both Vulkan and VAAPI using this PR: I'm seeing an about 1ms higher average host processing latency when using Vulkan (3.2-3.4ms) compared to VAAPI (2.2-2.4ms). There is also almost no difference on host processing latency for different VAAPI quality levels.

neatnoise:
@Kishi85 could you check if it resolves the memory leak issue on your side please? https://github.com/LizardByte/Sunshine/pull/5404

Kishi85:
Been running this for the last few days. No further issues on my end, everything seems to be working as intended.

PR #5495, opened 2026-08-11
fix(linux/pipewire): restore dummy buffer for software encoders and fix buffer ownership

luanweslley77:
## Description

the problem (`software encoder` + `KWin/portal` crashes at startup with `"Couldn't scale frame"`; and `SIGABRT` during teardown), the cause (regression from `PR #5360`, which removed the `dummy_img` allocation and added a destructor with `delete[]` despite lacking ownership), and what the PR does (restores the dummy buffer with `data_owned`; adds `NV12` support to the software encoder; adds a unit test).

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/linux/pipewire.cpp
+++ b/src/platform/linux/pipewire.cpp
@@ -128,11 +128,17 @@ namespace pipewire {
    */
   struct img_descriptor_t: public egl::img_descriptor_t {
     ~img_descriptor_t() override {
-      if (data) {
+      // Only free buffers this image actually owns. The memory-buffer capture
+      // path points img->data at the PipeWire staging vector (front_buffer),
+      // which is owned by pipewire_t -- deleting it here corrupts the heap.
+      if (data && data_owned) {
         delete[] data;
-        data = nullptr;
       }
+      data = nullptr;
+      data_owned = false;
     }
+
+    bool data_owned = false;  ///< Whether img->data is owned by this image and must be freed.
   };
 
   /**
@@ -430,13 +436,17 @@ namespace pipewire {
 
       struct spa_buffer *buf = stream_data.current_buffer->buffer;
       if (buf->datas[0].chunk->size != 0) {
-        auto *img_descriptor = static_cast<egl::img_descriptor_t *>(img);
+        auto *img_descriptor = static_cast<img_descriptor_t *>(img);
         fill_img_metadata(img_descriptor, buf);
         if (buf->datas[0].type == SPA_DATA_DmaBuf) {
           fill_img_dmabuf(img_descriptor, buf, stream_data);
         } else {
           img->data = stream_data.front_buffer->data();
+          img_descriptor->data_owned = false;
           img->row_pitch = stream_data.local_stride;
+          // NV12 is the only 1-byte-per-pixel format delivered on the memory
+          // path; every other negotiated format is packed 4 bytes per pixel.
+          img->pixel_pitch = (stream_data.format.info.raw.format == SPA_VIDEO_FORMAT_NV12) ? 1 : 4;
         }
       }
 
@@ -937,6 +947,7 @@ namespace pipewire {
       img->sequence = 0;
       img->serial = std::numeric_limits<decltype(img->serial)>::max();
       img->data = nullptr;
+      img->data_owned = false;
       std::fill_n(img->sd.fds, 4, -1);
 
       return img;
@@ -1061,7 +1072,20 @@ namespace pipewire {
      * @return Capture status reported to the streaming pipeline.
      */
     int dummy_img(platf::img_t *img) override {
-      // Empty images are recognized as dummies by the zero sequence number
+      // Software encoders convert the dummy image immediately; provide a valid
+      // (black) buffer instead of leaving img->data null, which makes sws fail
+      // with EINVAL. The buffer is new[]-allocated and marked as owned so the
+      // destructor releases it.
+      if (img->data == nullptr) {
+        const auto w = img->width;
+        const auto h = img->height;
+        if (w > 0 && h > 0) {
+          img->data = new uint8_t[static_cast<size_t>(w) * h * 4]();  // NOSONAR(cpp:S5025) - buffer is owned by the image and freed by img_descriptor_t's destructor
+          static_cast<img_descriptor_t *>(img)->data_owned = true;
+          img->row_pitch = w * 4;
+          img->pixel_pitch = 4;
+        }
+      }
       return 0;
     }
 
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -3,6 +3,7 @@
  * @brief Definitions for video.
  */
 // standard includes
+#include <algorithm>
 #include <array>
 #include <atomic>
 #include <bitset>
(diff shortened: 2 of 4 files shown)


src/platform/linux/pipewire.cpp
```diff
@@ -1061,7 +1069,20 @@ namespace pipewire {
...
+      if (img->data == nullptr) {
+        const auto w = img->width;
+        const auto h = img->height;
+        if (w > 0 && h > 0) {
+          img->data = new uint8_t[static_cast<size_t>(w) * h * 4]();  // NOSONAR(cpp:S5025) - buffer is owned by the image and freed by img_descriptor_t's destructor
+          static_cast<img_descriptor_t *>(img)->data_owned = true;
+          img->row_pitch = w * 4;
+          img->pixel_pitch = 4;
+        }
+      }
```
Kishi85, line 1088: If this is really necessary then other (linux) capture methods would require it as well but both kmsgrab (see [1](https://github.com/LizardByte/Sunshine/blob/3cba9baebac882b336be3ebe129ee612cb189853/src/platform/linux/kmsgrab.cpp#L1713) and [2](https://github.com/LizardByte/Sunshine/blob/3cba9baebac882b336be3ebe129ee612cb189853/src/platform/linux/kmsgrab.cpp#L1791)) and wlgrab (see [3](https://github.com/LizardByte/Sunshine/blob/3cba9baebac882b336be3ebe129ee612cb189853/src/platform/linux/wlgrab.cpp#L147)) do not need this and software encoders still work properly with both so this is cannot be the root cause as you describe your comment on top.

This change to `pipewire.cpp` therefore is just working around the root cause of the issue again (which might be somewhere else, even in the software encoder). Still, the workaround is better implemented this time as it hopefully won't leak memory (which definitely should be ensured before merging this so we don't fall back to a state before #5360). 

A proper fix would still be preferrable keeping same logic for `dummy_img()` intact between all linux capture methods for ease of future maintenance.

src/platform/linux/pipewire.cpp
```diff
@@ -128,11 +128,17 @@ namespace pipewire {
    */
   struct img_descriptor_t: public egl::img_descriptor_t {
     ~img_descriptor_t() override {
-      if (data) {
+      // Only free buffers this image actually owns. The memory-buffer capture
+      // path points img->data at the PipeWire staging vector (front_buffer),
+      // which is owned by pipewire_t -- deleting it here corrupts the heap.
+      if (data && data_owned) {
```
Kishi85, line 134: Makes sense to check for ownership here  to not corrupt the heap.

We didn't find a potential trigger where this is needed back in #5360 but I'm unsure if software encoding was extensively tested back then (as this might be the only trigger.

src/platform/linux/pipewire.cpp
```diff
@@ -430,12 +436,13 @@ namespace pipewire {
 
       struct spa_buffer *buf = stream_data.current_buffer->buffer;
       if (buf->datas[0].chunk->size != 0) {
-        auto *img_descriptor = static_cast<egl::img_descriptor_t *>(img);
+        auto *img_descriptor = static_cast<img_descriptor_t *>(img);
```
Kishi85, line 439: Looks like this was missed in #5360. Good catch!

luanweslley77:
Thanks for the review — and agreed on both the destructor ownership and the
`fill_img()` points. For the dummy image, I dug into what you linked and I think
I can show where the root cause actually is, because it ends up supporting your
preference to keep `dummy_img()` as a no-op.

The reason kmsgrab and wlgrab work with a no-op `dummy_img()` is not the dummy
itself — it's that their `alloc_img()` already hands out a buffer:

```cpp
// kmsgrab.cpp:1702
img->data = new std::uint8_t[height * img->row_pitch];

// wlgrab.cpp:351
img->data = new std::uint8_t[height * img->row_pitch];
```

So when `validate_config()` runs `alloc_img()` → `dummy_img()` → `convert()`,
the image already has valid data. PipeWire's `alloc_img()` is the odd one out:

```cpp
// pipewire.cpp:939
img->data = nullptr;
```

That's why the same no-op dummy breaks software encoding **only** on PipeWire
captures (kwin/portal): the software encoder's `sws_scale_frame()` gets a null
pointer and fails with EINVAL. The `ValidateEncoder/software` gtest reproduces
it — it fails on master and passes with the fix.

To keep `dummy_img()` a no-op everywhere (which I agree is the right thing for
maintenance), the fix would move into `alloc_img()` instead: allocate the
buffer there like kmsgrab/wlgrab, and have `fill_img()`'s memory path copy the
frame into that owned buffer instead of pointing `img->data` at the staging
vector. That removes the `front_buffer` aliasing entirely (no heap corruption
by construction), and there's no leak (the buffer is owned by the image,
recycled through the pool, and freed by the destructor).

One trade-off to be transparent about: that means one extra `memcpy` per frame
in the memory-buffer path (the staging copy already exists in `on_process`).
For 1080p60 NV12 that's roughly 180 MB/s of additional copying — a couple
percent of CPU on software encoding. If that cost is acceptable, I'll rework
the PR along those lines. If you'd rather keep the allocation in
`dummy_img()` as it is now (which doesn't add per-frame copies and doesn't
leak), the current version already addresses the crash on master.

---

Kishi85:
> Thanks for the review — and agreed on both the destructor ownership and the `fill_img()` points. For the dummy image, I dug into what you linked and I think I can show where the root cause actually is, because it ends up supporting your preference to keep `dummy_img()` as a no-op.
> 
> The reason kmsgrab and wlgrab work with a no-op `dummy_img()` is not the dummy itself — it's that their `alloc_img()` already hands out a buffer:
> 
> ```c++
> // kmsgrab.cpp:1702
> img->data = new std::uint8_t[height * img->row_pitch];
> 
> // wlgrab.cpp:351
> img->data = new std::uint8_t[height * img->row_pitch];
> ```
> 
> So when `validate_config()` runs `alloc_img()` → `dummy_img()` → `convert()`, the image already has valid data. PipeWire's `alloc_img()` is the odd one out:
> 
> ```c++
> // pipewire.cpp:939
> img->data = nullptr;
> ```
> 
> That's why the same no-op dummy breaks software encoding **only** on PipeWire captures (kwin/portal): the software encoder's `sws_scale_frame()` gets a null pointer and fails with EINVAL. The `ValidateEncoder/software` gtest reproduces it — it fails on master and passes with the fix.
> 
> To keep `dummy_img()` a no-op everywhere (which I agree is the right thing for maintenance), the fix would move into `alloc_img()` instead: allocate the buffer there like kmsgrab/wlgrab, and have `fill_img()`'s memory path copy the frame into that owned buffer instead of pointing `img->data` at the staging vector. That removes the `front_buffer` aliasing entirely (no heap corruption by construction), and there's no leak (the buffer is owned by the image, recycled through the pool, and freed by the destructor).
> 
> One trade-off to be transparent about: that means one extra `memcpy` per frame in the memory-buffer path (the staging copy already exists in `on_process`). For 1080p60 NV12 that's roughly 180 MB/s of additional copying — a couple percent of CPU on software encoding. If that cost is acceptable, I'll rework the PR along those lines. If you'd rather keep the allocation in `dummy_img()` as it is now (which doesn't add per-frame copies and doesn't leak), the current version already addresses the crash on master.

Thanks for clarifying the differences. I wasn't aware that both kmsgrab and wlgrab handle the alloc earlier already.

In that case I'm good with your PR changes on `pipewire.cpp` the way they are right now. No need to rework that especially when that would add unnecessary overhead. 

Also feel free to mark my earlier comments as resolved (for some reason github won't allow me to do that myself due to missing permissions).

P.S.: I'll apply this PR to my personal setup build and do some testing

UPDATE: No issues on my system running this so far.

Kishi85:
@ReenigneArcher This is (at least according to my testing) good to merge and will restore using software encoding for pipewire-based methods

ReenigneArcher:
Looks like it's segfaulting during unit tests.

luanweslley77:
On the unit test segfault in CI:

I asked my agent to analyze it, and it found something that seems to explain what's happening: a possible double-free in the new test. `set_frame()` takes ownership of the frame (per its docs), but the test also calls `av_frame_free(&frame)` — which would mean the frame gets freed twice. The failing run (c9e11a8) does point at `SoftwareEncoderConversion.Bgr0AndNv12` on every platform (Linux: `Aborted (core dumped)`; Windows: exit 127), and that would also explain the coverage jobs/codecov failing downstream. So this seems like a likely explanation, but we're still validating it.

As for why I didn't see this locally: the test never crashed on my machine, probably because glibc's tcache hides double-frees (the freed chunk stays reusable), while the CI allocators detect them and abort. It's reproducible locally with `GLIBC_TUNABLES=glibc.malloc.tcache_count=0` → SIGABRT, which matches.

What I'm considering: not freeing the frame in the test (the device owns it), and using the real class from `video.h` instead of the mirror declaration in the test (I think it was an ODR violation — `apply_colorspace` was missing). While testing I also noticed the `row_pitch == width` NV12 detection could break with padded strides, and that `reinit_sws()` may drop the colorspace set by `apply_colorspace()`

@ReenigneArcher I'm confident everything is in order now. Could you please trigger the CI pipeline once more?

ReenigneArcher:
- https://github.com/LizardByte/Sunshine/actions/runs/32138214340/job/95715577783?pr=5495#step:11:1669
- https://github.com/LizardByte/Sunshine/actions/runs/32138214340/job/95715578012?pr=5495#step:12:2397

PR #5510, opened 2026-08-15
fix(macos): recover capture after the display sleeps

### LINKED ISSUE #5509: macOS: display sleep permanently stalls capture and the stream cannot recover

Present in current `master` (`cf52f4b6`), which I built from source to investigate.

### Describe the Bug

On macOS, putting the display to sleep while a stream is running stalls capture
permanently. The client stays connected but never receives another frame, and waking the
display does not bring it back. Only quitting and relaunching Sunshine restores streaming.

To reproduce:

1. Start a stream
2. Put the display to sleep — a hot corner set to *Put Display to Sleep* does it instantly,
   but idle sleep or closing the lid gets there too
3. Wake the display
4. The stream never resumes. Reconnecting does not help.

**Cause**

`src/platform/macos/display.mm`, in `capture()`:

```objc
// FIXME: We should time out if an image isn't returned for a while
dispatch_semaphore_wait(signal, DISPATCH_TIME_FOREVER);

return capture_e::ok;
```

That semaphore is not "a frame arrived" — `-[AVVideo captureOutput:didOutputSampleBuffer:fromConnection:]`
only signals it when the frame callback returns `false`, i.e. when capture is deliberately
stopped. So the wait is for the entire capture session to end.

A sleeping display stops `AVCaptureSession` delivering sample buffers, and the session does
not resume when the display wakes. The capture thread is therefore parked for the lifetime
of the process, with nothing able to wake it.

Two more things turn that into a dead end rather than a hiccup:

- `capture()` can only ever return `capture_e::ok` — it has no path to `error` or `reinit`.
- Nothing listens for display sleep/wake or `CGDisplay` reconfiguration, so the session is
  never rebuilt when the display comes back.

And in `src/video.cpp`, the capture thread rebuilds the display on exactly one value:

```cpp
case platf::capture_e::reinit: {
  // ... refresh_displays / reset_display ...
}

case platf::capture_e::error:
case platf::capture_e::ok:
case platf::capture_e::timeout:
case platf::capture_e::interrupted:
  return;                       // ends the session
```

So the one value that would trigger recovery is the one the macOS backend never sends.

`dummy_img()` has the same unbounded wait, and there the semaphore genuinely does mean "one
frame arrived" — so encoder probing hangs indefinitely if the display happens to be asleep
at startup.

### Expected Behavior

When the display wakes, capture is rebuilt and the stream resumes on its own, without
restarting Sunshine.

### Additional Context

I have a patch for this and it is verified on hardware — PR to follow, and I'll link it
here.

Two details from writing it that may be useful to anyone else looking at this:

- **A frame-arrival timeout is the wrong shape.** `AVCaptureScreenInput` is change-driven,
  so a static desktop looks exactly like a stalled one and would be reinitialized
  needlessly. Waiting for a sleep-then-wake transition cannot produce that false positive.
- **Stopping the session before returning is not enough, and crashes.** The output stays in
  the capture session and in `AVVideo`'s three map tables, and `-[AVVideo dealloc]` releases
  those map tables *before* it stops the session — so the output is released while the
  session still holds it, and the process aborts in `objc_msgSend` during the next
  reinitialization. Abandoning a capture needs the same teardown the frame callback performs
  when it returns `false`.

With the fix, the log during recovery looks like this, and the client session continues
uninterrupted:

```
[11:31:42.133]: Info: Display [1] woke from sleep, reinitializing capture
[11:31:42.219]: Info: Configuring selected display (1) to stream
[11:31:42.229]: Info: Creating encoder [hevc_videotoolbox]
```

### Host Operating System

macOS

### Operating System Version

26.6.1 (25G76)

### Architecture

arm64

### Package

macOS installer (`Sunshine.app`), and a source build of `master` for the investigation

### GPU Type

Integrated (Apple silicon)

### GPU Model

Apple M1 Max

### GPU Driver/Mesa Version

n/a — macOS 26.6.1, VideoToolbox

### Capture Method

AVFoundation (`av_display_t`, the macOS default)

harrison001:
## Description

Display sleep can stall capture on macOS, and the stream never recovers.

The semaphore in `capture()` isn't signaled for every frame. `av_video.m` only signals it when the frame callback returns false, which normally happens when the session ends. If the display goes to sleep, buffers stop arriving and the capture thread can sit there forever.

Use a timed wait instead and check `CGDisplayIsAsleep`. If the display goes to sleep and later wakes up, stop the current capture and return `capture_e::reinit`. `video.cpp` treats the other return values as the end of the session, so this needs to be `reinit`.

A frame timeout doesn't work here because `AVCaptureScreenInput` is change-driven. An idle desktop may not produce frames either, so lack of frames alone doesn't tell us capture is stuck. Watching the display sleep/wake state avoids that ambiguity.

`dummy_img()` had the same unbounded wait.  There the semaphore really does mean one frame arrived, but encoder probing would hang forever if the display was already asleep at startup, so that wait is bounded now too.

Also use `stopCapture:` rather than calling `stopRunning()` directly. `stopRunning()` on its own leaves the output registered with the session and in `AVVideo`'s map tables. `dealloc` then releases those map tables before stopping the session, so the output gets released while the session still holds it.

No unit test for this one. It needs a real display going to sleep and waking up, which I can't reproduce in a test. Verified on my hardware instead: macOS 26.6.1, M1 Max, built-in display, streaming to Moonlight iOS. Capture rebuilt 86 ms and 62 ms after wake across two sleep/wake cycles, fresh IDR keyframe 264 ms after wake, one client session survived both cycles, no crash reports. I don't have a second display, so the external display path is untested.

### Issues Fixed or Closed

- Fixes #5509

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/src/platform/macos/av_video.h
+++ b/src/platform/macos/av_video.h
@@ -88,4 +88,15 @@ typedef bool (^FrameCallbackBlock)(CMSampleBufferRef);
  */
 - (dispatch_semaphore_t)capture:(FrameCallbackBlock)frameCallback;
 
+/**
+ * @brief Abandon a capture that has not ended on its own.
+ *
+ * The frame callback normally tears its own capture down by returning false. When the
+ * caller gives up on a capture that stopped delivering frames, such as after the display
+ * slept, this performs that teardown on its behalf.
+ *
+ * @param signal Semaphore previously returned by capture:.
+ */
+- (void)stopCapture:(dispatch_semaphore_t)signal;
+
 @end
--- a/src/platform/macos/av_video.h
+++ b/src/platform/macos/av_video.h
@@ -96,6 +96,7 @@ typedef bool (^FrameCallbackBlock)(CMSampleBufferRef);
  * slept, this performs that teardown on its behalf.
  *
  * @param signal Semaphore previously returned by capture:.
+ * @note This method waits for any in-flight frame callback to finish before returning.
  */
 - (void)stopCapture:(dispatch_semaphore_t)signal;
 
--- a/src/platform/macos/av_video.m
+++ b/src/platform/macos/av_video.m
@@ -5,6 +5,21 @@
 // local includes
 #import "av_video.h"
 
+/**
+ * @brief Private capture lifecycle helpers for AVVideo.
+ */
+@interface AVVideo ()
+
+/**
+ * @brief Tear down one capture after its callback queue has been serialized.
+ *
+ * @param connection Capture connection to tear down.
+ * @param signalCompletion Whether to wake the thread waiting for normal capture completion.
+ */
+- (void)finishCapture:(AVCaptureConnection *)connection signalCompletion:(BOOL)signalCompletion;
+
+@end
+
 @implementation AVVideo
 
 - (id)initWithDisplay:(CGDirectDisplayID)displayID frameRate:(int)frameRate {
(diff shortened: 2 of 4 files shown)


ReenigneArcher:

harrison001:
I am sorry, I skipped the template. I have fixed and marked AI usage as Heavy.

Worth adding that the first version of this patch was wrong. It detected the wake correctly and logged the right line, then killed the process ten seconds later .  The reason: `dealloc` frees the map tables before stopping the session. That's what `stopCapture:` is for. I found it by running it, not by reading it.

Please review and let me know if further changes are needed. Thanks.

ReenigneArcher:
> Please review and let me know if further changes are needed. Thanks.

I used GPT 5.6 to review and add one commit on top of your changes, with tests.

But I do not have a macOS device to properly test this all with, so I'd appreciate if you could. And please let me know if you disagree with any of the changes made. Thanks!

harrison001:
Tested on macOS 26.6.2, M1 Max, Apple clang.

Build is clean. Full suite: 431 tests, 428 passed, 3 skipped, 0 failed. Your three AVVideoTest cases pass.

Live test against a real client streaming the desktop:

- Display put to sleep mid-session, then woken. Log shows `Display [1] woke from sleep, reinitializing capture`, the encoder is rebuilt, the session continues, the process survives. That is the case this PR is about and it works.
- Display asleep, then client disconnects. Teardown completes, process survives.

I also looked at the `dispatch_sync` in `stopCapture:` for a self-deadlock. Both call sites in `display.mm` are on the capture thread, not on the sample buffer callback queue, so it can't deadlock there.

One build note, not caused by this PR. I first hit a SIGSEGV in `SystemTrayTest.LifecycleMenuAndStateTransitions`. The cause is Qt5 and Qt6 in the same process. `third-party/tray` does `find_package(Qt6 COMPONENTS Widgets Svg)` without REQUIRED, and on Homebrew `qtsvg` is a separate formula that does not live under the `qtbase` prefix, so Qt6 isn't found, tray silently falls back to Qt5, and the rest of Sunshine still links Qt6. The binary loads two QtCore and dies. Configuring with `-DQt6Svg_DIR=/opt/homebrew/opt/qtsvg/lib/cmake/Qt6Svg` drops every qt@5 reference and the suite passes. A stale `LDFLAGS`/`CPPFLAGS` pointing at qt@5 in my shell made it easier to hit. Might be worth making that `find_package` REQUIRED so it fails loudly instead of producing a mixed binary.

On the change itself, one gap. The loop only leaves when the display comes back:

```cpp
while (dispatch_semaphore_wait(signal, dispatch_timeout_from_now(capture_poll_interval)) != 0) {
  if (CGDisplayIsAsleep(display_id)) {
    display_slept = true;
  } else if (display_slept) {
    [av_capture stopCapture:signal];
    return capture_e::reinit;
  }
}
```

If the display sleeps and does not wake, neither exit is reachable, `videoThread.join()` never returns, and the watchdog in `session::join()` still fires after ten seconds. I hit that on 2026-09-07: the panel slept from a hot corner and did not come back until I closed and opened the lid, and Sunshine died with the same SIGTRAP this PR is meant to prevent.

I could not reproduce that on demand today. With a session running, the display woke by itself about five seconds after `pmset displaysleepnow` every time, so for that case I have the one crash report rather than a repeatable test.

What I run locally is a third exit: if `CGDisplayIsAsleep` stays true for more than three seconds, `stopCapture` and return `capture_e::error`, which ends the session and leaves the display asleep.

One thing worth knowing if you go that way: returning `reinit` there does not work. `reinit` goes back through `display()`, which calls `wake_display()`, so the display comes straight back on and you get a sleep/wake oscillation. I shipped `reinit` first and had to change it.

I can fold the third exit into this PR, or leave it for a separate one. Your call.

ReenigneArcher:
> I can fold the third exit into this PR, or leave it for a separate one. Your call.

Go ahead and add your suggested changes.

The QT thing might have been fixed by https://github.com/LizardByte/Sunshine/pull/5679 ?

PR #5529, opened 2026-08-19
feat(linux/input): optional absolute mouse as relative motion

danalec:
## Description

Moonlight TV clients (and other absolute-motion sources) send absolute mouse
positions. The Linux backend forwards them as absolute uinput events, which
compositors treat as absolute pointer motion.

COSMIC's magnifier does not track the cursor on absolute motion:
`PointerMotionAbsolute` moves the cursor but never calls `update_focal_point`
(pop-os/cosmic-comp#2760; same symptom class as #1418/VirtualBox). Relative
motion updates the zoom on every event.

This adds an `absolute_mouse_as_relative` option (disabled by default) that
converts absolute client coordinates into relative deltas:

- The first absolute event still places the cursor absolutely, anchoring the
  baseline.
- Subsequent events are emitted as relative deltas (dead reckoning), so the
  cursor arrives at the compositor as relative pointer motion.
- The KMS capture path publishes the cursor-plane position every frame; the
  internal estimate snaps to it while the client is idle, and a coordinate
  saturated at the client's own surface edge targets the matching host edge,
  so movement cannot stall mid-screen. On backends without cursor feedback
  the conversion works from dead reckoning alone.

Documented in `docs/configuration.md`.

## Type of Change

## Checklist

### AI Usage

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -691,6 +691,35 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### absolute_mouse_as_relative
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            When enabled, absolute mouse positions from Moonlight clients are converted to relative motion.
+            <br>
+            The first event still places the cursor absolutely; subsequent events are emitted as relative deltas,
+            using the cursor position captured from the KMS cursor plane (when available) to stay in sync.
+            <br>
+            This can be useful on compositors whose assistive features (e.g. screen magnifiers) only track relative
+            pointer motion.
+        </td>
+    </tr>
+    <tr>
+        <td>Default</td>
+        <td colspan="2">@code{}
+            disabled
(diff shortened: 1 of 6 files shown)


danalec:
Note for reviewers: the only failing check is `lizardbyte-gh-pages-sunshine` (Read the Docs). Looking at the RTD build history, this build has been failing for unrelated commits since this morning (e.g. 34132141, 34131657, 34131645 — none of them this PR), while `sunshinestream` (the other docs build) passes on this PR's commit. This looks like an upstream RTD/infra issue, not something introduced by this change.

ReenigneArcher:
@danalec It won't pass until I approve the workflows.

Anyway... Thank you for the PR submission, but it looks like you used AI to create this PR.

Kishi85:
This might also fix absolute touch input on KDE in a multi-monitor scenario as that currently sends the touch events to a fixed monitor (configurable in systemsettings) instead of selecting the correct monitor by absolute coordinates.

Issue #1671, opened 2023-09-24
Audio not switched for some applications


pizzapim:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your issue described in the documentation?

- [X] I have read the documentation

### Is your issue present in the nightly release?

- [X] This issue is present in the nightly release

### Describe the Bug

I have two profiles for two games, they are identical except for the executable that is launched. However, for one game the audio is correctly muted on the desktop and piped to the Moonlight client, while for the other it is not. The working game is Hogwards Legacy, while the audio for Cities Skylines plays on my desktop's speakers.

### Additional Context

Problem exists for Moonlight-qt client on both Linux and Windows

### Host Operating System

Windows

### Operating System Version

Windows 11 Enterprise 10.0.22621 Build 22621

### Architecture

64 bit

### Sunshine commit or version

0.20.0.17a320bfa5ccf724223c4bf3d157ec27eb33bbf4

### Package

Windows - installer

### GPU Type

AMD

### GPU Model

AMD Radeon RX 6700 XT

### GPU Driver/Mesa Version

31.0.12027.9001

### Config

```Shell
(empty)
```

### Apps

```json
{
    "env": {
        "PATH": "$(PATH);$(ProgramFiles(x86))\\Steam"
    },
    "apps": [
        {
            "name": "Desktop",
            "image-path": "desktop.png"
        },
        {
            "name": "Steam Big Picture",
            "detached": [
                "steam steam:\/\/open\/bigpicture"
            ],
            "image-path": "steam.png"
        },
        {
            "name": "Hogwards Legacy",
            "output": "",
            "cmd": "C:\\Games\\Hogwarts Legacy\\HogwartsLegacy.exe",
            "exclude-global-prep-cmd": "false",
            "elevated": "true",
            "image-path": "hogwards-legacy.png",
            "prep-cmd": [
                {
                    "do": "C:\\Users\\user\\Documents\\MonitorProfileSwitcher_v0700\\MonitorSwitcher.exe -load:C:\\Users\\user\\Documents\\MonitorProfileSwitcher_v0700\\Moonlight.xml",
                    "undo": "C:\\Users\\user\\Documents\\MonitorProfileSwitcher_v0700\\MonitorSwitcher.exe -load:C:\\Users\\user\\Documents\\MonitorProfileSwitcher_v0700\\PC.xml",
                    "elevated": "false"
                }
            ]
        },
        {
            "name": "Cities Skylines",
[...]

PR #5570, opened 2026-08-29
feat(linux): get _exact_ 4:4:4 via supersampling of 4:2:0

alanjds:
Adds `chroma_supersample` option, that changes the scaling of a 1080p to 4k from `GL_LINEAR` to `GL_NEAREST`.

It makes the client-side downscale to be _exactly_ like 4:4:4, not _a very good approximate_ with a small chroma error.

## Why

I tried to get rid of chroma aberrations on streaming by turning on 4:4:4. It worked, but I lost hardware acceleration on server and on client, because the GPUs simply cannot do it. (Strix Halo APU server, Intel client)

That lead to high CPU load on client, delays on server frames, decoding enqueueing, a mess.

After iterating, found the best compromise on my network to be spending a bit more bandwidth via streaming 4:2:0 1080p as 4k, then down to 1080p on client. This way everything runs on hardware and is _almost_ identical to 4:4:4.

Digging on why the "almost", found that Sunshine uses a filter that added some blurry, and when the upscale is a X2 multiplication, the "just linear" produces no artifacts. So let's apply it instead!

What follows is just lore on the details. 

## Background

Streaming at an integer multiple of the desktop resolution and letting the client downscale is a workaround discussed elsewhere and that keeps both ends in hardware. See [Moonlight issue #1671, "Recombined" 4:4:4](https://github.com/moonlight-stream/moonlight-qt/issues/1671) for the protocol-level version of the same idea (recombining supersampled 4:2:0 into 4:4:4 on the client)

However it here is cheaper to implement.

## Caveat: Applies only on _even integer_ multipliers

It only works good when the upscaled is _even times_ of the original. Non-integer or odd times are better with the current implementation. But even multiples are better with GL_NEAREST

Tests:

| Factor | Bilinear chroma error (max / mean) | Nearest chroma error (max / mean) |
|---|---|---|
| 2x (even) | 43 / 5.43 | **0 / 0** |
| 3x (odd) | 76 / 6.86 | **129 / 7.34** |
| 4x (even) | 91 / 10.92 | **0 / 0** |

## Measurements

From my own setup (Bazzite/AMD Strix Halo host, VAAPI encode, KMS capture; Bazzite/Intel
UHD 620 client, Moonlight Qt, `DRI_PRIME=0`, fullscreen 1080p), 1920x1080 desktop, 45fps,
100 Mbps cap, `vaapi_rc = vbr`, client fullscreen, steady state on a mostly-static
desktop:

| | 1080p H.264 4:4:4 (x264, sw) | 1080p→4K HEVC 4:2:0 (VAAPI, hw) |
|---|---|---|
| Host processing latency | 4.6 ms | 6.0 ms |
| Client decode time | 7.43 ms | 0.94 ms |
| Frame queue delay | 49.78 ms | 0.41 ms |
| Frames dropped (jitter) | 20.88% | 0.00% |
| Rendering frame rate | 19.50 / 45 | 44.98 / 44.98 |
| Client CPU | 54.2% | 30.4% |
| Upload | ~5.6 Mbps | ~14.4 Mbps |
| Host CPU | ~1.4 cores | negligible |

For reference, native 1080p 4:2:0 via VAAPI on the same host measured ~13 Mbps at a
40 Mbps cap. Looks like SW compresses better, ok, but the interesting here is that VAAPI does not gets _that worst_ from 1080p to 4k. Worsen just about 10%

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2129,6 +2129,44 @@ editing the `conf` file in a text editor. Use the examples as reference.
     </tr>
 </table>
 
+### chroma_supersample
+
+<table>
+    <tr>
+        <td>Description</td>
+        <td colspan="2">
+            When the stream resolution is an exact, even integer multiple of the capture resolution (2x, 4x, ...) in
+            both axes, use nearest-neighbour (point) sampling for the capture-to-stream upscale instead of the
+            default bilinear filter.
+            @note{This makes 4:2:0 chroma subsampling exact: every captured pixel is replicated into an identical
+            NxN block, so the encoder's 2x2 chroma averaging reproduces the capture's own chroma bit-for-bit, prior
+            to quantisation. This is a practical alternative to 4:4:4 encoding on GPUs whose hardware encoder has no
+            4:4:4 profile at all (e.g. AMD VCN, Intel Quick Sync). See
+            [Moonlight issue #1671](https://github.com/moonlight-stream/moonlight-qt/issues/1671) for the
+            protocol-level version of the same idea; this option makes the existing supersampling workaround
+            exact without any client-side changes.}
+            @note{The client (Moonlight) must present the stream at exactly 1/N of the stream resolution for the
+            property to hold end to end — fullscreen playback is the reliable way to get this; a window with
+            decorations, or any non-integer client-side scaling, silently breaks it.}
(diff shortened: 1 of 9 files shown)


ReenigneArcher:

PR #5704, opened 2026-09-13
fix(pipewire): extend KWin version check to handle 6.8 beta/RC versions

psyke83:
## Description

Ref: https://bugs.kde.org/show_bug.cgi?id=525619

## Type of Change

## Checklist

### AI Usage


psyke83:
@ReenigneArcher 

The KDE 6.8 beta is available via Arch's kde-unstable repo. I gave it a try and found a regression in screencasting compared to 6.7 that I reported upstream.

This PR doesn't necessarily need to be merged to master, but could you please kick off the runner for the PR so that the artifacts are publicly available? That would help the KDE developers to reproduce the issue so they don't need to recompile Sunshine themselves if they decided to investigate my linked bug report. Thanks.

ReenigneArcher:
> This PR doesn't necessarily need to be merged to master

If it doesn't can an issue with 6.7 it'd probably be good to get it in. I'm hoping to make another release here in the next few days addressing a lot of the bugs from the last release.

psyke83:
The KDE devs always seem to bump the micro value of their major.minor.micro versioning to major.(old)minor.80 for betas. My current KWin 6.8 package version is `6.7.90-1` and the same version is reported via DBUS. I don't think it will cause problems for stable release users if merged to master.

src/platform/linux/pipewire.cpp
```diff
@@ -848,7 +848,8 @@ namespace pipewire {
       // Ref: https://bugs.kde.org/show_bug.cgi?id=524129
       // Also negotiate variable rate for all other compositors. Mutter's variable rate pacing is superior.
       const static std::vector<int> kwin_version = get_running_kwin_version();
-      const static bool negotiate_variable_rate = kwin_version.empty() || (kwin_version[0] == 5 || (kwin_version[0] == 6 && kwin_version[1] < 8));
+      // Ensure that 6.8 beta/RC versions are covered to assist in troubleshooting screencast issues. Ref: https://bugs.kde.org/show_bug.cgi?id=525619
+      const static bool negotiate_variable_rate = kwin_version.empty() || (kwin_version[0] == 5 || (kwin_version[0] == 6 && kwin_version[1] < 8 && kwin_version[2] < 80));
```
Kishi85, line 852:
```suggestion
      const static bool negotiate_variable_rate = kwin_version.empty() || (kwin_version[0] == 5 || (kwin_version[0] == 6 && kwin_version[1] < 7 && kwin_version[2] < 80));
```

Shouldn't this be checking for <6.7.80 instead of 6.8.80?

Kishi85: Nevermind. I've missed the check ordering, so all good.

Kishi85: Note: This would break for older beta versions of KWin, e.g. 6.6.80 but it's unlikely that anyone is running this still.

psyke83: That's true, but I don't think we need to worry about changing this again for the time being. If we're still alive when KDE 7.x comes out, someone will definitely need to update it, though.

Kishi85: 7.x should be fine as this is checking for KWin 5.x.x or <6.[0-7].80 or am i missing something here?

The only KWin change requiring a logic update I'm seeing here would be a newer version requiring/recommending variable rate capture with Sunshine doing the pacing again.

psyke83: Actually, I'm not sure why I thought 7.x would be an issue. Regardless, this would be the corrected version:
```
const static bool negotiate_variable_rate = kwin_version.empty() || (kwin_version[0] == 5 || (kwin_version[0] == 6 && (kwin_version[1] < 7 || (kwin_version[1] == 7 && kwin_version[2] < 80))));
```

I don't think that anyone would realistically be running such older beta releases to be impacted by this, but I can file a follow-up PR if you guys think it's worth it even just for the sake of correctness.

ReenigneArcher: Might be best to make sure it's correct. I don't know what versions all the distros are providing and whether any are using beta or not.

PR #2424, opened 2024-04-15
feat(audio): custom surround-params

mariotaku:
## Description

This PR is made for improving webOS client support. On webOS 5.0 and above, if surround channel configuration is done properly, it can re-encode OPUS stream on-the-fly & send it to surround sound system through eARC/Optical.

Currently, webOS only handles 6-channels, 4 streams & 2 coupled streams configuration, and the channel must be ordered as `[0, 1, 4, 5, 2, 3]` which Sunshine doesn't have. It won't accept stream header either, which makes this PR necessary.

With this PR, if `surroundParams` is specified (formatted like `surround-params=` in RTSP handshake, e.g. `642014523`) in HTTP launch request, it will be applied, and during RTSP handshake, our custom surround-params `a=fmtp:97 surround-params=642014523` will be advertised twice first, followed by the original ones.

### Screenshot

Not applicable 

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Dependency update (updates to dependencies)
- [ ] Documentation update (changes to documentation)
- [ ] Repository update (changes to repository files, e.g. `.github/...`)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated the in code docstring/documentation-blocks for new or existing methods/components

## Branch Updates
LizardByte requires that branches be up-to-date before merging. This means that after any PR is merged, this branch
must be updated before it can be merged. You must also
[Allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I want maintainers to keep my branch updated

PR #5743, opened 2026-09-17
fix(linux/audio): capture in the canonical channel layout

Zeyckler:
## Description

`platf::microphone()` built the capture stream's `pa_channel_map` from the Opus channel mapping. That mapping is applied by the multistream encoder in `audio.cpp` and undone by the client when it decodes, so the two cancel out. Using it as the capture channel map makes PulseAudio reorder the PCM as well, and that second reordering is never undone.

It is invisible with the built-in mappings, which are all identities, so it went unnoticed until custom surround-params arrived in #2424. A client that requests its own mapping gets permuted channels. On 5.1 with the webOS client (`surroundParams=642014523`, mapping `0,1,4,5,2,3`): centre plays from rear left, rear left from centre, LFE from rear right, rear right from the subwoofer.

The WASAPI and macOS backends take the `mapping` argument and ignore it, capturing in the device's own layout, which is why only Linux hosts are affected.

This fills the map from the channel index instead, since `map_stereo`, `map_surround51` and `map_surround71` are all the identity over `speaker_e`. `position_mapping` gets an explicit `speaker::MAX_SPEAKERS` bound and out-of-range channel counts are rejected.

Clients that don't send `surroundParams` are unaffected: the resulting `pa_channel_map` is identical for every layout in `stream_configs`. I checked the other clients and none of them send it today; android, ios, qt, embedded and tizen send only the standard `surroundAudioInfo`.

Tested by building this branch and streaming 5.1 to an LG TV running moonlight-tv, over eARC to a 5.1 system, with a six-channel file playing one tone per channel. All six land on the correct speaker; stock master reproduces the scramble above. No PipeWire workaround was in place during the test. Note that the tested session negotiated the high quality 5.1 config (1536 kbps, 6 streams / 0 coupled); I have not verified the 256 kbps config.

The channel map construction is extracted into `fill_canonical_channel_map()` so it can be tested without a running PulseAudio server, and `tests/unit/platform/linux/test_audio.cpp` asserts the canonical order for 2, 6 and 8 channels, that the maps are valid, and that out-of-range counts are rejected. To be clear about what that buys: those tests lock in the invariant this PR establishes, but they do not reproduce the original bug, which needs a live capture stream with a client-supplied mapping. The evidence that the bug is fixed is the listening test above.

`mapping` is now unused and marked `[[maybe_unused]]`. It is dead in all three backends, so if you would rather remove it from the interface (`common.h:842`, the three overrides and the two call sites in `audio.cpp`) I am happy to do that instead.

### Issues Fixed or Closed

- Fixes the host side of https://github.com/mariotaku/moonlight-tv/issues/575

## Type of Change

## Checklist

### AI Usage


ReenigneArcher:

Issue #5750, opened 2026-09-19
Portal capture hangs on live client connection after loading restore token (succeeds during self-test) — GNOME headless Remote Login session


NovaStorm775:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your issue described in the documentation?

- [x] I have read the documentation

### Is your issue present in the latest beta/pre-release?

None

### Describe the Bug

**Description**

Portal-based screen capture (capture = portal) succeeds completely and consistently during Sunshine's own internal startup self-test — including full hardware encoder validation against the GPU — but deterministically hangs on the first live client connection from a Moonlight client, at the exact same point every time, regardless of encoder setting, portal daemon state, or restore token freshness. The Moonlight client eventually times out (Error 4).

Environment specifics: this occurs specifically when the GNOME session is a headless one created via GDM's "Remote Login" feature (RDP connecting to the GDM greeter and starting a fresh session, rather than sharing an already-active local session). The session has no physical display attached; Sunshine correctly finds and streams the session's virtual Mutter output (Meta-0) during self-test.

**Steps to reproduce:**

Start a GNOME session via GDM Remote Login (RDP) on a machine with no physical display active
Launch sunshine from a terminal inside that session, or from a program shortcut, with capture = portal and encoder = vulkan explicitly set in sunshine.conf
Observe Sunshine's own startup self-test succeed fully (see log below — finds the Meta-0 stream, validates h264_vulkan/hevc_vulkan/av1_vulkan against the GPU, reaches "Starting main loop")
From a separate device, connect via Moonlight and attempt to launch any application (including the default Desktop entry)
Observe the log stop advancing immediately after [portalgrab] Loaded portal restore token from disk, with no further output, until the client times out. 

**Things ruled out during troubleshooting, each independently tested:**

Firewall/network configuration (the hang occurs before any RTSP/streaming port would be involved — this is local capture negotiation only)
Stale restore token (reproduced immediately after a fresh interactive portal approval — see "Saved portal restore token to disk" in the log, ~63 seconds before self-test, following a manual xdg-desktop-portal/xdg-desktop-portal-gnome restart)
Encoder auto-detection churn (reproduced with encoder pinned explicitly to vulkan, eliminating the nvenc/vaapi/software cascade entirely — only one self-test cycle runs, and it succeeds)
[...]

PR #5762, opened 2026-09-20
feat(linux): promote portalgrab to default with fallback capture for setup

psyke83:
Promote Portal capture to the default option on Linux with fallback capture attempted during token setup.

New flow:

* Check for a Portal restore token on disk. If it exists, select Portal capture as before.
* If no restore token exists, queue a non-blocking TaskPool task to trigger token negotiation via the XDG dialog. Since Portal cannot be used yet, clear the effective capture selection so other capture methods can be attempted during the current session.
* Once token negotiation completes, restart Sunshine so subsequent sessions use Portal screencasting.

Example:

* Config: capture = portal, auto, or unset.
* KMS or KWin capture is available/usable.
* User connects -> the session uses KMS or KWin capture. The lack of an XDG icon, together with the fatal log message shown in the Web UI, indicates that Portal is still in setup mode.
* User sees the XDG Remote Desktop dialog and clicks Share.
* Sunshine detects that setup is complete and immediately restarts.
* All subsequent sessions use Portal screencasting.

Caveat:

* The simple token check does not guarantee that the saved token is valid, but the Web UI provides a "Reset XDG Portal Capture" button to handle stale tokens.

Bugfixes:

* Prevent blocking DBus calls from preventing Sunshine from shutting down when it is restarted while an XDG Remote Desktop dialog is still open.

## Type of Change

## Checklist

### AI Usage


Kishi85:
The only thing to consider here would be to prefer kwingrab if available (and functional, fallback to Portal if init fails) for KDE/Kwin as it would work out of the box without further setup.

psyke83:
> The only thing to consider here would be to prefer kwingrab if available (and functional, fallback to Portal if init fails) for KDE/Kwin as it would work out of the box without further setup.

That seems reasonable. I made the change, but it exposed a few issues, some of which were caused from my DRM worker security fix. The PR is still not ready as I need to test more thoroughly, but the latest refresh includes some changes to kwingrab and the (most likely obsolete) special handling in misc.cpp that I identified, so you might want to take a look at those parts to make sure you're happy with the changes when it comes time to review.

I'm happy with the general direction of the Portal capture fallback for this PR, but I'm not sure if we're ready to change the defaults, specifically because KWin itself still doesn't support HDR screencasts. https://invent.kde.org/plasma/kwin/-/merge_requests/8293

Kishi85:
> > The only thing to consider here would be to prefer kwingrab if available (and functional, fallback to Portal if init fails) for KDE/Kwin as it would work out of the box without further setup.
> 
> That seems reasonable. I made the change, but it exposed a few issues, some of which were caused from my DRM worker security fix. The PR is still not ready as I need to test more thoroughly, but the latest refresh includes some changes to kwingrab and the (most likely obsolete) special handling in misc.cpp that I identified, so you might want to take a look at those parts to make sure you're happy with the changes when it comes time to review.
> 

I'll have an in-depth look and test tomorrow when I have access to my test systems again.

> I'm happy with the general direction of the Portal capture fallback for this PR, but I'm not sure if we're ready to change the defaults, specifically because KWin itself still doesn't support HDR screencasts. https://invent.kde.org/plasma/kwin/-/merge_requests/8293

Agreed on that unfortunately. There's also some other pipewire-related issues (like #5750) to look at before the defaults should be changed to avoid unnecessary problems.

psyke83:
I've changed my mind on the promotion of Pipewire capture methods, but the fallback probing and misc fixes are worth keeping. Closing to be replaced by #5774.

